PR #53, opened 2026-04-15
Dashboards: fix Safari IntersectionObserver race leaving sections invisible

### LINKED ISSUE #52: cogni-portfolio dashboard: sections invisible on initial load due to IntersectionObserver + opacity:0 race
## Summary

The dashboard produced by `cogni-portfolio:portfolio-dashboard` renders as a nearly-blank page on initial load under certain conditions — content is present in the DOM but every `.section.reveal` element stays at `opacity: 0` indefinitely. The user sees only the sticky navigation and the page header; all entity counts, products, features, and taxonomy sections appear missing.

## Affected plugin / file

- **Plugin:** `cogni-portfolio`
- **Skill:** `portfolio-dashboard`
- **File:** `skills/portfolio-dashboard/scripts/generate-dashboard.py`
- **Version observed:** `0.9.7` (likely also present in earlier versions that use the same reveal animation)

## Root cause

The generated HTML uses a scroll-reveal pattern:

```css
.reveal {
  opacity: 0;
  transform: translateY(18px);
  transition: opacity 0.5s cubic-bezier(0.22,1,0.36,1), transform 0.5s cubic-bezier(0.22,1,0.36,1);
}
.reveal.visible { opacity: 1; transform: translateY(0); }
```

```js
var observer = new IntersectionObserver(function(entries) {
  entries.forEach(function(entry) {
    if (entry.isIntersecting) {
      entry.target.classList.add('visible');
      observer.unobserve(entry.target);
    }
  });
}, { threshold: 0.08, rootMargin: '0px 0px -40px 0px' });

document.querySelectorAll('.reveal').forEach(function(el) {
  observer.observe(el);
});
```

The problem: when the page opens via `file://` (which the `open <path>` step at the end of the skill produces), **some browsers — notably Safari on macOS — do not fire the initial `IntersectionObserver` callback for elements that are already in the viewport at the moment the observer is attached.** The callback only fires on the next scroll event, which may never happen if the user's first impression is \"there's nothing here\" and they close the tab.

Because there is no fallback, all `.reveal` elements stay at `opacity: 0` forever and the dashboard appears empty.

## Reproduction

1. Run any downstream portfolio skill to produce a project with products + features (or use any existing `cogni-portfolio/<slug>/` project with populated entities)
2. Run `/cogni-portfolio:portfolio-dashboard`
3. Let the skill pick or reuse a theme and generate the dashboard
4. The skill runs `open <output>/dashboard.html` which opens the file in the default browser via `file://`
5. **In Safari** (and occasionally in Chrome when restoring a backgrounded tab): the page shows only the header and nav. Products & Features section is blank. Entity Overview cards are blank. Taxonomy heatmap is blank.
6. Scroll the page manually — sections pop in one by one as the observer finally fires.

Expected: content visible immediately on load.
Actual: content invisible until the user scrolls.

## Diagnosis steps that confirmed the issue

- `grep -c '<div class=\"product-group\">' output/dashboard.html` → **13** (content is in the DOM)
- `grep -c 'class=\"feature-item\"' output/dashboard.html` → **208** (feature items are in the DOM)
- Inspecting CSS: `.reveal { opacity: 0; ... }` confirmed
- Adding `.reveal { opacity: 1; }` via devtools → **content immediately visible**

## Suggested fix

Any one of the following in `generate-dashboard.py`:

**Option A — the one-line fix:** Start `.reveal` elements at `opacity: 1` and only use the observer to add a progressive-enhancement `visible` class for the slide-up transform. Sections above the fold are visible immediately; sections below the fold still get the reveal animation when they enter the viewport.

```css
.reveal { opacity: 1; transform: none; transition: opacity 0.5s ..., transform 0.5s ...; }
.reveal:not(.visible) { /* optional progressive-enhancement rule */ }
```

**Option B — DOMContentLoaded fallback:** After attaching the observer, immediately walk the list and mark any element whose `getBoundingClientRect().top < window.innerHeight` as `.visible`. Handles the Safari case explicitly without changing the animation for off-screen sections.

```js
document.querySelectorAll('.reveal').forEach(function(el) {
  observer.observe(el);
  if (el.getBoundingClientRect().top < window.innerHeight) {
    el.classList.add('visible');
  }
});
```

**Option C — noscript / prefers-reduced-motion safety:** Add `@media (prefers-reduced-motion: reduce) { .reveal { opacity: 1; transform: none; } }` plus a `<noscript>` style block. Addresses the bug AND accessibility at once.

Recommendation: **Option B** — minimal change, preserves the animation intent, fixes Safari/file:// reliably, and also helps any browser that defers the initial observer callback.

## Workaround (for users who hit this before a fix ships)

Manual patch after running the skill:

```bash
python3 <<'PY'
from pathlib import Path
p = Path('cogni-portfolio/<slug>/output/dashboard.html')
html = p.read_text()
html = html.replace('.reveal {\n  opacity: 0;', '.reveal {\n  opacity: 1;')
html = html.replace('</body>', \"<script>document.addEventListener('DOMContentLoaded',function(){document.querySelectorAll('.reveal').forEach(function(e){e.classList.add('visible');});});</script></body>\")
p.write_text(html)
PY
```

## Environment

- macOS 15.4 (Darwin 25.4.0)
- Claude Code CLI
- cogni-portfolio plugin v0.9.7
- Observed while viewing `dashboard.html` in Safari (default browser)

## Related / blast radius

Any other cogni-plugin that follows the same \"design-variables → HTML dashboard\" pattern documented in `cogni-workspace/references/design-variables-pattern.md` could have the same bug if it copies the same reveal CSS/JS. Worth grepping for `opacity: 0` + `IntersectionObserver` across the other dashboard-producing skills (trend-report, scoring-ui, catalog) while fixing this one.

sdh07:
Closes #52.

## Summary
- Fix Safari (and occasional Chrome tab-restore) race where `.reveal` sections on `portfolio-dashboard` and `trends-dashboard` stay at `opacity: 0` on initial load because the IntersectionObserver does not fire for elements already in the viewport at attach time.
- After attaching the observer, immediately walk the `.reveal` list and force-mark any element whose `getBoundingClientRect().top < window.innerHeight` as `.visible`, unobserving it so the animation still fires normally on scroll for off-screen sections.
- Apply the same 6-line patch in both dashboard generators (`cogni-portfolio/skills/portfolio-dashboard/scripts/generate-dashboard.py` around line 3598 and `cogni-trends/skills/trends-dashboard/scripts/generate-dashboard.py` around line 2867).
- Bump patch versions: cogni-portfolio 0.9.7 → 0.9.8, cogni-trends 0.4.4 → 0.4.5, mirrored in `.claude-plugin/marketplace.json` so Claude Desktop picks up the update.

## Scope decisions
- **In scope:** exactly the two files the triage identified, plus the four version-metadata touch points (2× plugin.json, 2× marketplace.json entries). Per blast-radius check, no other dashboard uses the `.reveal` + IntersectionObserver pattern (cogni-wiki/wiki-dashboard, cogni-visual/render-html-slides, cogni-visual/enrich-report, and cogni-website/website-build are unaffected).
- **Out of scope (deferred):** (a) removing or softening `rootMargin: '0px 0px -40px 0px'` — the immediate-reveal walk makes the negative margin harmless for the initial paint, and removing it would change scroll-in timing for every reviewer's muscle memory; (b) switching to a CSS `@supports`-based no-JS fallback so `.reveal` is visible by default when JS is disabled — worthwhile but a separate concern; (c) adding a regression test harness that actually opens the generated HTML in a headless browser — the dashboards have no such harness today and adding one is a much larger change.
- **Format choices:** cogni-portfolio's script is a Python f-string so the inserted JS must use doubled braces (`{{` / `}}`); cogni-trends's script embeds the JS in a plain string so single braces are used.

## Test plan
- [ ] Regenerate a portfolio dashboard and open `dashboard.html` in Safari — all sections visible on first paint, no scroll required.
- [ ] Regenerate a trends dashboard and confirm the left nav entries and main panels render immediately in Safari.
- [ ] Scroll the regenerated dashboard and confirm off-screen sections still fade in with the original `.reveal.visible` transform — the patch must not disable the animation, only seed the initial viewport.
- [ ] Diff the generated HTML against a pre-patch baseline for a below-the-fold section; only the new immediate-reveal walk should differ.
- [ ] Verify `cogni-portfolio/.claude-plugin/plugin.json` reports `0.9.8`, `cogni-trends/.claude-plugin/plugin.json` reports `0.4.5`, and the corresponding entries in `.claude-plugin/marketplace.json` mirror those numbers.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -74,7 +74,7 @@
     {
       "name": "cogni-trends",
       "source": "./cogni-trends",
-      "version": "0.4.4",
+      "version": "0.4.5",
       "maturity": "preview",
       "description": "Strategic trend scouting and reporting pipeline. Combines the Smarter Service Trendradar (4-dimension structure) with the TIPS content framework (Trends, Implications, Possibilities, Solutions). Multilingual trend scouting, investment theme modeling, evidence-backed CxO reports, and reusable industry catalogs. European markets (DE/FR/IT/PL/NL/ES) + US/UK.",
       "keywords": [
@@ -90,7 +90,7 @@
     {
       "name": "cogni-portfolio",
       "source": "./cogni-portfolio",
-      "version": "0.9.7",
+      "version": "0.9.8",
       "maturity": "preview",
       "description": "Portfolio messaging and proposition planning for SMEs using the IS/DOES/MEANS (FAB) framework — market-specific value propositions with competitive analysis, customer profiling, and export-ready deliverables across eight pluggable industry taxonomies.",
       "keywords": [
--- a/cogni-portfolio/.claude-plugin/plugin.json
+++ b/cogni-portfolio/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "cogni-portfolio",
-  "version": "0.9.7",
+  "version": "0.9.8",
   "description": "Portfolio messaging and proposition planning for SMEs using the IS/DOES/MEANS (FAB) framework — market-specific value propositions with competitive analysis, customer profiling, and export-ready deliverables across eight pluggable industry taxonomies.",
   "author": {
     "name": "Stephan de Haas",
--- a/cogni-portfolio/skills/portfolio-dashboard/scripts/generate-dashboard.py
+++ b/cogni-portfolio/skills/portfolio-dashboard/scripts/generate-dashboard.py
@@ -3595,6 +3595,10 @@ function fmtCurrency(val, cur) {{
 
   document.querySelectorAll('.reveal').forEach(function(el) {{
     observer.observe(el);
+    if (el.getBoundingClientRect().top < window.innerHeight) {{
+      el.classList.add('visible');
+      observer.unobserve(el);
+    }}
   }});
 
   /* Nav: smooth scroll to sections and active state tracking */
--- a/cogni-trends/.claude-plugin/plugin.json
+++ b/cogni-trends/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "cogni-trends",
-  "version": "0.4.4",
+  "version": "0.4.5",
   "description": "Strategic trend scouting and reporting pipeline. Combines the Smarter Service Trendradar (4-dimension structure) with the TIPS content framework (Trends, Implications, Possibilities, Solutions). Multilingual trend scouting, investment theme modeling, evidence-backed CxO reports, and reusable industry catalogs. European markets (DE/FR/IT/PL/NL/ES) + US/UK.",
   "author": {
     "name": "Stephan de Haas",
--- a/cogni-trends/skills/trends-dashboard/scripts/generate-dashboard.py
+++ b/cogni-trends/skills/trends-dashboard/scripts/generate-dashboard.py
@@ -2864,7 +2864,13 @@ function initGraph() {
       }
     });
   }, { threshold: 0.08, rootMargin: '0px 0px -40px 0px' });
-  document.querySelectorAll('.reveal').forEach(function(el) { observer.observe(el); });
+  document.querySelectorAll('.reveal').forEach(function(el) {
+    observer.observe(el);
+    if (el.getBoundingClientRect().top < window.innerHeight) {
+      el.classList.add('visible');
+      observer.unobserve(el);
+    }
+  });
 })();
 
 // ============ SCROLL SPY for left panel ============


sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/53
**Perspectives consulted:** plugin-dev:plugin-validator (cogni-portfolio), plugin-dev:plugin-validator (cogni-trends)

**Summary:** All perspectives passed with zero findings; the Safari scroll-reveal fix is structurally sound across both affected plugins.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-dev:plugin-validator (cogni-portfolio) | ✅ pass |
| plugin-dev:plugin-validator (cogni-trends) | ✅ pass |

### Positive observations

- plugin.json is valid JSON and passes semver: 0.9.7 → 0.9.8 is a correct patch bump for a Safari bug fix (cogni-portfolio)
- marketplace.json mirrors the new 0.9.8 version (cogni-portfolio)
- generate-dashboard.py compiles cleanly under `python3 -m py_compile` (cogni-portfolio)
- Inserted JS correctly uses doubled braces (`{{` / `}}`) matching the surrounding f-string (cogni-portfolio)
- Fix preserves scroll-reveal for off-screen sections: `observer.observe(el)` is called unconditionally before the immediate-reveal check (cogni-portfolio)
- Immediate-reveal path calls `observer.unobserve(el)` to prevent duplicate processing (cogni-portfolio)
- plugin.json parses; version 0.4.5 is valid semver and a clean patch bump from 0.4.4 (cogni-trends)
- marketplace.json entry mirrors 0.4.5 (cogni-trends)
- generate-dashboard.py passes `python3 -m py_compile` after the change (cogni-trends)
- Edited JS block lives inside a plain triple-quoted string (not an f-string), so single braces are emitted verbatim (cogni-trends)
- Fix preserves scroll-reveal for off-screen sections (cogni-trends)
- Patch is minimal and scoped (7 additions / 1 deletion in one script plus a version bump) (cogni-trends)

### Recommendation

---
*Generated by `cogni-service:service-review` Phase 1 (verdict-only mode). This is an advisory gate — the human reviewer makes the final merge decision.*

Issue #289, opened 2026-05-23
cogni-knowledge F24–F26: M12 minor follow-ups (count drift, EUR-Lex curation, helper-script version resolution)


sdh07:
Lower-priority follow-ups surfaced by the M12 alpha gate. Detail in `cogni-knowledge/references/alpha-findings.md` §"M12 alpha re-run".

- [ ] **F24 — citation-count drift.** Same draft reported as 156 (composer return) / ~169 (manifest) / 168 (verify-v1) / 169 (verify-v2) / 165 (`pipeline-summary project`). Pin one authoritative count (manifest `len(citations)`). Symptom of the same independent-tokenization coupling as F22.
- [ ] **F25 — EUR-Lex ELI/landing URLs mis-resolve or summarise.** `eli/C/2025/6233` served General Product Safety Regulation text instead of the GPAI Code endorsement; both `reg/2024/1689` URLs returned a short WebFetch summary, not the full Act text. Pipeline behaved correctly (stored as-is, composer declined to cite the mismatch). Curation should prefer `artificialintelligenceact.eu` article pages over EUR-Lex landing URLs for normative text.
- [ ] **F26 — `knowledge-ingest` helper-script resolution picks lexically-first cached cogni-wiki version.** `resolve_wiki_ingest_scripts()` takes the first glob match (`0.0.16` on a multi-version dev cache), not the installed `0.0.45`. Sort by version or read the installed version. Only bites dev boxes with multiple cached versions.

Surfaced by the M12 alpha gate (2026-05-23). Full context: `cogni-knowledge/references/alpha-findings.md` §"M12 alpha re-run" and PR #285. Part of epic #264.

PR #290, opened 2026-05-23
cogni-knowledge 0.0.28: M12 gate blockers — F22 + F21 + F23 (Slice 10)

### LINKED ISSUE #286: cogni-knowledge F21: verify/revise wall-clock scales ~linearly with citation count (M12 C3 blocker)
**Blocks the v0.1.0 flip (M12 criterion C3: claim-verify < 5 min).**

At 169 citations each zero-network `wiki-verifier` pass took ~16–18 min and `revisor` ~23 min; the full verify→revise→verify→revise→verify loop is ~90 min. Zero-network is genuinely won vs cogni-claims (no re-fetch, beats the 20–30 min baseline), but a **single** verifier dispatch doing per-citation LLM scoring dominates wall-clock, and the 5-min target — calibrated for the ~37-citation contract example — is unreachable at realistic draft size.

### Primary fix: fan out the verifier

Verification is **embarrassingly parallel** — each citation's verdict is independent (it reads one cited page's `pre_extracted_claims` and one draft sentence; no cross-citation dependency). Composition can't be sharded that way (it needs whole-draft coherence), which is why the composer is a single dispatch but curate/fetch/ingest already fan out. So:

- `knowledge-verify` shards `citations[]` into N batches.
- Dispatches N `wiki-verifier` instances **in parallel**, each scoped to its citation subset (the agent already reads only the pages its citations reference).
- Merges the per-shard `verified[]` / `deviations[]` / `counts` fragments — the same merge pattern as `candidate-store` / fetch-manifest / ingest-manifest.

Wall-clock drops ~linearly with shard count **while preserving the LLM judgment**. Re-baseline the 5-min target per-shard once this lands.

### Complementary (not either/or)
- Deterministic script pre-filter: confirm `excerpt_quote` substring-presence at `excerpt_position` with no LLM, escalate only ambiguous cases to the model.
- Cap/segment citations per section.

Surfaced by the M12 alpha gate (2026-05-23). Full context: `cogni-knowledge/references/alpha-findings.md` §"M12 alpha re-run" and PR #285. Part of epic #264.

### LINKED ISSUE #287: cogni-knowledge F22: make draft_position a stable composer-emitted / verifier-consumed id (off-by-one drift) (M12 C4 blocker)
**Blocks the v0.1.0 flip (M12 criterion C4: every citation aligns to a claim).** This is the F20 "watch at M12" sentence-delimiter item, now confirmed as the dominant cause of `unsupported` verdicts.

The composer emits `section:sentence` positions (e.g. `06:09`); the verifier independently re-tokenizes the draft into sentences and the counters drift by one in some paragraphs (section-06 timeline/fines clusters). Citations get flagged `unsupported` because the pointer lands on the *neighbouring* sentence, not because the evidence is wrong. Sub-case: **phantom manifest entries** — the manifest declares a citation at a `draft_position` whose draft text carries a *different* wikilink (the killed round-2 revisor spent its whole run reconciling this).

**Fix:** make `draft_position` a stable identifier the composer writes and the verifier/revisor consume verbatim (rather than each re-deriving it from independent sentence tokenization). Pairs with the F20 tokenizer rewrite (abbreviation lookbehind + numbered-list continuation for `Article 53(1)(c)`, `10^25`, `Annex XI`, `Dr.`).

Surfaced by the M12 alpha gate (2026-05-23). Full context: `cogni-knowledge/references/alpha-findings.md` §"M12 alpha re-run" and PR #285. Part of epic #264.

### LINKED ISSUE #288: cogni-knowledge F23: revisor biases to citation-drop over re-alignment; 2-round cap doesn't converge
Observed at the M12 alpha gate. Round 1 applied 14 `rephrase` + **14 `drop`** of 28 deviations. A `drop` deletes the citation / rewrites the sentence as non-evidence-based, so the `unsupported` count falls partly by *eroding the evidence base* rather than correcting alignment — a quality regression hidden inside an improving metric. Combined with F22 the loop terminated at 10 unsupported on draft-v2 and would not reach 0 within the structural 2-round cap.

**Fix:** the revisor should prefer re-pointing to the correct on-page claim (cheap — claims are local to the cited page) before dropping the citation. Largely dissolves once F22 lands (most "misalignments" are off-by-one, not genuinely unsupported).

Surfaced by the M12 alpha gate (2026-05-23). Full context: `cogni-knowledge/references/alpha-findings.md` §"M12 alpha re-run" and PR #285. Part of epic #264.

sdh07:
## Summary

Clears the three structural blockers the first **M12 alpha gate** (run 2026-05-23) surfaced. The gate was **HELD** on **C3** (each `wiki-verifier` pass ~16–18 min at 169 citations ≫ 5 min target) and **C4** (verify→revise doesn't converge within the 2-round cap) — both rooted in **F20**: the composer, verifier, and revisor each independently re-tokenize the draft into `section:sentence` positions and drift by one on EU-AI-Act prose (`Article 53(1)(c)`, `10^25`, `Annex XI`, `Dr.`), so a citation is flagged `unsupported` because the pointer landed on the *neighbouring* sentence.

Target version **v0.0.28** (stays **incubating**).

- **F22 (`Closes #287`) — stable carried citation positions.** Each `citation-manifest.json` entry gains a stable `id` (`cit-001`, …) and `draft_sentence` (the cited sentence copied verbatim). The verifier scores `draft_sentence` **directly** against the cited claim and never re-tokenizes the draft — dissolving the C4 off-by-one. `draft_position` is demoted to a best-effort locator; `id` is the universal join key (verifier echo → orchestrator prune → revisor map); `draft_position_out_of_range` → `sentence_not_in_draft`. Schema stays `0.1.0` (additive). **The F20 tokenizer rewrite is deliberately not done** — `draft_sentence` makes it non-load-bearing.
- **F21 (`Closes #286`) — fan out the verifier.** `knowledge-verify` Step 3.1 is now shard → dispatch N `wiki-verifier` in parallel → merge, via the new `scripts/verify-store.py`. The verifier gains two optional, fully backward-compatible params (`CITATIONS_PATH` / `VERIFY_OUT_PATH`); omitting both is today's single-dispatch behaviour. `--shard-size` default 40 (169 citations ≈ 5 parallel shards). The orchestrator asserts `shards_merged == shard_count` so a crashed shard can't ship partial verification. C3 re-baselined as per-shard wall-clock. **No `fcntl.flock`** in `verify-store.py` — shards are partition-disjoint and `merge` is single-shot, so there is no shared-write contention (the one structural difference from `candidate-store.py`).
- **F23 (`Closes #288`) — revisor re-points before dropping.** The revisor exhausts on-page re-pointing first (scan all `pre_extracted_claims` on the cited page for a covering claim, rephrase + repoint `claim_id`); **drop is the last resort**. `fixes_summary` gains a `repoint` count so the metric distinguishes re-alignment from evidence erosion. The revisor locates sentences by exact-string-search of `draft_sentence` (F22) and preserves each entry's `id`.

Epic #264.

### ⚠️ Not in this PR

- **The M12 gate is NOT re-run here**, and the **v0.1.0 + Preview maturity flip is a separate, later landing** — only after a fresh `.alpha/` gate re-run shows C3 (per-shard verify < 5 min) and C4 (verify→revise converges; `drop` down / `repoint` up) green. We don't cross the 0.1.0 stability boundary speculatively.
- **#289 (F24–F26)** — citation-count drift, EUR-Lex curation, helper-script version resolution — stays out of this slice.
- The deterministic substring pre-filter ("complementary" in #286) stays a documented future option; fan-out alone is projected to clear C3.

## Test plan

- [x] `bash tests/test_verify_store.sh` (new) — shard split shape, subset validity + exact union, single-shard case, merge recount, idempotent re-shard, malformed-input rejection
- [x] Full suite green — `for t in tests/test_*.sh; do bash "$t"; done` → 19/19 pass (incl. updated `test_verify_contract.sh` + `test_compose_contract.sh`, and `test_skill_contracts.sh` audit-grep)
- [x] No stale `draft_position_out_of_range` in the live surface (only the new not-present assertion + historical F20 finding + changelog rename note)
- [ ] **Operator-run, gated next:** re-run the M12 alpha gate on a fresh `.alpha/` base — confirm C3 per-shard < 5 min and C4 convergence before the v0.1.0 bump

https://claude.ai/code/session_016Kszz4TH7P6xJuVdc1Fsiz

---
_Generated by [Claude Code](https://claude.ai/code/session_016Kszz4TH7P6xJuVdc1Fsiz)_


sdh07:
## Code review pass — fan-out hardening (`9d9ea6f`)

Ran a multi-angle correctness review over the Slice 10 diff. The fixes below land in a follow-up commit; full test suite green (19 files, 15 assertions in `test_verify_store.sh`). All findings clustered in the new `verify-store.py merge` path and the `knowledge-verify` prune — the LLM-prompt edits were clean.

### Fixed

1. **`merge` had no completeness check** — a missing / under-populated / duplicate shard fragment merged with `success:true` (only an *internal* `counts.total` tally was checked, and `shards_merged == shard_count` counts files, not citations). A crashed or truncated verifier shard would silently drop its `unsupported` citations so the revisor never sees them and finalize ships them unverified. **Fix:** `merge` now reads the shard inputs to learn the expected citation-id set and requires the recombined `verified+deviations` to equal it (count **and** id-set).
2. **`{None}`-key over-prune** in `knowledge-verify` Step 3.2 — a deviation missing its `id` put `None` into the prune set, deleting *every* id-less citation from the manifest. **Fix:** filter `None` out of the stale-id set.
3. **Uncaught `AttributeError`** (raw traceback instead of the JSON envelope) in `merge`/`shard` on non-dict JSON — a non-object manifest, fragment, or verdict entry. Broke the envelope contract the orchestrator parses. **Fix:** `isinstance(..., dict)` guards at all three spots.
4. **Stale validator** — `knowledge-compose` Step 5 (the only deterministic gate on the manifest) still required only `draft_position`/`wiki_slug`/`claim_id`; F22 added `id`+`draft_sentence` but left the check behind, so the composer could omit the new load-bearing fields and still report success. **Fix:** the assert now requires `id` + `draft_sentence`.
5. **`unsupported` mis-filed into `verified[]`** escaped the revisor (it only triages `deviations[]`); the count tally couldn't see it. **Fix:** `merge` rejects it.
6. **Duplicate `id` across fragments** silently merged (inflated counts, double verdict). **Fix:** `merge` rejects it.
7. **Zero-citation manifest** → 0 shards → `merge` hard-errored "no fragments". **Fix:** Step 2 aborts cleanly, matching the non-empty-`citations[]` precondition.
8. **Non-atomic manifest rewrite** in Step 3.2 (`write_text` truncate-in-place) could corrupt the canonical manifest on interrupt (pre-existing; fixed in the touched block via temp + `os.replace`).
9. **Revisor** Phase 2 was underspecified when `draft_sentence` isn't found in the draft — now drops the entry.

### Noted, not fixed
- 2-digit shard zero-pad mis-orders `merge` above 100 shards (>4000 citations) — order is non-load-bearing (everything keyed by `id`); latent/cosmetic.
- Step 3.2 prune doesn't bump `draft_version` — M9 uses the manifest's `wiki_slug` as source of truth (no id cross-join against `verify-vN.json`), so no impact path; pre-existing.

### Clean (checked, refuted)
Glob version-namespacing (`verify-shard-*-v1.json` doesn't match `*-v10`/`v11`; `shard-*` doesn't match `verify-shard-*`); the no-flock design (verifiers write disjoint paths, `merge` is single-shot); partition slicing; bash heredoc interpolation in the tests.

---

PR #313, opened 2026-05-25
cogni-visual 0.16.23: surface render-html-slides tier-0 fallback reason (#241)

### LINKED ISSUE #241: cogni-visual: render-html-slides silently falls back to tier-0 when COGNI_WORKSPACE_ROOT is unexported
## Symptom

When invoking `cogni-visual/skills/render-html-slides/scripts/generate-html-slides.py --theme-slug cogni-work` from a shell where `COGNI_WORKSPACE_ROOT` is set but **not exported** (e.g. `VAR=value` on a separate line followed by a later `python3 …` invocation), the script silently falls back to the tier-0 code path and returns `tokens_css_imported: false` with no warning.

The fallback itself is correct per the documented backcompat contract — every existing tier-0 caller flows through this path. But the **silent** part is a footgun: the caller asked for tier-aware behaviour via `--theme-slug`, and the only signal that it didn't get it is the boolean in the output JSON envelope.

## Reproduction

```bash
# Variable set in the shell but never exported to the child process:
COGNI_WORKSPACE_ROOT="/Users/me/insight-wave/cogni-workspace"

python3 .../generate-html-slides.py \
  --slide-data slide-data.json \
  --design-variables design-variables.json \
  --output out.html \
  --theme-slug cogni-work
# → {"status": "ok", ..., "tokens_css_imported": false}   ← silent fallback
```

Compare with the working invocation:

```bash
COGNI_WORKSPACE_ROOT=".../cogni-workspace" \
python3 .../generate-html-slides.py --theme-slug cogni-work …
# → {"status": "ok", ..., "tokens_css_imported": true}
```

## Root cause

`resolve_themes_dir` (lines 79–107 in `scripts/generate-html-slides.py`) consults `os.environ.get("COGNI_WORKSPACE_ROOT")` and, when that returns None, walks up from the script's directory looking for a sibling `cogni-workspace/themes`. Because the script lives in the plugin cache (`~/.claude/plugins/cache/insight-wave/cogni-visual/…/skills/…/scripts/`), the walk-up never finds the workspace themes dir. So `themes_dir` becomes `None`, `resolve_tokens_css` short-circuits to `None`, and the tier-aware path is skipped without any user-visible signal.

## Suggested fix

Two complementary nudges, neither breaks the backcompat contract:

1. **Diagnostic field in the output envelope**. When `--theme-slug` is set but `resolve_themes_dir` returns `None`, populate a `theme_slug_resolution` key in the JSON output:
   ```json
   "theme_slug": "cogni-work",
   "tokens_css_imported": false,
   "theme_slug_resolution": "themes_dir_unresolved: $COGNI_WORKSPACE_ROOT unset, walk-up found no cogni-workspace/themes"
   ```
   Same key reports `manifest_missing`, `tokens_tier_absent`, `tokens_css_missing` in the analogous `resolve_tokens_css` short-circuits, so callers can distinguish "theme is tier-0" from "I couldn't find it" without re-walking the resolution themselves.
2. **`--verbose` flag** that additionally writes a one-line diagnostic to stderr (`themes_dir=…`, `manifest_path=…`, `tokens_css=…`). Off by default — preserves the stdout JSON contract.

Both fixes leave the silent-fallback semantics intact for callers that don't care, and give callers that do care a clean way to detect it.

## How discovered

Filed during the cogni-work canary smoke test (issue #240, post-merge). First invocation set `COGNI_WORKSPACE_ROOT` as a plain shell variable without export, got `tokens_css_imported: false`, briefly suspected the importer / mapping doc / manifest before realising the env var hadn't propagated.

## Severity

S4-enhancement — current behaviour is documented-correct and harmless when the caller checks the boolean, but the failure mode is non-obvious. Worth a small DX patch.

sdh07:
## Summary

Resolves the DX footgun in #241: `render-html-slides/scripts/generate-html-slides.py` silently fell back to the tier-0 inline-tokens path when `--theme-slug` was set but tier-1 `tokens.css` couldn't be resolved (e.g. `COGNI_WORKSPACE_ROOT` set but not *exported*, so the child process never sees it and the plugin-cache walk-up finds no `cogni-workspace/themes`). The only prior signal was `tokens_css_imported: false`.

- **`theme_slug_resolution` envelope field** names *why* the tier-1 path was skipped, letting callers distinguish "this theme is tier-0" from "I couldn't find the workspace": `imported`, `themes_dir_unresolved` (with a human hint), `manifest_missing`, `manifest_unreadable`, `tokens_tier_absent`, `tokens_css_missing`. `null` when `--theme-slug` is omitted.
- **`--verbose` flag** writes a one-line `themes_dir=… manifest_path=… tokens_css=… resolution=…` diagnostic to stderr. Off by default — stdout stays JSON-only.
- Internals: `resolve_tokens_css` → `resolve_tokens_css_with_reason` returning `(path, reason)`; the old name had a single in-file caller and is removed (no shim).

Silent-fallback semantics and byte-identical HTML output are **unchanged** for callers that don't inspect the field.

Version bumped `cogni-visual` `0.16.22 → 0.16.23` (patch, additive DX) in `plugin.json` + `marketplace.json` mirror.

## Test plan

- [x] `python3 cogni-visual/skills/render-html-slides/evals/run.py` → `{"passed": true, "total": 3, "failed": 0}` (each of the 3 cases now also asserts `theme_slug_resolution`: `null` / `imported` / `manifest_missing`)
- [x] Forced `themes_dir_unresolved` via bogus `--themes-dir` → envelope reports the reason + hint; `--verbose` line on stderr
- [x] Tier-0 `_template` → `manifest_missing`; working `cogni-work` path → `imported`, `tokens_css_imported: true`
- [x] stdout remains JSON-only with `--verbose` (diagnostic on stderr)
- [x] `python3 -m json.tool` valid for `plugin.json` and `marketplace.json`

Closes #241

https://claude.ai/code/session_01UzTMbbKkExSexd1vCyYqpE

---
_Generated by [Claude Code](https://claude.ai/code/session_01UzTMbbKkExSexd1vCyYqpE)_


sdh07:
## Review — verified, safe to merge, with polish items

Functionally exercised the branch's script end-to-end (temp copy, all reason codes fire correctly): `cogni-work`→`imported`/`true`, `_template`→`manifest_missing`/`false`, bogus dir→`themes_dir_unresolved`+hint/`false`, no-slug→`null`; `--verbose` keeps stdout pure JSON with the diagnostic on stderr. Rename is complete (zero dangling `resolve_tokens_css` refs — the no-shim decision is safe since the eval tests end-to-end and never imports the function). Versions consistent across `plugin.json` / `marketplace.json` / README tree line; HTML output byte-identical. No correctness issues found.

### Polish items for the remote to address

1. **Add a regression case for `themes_dir_unresolved` — the actual #241 footgun.** The eval extends the 3 existing cases (`null` / `imported` / `manifest_missing`) but the precise condition the issue was filed about is only manually verified. Add a 4th case to `evals/run.py`:
   ```python
   def case_4_themes_dir_unresolved():
       html, status = run_render(theme_slug="cogni-work", themes_dir="/nonexistent/bogus")
       # assert status.get("tokens_css_imported") is not True
       # assert (status.get("theme_slug_resolution") or "").startswith("themes_dir_unresolved")
   ```
   Note the **`startswith`** — exact-equality would silently miss because of point 2.

2. **Prefix-match contract is a sharp edge.** `theme_slug_resolution` returns bare codes for every case *except* `themes_dir_unresolved`, which appends a human hint — and that's the one code a consumer is most likely to branch on. A naive `== "themes_dir_unresolved"` silently fails. Two options:
   - **Preferred:** keep `theme_slug_resolution` a bare code always, add an optional `theme_slug_resolution_detail` for the hint (machines exact-match, humans read the detail).
   - **Minimum:** leave as-is (it mirrors the issue's example and SKILL.md documents prefix-matching), but make the prefix-match requirement loud in the SKILL.md consumer note.

3. **Minor (optional):** `theme_slug_missing` is unreachable from the CLI (resolver only called inside `if args.theme_slug:`) — harmless defensive branch, just noting it can't appear in the envelope. And the `--verbose` line recomputes `manifest_path` independently of the resolver, which could drift if the resolver's path construction changes.

Items 1 and 2 are the ones worth doing before merge; item 3 is cosmetic.

sdh07:
Addressed in 140e7eb.

**Item 1 — regression case added.** New `case_4_themes_dir_unresolved` in `evals/run.py` exercises the precise #241 footgun: `--theme-slug cogni-work` with a bogus `--themes-dir` (which `resolve_themes_dir` rejects → `None`). Eval is now 4/4 green. It asserts `tokens_css_imported is not True`, `theme_slug_resolution == "themes_dir_unresolved"`, and a non-empty detail hint.

**Item 2 — went with your Preferred option.** `theme_slug_resolution` is now **always a bare, exact-matchable reason code** (no appended hint on any case). The human hint moved to a companion `theme_slug_resolution_detail` field (`null` unless populated — today only for `themes_dir_unresolved`). This removes the prefix-match sharp edge entirely; machines exact-match the code, humans read the detail. SKILL.md updated to document the bare-code contract and the detail field, so case_4 can exact-match rather than `startswith`.

Footgun envelope now:
```json
"tokens_css_imported": false,
"theme_slug_resolution": "themes_dir_unresolved",
"theme_slug_resolution_detail": "$COGNI_WORKSPACE_ROOT unset / not a dir, walk-up found no cogni-workspace/themes"
```

**Item 3 — left as-is per your "cosmetic" call.** `theme_slug_missing` remains a defensive (CLI-unreachable) branch, and the `--verbose` line still recomputes `manifest_path` independently. Happy to fold `manifest_path` into the resolver's return if you'd rather not carry the duplication — just say the word.

---

Issue #311, opened 2026-05-25
[Task] cogni-knowledge: live German end-to-end bake-in run (Slice 13 exit verification)


sdh07:
## What

Run a **live German (`--output-language de`) end-to-end pipeline** (`knowledge-plan` → `knowledge-curate` → `knowledge-fetch` → `knowledge-ingest` → `knowledge-compose` → `knowledge-verify` → `knowledge-finalize`) against a fresh `.alpha/` base and confirm the Slice 13 (v0.1.4) exit criterion holds in practice — not just under contract/unit coverage.

## Why

Slice 13 (#303 → #301 → #300, landed in #310) fixed the three German/localized-output bugs surfaced by the **first** real DACH run ("Lean Canvas für Insight-Wave"). The fixes shipped on contract + unit coverage (consistent with Slices 3–11); the helper logic is now unit-tested in `_knowledge_lib.py` (`slugify`, `ref_heading`, `first_url`, `md_link_dest`, `strip_reference_section`, `renumber_inline_citations`). But the **full pipeline has not been re-run in German since the fixes**. This live run is the natural bake-in signal that the Phase 6 gate (cogni-research retirement at v1.0.0) ultimately waits on — Phase 6 stays gated on ≥ 2 minor versions of clean real v0.1.x usage.

## Exit criterion (confirm all)

A German run deposits a synthesis page with:
- [x] **One** localized reference list — `## Referenzen`, no duplicate `## References` (#301)
- [x] **Clickable numbered** inline citations — `<sup>[N](url)</sup>`, no raw `[[sources/<slug>]]` in prose; wikilinks only in the reference list (#300)
- [x] **Clean transliterated slugs** — source + synthesis slugs like `lean-canvas-fuer-insight-wave`, not `...f-r...` (#303)
- [x] Body inline `[N]` numbering matches the deposited reference list (renumber pass), including after any revisor full-source-drop
- [x] German body + section headings with proper encoding (`ä/ö/ü/ß`), no ASCII-folded prose

## Context

- Epic: #264 (Phase 5 v0.1.x bake-in)
- Roadmap: `cogni-knowledge/references/absorption-roadmap.md` → "Phase 5 v0.1.x bake-in" / Slice 13
- Deferred from: #310 (the v0.1.4 landing); review finding **B**
- `.alpha/` fixtures are gitignored / local-only

sdh07:
[...]

PR #316, opened 2026-05-25
cogni-knowledge 0.1.6: Slice 15 cost/wall-clock (#299, #305)

sdh07:
## Summary

Slice 15 of the cogni-knowledge Phase 5 v0.1.x bake-in (epic **#264**) — two cost / wall-clock wins for the inverted pipeline. Maturity stays **Preview**; no schema bump. Closes **#299**, **#305**.

- **#299 — one-wave curator fan-out.** `knowledge-curate` Step 3 now dispatches all N sub-questions in a single assistant message of N `Task(source-curator)` calls instead of waves of ≤3 (which ran a 6-SQ plan as two sequential waves). The limit is *derived* — `knowledge-plan` caps a plan at 3–7 sub-questions so N ≤ 7, peak concurrent web calls = N (each curator's WebSearch/WebFetch are sequential within itself), and the verifier fan-out already runs at this scale.
- **#305 — incremental verify + patch-in-place revisor.** Three coupled changes cut the verify→revise loop cost:
  - **Patch-in-place revisor** — the orchestrator `cp`s `draft-v{N+1}` and `revisor` (now with `Edit`) patches only the changed sentences instead of regenerating the ~5k-word draft, keeping untouched sentences byte-identical.
  - **Incremental re-verify** — `verify-store.py` gains `shard --only-ids`, `merge --manifest`, `merge --carry-forward-from`; round ≥1 re-scores only the touched delta and carries the rest forward, so the canonical `verify-vN.json` stays complete while the shards shrink.
  - **Deterministic substring pre-filter** — new `verify-store.py prefilter` (+ stdlib `_knowledge_lib.parse_pre_extracted_claims`) marks trivially-`verbatim` citations without an LLM call.

## Review hardening (second commit)

A multi-angle code review of the prefilter found it could emit a wrong-too-strong `verbatim` (silently skipping verification). The fail-safe guarantee now actually holds:

- Parser drops YAML block-scalar headers (`>`/`|`) instead of capturing the bare indicator (which substring-matched every `<sup>` marker); prefilter requires a **substantial** (≥24-char) needle.
- Prefilter reads the current draft (`--draft`) and only asserts `verbatim` on an **in-draft** sentence (`sentence_not_in_draft` parity), **NFC-normalizes** both sides (multilingual matches), and treats a duplicate `claim_id` on a page as ambiguous.
- `DELTA_IDS` derives from a **deterministic** pre-revisor manifest-snapshot diff, not the LLM's self-reported `fixes_applied`; `knowledge-verify` runs `shard` every round so a stale numbered fragment can't leak into `merge`.
- `merge --manifest` rejects null/duplicate manifest ids; `--carry-forward-from` rejects a duplicate-id prior file; `_unquote_scalar` decodes double-quoted values via `json.loads`.

## Test plan

- [x] `bash tests/test_verify_store.sh` — `shard --only-ids`, `prefilter` (match / cross-lang / unparseable / **block-scalar / short-needle / stale-sentence / NFC-NFD / duplicate-claim** false-positive guards), `merge --manifest` conservation, `--carry-forward-from` (delta + carry == manifest, empty-delta, **duplicate-manifest-id / duplicate-prior-id** rejections), reshard-preserves-prefilter
- [x] `bash tests/test_knowledge_lib.sh` — `parse_pre_extracted_claims` units (block-scalar dropped, inline comment, column-0 bullets, malformed → `[]`)
- [x] `bash tests/test_verify_contract.sh` + `tests/test_skill_contracts.sh` — revisor `Edit`/patch-in-place, `cp` substrate, deterministic-diff DELTA, `--draft`, always-run-shard, one-wave fan-out
- [x] Full suite: **20/20 test files green**
- [ ] Live end-to-end convergence run — **not** in this PR; folds into the open #311 German bake-in (consistent with how Slices 3–14 shipped, on contract + deterministic-gate coverage)

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01YFEW1GVBh1vC2AJnUk4meA)_


sdh07:
## Local `/review` verdict — ✅ safe to merge (2 minor follow-ups)

Reviewed the full logic of `verify-store.py`, `_knowledge_lib.parse_pre_extracted_claims`, `agents/revisor.md`, `skills/knowledge-verify/SKILL.md`, `skills/knowledge-curate/SKILL.md`, the test additions, and the version bumps. The implementation is correct and defensively written. Findings below are **one LOW + one NIT** — neither blocks merge.

### Strengths worth keeping
- **`DELTA_IDS` derived deterministically from a pre-revisor manifest snapshot diff, not the revisor's `fixes_applied`** — an LLM under-reporting a fix can't silently carry a stale verdict forward. This is the right call.
- **Pre-filter is strictly fail-safe**: ≥24-char needle floor, block-scalar headers dropped, NFC-normalized both sides, ambiguous `claim_id` dropped, and it re-reads the draft to enforce the `sentence_not_in_draft` staleness guard before classifying. Only ever asserts `verbatim`; never a deviation/drop.
- **Conservation is rigorous**: `merge --manifest` validates against the current manifest id-set; rejects null/duplicate manifest ids, duplicate prior-file ids, and a manifest id with no prior verdict.
- **Idempotent reshard scoped to numbered fragments** (`verify-shard-[0-9]*`) so the prefilter fragment written earlier in the same round survives.
- Test coverage hits the real risk surface (block-scalar / short-needle / stale-sentence false-positive guards, NFC-vs-NFD match, cross-language fall-through, "prefilter never emits a deviation").

### Findings

**[LOW] Substring pre-filter can over-assert `verbatim` on a sentence that quotes the excerpt but adds an unsupported qualifier/negation.**
The match is `excerpt_quote in draft_sentence` — "contains", not "is". A sentence like *"Contrary to early drafts, the regulation applies to high-risk systems deployed in the EU **only after 2027**."* would substring-match a ≥24-char excerpt yet carry an unsupported scope, and skip the LLM that would flag it. The pipeline's own generators don't produce adversarial negations, so real-world frequency is low — but it's a genuine semantic gap in a zero-LLM auto-pass.
*Suggested tightening:* require the needle to cover essentially the whole sentence — e.g. `draft_sentence` stripped of its `<sup>[N](url)</sup>` marker and trailing punctuation ≈ the needle (or needle length ≥ ~90% of the stripped sentence). That turns "contains the quote" into "is the quote", which is what `verbatim` should mean. Add a test case for the qualifier-wrapped exact quote.

**[NIT] The `.citation-manifest.pre-r<ROUND>.json` snapshots aren't cleaned up.**
`knowledge-finalize` clears `verify-shards/` but these snapshots linger in `.metadata/` root. Harmless; either add them to finalize's scratch-removal step or note them as intentional debug artifacts.

### Notes
- Versions bumped in **both** `plugin.json` and `marketplace.json` to 0.1.6 ✅
- No live convergence run in this PR (deferred to the open #311 German bake-in) — consistent with how Slices 3–14 shipped, on contract + deterministic-gate coverage.

**Recommendation:** address the LOW prefilter tightening (it's a small, well-contained change with a clear test), keep or note the NIT, then squash-merge.

_Posted from a local `/review` pass for remote polish._

sdh07:
Both addressed in `28396e9`.

**[LOW] prefilter `verbatim` now means "is the quote", not "contains it".** The match strips the inline `<sup>[N](url)</sup>` marker(s) (new `_knowledge_lib.strip_inline_citation_markers`) and requires the needle to cover ≥ `PREFILTER_COVERAGE_RATIO` (0.9) of the resulting sentence. Your exact example — *"Contrary to early drafts, … shall be considered high-risk **only after 2027**."* — now drops below the coverage bar and falls through to the LLM. Added a test case for the qualifier-wrapped exact quote (`cit-qual`), and reworked the existing prefilter fixtures to genuine whole-sentence matches.

**[NIT] snapshots cleaned up.** `knowledge-verify` Step 4 now `rm -f`s the `.metadata/.citation-manifest.pre-r*.json` diff snapshots once the run is validated.

Full suite green (20 files). Ready for squash-merge.

---

sdh07:
## Re-review: ✅ both findings resolved — approving for merge

- **[LOW]** Prefilter tightened to whole-sentence verbatim (`strip_inline_citation_markers` + 0.9 coverage gate). Verified empirically: the qualifier-wrapped exact quote (`…only after 2027`) now falls through to the LLM; only the whole-sentence quote auto-passes.
- **[NIT]** Per-round `.citation-manifest.pre-r*.json` snapshots cleaned up in Step 4.

Nice touch: the new test encodes the exact qualifier-wrapped case, and the dependent NFC / duplicate-id tests were *adjusted* (not deleted) to stay valid under the new coverage rule. No new findings. Squash-merging.

Issue #324, opened 2026-05-27
[Bug] cogni-knowledge: wiki index.md one-liners truncated mid-word — summary authoring is script-stored, not LLM-bounded


sdh07:
## Context
Surfaced during the #311 live German bake-in run. The wiki **index.md** category entries are cut off **mid-word** — they read "abgehackt". Example observed:

```
[[art-10-ki-vo-daten-und-daten-governance-ki-verordnung]] — Art. 10 KI-VO legt Daten-Governance-Pflichten für Hochrisiko-KI-Systeme fest: Qualitätsanforderungen an Trainings-/Validierungs-/Testdaten, Bias-Korrektur und Ausnahme für Sonderka
```

("…Sonderka" = truncated mid-word, should be "Sonderkategorien von Daten".)

## Root cause (two layers)
1. **Mechanical hard-truncation.** `knowledge-ingest` Step 4.2 hands a free-text `--summary` to `wiki_index_update.py`, which stores the string **verbatim**. The summary gets clipped to ~180 chars by a hard slice (`[:180]`) with **no word/sentence-boundary awareness** → mid-word cut. (In this run the orchestrator-side loop did the `[:180]`; but the contract itself only says "≤180 chars" and nothing enforces a clean boundary, so the same artifact appears whenever the upstream summary exceeds the cap.)
2. **Design question (the substantive one).** `index.md` is **reader-facing prose** AND the **dominant signal `wiki-coverage.py` reads for read-before-web (P1.3, #309)** — coverage scoring leans on `title` + the index one-liner. So the one-liner quality is load-bearing twice over. A script can *store* the line, but **authoring** a complete, self-contained ≤N-char thematic summary is an LLM task (or, at minimum, the truncation must fall on a word/sentence boundary with an ellipsis). Right now neither holds.

## Proposed fix (options)
- **Minimal (deterministic):** make the summary cut **word/sentence-boundary aware** with a trailing `…` — e.g. a `_knowledge_lib.clamp_summary(text, max_len)` helper that never splits a token; `wiki_index_update.py` could enforce it defensively too. Removes the "abgehackt" artifact for ~zero cost.
- **Better (LLM-authored):** have `source-ingester` emit a purpose-built **complete** one-line index summary (≤ N chars by construction, not a truncation of a longer abstract) as a dedicated field, distinct from any longer page abstract. The orchestrator passes that field straight through — no clipping.
[...]

Issue #323, opened 2026-05-27
[Change] cogni-knowledge: investigate parallel fan-out / batch-size boundaries (knowledge-ingest --batch-size, curate wave)


sdh07:
## Context

Surfaced during the #311 live German bake-in run. `knowledge-ingest` dispatches `source-ingester` in **sequential batches of `--batch-size` (default 8)**. On a 67-source run that is **9 sequential waves**, each of which drains the concurrency pool to zero and waits for its slowest agent before the next wave starts (a per-wave barrier). Observed: each wave of 8 takes ~100–145 s (bounded by the slowest ingester), so the barrier overhead is real.

The user has observed fan-outs of **40+ agents in parallel** working fine in Claude Code. So the question is whether our conservative batch-of-8 is leaving wall-clock on the table.

## Question to answer

1. **What is Claude Code's actual concurrent-subagent ceiling?** Issuing N `Task` calls in one message: how many run truly concurrently vs. queued? Does it vary by model / settings (`Task` concurrency env var)? Empirically measure (e.g. 8 vs 16 vs 32 vs 48 in one message; timestamp start/finish).
2. **Where do returns / observability degrade?** Large single-message fan-out returns a large tool-result block to the orchestrator — at what fan-out does orchestrator context or readability degrade? (Total return volume is batch-size-independent, but *per-message* volume is not.)
3. **What is the right default for `knowledge-ingest --batch-size`?** Likely raise it well above 8 (or make it "one wave, capped at the CC ceiling"), mirroring the #299 one-wave curate fan-out (which is already N≤7 = single wave). Note ingest agents are heavier than curators (each spawns a `claim-extractor`), so the ceiling may differ.
4. Should the cap be **derived** (read a CC concurrency setting) rather than a guessed constant — same posture as #299's "derived, not guessed" ceiling?

## Scope
- `knowledge-ingest` SKILL.md Step 3 (batch dispatch + the per-batch barrier).
- Possibly `knowledge-curate` (already one-wave, N≤7) and `knowledge-verify` (sharded fan-out) for consistency.
- A short empirical write-up of the measured CC ceiling → folded into the eventual default.

## Relation to epic
Epic #264 (cogni-knowledge). Bake-in finding from #311 — performance/ergonomics, not correctness. Not a Phase-6 cutover blocker; schedule independently.

PR #329, opened 2026-05-27
cogni-knowledge 0.1.9: serialize citation-manifest.json off the LLM (#325)

### LINKED ISSUE #325: [Bug] cogni-knowledge: wiki-composer emits invalid JSON in citation-manifest.json when draft_sentence contains a double-quote (breaks knowledge-verify)
## Severity
**High — correctness blocker for non-trivial prose.** Surfaced live in the #311 German bake-in. `knowledge-verify` could not even shard: `citation-manifest.json` failed `json.loads` with `Expecting ',' delimiter`. The whole verify→revise→finalize tail is dead until the manifest is hand-repaired.

## Root cause
`wiki-composer` writes `citation-manifest.json` with **`draft_sentence` values that contain unescaped `"` characters** — i.e. the manifest is being assembled as hand-templated/string-built JSON rather than via a real serializer (`json.dumps`). Any sentence whose prose contains a straight ASCII double-quote breaks the file.

In this run it was a German typographic quote pair that closed with a straight `"`:

```
"draft_sentence": "… bei Systemen, die ein „Profiling natürlicher Personen" durchführen: …<sup>[3](…)</sup>",
                                                              ^ unescaped " → invalid JSON
```

4 of 109 citation entries were corrupted this way. The composer's own Step-5 self-check returned `ok: true` ("citation manifest confirmed … verified"), so **the agent's internal validation did not actually `json.loads` the file it wrote** — a second gap.

## Why it matters
- German (and FR/IT/ES/PL) prose routinely contains quotation marks; English prose with any quoted term does too. This is not an edge case — it will hit most real drafts.
- The composer agent already correctly uses `json.dumps` / `ensure_ascii=False` for *source* pages (the 67 `source-ingester` pages all parsed fine). The composer's *manifest* write is the regression.

## Fix
1. **`wiki-composer`: serialize `citation-manifest.json` via `json.dumps` (the whole object), never hand-built string JSON.** Each `draft_sentence` must be a properly escaped JSON string. (The source-ingester pattern is the reference — it does this correctly.)
2. **Make the composer Step-5 self-check actually parse what it wrote** — `json.loads(manifest)` + assert `len(citations)` + assert each `draft_sentence` is a substring of the draft. A write that doesn't round-trip must return `ok:false, error: write_failed`, not `ok:true`.
3. Add a contract test: a `draft_sentence` containing `"` (straight quote) and a German `„…"` pair must round-trip through the manifest writer.
4. Consider a defensive guard in `knowledge-compose` Step 5 (orchestrator) — it already `json.loads` the manifest, but the dispatched-agent path bypassed it in this run; ensure the orchestrator's own validation is the authoritative gate.

## Workaround used in the bake-in
Deterministic repair: re-escape every string value in the pretty-printed manifest via `json.dumps` (0 backslashes present → safe). All 109 `draft_sentence`s then matched the draft body verbatim and verify proceeded. Broken original saved as `.citation-manifest.broken.json` for evidence.

## Relation to epic
Epic #264. Bake-in finding from #311. This one is **correctness**, not ergonomics — arguably should gate the #311 'core exit criteria' verdict (the German run does not pass end-to-end on stock v0.1.8 without this fix). Related bake-in findings: #323 (batch boundaries), #324 (index truncation).

sdh07:
## Summary

Fixes #325 — the #311 German bake-in's correctness blocker. `wiki-composer` has no `Bash` and **hand-typed** `citation-manifest.json` into the `Write` tool, so an unescaped straight `"` inside a `draft_sentence` (a German `„Profiling natürlicher Personen"` pair, 4 of 109 entries) broke `json.loads` and `knowledge-verify` couldn't even shard — the whole verify→revise→finalize tail died.

The fix takes JSON construction off the LLM entirely:

- **`wiki-composer`** now writes a raw-text **citation-records** file (`.metadata/citation-records-vN.txt`, one labeled `- id:` block per citation, sentence verbatim/unescaped) through the byte-safe `Write` channel. Its tools are **unchanged** (`Read`/`Write`/`Glob`/`Grep` — no `Bash`); it never touches JSON.
- **NEW `scripts/citation-store.py build`** — run by `knowledge-compose` Step 4.5 (the orchestrator already has `Bash`). It parses the records via the new `_knowledge_lib.parse_citation_records()`, `json.dumps` the manifest (`ensure_ascii=False` — escaping owned by the serializer), asserts every `draft_sentence` is a verbatim NFC substring of the draft, and round-trips the file it wrote (`json.loads` + count). Any failure → `success:false` so the orchestrator stops instead of shipping a broken manifest.
- **`knowledge-compose` Step 5** re-asserts every `draft_sentence` is in the draft (NFC) — the authoritative gate (issue #4).

### How #325's requirements map
- **#1** (serialize via `json.dumps`, never hand-built): `citation-store.py build` is the only writer of the manifest.
- **#2** (self-check actually parses what it wrote): JSON round-trip + count + substring check; non-round-trip / stale-sentence → `error: write_failed`.
- **#3** (contract test for a `"` / German-pair `draft_sentence`): `tests/test_citation_store.sh`.
- **#4** (orchestrator's validation is the authoritative gate): the orchestrator both **builds** (Step 4.5) and **re-validates** (Step 5).

Out of scope: bake-in findings #323 / #324.

## Files
- NEW `cogni-knowledge/scripts/citation-store.py`, `cogni-knowledge/tests/test_citation_store.sh`
- `_knowledge_lib.py` (`parse_citation_records`), `agents/wiki-composer.md`, `skills/knowledge-compose/SKILL.md`, `tests/test_compose_contract.sh`
- version 0.1.8 → 0.1.9 (`plugin.json` + `marketplace.json`); docs (`CLAUDE.md`, `README.md`, `references/inverted-pipeline.md`)

## Test plan
- [x] `for t in tests/test_*.sh; do bash "$t"; done` — all 22 green (incl. new round-trip + negative + edge cases)
- [x] Direct repro: build a records file with the German `„…"` pair → `citation-store.py build` → output `json.loads` cleanly AND `verify-store.py shard` accepts it (the exact path #325 broke)
- [x] Regression: a tampered `draft_sentence` (not in draft) is caught by both the build self-check (`write_failed`) and the Step-5 guard; empty citations still WARN (not fail)
- [ ] Live German end-to-end re-run (#311 core-exit-criteria) on the unblocked pipeline

Closes #325

---
_Generated by [Claude Code](https://claude.ai/code/session_01JF2AJUzMnPuCsQzgGkrBMG)_


sdh07:
## Review — compose-path fix is solid; revisor carries the same bug downstream

**Approve with follow-up.** The #325 compose-path fix is correct, robust, and well-tested — safe to merge.

### Strengths
- Root cause fixed: JSON construction moved off the LLM. Composer writes a raw-text `citation-records-vN.txt` via the byte-safe `Write` channel (tools unchanged, no `Bash`); `citation-store.py build` (orchestrator Step 4.5, where `Bash` already exists) `json.dumps` the manifest with `ensure_ascii=False` and self-checks (substring-in-draft + round-trip `json.loads` + count).
- `test_citation_store.sh` covers the exact bake-in cases: ASCII `"`, the German `„Profiling natürlicher Personen"` pair, a backslash, `claim: null` → JSON `null`, empty/missing sentence, duplicate id, missing files, long-key aliases, U+2028 (split on `\n`, not `splitlines()`), and **downstream acceptance by `verify-store.py shard`** — the precise path #325 broke.
- Conventions honored: stdlib `{success,data,error}` envelope; version bumped in both plugin.json and marketplace.json; docs updated across CLAUDE.md / README / inverted-pipeline.md / absorption-roadmap.

### 🔴 Material gap — the `revisor` has the identical latent bug
`agents/revisor.md` has `tools: ["Read","Write","Edit","Glob","Grep"]` (no `Bash`) and Phase 3 Step 3 still *"Compose … `Write` to `citation-manifest.json`"*. On a revise round it **rephrases** sentences (German → `„…"` pairs) and hand-types the updated `draft_sentence` into JSON. So round 0 (compose) is safe, but round ≥1 (revise) can re-break the manifest → `verify-store.py shard` fails `json.loads` again → the verify→revise→finalize loop dies with the same symptom. Since #325 exists to unblock the German tail, this should be tracked before calling #311 green — the same `records → citation-store.py build` mechanism (or an orchestrator-side rebuild after each revise round) applies.

### 🟡 Minor — composer records read-back dropped; count not reconciled
The old "Read it back to confirm persistence" was removed for the records file (the `write_failed` envelope text still references it). A phantom-empty records write → `build` parses `[]` → a *valid empty* manifest → `success` count 0 → Step 5 `WARN` (not fail) → pipeline proceeds silently with zero citations; the composer's claimed `citations` is never compared to `data.citations_count`. Suggest restoring the records read-back and/or warning on `composer_count != citations_count`.

sdh07:
Thanks for the review — both points addressed:

**🔴 Revisor (fixed in this PR, commit `2df22c9`).** The revisor carried the identical bug, confirmed: `agents/revisor.md` (no `Bash`) hand-typed the rewritten `citation-manifest.json` each revise round, so a rephrased German `„…"` sentence would re-break `json.loads`. Applied the same `records → citation-store.py build` mechanism to the revise path:
- The revisor now `Edit`s the draft in place and writes a raw-text `citation-records-v{N+1}.txt` (tools unchanged — still no `Bash`); it never builds JSON.
- `knowledge-verify` runs `citation-store.py build` after the revisor returns to serialize + self-check the manifest (the substring-in-draft check also catches a revisor whose `Edit` didn't land the rephrased sentence verbatim), then derives `DELTA_IDS` from the manifest diff as before.
- Verified end-to-end: a `„hochriskant"` rephrase round-trips through the build at `draft_version 2`. `test_verify_contract.sh` asserts the new contract.

**🟡 Records read-back + count reconciliation (commit `54ecc53`).** Restored a records read-back in the composer's Phase 3 (the `write_failed` reason text referenced one that no longer ran), and `knowledge-compose` now reconciles the composer's claimed count against the authoritative `citation-store.py build` count, surfacing a `⚠ citation count mismatch` on a phantom/truncated records write.

All 22 contract/unit tests green. The only red check is `cla` (CLA signature gate), which needs a human signature — no code change on my side.

---

sdh07:
## Re-review — all findings resolved ✅

Ran the three affected suites against the branch (`test_citation_store.sh`, `test_verify_contract.sh`, `test_compose_contract.sh`) — all green. **Approve — ready to merge.**

### 🔴 → ✅ Revisor latent bug (`2df22c9`, `a076ce6`)
The revisor no longer hand-builds JSON. It now writes `citation-records-v{N+1}.txt` (same raw-text `- id:` format as the composer, tools unchanged), and `knowledge-verify` Step 3.3's `ok:true` branch runs `citation-store.py build` to serialize + self-check the manifest — same substring-in-draft + round-trip guard, surfacing `write_failed`/`duplicate_id` and **stopping** (the prior `verify-vN.json` stays the latest valid trail). `DELTA_IDS` is now diffed against the **rebuilt** manifest, keeping incremental re-verify consistent. A rephrased German `„…"` sentence can no longer re-break `json.loads` at round ≥1.

### 🟡 → ✅ Records read-back + count reconciliation (`54ecc53`)
Composer and revisor both read-back-verify the records file (catches a phantom/truncated write). Step 4.5 reconciles the composer's claimed `citations` vs the authoritative `data.citations_count` with a `⚠ citation count mismatch` line — soft, not hard-fail.

### 🟢 → ✅ id-less blocks (`6300e63`)
`parse_citation_records` no longer silently drops an id-less block — `citation-store.py`'s new `empty_id` guard surfaces it as `write_failed`. Bonus: `duplicate_id` guard, `OSError` handling on `atomic_write`, explicit empty-`draft_sentence` guard.

The German verify→revise→finalize tail is now quote-safe end-to-end (compose round 0 **and** the revise rounds). Merging.

Issue #344, opened 2026-05-28
[Change] cogni-knowledge: make concept/entity pages citable (composer cites them, verifier scores distilled_claims)


sdh07:
**Plugin:** cogni-knowledge v0.1.13 (Preview)
**Marketplace:** insight-wave
**Epic:** #264 · **Adjacent:** #336.

## Summary
Concept/entity pages are read by `wiki-composer` as **framing only** (#336) — they're absent from the citation manifest and not scored by `wiki-verifier`. The composer can use them to shape the arc, but every inline `[N]` must still resolve to a source's `pre_extracted_claims:`. This is the right Phase-1 posture (keeps Phases 5/6/7 byte-stable), but it means cross-source-distilled facts can only be cited by enumerating their backing sources, not by citing the synthesis directly.

## Why it matters
When a fact is asserted by 5 sources and distilled onto one concept page, the composer must either cite all 5 (verbose) or pick one (loses the cross-source signal). Citing the concept page itself would carry the right epistemic weight ("5 sources agree") while still letting the verifier check the assertion.

## Approach
- Extend `wiki-verifier` to accept a citation whose `wiki_slug` resolves to a concept/entity page and score `draft_sentence` against the page's `distilled_claims[].text` (vs `pre_extracted_claims[].text` on sources).
- Extend `citation-store.py` to accept concept/entity slugs in records (validate via the same on-disk existence check).
- Extend `cycle-guard.py` to be aware of concept→source lineage so a synthesis citing a concept derived from its own sources doesn't loop.

Bigger contract extension — defer until #311 produces real evidence that concept-citing is needed.

## Sources
`cogni-knowledge/agents/wiki-verifier.md`, `scripts/citation-store.py`, `scripts/cycle-guard.py`, `agents/wiki-composer.md` (the framing-only restriction).

Issue #341, opened 2026-05-28
[Change] cogni-knowledge: re-narrate concept/entity ## Summary across runs (Phase-1 was first-writer-wins)


sdh07:
**Plugin:** cogni-knowledge v0.1.13 (Preview)
**Marketplace:** insight-wave
**Epic:** #264 · **Adjacent:** #335 (contradiction surfacing — natural co-implementation), #336.

## Summary
Phase-1 distillation (#336) deliberately keeps the `## Summary` machine block first-writer-wins on update: only the `## Claims` / `## Related` / `## Sources` machine blocks grow across runs. Run N+1's new evidence never refreshes the prose narrative; a concept page can list 20 distilled claims while its summary still reflects only the first run's framing.

## Why it matters
Compounding right now is structural (the lists accrete) but not narrative (the prose doesn't integrate new evidence). For a reader of the wiki, the `## Summary` is the entry point — a stale summary undersells what the page now contains. This is also where contradictions naturally surface ("source A asserts X; source B asserts ¬X — see Claims").

## Approach
Add an optional step to `knowledge-distill` (or a dedicated `knowledge-rerationate` pass) that — for each `updated` action — dispatches an agent to rewrite the `## SUMMARY` machine block given the existing summary + the merged `distilled_claims[]`. The other machine blocks stay deterministic. Human `## Notes` still preserved by sentinel splice. Fail-soft: an agent failure leaves the existing summary in place. Pair with **#335** so contradictions can surface in the same pass.

## Sources
`cogni-knowledge/scripts/concept-store.py::_merge_one` (summary reuse), `references/inverted-pipeline.md` Phase 4.5 ("Phase-1 does not re-narrate"), #335.

Issue #338, opened 2026-05-27
[Change] cogni-knowledge: knowledge-finalize doesn't refresh open_questions.md — research data-gaps don't flow back to the wiki backlog


sdh07:
**Plugin:** cogni-knowledge v0.1.11 (Preview)
**Marketplace:** insight-wave
**Epic:** #264

## Summary
cogni-wiki maintains `wiki/open_questions.md` — the persistent "data gaps" backlog from the Karpathy pattern — and rebuilds it via `rebuild_open_questions.py` at the end of **every** `wiki-lint` dispatch (`cogni-wiki/CLAUDE.md` §"open questions", v0.0.30). cogni-knowledge's `knowledge-finalize` Step 10.5 runs cogni-wiki helpers directly at script level but calls **only** `wiki_index_update.py` + `config_bump.py` + `rebuild_context_brief.py` (confirmed in `skills/knowledge-finalize/SKILL.md` Steps 7–10.5). It **never** calls `rebuild_open_questions.py`, so the deposited base's gap backlog never reflects what a knowledge run revealed.

## Why it matters (and why it's a cheap, high-signal win)
cogni-knowledge has the **richest possible data-gap signal in the whole monorepo**, and currently throws it away at deposit:
- `plan.json` sub-questions (the explicit research decomposition), and
- `wiki-coverage.py`'s per-sub-question `uncovered` / `partial` verdicts (#309 P1.3 read-before-web, already computed and written to `<project>/.metadata/wiki-coverage.json`).

These are exactly the "what does this base still not answer" gaps `open_questions.md` exists to track — but they stay trapped in the project's `.metadata/` and never reach the wiki backlog.

## Options
- **(a)** Add `rebuild_open_questions.py --wiki-root "$WIKI_ROOT"` to `knowledge-finalize` Step 10.5 (alongside the existing `lint_wiki.py --fix=all` / `health.py` gate), so finalize maintains the backlog the same way a `wiki-lint` run would.
- **(b) Go further:** seed `open_questions.md` from the run's `uncovered` / `partial` sub-questions (the `wiki-coverage.json` verdicts) via the script's existing `--findings -` stdin contract, so the backlog captures research-time gaps a deterministic lint pass can't see.

Lowest-cost item in the audit — finalize already resolves the `wiki-lint` script dir for the Step 10.5 gate, so this is one extra script call on a path that's already wired.

## Sources
[...]

PR #357, opened 2026-05-29
feat(cogni-knowledge,cogni-wiki): stream research-time gaps into open_questions.md (#354)

### LINKED ISSUE #354: [Change] cogni-knowledge + cogni-wiki: stream research-time gaps from wiki-coverage.json into open_questions.md (Option b deferred from #338)
**Epic**: #264 · **Depends on**: #338 (Phase 1 — Option a; landed v0.1.19).

#338 Option (a) wired `rebuild_open_questions.py` into `knowledge-finalize` Step 10.5 sub-step 5 so `wiki/open_questions.md` tracks finalize-time **deterministic lint** state. It stopped short of Option (b) — seeding the backlog from **research-time gaps** (`<project>/.metadata/wiki-coverage.json::sub_questions[].coverage_verdict ∈ {uncovered, partial}`). Option (b) is structurally blocked on a cogni-wiki side change.

## The block

`rebuild_open_questions.py::_flatten` drops any finding whose `class` is not in `TRACKED_CLASSES`. The seven existing classes (`no_sources`, `synthesis_no_wiki_source`, `orphan_page`, `stale_page`, `stale_draft`, `claim_drift`, `reverse_link_missing`) are all derived from `lint_wiki.py` findings on **existing** pages. A research-time gap is a finding about a **missing** page — semantically a new class (proposed: `research_uncovered`, `research_partial`).

## Two-plugin change

1. **cogni-wiki** (`rebuild_open_questions.py`):
   - Add `research_uncovered` + `research_partial` to `SECTIONS` with new section headers (e.g. `Research-time gaps — uncovered`, `Research-time gaps — partial`).
   - Extend `TRACKED_CLASSES` and the `HEADER_TO_CLASS` reverse-map.
   - Add `finalize` to `CLOSING_OPS` (currently `("update", "ingest", "re-ingest", "synthesis")`) so a later finalize that covers a sub-question credits the close. **Note:** #338 left close-attribution for the deterministic path falling back to a bare `closed <date>` precisely because `finalize ∉ CLOSING_OPS` today; this is the fix.
   - Bump cogni-wiki patch version; CHANGELOG entry.
2. **cogni-knowledge** (`knowledge-finalize` Step 10.5 sub-step 5):
   - Read `<project>/.metadata/wiki-coverage.json` if present.
   - Map each `coverage_verdict ∈ {uncovered, partial}` sub-question to a `{class, page, message}` entry.
   - Merge with lint output and stream via `--findings -` to bypass the internal `lint_wiki.py` subprocess.

## Open contract questions (need decisions before either plugin changes)

- **`page` identity**: a research-time finding is about a *missing* page, so it has no on-disk slug. `parse_existing` expects a backtick-quoted page slug. The synthesis slug is wrong (multiple sub-questions per synthesis); `sq:<sq_id>` is novel and `parse_existing`'s regex would need to accept it.
- **File layout**: one merged `open_questions.md`, or a separate per-research `open_questions_research.md`?
- **Close-attribution**: `attribute_close` walks `wiki/log.md` for `op ∈ CLOSING_OPS` whose `rest` contains the page slug — but `sq:sq-04` does not appear in any log line by default. How does a *future* finalize covering a previously-uncovered sub-question get credited?

Not just a stream-in — those three contract questions need explicit answers first.

https://claude.ai/code/session_01LY9X91nfA6j3c7eEhLWjMY

sdh07:
## Summary

Closes #354 — Option (b) of the #338 open-questions backlog work.

`knowledge-finalize` Step 10.5 sub-step 5 now streams research-time gaps from
`<project>/.metadata/wiki-coverage.json` into the cross-session
`wiki/open_questions.md` backlog. Sub-questions with `coverage_verdict ∈
{uncovered, partial}` appear as `- [ ] \`sq:<sq_id>\` — …` rows under two new
sections (`Research-time gaps — uncovered`, `Research-time gaps — partial`). A
later finalize whose synthesis covers a previously-uncovered sub-question
credit-closes the row as `closed YYYY-MM-DD by finalize` — both via the new
`sqs=<sq-01,sq-04>` suffix on the `wiki/log.md` finalize line and via `finalize`
joining `CLOSING_OPS` in `rebuild_open_questions.py`.

The backlog now compounds the research pipeline's pending sub-questions, not just
lint findings about existing pages — defending `differentiation-thesis.md`'s "the
wiki compounds" posture instead of letting the signal die in `.metadata/`.

Single PR; two plugins; additive (zero regression on the seven existing tracked
classes).

## Open contract questions — resolved

1. **`page` identity for a research-time gap?** → `sq:<sq_id>` (synthetic). The
   existing backtick-token regex already matches it (`:` is not excluded), so the
   wire format is unchanged; `_flatten` accepts an `id` field (falling back to
   `page`).
2. **One file or split?** → One `wiki/open_questions.md`. Two sections at the tail.
3. **How does a future finalize get credited?** → `finalize` joins `CLOSING_OPS`;
   the finalize log line carries `sqs=…`; `attribute_close()`'s existing substring
   scan finds it (stripping the `sq:` prefix from the needle).

### cogni-wiki 0.0.48 → 0.0.49
- `wiki-lint/scripts/rebuild_open_questions.py`: two deposit-only tracked classes
  (`research_uncovered`, `research_partial`) rendered as two tail sections;
  `finalize` added to `CLOSING_OPS`; `attribute_close` strips the `sq:` prefix;
  `_flatten` accepts an `id` field; `reconcile` returns per-class
  `opened_by_class`/`closed_by_class`; `--findings -` also unwraps a
  `{success, data}` envelope. Seven existing classes, regexes, and 90-day trim
  byte-unchanged.
- `wiki-lint/scripts/lint_wiki.py`: one cross-ref comment (these are NOT lint
  classes).
- `wiki-lint/SKILL.md`: documents the two new sections + deposit-only nature.

### cogni-knowledge 0.1.20 → 0.1.21
- `skills/knowledge-finalize/SKILL.md`: Step 10 appends a conditional `sqs=…`
  suffix to the `wiki/log.md` line; Step 10.5 sub-step 5 pipes a merged
  lint+research-gap payload through `rebuild_open_questions.py --findings -`;
  Step 11 surfaces a `(lint=L, research=R)` split; new `--no-research-gaps` flag.
- `scripts/_knowledge_lib.py`: `load_wiki_coverage_findings()` +
  `gap_sq_ids_from_coverage()` (sharing `_iter_coverage_gaps`), fail-soft on
  missing/malformed manifests, regex-unsafe sq_ids dropped.
- `scripts/build_open_questions_payload.py` (new): merges `lint_wiki.py` output
  with the gap findings into one `{success, data, error}` envelope; a lint crash
  degrades fail-soft rather than blocking.

## Test plan

- [x] `bash cogni-wiki/tests/test_open_questions_research_gaps.sh` (T1 baseline
      regression, T2 gaps open, T3 credit-close, T4 reopen, T5 trim, T6 envelope)
- [x] `bash cogni-wiki/tests/test_open_questions.sh` (existing #338 regression)
- [x] `bash cogni-knowledge/tests/test_load_wiki_coverage_findings.sh`
- [x] `bash cogni-knowledge/tests/test_build_open_questions_payload.sh`
- [x] `bash cogni-knowledge/tests/test_finalize_contract.sh` (extended with #354 asserts)
- [x] Full suite: **49/49** passing across both plugins.
- [x] End-to-end cross-plugin smoke: gaps open in dispatch 1
      (`opened_by_class={research_uncovered:1, research_partial:1}`), then
      credit-close `by finalize` in dispatch 2 after re-curation.

## Manifests

- `cogni-wiki/.claude-plugin/plugin.json` → `0.0.49`; `cogni-knowledge` → `0.1.21`;
  both mirrored in `.claude-plugin/marketplace.json`; CHANGELOGs + CLAUDE.md +
  reference notes updated.

## Out of scope (explicit deferrals)

- Re-narrating concept/entity `## Summary` blocks on a finalize that closes a gap (#341).
- Citing concept/entity pages directly in syntheses (#344).
- Concurrency stress tests beyond the existing lock contract.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FB44kdvs3ZXYDSomfAJEeD)_


sdh07 (commented):
## Review — PR #357 (closes #354)

Solid two-plugin bundle. The contract design (synthetic `sq:<sq_id>` ids, `finalize` joining `CLOSING_OPS`, the merge-helper envelope) is well-thought-out and the tests cover the happy paths plus the credit-close two-dispatch dance (T2→T3→T4) end-to-end. The CHANGELOG / CLAUDE.md / reference-doc updates are thorough. Findings below ordered by severity.

### 🔴 Must fix before merge

**1. `KNOWLEDGE_SCRIPTS` is not exported into Step 10's heredoc → `sqs=` suffix silently never written**

`cogni-knowledge/skills/knowledge-finalize/SKILL.md` Step 10 (lines ~598–603):

```bash
GAP_SQS=$(PROJECT_PATH="$PROJECT_PATH" python3 -c '
import os, sys
sys.path.insert(0, os.environ.get("KNOWLEDGE_SCRIPTS", ""))
from _knowledge_lib import gap_sq_ids_from_coverage
...
' 2>/dev/null || true)
```

Every other subprocess in this SKILL explicitly prefixes `KNOWLEDGE_SCRIPTS="${CLAUDE_PLUGIN_ROOT}/scripts"` per-call (lines 129, 146, 178, 197, 248). Step 10 breaks that convention — it only inlines `PROJECT_PATH` and relies on `KNOWLEDGE_SCRIPTS` being inherited from the parent shell, which is not guaranteed by the SKILL's environment contract.

When `KNOWLEDGE_SCRIPTS` is unset, `os.environ.get(..., "")` returns `""`, `sys.path.insert(0, "")` makes only the CWD importable, the `_knowledge_lib` import fails, stderr is swallowed by `2>/dev/null`, and `|| true` returns `GAP_SQS=""`. The `if [ -n "$GAP_SQS" ]` branch then falls through to the **no-suffix** log line on every run — silently disabling the entire credit-close mechanism the PR ships.

The bug is invisible in tests (test_finalize_contract.sh only greps for the string `sqs=%s` in the SKILL.md), and `test_open_questions_research_gaps.sh` writes its own `sqs=…` line manually — neither exercises the Step 10 bash heredoc.

**Fix**: inline the env var like the rest of the SKILL:
```bash
GAP_SQS=$(KNOWLEDGE_SCRIPTS="${CLAUDE_PLUGIN_ROOT}/scripts" PROJECT_PATH="$PROJECT_PATH" python3 -c '
import os, sys
sys.path.insert(0, os.environ["KNOWLEDGE_SCRIPTS"])
...
```
(Also drop the `.get(..., "")` fallback — failing loud on a misconfigured environment is preferable to silently disabling the feature.)

---

**2. `attribute_close` substring scan can false-positive across sq_ids**

`cogni-wiki/skills/wiki-lint/scripts/rebuild_open_questions.py:282`:

```python
needle = page_slug[3:] if page_slug.startswith("sq:") else page_slug
...
if needle in m.group("rest"):
    return m.group("op")
```

`needle = "sq-04"` matches as a substring of `"sq-040"`, `"sq-041"`, `"sq-04-foo"`, etc. Concretely, with the bare CSV form `sqs=sq-040,sq-01` in an earlier finalize line, a later close of `sq:sq-04` would be falsely credited to that earlier finalize. Same problem for any `sqs=sq-4…` containing `sq-4`.

The risk is low today (`knowledge-plan` emits `sq-01..sq-NN` and projects rarely cross ~10 sub-questions) but the surface is permanent and the failure mode is silent — wrong "by finalize" attribution.

**Fix**: parse the `sqs=` value as a CSV list and check exact membership, e.g.

```python
m_sqs = re.search(r'\bsqs=([^\s|]+)', m.group("rest"))
if m_sqs and needle in m_sqs.group(1).split(","):
    return m.group("op")
```

A simple regex `\b<needle>\b` on `rest` would also do, but the explicit `sqs=` parse is more defensible (it also rejects accidental sq_id occurrences inside `slug=…` or `project=…`).

---

**3. Build-helper hard-crash path silently mass-closes all open items**

`cogni-knowledge/scripts/build_open_questions_payload.py:118-122` — the outer exception handler emits `{"success": false, "data": {}, "error": "..."}` and `sys.exit(1)`. The SKILL captures this into `OPEN_Q_PAYLOAD` without checking the build-helper's exit code (only the rebuild's is captured into `OPEN_Q_EXIT`).

Downstream in `rebuild_open_questions.py:198-204`:

```python
envelope_data = payload.get("data") if isinstance(payload, dict) else None
if isinstance(envelope_data, dict) and any(
    k in envelope_data for k in ("errors", "warnings", "info")
):
    payload = envelope_data
```

With `data = {}` (empty dict, truthy), the `any(...)` check is False → payload is NOT unwrapped → `_flatten(payload)` reads `payload.get("errors") or []` = `[]` → `reconcile()` sees `incoming = []` → **every previously-open item is closed** as "no longer in incoming." The rebuild reports `success: true` with `closed = <previous open count>`, so the Step 11 summary surfaces a ✓ tick rather than the documented ⚠ failure.

This contradicts the SKILL's own fail-soft contract (sub-step 5 is supposed to surface the failure loudly, not silently close the backlog).

**Fix**: pick one of:

- (a) In `rebuild_open_questions.py::load_findings`, when the payload looks enveloped (has `success` and `data` keys), treat `success: false` as a hard failure — raise so the script exits non-zero and `OPEN_Q_EXIT` fires the documented ⚠ surface.
- (b) In the SKILL bash, capture the build-helper exit code: `OPEN_Q_PAYLOAD=$(... build_open_questions_payload.py ...) || OPEN_Q_PAYLOAD_EXIT=$?` and short-circuit to the ⚠ surface when non-zero.

(a) is preferable because it also defends against any future caller of `--findings -` that pipes a `{success: false}` envelope in.

### 🟡 Should address

**4. Missing test coverage for the failure-surface contract**

Neither `test_build_open_questions_payload.sh` nor `test_open_questions_research_gaps.sh` covers what happens when the build helper hits its hard-crash path — directly tied to finding (3). Suggest a T7 in the wiki test that pipes `{"success": false, "data": {}, "error": "boom"}` and asserts the rebuild surfaces a failure (not a silent mass-close).

**5. `test_build_open_questions_payload.sh` case (c) — `RC=$?` is unreachable**

`cogni-knowledge/tests/test_build_open_questions_payload.sh:65-67`:

```bash
OUT=$(python3 "$HELPER" ... --wiki-lint "$LINT_FAIL")
RC=$?
assert_env c "$OUT" "..." "..."
[ "$RC" = "0" ] || { red "FAIL: c exit code should be 0, got $RC"; errors=$((errors + 1)); }
```

With `set -eu` at the top of the file, a non-zero exit from the command substitution aborts the script before `RC=$?` runs. So the assertion `[ "$RC" = "0" ]` is dead code — it can never fail. Either drop `set -e` around this block, or capture with `OUT=$(... ) || RC=$?`.

**6. SKILL Step 10.5 sub-step 5 — build-helper stderr leakage into `OPEN_Q_PAYLOAD`**

Per the comment at line 696, the intent is "Capture stdout ONLY — no 2>&1." That holds for `rebuild_open_questions.py` itself, but `build_open_questions_payload.py::_run_lint` invokes lint_wiki with `capture_output=True` — meaning lint stderr stays inside the Python process and is dropped on success but lost on failure (only summarized in `meta.degraded`). Not a bug, just worth a one-line note that the helper deliberately swallows lint stderr (the operator-visible trace lives on cogni-wiki's terminal during its own dispatches, not here).

### 🟢 Nice-to-have / minor

- **`_sanitize_gap_message`** (`_knowledge_lib.py:166`) strips only backticks and `~~`. A `query` containing `]` or `(` could break markdown rendering of the `- [ ] \`sq:…\` — msg` line if it ever lands at a column where it forms an unintended link. Low-probability given current `plan.json` content but easy to harden: `re.sub(r'[][\\\\`~]', '', msg)`.
- **`_plan_message_index` truncation off-by-one**: `query[:139] + "…"` gives 140 chars including the ellipsis. If the intent was "140 chars including ellipsis," it's correct; if "140 chars of query + ellipsis," it's one short. Worth a comment either way.
- **`build_open_questions_payload.py` timeout=120s** on `subprocess.run`. Reasonable for typical wikis, but worth noting that on >1000-page wikis lint_wiki can exceed this. The fail-soft path catches it (timeout raises `SubprocessError` → degraded), so this is not blocking — just a future tuning knob.
- **Long single-line CLAUDE.md updates** (cogni-knowledge:86, cogni-wiki:796): the existing table-cell paragraphs are already long, but the new clauses push some of them well past 1k chars. No semantic issue, just hard to diff. Could be split into two sentences across lines.

### Summary

The contract is sound and the test coverage on the wiki side (T1–T6) is good. The three 🔴 findings are all in the *interaction surface* between the new merge helper and the existing SKILL/script — none invalidate the design, but (1) silently breaks the credit-close feature in production, (2) is a permanent latent misattribution risk, and (3) is a silent data-quality regression on a failure path the docs claim is fail-soft. Recommend resolving 1–3 before merge; 4–6 are nice follow-ups.

Verdict: **needs_revision**.

sdh07:
Thanks for the thorough review — all three 🔴 findings were real bugs in the merge-helper ↔ SKILL/script interaction surface. Fixed in c527c4d.

**🔴 1 — `KNOWLEDGE_SCRIPTS` not exported into Step 10's heredoc.** Confirmed: the suffix was silently never written when the var wasn't inherited. Inlined `KNOWLEDGE_SCRIPTS="${CLAUDE_PLUGIN_ROOT}/scripts"` like every other subprocess in the SKILL, and dropped the `.get(..., "")` fallback so a misconfigured env fails loud rather than silently disabling credit-close.

**🔴 2 — `attribute_close` substring false-positive (`sq-04` ⊂ `sq-040`).** Took your preferred fix (a): research-gap ids now parse the `sqs=` CSV and check **exact** membership (`re.search(r"\bsqs=([^\s|]+)", rest)` → `split(",")`). Lint page slugs keep the long-standing substring scan untouched, so the seven existing classes are byte-for-byte unchanged. New **T8** asserts `sqs=sq-040` does not credit-close `sq:sq-04`.

**🔴 3 — `{success:false, data:{}}` crash payload → silent mass-close.** Confirmed the exact trace you laid out (empty truthy dict → not unwrapped → `_flatten` sees `[]` → reconcile closes everything under a ✓). Took fix (a): `load_findings` now treats any enveloped `success:false` as a hard `RuntimeError` → exit 1 → the SKILL's documented ⚠ surface fires. This also defends any future `--findings -` producer. New **T7** seeds an open item, pipes the failure envelope, and asserts non-zero exit + backlog untouched.

**🟡 4** — covered by T7 (failure-surface) + T8 (substring guard).
**🟡 5** — fixed: `RC=$?` was indeed dead under `set -e`; switched to `OUT=$(...) || RC=$?` with a comment.
**🟡 6** — added a docstring note that `_run_lint`'s `capture_output=True` deliberately swallows lint stderr.
**🟢** — added the 140-char truncation comment and the timeout note context.

One 🟢 I **skipped** (happy to revisit): stripping `]`/`(` in `_sanitize_gap_message`. `OPEN_LINE_RE` captures the message as `.*`, so those chars don't break line parsing, and a stray `[text](url)` link needs both `]` and `(` adjacent — which the `- [ ] \`sq:…\` — ` prefix shape never produces. Stripping parens/brackets would degrade legitimate query text (e.g. "Article 6 (Annex III)") for a render risk I couldn't construct. The two delimiter chars that *can* break the regexes (backtick, `~~`) are still stripped. If you'd prefer the belt-and-suspenders strip anyway, it's a one-liner.

Full suite green (49/49) after the fixes.

---

sdh07 (commented):
## Re-review — PR #357 (commit `c527c4d`)

Re-ran the audit against the pushed fixes. All three 🔴 findings are resolved and the 🟡 follow-ups are addressed. Tests pass locally on the PR branch.

### 🔴 Must-fix — all resolved

**1. `KNOWLEDGE_SCRIPTS` inlined in Step 10's heredoc** ✅

`cogni-knowledge/skills/knowledge-finalize/SKILL.md` Step 10 now reads:

```bash
GAP_SQS=$(KNOWLEDGE_SCRIPTS="${CLAUDE_PLUGIN_ROOT}/scripts" PROJECT_PATH="$PROJECT_PATH" python3 -c '
import os, sys
sys.path.insert(0, os.environ["KNOWLEDGE_SCRIPTS"])
from _knowledge_lib import gap_sq_ids_from_coverage
print(",".join(gap_sq_ids_from_coverage(os.environ["PROJECT_PATH"])))
' 2>/dev/null || true)
```

Both fixes applied as recommended: env var inlined like every other subprocess in this SKILL, and the `.get(..., "")` fallback dropped in favour of a direct `os.environ["KNOWLEDGE_SCRIPTS"]` subscript so a misconfigured env now fails loud rather than silently disabling the credit-close mechanism.

**2. `attribute_close` exact-CSV match for sq_ids** ✅

`cogni-wiki/skills/wiki-lint/scripts/rebuild_open_questions.py` (the new `attribute_close`):

```python
is_gap = page_slug.startswith("sq:")
needle = page_slug[3:] if is_gap else page_slug
...
rest = m.group("rest")
if is_gap:
    m_sqs = re.search(r"\bsqs=([^\s|]+)", rest)
    if m_sqs and needle in m_sqs.group(1).split(","):
        return m.group("op")
elif needle in rest:
    return m.group("op")
```

Exact-membership check against the parsed `sqs=` CSV (suggested fix path). Lint slugs keep the long-standing substring scan. The docstring was also extended to call out the load-bearing exact-check rationale, which is the right level of WHY for a non-obvious invariant.

New T8 in `cogni-wiki/tests/test_open_questions_research_gaps.sh` directly exercises this: it writes a `sqs=sq-040,sq-99` log line, drives `sq:sq-04` open then closed, and asserts the close lands with **no** `by finalize` attribution (line ~167 of the test). Test passes.

**3. `{success: false}` envelope now hard-fails the rebuild** ✅

`rebuild_open_questions.py::load_findings`:

```python
if isinstance(payload, dict) and "success" in payload and "data" in payload:
    if not payload.get("success"):
        raise RuntimeError(
            f"--findings -: upstream payload reported failure: "
            f"{payload.get('error', '')}"
        )
    payload = payload.get("data") or {}
```

Took option (a) — the more defensible path that also defends any future `--findings -` producer. The silent mass-close path through `_flatten(payload={})` is no longer reachable; the helper's hard-crash path now propagates to `OPEN_Q_EXIT != 0` and the documented ⚠ surface fires in Step 11.

### 🟡 Should-address — resolved

**4. Failure-surface test** ✅ — T7 in `test_open_questions_research_gaps.sh` pipes `{"success": false, "data": {}, "error": "boom"}`, asserts `RC != 0`, and verifies the previously-open `sq-04` item is **not** silently closed. Exactly the missing coverage from the prior review.

**5. Dead `RC=$?` in test_build_open_questions_payload.sh case (c)** ✅ — Now uses `OUT=$(...) || RC=$?` (line 575 of the test) with an inline comment explaining the `set -e` interaction. The exit-code assertion is now actually live.

**6. Build-helper stderr-swallow note** ✅ — The `_run_lint` docstring in `build_open_questions_payload.py` (lines 313-318) now explicitly documents the deliberate `capture_output=True` stderr-swallow and points to where lint's operator-visible trace lives.

### 🟢 Nice-to-have notes

- The query-truncation comment in `_plan_message_index` now reads `# Clamp to 140 chars *including* the ellipsis (139 of query + "…").` — intent recorded, no off-by-one ambiguity left.
- Additional sanitization (broader char class) and the lint-timeout knob were marked nice-to-have and were not blocking; both can land as follow-ups.

### Local test runs

All four affected test suites pass on the PR branch:

- `cogni-knowledge/tests/test_load_wiki_coverage_findings.sh` — 5/5 PASS
- `cogni-knowledge/tests/test_build_open_questions_payload.sh` — 5/5 PASS (case c now actually asserts RC=0)
- `cogni-wiki/tests/test_open_questions_research_gaps.sh` — 8/8 PASS (T7 + T8 are the new coverage)
- `cogni-knowledge/tests/test_finalize_contract.sh` — PASS

### Verdict

**approved** — the three production-impact findings (silent credit-close disable, latent misattribution, silent mass-close on failure) are all closed with test coverage that would catch a regression. Safe to merge.

PR #45, opened 2026-04-15
cogni-portfolio: add fr/it/nl/es to regions.json

sdh07:
Closes #41.

## Summary
- Add four single-country market entries (fr, it, nl, es) to `cogni-portfolio/skills/portfolio-setup/references/regions.json`, matching the structure of the 8 entries added in PR #40.
- Each entry has name, scope, currency (EUR), locale, timezone, and a 3-body `regulatory_bodies` array covering data protection + telecom/media + sector-specific cyber regulator.
- Bump `cogni-portfolio` patch version from 0.9.6 to 0.9.7 in both `plugin.json` and `marketplace.json`.

## Scope decisions
- In: the four missing European markets already supported by cogni-research and cogni-trends, plus mirrored version bump. This closes the full cross-plugin drift identified in the PR #40 review.
- Out: no script changes — `validate-entities.sh` derives `VALID_REGIONS` from `regions.json` at runtime since PR #40, so the whitelist picks up the new entries automatically.
- Out: no documentation sweep of downstream plugins (cogni-research, cogni-trends) — they already support these markets; this PR only catches cogni-portfolio up.
- Regulator triads chosen per issue guidance: FR = CNIL, ARCEP, ANSSI; IT = Garante Privacy, AGCOM, ACN; NL = Autoriteit Persoonsgegevens, ACM, NCSC-NL; ES = AEPD, CNMC, INCIBE.

## Test plan
- [x] `python3 -m json.tool cogni-portfolio/skills/portfolio-setup/references/regions.json` parses without error.
- [x] `jq '.plugins[] | select(.name=="cogni-portfolio") | .version' .claude-plugin/marketplace.json` returns `"0.9.7"` and matches `cogni-portfolio/.claude-plugin/plugin.json`.
- [x] Spot-check that each new entry has exactly the six required fields (name, scope, currency, locale, timezone, regulatory_bodies) and a 3-element `regulatory_bodies` array.
- [ ] A throwaway market entity file with `"region": "fr"` (and it/nl/es) passes `cogni-portfolio/scripts/validate-entities.sh` without `Invalid region` errors. (VALID_REGIONS is derived from regions.json at runtime — confirmed via direct inspection; end-to-end market-entity run left for reviewer.)

Issue #35, opened 2026-04-15
cogni-research: decouple report length from research depth (target_words)


sdh07:
## Problem

In cogni-research, `report_type` bundles three independent concerns: research depth (pipeline shape + sub-question count), review iteration cap, and target word count. Choosing `deep` auto-commits the writer to an 8,000-word floor, but professional deep-research norms place the sweet spot for focused deep dives at ~2,000–5,000 words. Users who want a deep tree (10–20 leaf nodes, recursive deep-researchers) and a crisp 5K deliverable have no way to express that today without setting `allow_short: true`, which is a crude opt-out of all expansion gates rather than a lower floor.

## Proposal

Introduce `target_words` as an independent field in `project-config.json`, defaulted per depth mode. Reduce the default for `deep` from 8000 to 5000. Users who want the old behaviour explicitly set `target_words: 8000`. Length becomes a knob, depth stays a structural choice.

### New defaults

| Depth (`report_type`) | Old default words | New default words |
|---|---|---|
| basic     | 3000  | 3000 (unchanged) |
| detailed  | 5000  | 5000 (unchanged) |
| **deep**  | **8000** | **5000** |
| outline   | 1000  | 1000 (unchanged) |
| resource  | 1500  | 1500 (unchanged) |

### New UX in research-setup

Configuration Menu adds a Length row alongside Depth:

```
Depth:  basic | detailed | deep | outline | resource
Length: brief (1.5K) | standard (3K) | deep-dive (5K) | comprehensive (8K) | whitepaper (12K)
```

Length is optional. If omitted, it derives from depth via the new default table. Length accepts a named preset OR an explicit integer (`target_words: 6500`). Detection keywords: "5K", "~8000 words", "short", "long-form", "whitepaper".

### Interaction matrix

| Request | report_type | target_words | Outcome |
|---|---|---|---|
| "deep research on X"              | deep     | 5000  | Deep tree, crisp 5K deliverable |
| "deep research, whitepaper"       | deep     | 12000 | Deep tree, 12K reference doc |
| "detailed but short"              | detailed | 3000  | Multi-section but brief |
| "basic, make it thorough"         | basic    | 5000  | Single-pass, 5K prose |

## Affected components

- `skills/research-setup/SKILL.md` — Configuration Menu, Step 1 extraction
- `scripts/initialize-project.sh` — `--target-words` flag, persist to project-config
- `skills/research-report/references/report-types.md` — word counts become defaults, not floors-by-type
[...]

PR #420, opened 2026-06-01
cogni-knowledge: ingest-wave integrity sweep + pre-write guard

### LINKED ISSUE #413: [Bug] cogni-knowledge: concurrent source-ingester wave cross-contaminates page content/id (one page gets another's body + frontmatter)
## Context

Found during the first real cogni-knowledge dogfooding run (EU AI Act for SMEs, DE/DACH, v0.1.47 — the #264 Phase-6 human-gate run). `knowledge-ingest` ran 57 sources in 3 fan-out waves (batch-size 25). The Step 10.5 conformance gate (`health.py`) flagged **1 `id_mismatch` error**.

## Symptom

The page `wiki/sources/overview-of-the-code-of-practice-eu-artificial-intelligence-act.md` (source #45, batch 2) contains the **entire content of a different source** (#35, `eu-ai-act-2026-…-zwischenstand-tuev`, also batch 2) — identical `id:`, `title:`, AND `sources:` URL:

```
# file: overview-of-the-code-of-practice-eu-artificial-intelligence-act.md
id: eu-ai-act-2026-was-sich-seit-inkrafttreten-veraendert-hat-ein-zwischenstand-tuev
title: "EU AI Act 2026: Was sich seit Inkrafttreten verändert hat – ein Zwischenstand | TÜV"
sources: ["https://consulting.tuv.com/aktuelles/ki-im-fokus/eu-ai-act-2026-zwischenstand"]
```

The real "Code of Practice overview" content (source #45) was **lost** — overwritten by / cross-written with source #35's content. Both ingesters were dispatched with their own explicit, distinct `SLUG`, `URL`, and `BATCH_OUTPUT_PATH` in the same single-message fan-out wave, yet one page ended up with the other's payload. 1 of 57 pages affected.

## Impact

- A `type: source` page carries the wrong body, claims, and provenance under the right filename → the composer can cite the wrong source under a slug, and verify scores against the contaminated claims (a citation can pass verify while pointing at the wrong source's claim).
- Surfaces as a `health.py` `id_mismatch` **error** that fails the finalize Step 10.5 conformance gate (loud, non-fatal). On an autonomous run this silently ships a corrupted page.
- Only reproduces at fan-out scale (concurrent wave), which is why smaller test runs (≤ a few sources) never caught it.

## Hypothesis

A race in the concurrent `source-ingester` wave — content/frontmatter cross-talk or a mis-resolved output path between two agents running in the same wave. Each ingester reads its own cached body via `fetch-cache.py` and writes via `atomic_write_text` to a unique `wiki/sources/<slug>.md`, so the crossover is upstream of the atomic write (wrong body loaded, or wrong slug→path binding under concurrency).

## Suggested investigation

- Have `source-ingester` assert `frontmatter.id == basename(output_path)` and `frontmatter.sources[0] == URL` (the dispatched URL) before the atomic write — fail fast in-agent.
- Audit whether anything in the ingest path resolves the slug/URL from shared/ambient state rather than strictly from its own dispatch parameters.
- Consider an orchestrator-side post-wave integrity sweep (id == filename, sources[0] == dispatched URL) before the index/backlink step.

## Repro

DE/DACH, 57 sources, batch-size 25, single-message fan-out per batch. 1 cross-contaminated page in batch 2.

Relates to #264 / #388. See also `references/fan-out-concurrency.md` (#323 raised the batch default to 25 — this is a concurrency-correctness counterweight to that throughput change).

sdh07:
Closes #413

## Problem

In the first real `cogni-knowledge` dogfooding run (EU AI Act for SMEs, DE/DACH — the #264 Phase-6 human-gate run), `knowledge-ingest` ran 57 sources in 3 fan-out waves (batch-size 25). One on-disk page — source #45 — ended up carrying the **entire payload of a different source** (#35): its `id:`, `title:`, `sources:` URL, claims, and body. #45's real content was lost.

The conformance gate caught it as a `cogni-wiki health.py` `id_mismatch` error — but only at `knowledge-finalize` Step 10.5, far downstream. On an autonomous run this silently ships a corrupted `type: source` page the composer can cite and the verifier scores against contaminated claims before anything notices.

**Root cause:** `source-ingester` resolves `SLUG`/`URL`/`TITLE_HINT` strictly from its own dispatch parameters — there is no ambient/shared-state path that could cross-resolve them. The contamination is **non-deterministic LLM attention cross-talk** between two `source-ingester` dispatches in the same single-message fan-out wave: the agent handling #45 composed its page from #35's fetched body + frontmatter. This cannot be prompted away, so the fix is deterministic post-hoc verification against the orchestrator's authoritative dispatch record, plus a cheap in-agent fail-fast. Batch-size stays 25 (the #323 throughput calibration is correct; this is its correctness counterweight, not a rollback).

## Approach — two layers + quarantine

**Layer 1 (load-bearing) — orchestrator post-wave sweep.** New `scripts/ingest-integrity.py sweep` reads each dispatched `wiki/sources/<slug>.md` and asserts `id == slug` AND `normalize_url(sources URL) == normalize_url(dispatched URL)`, comparing against the orchestrator's authoritative `(slug, url)` dispatch record — never the agent-returned batch JSON (which can itself be contaminated). `knowledge-ingest` gains:
- Step 3 persists `.metadata/.ingest.dispatch.<NNN>.json` at dispatch time.
- New **Step 3.5** (after the 3.4 merge, before Step 4) runs the sweep over the batch's ok-true subset, quarantines each violation to `.metadata/quarantine/<slug>.md`, drops it from `ingested[]` into `skipped[]` with `reason: integrity_mismatch`, and excludes it from Step 4/4.5. The freed slug + URL stay **re-dispatchable** on a plain re-run.
- Step 6 prints a loud `⚠ Integrity:` line only when violations occurred.

The detector is **read-only** (detect-in-script / act-in-orchestrator, cf. cogni-wiki's `backlink_audit.py`), reusing `_knowledge_lib.normalize_url`/`first_url`/`_FRONTMATTER_RE`/`_unquote_scalar`; the frontmatter read mirrors `wiki-coverage.py::_page_title_tags`.

**Layer 2 (in-agent fail-fast) — `source-ingester` Phase 3.** A pre-write assertion folds into the existing `atomic_write_text` wrapper: parse the composed page's frontmatter and refuse to write (exit 3, emit a Phase 4 `integrity_mismatch` skip) unless `id == SLUG`, the `sources:` URL normalizes to `URL`, and the path stem equals `SLUG`. Stops most cross-writes before they hit disk; Layer 1 is the deterministic backstop the LLM can't defeat.

## Verification

- New `tests/test_ingest_integrity.sh` — positive / `id_mismatch` (both legs) / `url_mismatch` / `page_missing` / URL-normalization / stdin-dispatch — all green.
- Extended `tests/test_ingest_contract.sh` (Step 3.5 + sweep script + quarantine + `integrity_mismatch`; agent pre-write assertion + `sys.exit(3)`).
- Full `cogni-knowledge/tests/*.sh` suite green (no Step-renumbering regressions). No `#NNN` leaked into SKILL/agent prose (`feedback_no_issue_refs_in_skills`).

Bumps cogni-knowledge `0.1.49 → 0.1.50` (plugin.json + marketplace.json).

https://claude.ai/code/session_011YrnwnS9K7bjBG1hfATjoL

---
_Generated by [Claude Code](https://claude.ai/code/session_011YrnwnS9K7bjBG1hfATjoL)_


sdh07:
## Code review — verdict: needs revision (one targeted fix)

Reviewed statically against the full diff. The detector script and the in-agent guard are solid and merge-ready; there is **one correctness gap in the orchestration glue** that should be fixed before merge, plus two non-blocking follow-ups.

### 🟠 Major — Step 3.5 sweep-input filter is keyed on the agent-returned URL (the field contamination corrupts)

`knowledge-ingest` SKILL Step 3.5 step 1 says:

> keep only the entries whose URL the Step 3.4 merge just landed in `ingested[]` this run (the ok-true subset)

`ingested[].url` comes from the agent's **batch JSON**, which is exactly what cross-talk can corrupt. Under the **reported signature** (the #45 page carries #35's `id` *and* `sources` URL — the agent adopted #35's identity wholesale), that agent's batch JSON most likely also emits `url=#35`. Then:

- `#45-url ∉ ingested[]` → the `{slug:#45, url:#45}` dispatch entry is **filtered out of the sweep input** → the contaminated `#45.md` page is **never swept** and escapes detection.

So whether the contaminated page is caught depends on whether the confused agent happened to echo its *own* dispatched URL in its batch JSON — nondeterministic, and adversarial to the exact threat this PR targets.

**Fix (SKILL prose only — the detector script needs no change):** key the ok/skip decision by **dispatch index**, not by URL membership. The orchestrator already persists `.ingest.dispatch.<NNN>.json` (positional) and writes `.ingest.batch.<NNN>.<ii>.json` per source — pair them by `<ii>`: include dispatch entry `i` in the sweep iff batch result `i` was `ok:true`. (Equivalent safe formulation: sweep the **full** dispatch table and suppress `page_missing` only for slugs whose batch result was an index-confirmed legitimate skip.) The index is orchestrator-owned and contamination-proof.

### 🟡 Minor (follow-up, not blocking) — body-only cross-talk slips through

The sweep checks `id` + `sources` URL only. If an agent keeps its own `id`/`sources` frontmatter but writes a sibling's **body + claims**, both legs pass. The deterministic closure already exists conceptually in `source-ingester.md`: assert the page's `content_hash` equals `fetch-cache.py fetch --url <dispatched-url>` → `entry.content_hash`. Worth a follow-up issue; the observed bug is id+url-crossed, so not required here.

### 🟡 Nit — no automated guard for the Step 3.5 orchestration glue

The URL filter / quarantine / manifest-rewrite / re-dispatchability path is LLM prose (matches house style), but combined with the major finding it means the fragile filter has no regression guard. The manual quarantine round-trip from the plan would surface it.

### ✅ Strong

- `ingest-integrity.py` is a clean read-only detector (detect-in-script / act-in-orchestrator, cf. `backlink_audit.py`); standard envelope; exits 0 with violations as data.
- Shared `_knowledge_lib.extract_page_id_and_url` removes script/agent drift; `^id`/`^sources` anchors correctly skip indented `pre_extracted_claims:` entries.
- Verifies against the orchestrator dispatch record, not the agent batch JSON (the right instinct — the major finding is just that the *filter* reintroduces batch-JSON trust).
- Good script test coverage (both-legs `id_mismatch`, `url_mismatch`, `page_missing`, URL-normalization, stdin).
- Version + marketplace mirrored (0.1.49→0.1.50); no issue refs in SKILL/agent prose; reference doc updated; batch-size 25 correctly preserved.

**Summary:** re-key the Step 3.5 ok-true filter from URL-membership to dispatch index, and the load-bearing layer becomes deterministic against the exact contamination it targets.

_Reviewed by Claude Code (local). Static review — branch not checked out._

sdh07:
Good catch on the major finding — fixed in b99f2c0.

The Step 3.5 sweep input was built by "URL ∈ `ingested[]`", which re-trusts the agent batch JSON the contamination corrupts. Under the reported signature, source #45's agent echoes #35's `sources` URL, so `u45 ∉ ingested[]` → the `{slug:#45, url:#45}` dispatch entry is filtered out → the contaminated page is never swept. Exactly the threat the load-bearing layer is supposed to be deterministic against.

Re-keyed to the **per-source dispatch index**: dispatch entry `i` is swept iff its batch-result file `.ingest.batch.<NNN>.<i>.json` reported `ok: true`. Step 3.1b now states the dispatch table is ordered by the same `<NN>` the orchestrator assigned to each `BATCH_OUTPUT_PATH`, so the pairing is well-defined and contamination-proof — no agent-populated field is trusted. A contract assertion (`per-source index`) locks the glue against a silent revert to URL-membership. Detector script unchanged, as you noted.

Follow-ups:
- **Body-only cross-talk** (🟡): agreed this slips through an id+url-only check; the deterministic closure is the `content_hash` == `fetch-cache.py fetch` cross-check you mention. Filing as a separate issue rather than scoping it here, since the observed bug is id+url-crossed.
- **Glue regression guard** (🟡): partially addressed by the `per-source index` contract assertion above; the full quarantine round-trip stays a manual/local check per house style (the orchestration is LLM prose).

---

sdh07:
## Re-review — ✅ approve, merge-ready

The major finding is resolved correctly in `b99f2c0a3`.

- **Step 3.1b** now orders the persisted dispatch table by the same per-source index `<NN>` as the batch output paths, so dispatch entry `i` ↔ `.ingest.batch.<NNN>.<i>.json` describe the same source — a well-defined, orchestrator-owned pairing.
- **Step 3.5.1** keys the sweep input by **per-source index** (`keep entry i iff .ingest.batch.<NNN>.<i>.json reported ok:true`), explicitly forbids URL-membership filtering, and documents why in-prose (the contaminated-source-echoes-sibling-URL escape). Only `ok:false` indices are excluded, so the false-`page_missing` concern stays handled without trusting any agent-populated field.
- A contract assertion (`per-source index`) locks the glue against a silent revert.

Detector script + pre-write guard unchanged, as expected — the gap was purely orchestration glue and it's now contamination-proof against the exact threat the load-bearing layer targets.

Follow-ups both handled: body-only cross-talk filed as #421; the glue is partially guarded by the new contract assertion (full quarantine round-trip stays a manual/local check per house style).

No remaining blockers. _(Static re-review — branch not checked out; the diff reads correctly and the new assertion passes as written.)_

Issue #407, opened 2026-06-01
[Change] cogni-knowledge: make sub-questions first-class wiki nodes with bidirectional sq↔finding backlinks


sdh07:
## Context

Surfaced during the first real cogni-knowledge dogfooding run (EU AI Act for SMEs, DE/DACH, the #264 Phase-6 human-gate run). The pipeline already records the **forward** link from a finding to the sub-question it answers, but the relationship is not a browsable bidirectional graph in the wiki.

## Current state

- Every candidate in `candidates.json` carries `sub_question_refs: [sq-NN, ...]`; this propagates to each extracted claim (`pre_extracted_claims[].sub_question_refs`) at `knowledge-ingest`.
- `knowledge-ingest` files each source under its sub-question's `theme_label` as the `wiki/index.md` category, so the index reads thematically.

## The gap

There is **no first-class sub-question node** in the wiki, and **no `[[wikilink]]` backlink graph** between a sub-question and its finding pages. The sq↔finding linkage lives only in `.metadata/*.json` and in index headings — it is invisible to the wiki's forward→reverse link contract (`SCHEMA.md` R3) and to backlink-graph tooling (`wiki-dashboard`, `backlink_audit.py`, graph).

Consequences:
- Cannot open a "sq-03: Pflichten nach Risikoklasse" page and see its findings backlinked.
- A source page does not `[[link]]` back to the research question(s) it answers.
- The research-question structure of a run is lost once the project `.metadata/` is archived; only the wiki survives long-term, and the wiki doesn't hold it.

## Proposed change

Make sub-questions first-class, citable wiki nodes so the question→finding relationship becomes part of the Karpathy backlink graph:

1. At `knowledge-finalize` (or `knowledge-ingest`), emit one wiki page per sub-question (candidate type: a new `question` type, or reuse `note`/`concept`) carrying `query`, `theme_label`, and `[[wikilinks]]` to the source/synthesis pages that answer it.
2. Have ingested source pages `[[backlink]]` to their sub-question node(s) (driven by the `sub_question_refs` already present), satisfying SCHEMA.md R3 forward→reverse.
3. Surface the question nodes in `wiki/index.md` (they can become the thematic category anchors that `theme_label` currently only approximates).

## Out of scope / open questions

- Whether sub-question nodes should persist across runs (a recurring theme answered by multiple projects → one node accumulating backlinks) or be per-project. This couples to `topic_lineage.covered_themes` in `binding.json`.
[...]

PR #430, opened 2026-06-02
Close question→synthesis loop: link syntheses to question nodes (#410)

### LINKED ISSUE #410: cogni-knowledge: synthesis→question links (finalize + compose) for type: question nodes
Follow-up to #407 / #408.

## Context

#408 added `type: question` wiki nodes with the **sq↔finding** backlink pair (question `## Findings` `[[source]]` forward links + a `source→question` reverse link backfilled via `backlink_audit.py --apply-plan`), satisfying SCHEMA `R1`. The `synthesis→question` relationship was explicitly **out of scope for v1**.

## What's deferred

Two related additions so the question node participates in the rest of the pipeline's graph, not just at ingest time:

1. **`knowledge-finalize`:** when a verified synthesis is deposited, link it to the research-question node(s) it answers — i.e. emit a `synthesis→question` `[[link]]` (and its reverse) so a reader on a question page can reach the synthesis that ultimately answered it, closing the loop `question → findings → synthesis`.
2. **`knowledge-compose`:** allow the composer to cite a question node as cross-source framing where useful (analogous to how it already reads prior `wiki/syntheses/*.md`).

## Considerations

- Reuse the existing `backlink_audit.py --apply-plan` mechanism (same shape Step 4.5 already uses) rather than inventing a new linker.
- Mind the `R2_synthesis_wiki_source` rule — synthesis pages are leaves on the citation graph; decide deliberately whether the synthesis→question edge wants a reverse link or follows the synthesis-is-a-leaf exemption.
- cycle-guard interaction: confirm a synthesis→question edge can't create a self-citing loop the guard should catch.

Not blocking — #408 fully resolves #407's core ask.

sdh07:
Closes #410

Resolves the two deferred items of #410 (follow-up to #407/#408). Question nodes (`type: question`, `wiki/questions/<slug>.md`) previously dead-ended at the sources that answered them — a reader on a question page could not reach the verified synthesis that *ultimately* answered the question. This closes the `question → findings → synthesis` loop.

## Part 1 — `knowledge-finalize`: synthesis → question links

- **`knowledge-ingest` Step 4.5.1** now persists the `question-store.py emit` envelope's `data.questions[]` to `.metadata/question-manifest.json` — **unconditionally** (NOT gated on the `theme_bindings[]`-empty skip, which would have dropped it on legacy/no-theme projects). This is the phase handoff finalize reads.
- **`knowledge-finalize` Step 4.7** (new) reads that manifest into `QUESTION_SLUGS_CSV`; the Step 5/6 subprocess appends a trailing `## Research questions` body section of **bare** `[[<question-slug>]]` forward links, existence-gated by the same `is_file()` check the reference-row backlinks already use (so a question page deleted between ingest and finalize never becomes a `broken_wikilink`).
- The reverse `question→synthesis` link needs **no new code**: the existing Step 10.5 `lint --fix=reverse_link_missing` gate backfills `[[<synthesis-slug>]]` into each linked question page's `## See also`, exactly as it already de-orphans cited sources.
- `--no-question-links` opt-out; Step 11 surfaces the link count.

**Safety invariants (verified against the code):**
- **cycle-guard:** the forward links live in the page **body** — never in `sources:` frontmatter, never in the citation manifest — the only two places `cycle-guard.py` walks. So they cannot create a self-citing loop. No cycle-guard change.
- **R2:** body `[[wikilinks]]` are pure `R1_bidirectional_wikilink` → they get a reverse link (which is exactly what #410 asks for). `R2_synthesis_wiki_source` exempts only `sources:` frontmatter citations, not body links.

## Part 2 — `knowledge-compose` / `wiki-composer`: framing-only read

- **`wiki-composer` Phase 0** gains a question-node read step (`Glob wiki/questions/*.md`, topic-matched, lazy) — read as **framing-only orientation**, with an explicit **NEVER-cite** guard: a question node has no claim block, so an inline citation would score `unsupported` by `wiki-verifier`. The inverse of the distilled-page citation rule.
- **`knowledge-compose`** read-scope prose documents the same.
- Prompt-surface only — no script/manifest/citation-format/verifier change.

## Version, docs, tests

- Version bump `0.1.54 → 0.1.55` (`plugin.json` + `marketplace.json`).
- `CLAUDE.md`: four row updates (`knowledge-ingest`, `knowledge-finalize`, `knowledge-compose`, `wiki-composer`), `#410`-tagged.
- Grep-contract guards added to `test_{ingest,finalize,compose}_contract.sh`.
- Frontmatter `description:` fields untouched → no README §4 drift.

## Verification

- `test_{ingest,finalize,compose}_contract.sh` + `test_skill_contracts.sh` — all green; full `tests/` suite green (0 failing suites).
- The new Step 4.7 CSV-builder and the Step 5/6 append logic were exercised against a synthetic fixture: the existence gate correctly drops a missing question page, and a missing/empty manifest fails soft to an empty CSV (byte-identical legacy deposit).
- `question-store.py` is unchanged (`emit --help` byte-stable).

Remaining manual check (live `.alpha/` base, gitignored): end-to-end `ingest → finalize` confirming the synthesis `## Research questions` section, the question-page `## See also` reverse links after Step 10.5, and `wiki-health` reporting 0 errors; plus `doc-audit` on cogni-knowledge.

https://claude.ai/code/session_01GVBiXFRrqdKpcvRn1PWckY

---
_Generated by [Claude Code](https://claude.ai/code/session_01GVBiXFRrqdKpcvRn1PWckY)_


sdh07:
## Review — #410 question→synthesis loop

Reviewed against the approved plan; checked out the branch and ran the full `cogni-knowledge/tests/` suite (**39/39 green**), and traced the load-bearing claims in the cogni-wiki source.

### Correctness — verified in code
- **Discovery is robust.** Persisting the slug map at *emit time* (`question-manifest.json` handoff) sidesteps re-derivation drift — a later `knowledge-distill` page can't shift `-q` disambiguation, because finalize reads ground truth rather than recomputing.
- **Reverse-link survival across re-merges holds.** `_append_see_also` (`lint_wiki.py:653`) appends a fresh `## See also` **at EOF** — after the question page's `## Notes`. `question-store.py` captures everything from the `## Notes` marker to EOF as the preserved `notes_block`, so the `[[synthesis]]` reverse link is retained on a future enrich-merge. ✅
- **cycle-guard / R2.** Forward links live in the page **body**, never in `sources:` or the citation manifest — the only surfaces `cycle-guard.py` walks. `R2_synthesis_wiki_source` exempts only `sources:`-field citations, so the body link is pure R1 and correctly earns a reverse. No cycle-guard change needed. ✅
- **Existence-gated write** (`is_file()` per `[[slug]]`) prevents a `broken_wikilink` if a question page was deleted between ingest and finalize. ✅
- **Framing-only Part 2** is the right call — question nodes carry no claim block, so the explicit never-cite guard prevents an `unsupported` verdict. ✅
- **Fail-soft** on missing/empty/unparseable manifest and `--no-question-links` → byte-identical legacy deposit. ✅

### Findings
**🟢 Optional nits (non-blocking):**
- **Defensive dedupe:** within one `emit` run `resolve_slug` guarantees distinct slugs, so the manifest can't yield a duplicate `[[slug]]` bullet today; a one-line `dict.fromkeys(...)` over `question_links` would be cheap insurance against a hand-edited manifest, but isn't required.
- **Trailing section placement:** `## Research questions` sits *after* `## References`. The one real consumer (`wiki-contradictor`) strips the ref section and its assertive-sentence pre-filter never matches bare `[[slug]]` bullets, so it's safe in practice — just a note for any future synthesis post-processor.

### Tests & hygiene
Grep-contract guards added to all three contract suites in house style; version bumped `0.1.54 → 0.1.55` in both `plugin.json` and `marketplace.json`; CLAUDE.md four rows updated + `#410`-tagged; frontmatter `description:` untouched → no README §4 drift (final `doc-audit` worth running, already listed as a manual check).

### Verdict
**Safe to merge.** Logic matches the approved plan, the load-bearing reverse-link re-merge survival checks out, tests green. The remaining live-`.alpha` end-to-end check (synthesis `## Research questions` section + question-page `## See also` + `wiki-health` 0 errors) is the right pre-merge confirmation, since the contract tests can't exercise the runtime deposit.

🤖 Reviewed via Claude Code

sdh07:
Thanks for the thorough trace — both nits addressed:

- **Nit 1 (defensive dedupe):** applied in `35b44c3`. Added a `seen=set()` guard over `question_links` (the same idiom the reference-list dedupe uses earlier in the same subprocess), so a hand-edited `question-manifest.json` with a repeated slug can't yield a duplicate `[[slug]]` bullet — and it skips a redundant `is_file()` stat on a repeat.
- **Nit 2 (`## Research questions` after `## References`):** left as-is intentionally. As you noted, `wiki-contradictor` strips the ref section and its assertive-sentence pre-filter never matches bare `[[slug]]` bullets, so it's safe today; appending after the (auto-generated, last) references block keeps the deposit logic a single trailing append. Noted for any future synthesis post-processor.

Full `tests/` suite + breadcrumb guard still green. The live-`.alpha` end-to-end deposit check remains the right pre-merge confirmation.

---

Issue #415, opened 2026-06-01
[Bug] cogni-knowledge: source page frontmatter id crossed with another source's slug (wiki-health id_mismatch)


sdh07:
## Context

Surfaced by `wiki-health` during the #407 live dogfood re-ingest of the EU-AI-Act-for-SMEs base (`.alpha/eu-ai-act-sme/`, gitignored). It is **unrelated to #407** — the question-node emission was clean (0 type errors); this is a pre-existing source-page defect from the original ingest of that base.

## The bug

`wiki-health` reports exactly **one** `id_mismatch` error:

- **File:** `wiki/sources/overview-of-the-code-of-practice-eu-artificial-intelligence-act.md`
- **Frontmatter `id:`** `eu-ai-act-2026-was-sich-seit-inkrafttreten-veraendert-hat-ein-zwischenstand-tuev`

i.e. one source page's frontmatter `id:` carries a **different source's slug** (the TÜV "2026 Zwischenstand" page). The `id:` should equal the filename slug (`overview-of-the-code-of-practice-…`). Both are real, distinct sources ingested in the same 57-source wave.

## Why it matters

- It breaks `wiki-health` (1 hard error → blocks `wiki-lint` without `--ignore-health`).
- A crossed `id:` is the slug other pages address via `[[wikilink]]` / `wiki://`, so any citation or backlink resolving by that id would hit the wrong page.

## Root cause — unconfirmed

Observed, not root-caused. Most plausible: a **`SLUG` / id crossing between two `source-ingester` dispatches in the same one-wave fan-out** (the orchestrator resolves slugs in Step 1.2 and threads `SLUG=` per dispatch; the ingester sanity-checks the grammar but does not re-derive). Could also be a one-off local artifact. Needs investigation:

1. Confirm whether the on-disk page body/claims belong to the `overview-of-the-code-of-practice` source or the TÜV source (which one got mis-id'd).
2. Audit `source-ingester` + `knowledge-ingest` Step 1.2/3 for any path where a per-dispatch `SLUG` env var could be crossed across a wave.
3. If reproducible, add a contract guard; if a one-off, a `wiki-lint --fix=frontmatter_defaults`-style id reconciliation is the local cleanup.

Relates to epics #264 / #388.

PR #442, opened 2026-06-02
[cogni-service remote-test] cogni-knowledge: Phase 3.5 post-write id read-back in source-ingester

### LINKED ISSUE #417: [Enhancement] cogni-knowledge: source-ingester post-write id==filename self-check (prevent crossed-id recurrence)
## Context

Deferred follow-up from #415 / PR #416. #415 was a `wiki-health` `id_mismatch`: a `wiki/sources/<slug-A>.md` page whose frontmatter `id:` carried a *different* source's slug (`<slug-B>`) from the same 57-source ingest wave. PR #416 shipped the **cleanup** side (cogni-wiki `wiki-lint --fix=frontmatter_defaults` now reconciles a crossed `id:` to the filename stem). This issue tracks the **prevention** side, which #415 explicitly deferred as an unconfirmed one-off.

## Root cause (from #415 investigation)

`source-ingester` (`cogni-knowledge/agents/source-ingester.md`) writes **both** the page filename `wiki/sources/<slug>.md` (Phase 3, line ~137) and the frontmatter `id: <slug>` line (line ~92) from a **single** `SLUG` prompt parameter the orchestrator resolves once per URL (`knowledge-ingest` SKILL Step 1.2). `SLUG` arrives as **prompt text**, not an OS env var, so there is no env-var race.

A Step-1.2 misassignment would make filename **and** id both wrong → no mismatch. The only way filename ≠ id is the **LLM emitting the `id:` frontmatter line differently from the Write path it used** — an authoring slip inside the agent's Phase 3. The agent only sanity-checks the slug grammar (`[a-z0-9][a-z0-9-]{0,79}`); it does not assert `id == filename` after writing.

## Proposed guard

Add a fail-soft Phase-3 post-write self-check in `source-ingester.md`: after `atomic_write_text`, read back the just-written page and assert its frontmatter `id:` equals the filename stem (== the `SLUG` it was given). On mismatch, either correct the `id:` line deterministically (it must equal `SLUG`) or surface `reason: id_self_check_failed` in the batch envelope so the orchestrator can re-dispatch. Keeps the filename authoritative and closes the recurrence path at the source without relying on a later `wiki-lint --fix` sweep.

## Why deferred / priority

S4 — the crossing is a single observed occurrence in a local (gitignored) base, not reproduced. File now so it isn't lost; promote if it recurs in a future bake-in wave.

Relates to #415, #416, epics #264 / #388.

sdh07:
## Summary

- Adds an 8-line Phase 3.5 post-write read-back block inside the `source-ingester` Phase 3 Python write wrapper: after `atomic_write_text` completes, re-reads the page from disk and asserts the on-disk `id:` equals `SLUG`. On mismatch, emits `post_write_check_failed` to stderr and exits 4 (distinct from the pre-write guard's exit 3).
- Documents `reason: post_write_id_check_failed` in the Phase 4 skip-reason prose and adds an exit-4 bullet to the Failure-mode invariants.
- Adds two `assert_grep` assertions to `test_ingest_contract.sh` covering `sys.exit(4)` and `post_write_id_check_failed`.
- Bumps `cogni-knowledge` version `0.1.59 → 0.1.60` in `plugin.json` and `marketplace.json`.

Refs #417

## Assumptions (plan open questions)

**Open question 2 — exit-4 semantics:** soft path chosen. The page is already on disk when exit 4 fires (the pre-write Guard 1 passed and `atomic_write_text` completed), so a hard abort would leave the dispatch state unaccounted. Exit 4 emits `ok: false, reason: post_write_id_check_failed`; Step 3.5 sweep is the deterministic backstop that quarantines it on re-run.

**Open question 3 — URL re-check scope:** `id:`-only. The URL is already covered by the pre-write assertion (Guard 1); the post-write check closes only the residual risk of a bug in Guard 1's own Python code path on the `id:` field, which was the specific field that drifted in #415.

## Note

This PR was filed autonomously by a remote routine as a step-3→4 feasibility test of the cogni-service local→remote workflow and is safe for the maintainer to close. It does not supersede the maintainer's existing decision to close #417 as resolved by #413/#420 — it implements the plan that was approved as a workflow test regardless of issue disposition.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PGPj4q8M1LkRqEFWHFAR7Q)_

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -326,7 +326,7 @@
     {
       "name": "cogni-knowledge",
       "source": "./cogni-knowledge",
-      "version": "0.1.59",
+      "version": "0.1.60",
       "maturity": "preview",
       "description": "Wiki-first research that compounds. Binds each run to a cogni-wiki knowledge base; future runs read what prior runs filed before hitting the web. v0.1.0 inverted pipeline (plan → curate → fetch → ingest → distill → compose → verify → finalize) with zero-network, citation-consistent claim verification — opt-in live-source resweep via `knowledge-refresh --resweep`.",
       "keywords": [
--- a/cogni-knowledge/.claude-plugin/plugin.json
+++ b/cogni-knowledge/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "cogni-knowledge",
-  "version": "0.1.59",
+  "version": "0.1.60",
   "description": "Wiki-first research that compounds. Binds each run to a cogni-wiki knowledge base; future runs read what prior runs filed before hitting the web. v0.1.0 inverted pipeline (plan → curate → fetch → ingest → distill → compose → verify → finalize) with zero-network, citation-consistent claim verification — opt-in live-source resweep via `knowledge-refresh --resweep`.",
   "author": {
     "name": "Stephan de Haas",
--- a/cogni-knowledge/agents/source-ingester.md
+++ b/cogni-knowledge/agents/source-ingester.md
@@ -152,6 +152,18 @@ if obs_id != slug or normalize_url(obs_src) != normalize_url(url) or Path(os.env
     sys.stderr.write(f"integrity_mismatch: id={obs_id!r} slug={slug!r} src={obs_src!r} url={url!r}\n")
     sys.exit(3)
 atomic_write_text(Path(os.environ["PAGE_PATH"]), page)
+# Phase 3.5 post-write read-back — thin guard against a bug in the pre-write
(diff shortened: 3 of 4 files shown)


sdh07:
Closing — this was an autonomous cogni-service remote-impl-test artifact (its own body flags it as safe to close). Issue #417 is already CLOSED, resolved by #413/#420 (the Step 3.5 integrity sweep + the pre-write `id`/`url` guard already cover the crossed-id case). The PR is also stale against main (bumps to 0.1.60 while main is now 0.1.61 after #445, and overlaps source-ingester.md + test_ingest_contract.sh). No unique value to merge.

PR #443, opened 2026-06-02
cogni-knowledge: catch body-only ingest cross-talk via content_hash leg (#421)

### LINKED ISSUE #421: cogni-knowledge: ingest integrity sweep should also catch body-only cross-talk via content_hash
## Context

Follow-up from the #413 review (PR #420). The Step 3.5 post-wave integrity sweep (`ingest-integrity.py`) and the `source-ingester` Phase 3 pre-write assertion catch a cross-contaminated page by comparing its frontmatter `id:` + first `sources:` URL against the orchestrator's authoritative dispatch record. That covers the **observed** signature from the #264 dogfooding run, where source #45's page adopted source #35's identity wholesale (`id`, `title`, `sources` URL, claims, and body).

## Gap

The id+url check has a residual blind spot: if an ingest-wave agent keeps its **own** dispatched `id:` / `sources:` frontmatter but writes a **sibling's body + `pre_extracted_claims:`**, both legs of the sweep pass and the body-only contamination ships undetected. The composer would then cite a page whose claims don't match its stated source.

## Proposed fix (deterministic, no new trust surface)

The closure already exists conceptually in `source-ingester.md`: the page frontmatter carries `content_hash` = the fetched source body's provenance hash (`fetch-cache.py fetch --url <dispatched-url>` → `entry.content_hash`). Extend the integrity check to assert the page's `content_hash` equals the cache entry's `content_hash` for the **dispatched** URL:

- In `ingest-integrity.py sweep`: for each dispatched entry, additionally compare the page's frontmatter `content_hash` against `fetch-cache.py fetch`'s `entry.content_hash` for that entry's URL. This requires passing `--knowledge-root` so the sweep can read the cache. A mismatch is a new `body_mismatch` / `content_hash_mismatch` violation reason.
- Optionally mirror the same assertion in the `source-ingester` Phase 3 pre-write guard.

Note `source-ingester.md` already documents that `content_hash` is the **fetched body's** hash, not `hash(<on-disk page>)` — the on-disk body diverges by design once backlink trailers are appended — so the comparison must be cache-`entry.content_hash` vs page-frontmatter-`content_hash`, never a recomputed hash of the on-disk markdown.

## Scope / priority

Non-blocking; the observed bug is id+url-crossed and is covered by PR #420. This hardens the sweep against the narrower body-only variant. Low priority until/unless a body-only case is observed.

sdh07:
## Context

cogni-knowledge's Phase-4 ingest fans out many `source-ingester` agents in **one** wave (`--batch-size`, default 25). Non-deterministic LLM attention cross-talk in that wave can make agent A's `wiki/sources/<A>.md` page adopt sibling B's content. Two deterministic guards already defend against this, both comparing the on-disk page to the orchestrator's **authoritative dispatch record** (never the agent-returned batch JSON, which can itself be contaminated):

- **Script sweep** — `ingest-integrity.py sweep` (Step 3.5), the load-bearing backstop.
- **Agent pre-write guard** — `source-ingester.md` Phase 3, the fail-fast first line.

Both assert frontmatter `id == dispatched slug` **and** first `sources:` URL normalizes to the dispatched URL — catching the loud signature where a page's `id:`/`sources:` were crossed wholesale.

### Residual blind spot (#421)

If a contaminated agent keeps its **own** dispatched `id:`/`sources:` but emits a **sibling's body + `pre_extracted_claims:`** — and, riding along with that frontmatter, the sibling's `content_hash:` line — both existing legs pass and the body-only contamination ships. The composer would later cite a page whose claims don't match its stated source.

## Fix

Add a third, deterministic, **additive** leg: assert the page's frontmatter `content_hash` equals the fetch-cache entry's `content_hash` for the **dispatched** URL (the authoritative provenance hash of the fetched body, `sha256:<hex>`). It compares cache-`entry.content_hash` vs page-frontmatter-`content_hash` — **never** a recomputed hash of the on-disk markdown, which diverges by design once the `# <title>` H1 and `## See also` backlink trailers are appended.

Implemented on **both** surfaces (sweep + agent mirror), both reading frontmatter through one new shared `_knowledge_lib` parser so they cannot drift — exactly how the existing `id:`/`sources:` guard is split.

## Changes

- **`_knowledge_lib.py`** — `extract_page_content_hash(page_text)` beside `extract_page_id_and_url`, mirroring its idiom (reuses `_FRONTMATTER_RE`/`_unquote_scalar`; `""` on absence so the leg fail-safe-skips).
- **`ingest-integrity.py`** — optional `--knowledge-root` enables the leg; `_cache_content_hash` shells to `fetch-cache.py fetch` (owns the cache layout — not re-inlined). Cache miss / non-zero exit / empty hash on either side ⇒ leg skipped (never a false positive). New `reason: content_hash_mismatch`, precedence **id > url > content_hash**. `_violation` gains `observed_content_hash`/`expected_content_hash`/`content_hash_ok` (defaulted so the no-`--knowledge-root` path is byte-compatible with the prior id/url-only sweep).
- **`source-ingester.md`** — Phase 3 guard threads `CONTENT_HASH` (the Phase-1 `entry.content_hash`) and adds the leg (`obs_ch and obs_ch != CONTENT_HASH` ⇒ `sys.exit(3)` / `integrity_mismatch`). Stated semantically (no breadcrumbs).
- **`knowledge-ingest/SKILL.md`** — Step 3.5 passes `--knowledge-root <KNOWLEDGE_ROOT>` (already in scope); documents `content_hash_mismatch` + the body-only variant. Quarantine sub-steps unchanged (reason-agnostic).
- **Versioning** — `plugin.json` + `marketplace.json` `0.1.59 → 0.1.60`; `CLAUDE.md` rows updated.

## Out of scope (acknowledged residual)

The check catches the body-only variant **only when the contaminated `content_hash:` line bled along with the body**. A page that keeps the *correct* `content_hash` but a sibling's body cannot be caught without recomputing the on-disk hash — impossible by design (title/backlink trailers + `excerpt_position` anchoring). This is exactly the limitation #421 states.

## Tests

- `test_ingest_integrity.sh` — new cases: `content_hash_mismatch` (id+url match, hash diverges), positive (hash match), cache-miss (leg skips), backwards-compat (no `--knowledge-root`).
- `test_knowledge_lib.sh` — `extract_page_content_hash` unit cases (quoted / unquoted-with-comment / absent / no-frontmatter), plus a companion `extract_page_id_and_url` case.

### Verification run locally

- `ingest-integrity.py sweep --help` shows `--knowledge-root` as optional ✅
- `test_ingest_integrity.sh` ✅ · `test_knowledge_lib.sh` ✅ · `test_ingest_contract.sh` ✅ · `test_fetch_cache.sh` ✅ · `test_skill_contracts.sh` ✅
- `scripts/check-breadcrumbs.py` — 0 new violations ✅
- Backwards-compat: no-`--knowledge-root` sweep over the new fixtures → zero violations, unchanged envelope shape ✅

Closes #421

https://claude.ai/code/session_01MHhozEhrAMbALscay91kr2

---
_Generated by [Claude Code](https://claude.ai/code/session_01MHhozEhrAMbALscay91kr2)_


sdh07:
## Review verdict: approve with one minor follow-up

Reviewed the `content_hash` leg (Closes #421). Correct, fail-safe, backwards-compatible, conventions-clean (additive JSON envelope, stdlib-only, version mirrored to marketplace.json, no breadcrumbs in SKILL/agent files). The acknowledged residual is exactly what #421 states. One actionable gap for the remote session to fix before marking ready:

### 🔧 Fix: contract-guard the new `--knowledge-root` flag

`test_ingest_contract.sh` asserts the Step 3.5 sweep is named and calls `ingest-integrity.py` (lines 125–134), but was **not** updated to guard the new leg. The functional test (`test_ingest_integrity.sh`) proves the *script* honors `--knowledge-root`, but **nothing asserts the orchestrator actually passes it**. So a future edit could drop `--knowledge-root` from `SKILL.md`, silently disabling the entire content_hash leg, and CI would stay green — the exact regression class these contract tests exist to prevent (cf. how #413 guarded the original sweep invocation).

Please add to `cogni-knowledge/tests/test_ingest_contract.sh`:

```bash
assert_grep 'knowledge-root' "$INGEST" "knowledge-ingest: Step 3.5 sweep passes --knowledge-root to enable the content_hash leg"
assert_grep 'content_hash_mismatch' "$INGEST" "knowledge-ingest: documents the content_hash_mismatch reason"
```

and alongside the existing `extract_page_id_and_url` lib guard (line ~250):

```bash
assert_grep 'def extract_page_content_hash' "$LIB" "_knowledge_lib: defines extract_page_content_hash shared by sweep + Phase-3 guard"
```

### Nitpicks (non-blocking — fix only if convenient)

- **Agent/script asymmetry:** the script skips the leg when the cache hash is empty, but the agent guard (`obs_ch and obs_ch != content_hash`) trips when the page has a hash even if `CONTENT_HASH` is empty. Only reachable in the degenerate empty-body case (where the agent templates `content_hash: ""` too), so harmless — noting for completeness.
- **`subprocess.run` in `_cache_content_hash` has no `timeout=`.** `fetch-cache.py fetch` does no network (local JSON read), so a hang is implausible; leaving it is fine.

Everything else — fail-safe skip logic, reason precedence (id > url > content_hash), root separation, the black-box `fetch-cache.py store` test hygiene, and the symmetric shared-parser design — looks right. 👍

PR #472, opened 2026-06-04
cogni-knowledge: honest pdf_render_unavailable reason for saved-but-unrenderable PDFs

### LINKED ISSUE #458: [Bug] cogni-knowledge: source-curator drops text-layer PDFs as pdf_extraction_failed when poppler is absent
**Plugin:** cogni-knowledge v0.1.65
**Marketplace:** insight-wave
**Environment:** macOS (Darwin 25.4.0) / Claude Code

### What happened
During `knowledge-curate`, every PDF candidate was recorded as
`fetch.status: unavailable` with `reason: pdf_extraction_failed`. The
`source-curator` PDF branch (agents/source-curator.md §"PDF branch") relies on the
Read tool transcribing the WebFetch-saved PDF via vision, but the Read tool's PDF
rendering requires `pdftoppm` (poppler-utils). On a host without poppler — and where
Homebrew is also absent, so `brew install poppler` is not even available — the branch
fails for ALL PDFs, including plain text-layer PDFs that need no rasterization/OCR.

Three high-value EU authority PDFs were wrongly dropped in a real run:
- ENISA Managed Security Services Market Analysis (June 2025)
- EC Cloud Sovereignty Framework v1.2.1 (Oct 2025)
- European Parliament ITRE briefing (European software & cyber dependencies)

All three were recoverable: `pip install pdfminer.six` + `pdfminer.high_level.extract_text`
pulled clean text from each (ENISA ~14k words) with no poppler involved.

### Expected behavior
Text-layer PDFs should be extracted without poppler. The PDF branch should fall back to
a pure-Python text extractor (e.g. pdfminer.six / pypdf) before recording
`pdf_extraction_failed`. Only genuinely image-only PDFs (no text layer) should fall
through to `pdf_extraction_failed` (which is correctly non-cobrowse-eligible).

### Steps to reproduce
1. On a host without poppler/pdftoppm installed.
2. Run `knowledge-plan` -> `knowledge-curate` on a topic whose authority sources include
   institutional PDFs (e.g. ENISA reports).
3. Observe every `.pdf` (and PDF-by-content-type) candidate recorded as
   `fetch.status: unavailable`, `reason: pdf_extraction_failed`.

### Error output / logs
```
pdftoppm is not installed. Install poppler-utils (e.g. `brew install poppler`
or `apt-get install poppler-utils`) to enable PDF page rendering.
```
(Returned by the Read tool when handed the WebFetch-saved PDF; the curator then
records `pdf_extraction_failed`.)

### Additional context
- Two issues compound: (1) the only PDF path is poppler-dependent, and (2) the curator's
  returned summary attributed the failure to "pdftoppm not available" rather than to the
  missing saved-file/extraction step, which is misleading during triage.
- Suggested fix: add a pure-Python text-extraction fallback in the PDF branch
  (pdfminer.six or pypdf, stdlib-friendly install) ahead of the poppler/vision path, and
  reserve `pdf_extraction_failed` for PDFs with no recoverable text layer.
- A `pdf_extraction_failed` miss is marked `cobrowse_eligible: false`, so without this
  fallback these sources are terminally lost on poppler-less hosts even though their text
  is trivially extractable.

sdh07:
Closes #458.

## Summary

The `source-curator` PDF branch collapsed **two distinct failures** into one terminal `pdf_extraction_failed` (`cobrowse_eligible: false`, dropped forever):

1. WebFetch surfaced **no saved-file path** (the EUR-Lex ELI/landing case — a summary, not the PDF binary). Read was never reached. *This is genuinely terminal.*
2. A file **was** saved, but the **Read tool couldn't render it** in this runtime. *This is environmental, not URL-fatal.*

Case (2) was mislabeled as `pdf_extraction_failed` and blamed on `pdftoppm`/poppler in the agent's summary. But **poppler is the Read tool's own page→image rasterization dependency — it appears nowhere in cogni-knowledge** (`grep -rni poppler cogni-knowledge/` → zero hits). It is not ours to work around.

## Approach — curator-side honesty, not local extraction

Per the design rule, **PDFs are read via the Read tool only — no homegrown or external parser**. So this fix does not add any text extractor; it makes the failure reporting honest and un-conflated:

- **`fetch-cache.py`** — add `pdf_render_unavailable` to `VALID_REASONS`, classed environmental / operator-actionable (mirrors `cobrowse_unavailable`).
- **`agents/source-curator.md`** — the PDF branch now handles "saved file but Read can't render it" explicitly: record `pdf_render_unavailable`, **never** attribute the failure to an external binary, and **never** attempt local PDF text extraction. `pdf_extraction_failed` is narrowed to "no saved-file path at all".
- **`references/fetch-cache-design.md`** — both reasons documented; `cobrowse_eligible: false` for both (cobrowse can't render PDF text either), but `pdf_render_unavailable` is **not** URL-fatal — re-run where the Read tool can render PDFs and the URL resolves.
- **`tests/test_ingest_contract.sh`** — assert the new token in `VALID_REASONS` and that the curator records `pdf_render_unavailable` + forbids local extraction.
- Version bump `0.1.70 → 0.1.71` (`plugin.json` + `marketplace.json`).

## Scope note

The bug report's original "Expected behavior" asked for a pure-Python text-extraction fallback (pdfminer/pypdf). That was **declined by the maintainer** in favor of keeping the Read tool as the sole PDF mechanism (no external packages, no homegrown parser). This PR therefore does **not** recover text locally on a host that lacks PDF-rendering support — it makes such a host report an honest, operator-actionable reason instead of silently dropping the source. Recovering the three EU authority PDFs is a matter of running where the Read tool can render PDFs.

## Test plan

- [x] `bash cogni-knowledge/tests/test_ingest_contract.sh` — ALL PASS (new `pdf_render_unavailable` + no-local-extraction assertions included).
- [x] `bash cogni-knowledge/tests/test_fetch_cache.sh` — all pass (reason gate).
- [x] `fetch-cache.py store --reason pdf_render_unavailable` accepted at the CLI gate (`success: true`).
- [x] `plugin.json` + `marketplace.json` parse; both at `0.1.71`.
- [x] Maintainer-breadcrumb guard clean; no `#NNN` ref in the agent file.


sdh07:
**Minor finding (non-blocking) — stale "full vocabulary" enumeration**

`cogni-knowledge/agents/source-fetcher.md:105` claims to list *the full closed vocabulary* but now omits `pdf_render_unavailable` (added to `VALID_REASONS` in this PR):

> The full closed vocabulary (`webfetch_*`, `pdf_extraction_failed`, `cobrowse_unavailable`, `cobrowse_failed`, `cache_write_failed`) lives in `references/fetch-cache-design.md` …

It already enumerates `pdf_extraction_failed` — which, like the new token, is `cobrowse_eligible: false` and never actually reaches source-fetcher — so by the same completeness logic the new token belongs in this list too.

Functionally harmless (the reference file stays the single source of truth, and source-fetcher never sees either PDF reason), but it is exactly the kind of inline-list drift this PR otherwise eliminates. Suggested one-token fix:

```
The full closed vocabulary (`webfetch_*`, `pdf_extraction_failed`, `pdf_render_unavailable`, `cobrowse_unavailable`, `cobrowse_failed`, `cache_write_failed`) lives in `references/fetch-cache-design.md` §"Reason semantics" — single source of truth.
```

Everything else LGTM — tests pass, both manifests at `0.1.71`, `grep -rni poppler cogni-knowledge/` confirms zero hits, and the no-local-extraction guard is correctly asserted. Approve with this optional cleanup.

sdh07:
Addressed in 0e4b7b22 — added `pdf_render_unavailable` to the inline closed-vocabulary list at `source-fetcher.md:105`, so it now mirrors `references/fetch-cache-design.md` §"Reason semantics" (the single source of truth, which already documents the token at line 75). No version bump: cosmetic doc realignment within the same unmerged PR, both manifests stay at `0.1.71`.

That clears the one optional cleanup from your approve — PR is ready for merge.

PR #479, opened 2026-06-04
cogni-wiki: raw_citation_depth lint auto-fix for depth-wrong ../raw/ citations

sdh07:
Closes #478.

Follow-up to #459 / PR #475 (the `../raw/` → `../../raw/` depth fix + health.py resolve-from-page-location). PR #475 made `health.py` resolve raw-source citations from the page's actual on-disk location, so every already-ingested page still carrying a depth-wrong `../raw/<file>` citation now flips to a health ERROR — and `wiki-lint` refuses to run while `health.errors > 0`. Users on existing wikis would otherwise be stuck doing the repair by hand.

## Summary
- New deterministic `raw_citation_depth` fix class in `lint_wiki.py`'s existing `--fix` framework. For each `sources:` entry and each body `## Sources` link whose `../raw/<tail>` resolves (from the page's on-disk parent) outside `raw/`, rewrite to `../../raw/<tail>` — but only when `<wiki-root>/raw/<tail>` actually exists (skip-don't-guess otherwise). Subdirectory tails are preserved (computed from the original string, not `Path.name`).
- Class added to `FIX_CLASSES` (auto-propagates to `FIX_CHOICES`, argparse `choices`, and `--fix=all`); dispatch wired into the in-process `_wiki_lock` block alongside the other page-body fixers, reading each page fresh from disk so it composes with them.
- Body rewrites are restricted to the `## Sources` section (until the next `##` heading) so prose links are never touched.
- SKILL.md: new fix-class table row, `--fix=` param entry, frontmatter description, and the count/locking-paragraph updates.
- CHANGELOG `0.0.58` entry warning existing-wiki users that the hardened `missing_source` check now surfaces depth-wrong citations as health ERRORs, and to run `wiki-lint --fix=raw_citation_depth` (or `--fix=all`) once to repair.
- Version bump `0.0.57` → `0.0.58` in plugin.json and marketplace.json.

## Scope decisions
- Implemented the issue's **option (a)** — a fix class in the existing `--fix` framework — over a standalone `migrate-raw-citation-depth.py` driver, per triage's preferred approach and to reuse the locking, dry-run, fail-soft, and stats plumbing already in place.
- The fixer iterates all pages directly via `slug_index` (like `frontmatter_defaults`), not a lint warning — lint emits no such class; this is a pre-health repair that runs regardless of warnings to clear the health errors.
[...]

PR #481, opened 2026-06-04
cogni-wiki: simplify raw_citation_depth fixer (cleanup on #479)

sdh07:
Quality-only cleanup (`/simplify`) stacked on top of #479's `raw_citation_depth` lint auto-fix. No behaviour change; reviewed for reuse / simplification / efficiency / altitude, then applied the high-confidence findings.

## What changed
- **Reuse** — `_rewrite_body_sources` (new in #479) and the existing `_append_see_also` each hand-rolled the same "find a `## <heading>` section body, from just after the heading line to the next `##` or EOF" logic. Extracted a shared `_find_section(text, heading)` helper; both call sites now use it, so the section-bounds rule lives in one place.
- **Efficiency** — `fix_raw_citation_depth` now early-outs on pages whose text has no `../raw/` substring, before the frontmatter split and the `sources:` / `## Sources` scans. The depth regex can never match without that substring, so the common no-raw-citation page costs one `in` check instead of a full parse.

## Findings reviewed but not applied
- **Altitude** (all acceptable): the `../raw/` → `../../raw/` one-level correction is a hard schema invariant (pages live two deep since schema 0.0.5, enforced by `fail_if_pre_migration`); the regex's single-`..` scope matches exactly the known migration history; the "does the corrected path exist" check intentionally diverges from `health.py`'s detector (detector vs. repairer); the `## Sources` body-scope is the right altitude.
- The no-frontmatter `else` branch was flagged as possibly redundant, but unlike `fix_frontmatter_defaults` this fixer does real work on a body `## Sources` without frontmatter (reachable under `--ignore-health`), so it was kept.
- A per-`tail` `exists()` cache and a one-line `_sources_field_idx` extraction were judged too marginal to be worth the added surface.

## Test plan
- `bash cogni-wiki/tests/test_lint_fix.sh` passes (incl. the `reverse_link_missing` cases that exercise the refactored `_append_see_also`, and all `raw_citation_depth` cases).
- `bash cogni-wiki/tests/test_lint_health_partition.sh` passes.

https://claude.ai/code/session_01RyE6Vor7mkCwypeMykJxTe

---
_Generated by [Claude Code](https://claude.ai/code/session_01RyE6Vor7mkCwypeMykJxTe)_


sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/481

**Summary:** Single perspective passed with no findings; cogni-wiki v0.0.59 is a clean behaviour-preserving simplify cleanup with consistent versioning and a passing test suite. Conflicts with main (from #479) were resolved by keeping this PR's shared-helper refactor and early-out on top of #479's merged base.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- Version consistency verified: plugin.json (0.0.59) and marketplace.json cogni-wiki mirror (0.0.59) match, bumped from 0.0.58. (plugin-validator)
- _find_section is a faithful byte-for-byte extraction of the section-bounds regex used by both _append_see_also and _rewrite_body_sources. (plugin-validator)
- Helper genuinely shared by both fixers (See-also path + Sources path), satisfying the reuse claim. (plugin-validator)
- raw_citation_depth early-out is provably behaviour-preserving — a page without '../raw/' is a guaranteed no-op. (plugin-validator)
- fix_raw_citation_depth still registered in FIX_CLASSES and dispatched inside the shared _wiki_lock. (plugin-validator)
- Full test_lint_fix.sh passes including new inline-list/single-scalar coverage. (plugin-validator)
- Plugin manifest well-formed; stdlib-only; no schema bump needed (write-path internals only). (plugin-validator)
- Changed file set is exactly the five declared files with no scope creep; CLAUDE.md v0.0.59 changelog accurate. (plugin-validator)

### Recommendation

---

sdh07:
### Code review

Three low-severity findings from a high-effort pass on the `_find_section` refactor and early-out. No correctness bugs — the logic is sound — but worth tidying before this pattern propagates to new call sites.

---

**1. Return-type annotation is wrong (`-> tuple` can return `None`)**
`cogni-wiki/skills/wiki-lint/scripts/lint_wiki.py`, line 666

```python
def _find_section(text: str, heading_name: str) -> tuple:  # ← should be tuple | None
```

Both call sites already guard against `None` before unpacking, so there's no runtime hazard today. But the annotation is factually wrong — mypy/pyright will flag any caller that doesn't check for `None`, and a future caller who trusts the annotation and writes `a, b = _find_section(...)` directly will get a `TypeError`. Fix: `-> tuple[int, int] | None`.

---

**2. Inconsistent `None`-check between the two call sites**
Lines 692 vs 1000

```python
# _append_see_also (line 692)
if span:               # implicit truthiness — None → falsy, (a,b) → truthy

# _rewrite_body_sources (line 1000)
if span is None:       # explicit sentinel check
```

Both happen to work today (a non-empty tuple is always truthy). The inconsistency becomes a maintenance hazard once more call sites exist. Suggest standardising on `if span is not None:` / `if span is None:` throughout — explicit sentinel checks are clearer when the function's return contract is `value | None`.

---

**3. `re.compile()` called on every `_find_section` invocation — breaks the module-level regex pattern**
Lines 672–674

```python
def _find_section(text: str, heading_name: str) -> tuple:
    pat = re.compile(
        rf"^##\s+{re.escape(heading_name)}\s*$", re.MULTILINE | re.IGNORECASE
    )
```

The file compiles all its regexes as module-level constants (`REVERSE_LINK_SOURCE_RE`, `TAG_TYPO_MSG_RE`, `RAW_DEPTH_RE`, etc., lines 134–145). `_find_section` is called once per page per fix pass — so for a 500-page wiki running `--fix=all`, this is 1 000+ redundant compile calls (two fixers × page count). Since the heading names are known at call time, one option is to cache compiled patterns in a module-level dict keyed by heading name, or pass a pre-compiled pattern as the argument.

---

No action needed before merging — all three are clean-up candidates for a follow-up. Marking informational.

---

PR #847, opened 2026-06-17
cogni-knowledge: recompute excerpt_position at ingest, stop silent claim drops

### LINKED ISSUE #836: [Bug] cogni-knowledge: claim-extractor silently drops claims on excerpt_position offset drift
## Summary
The Phase-4 `claim-extractor` contract carries an `excerpt_position` (a character offset) per claim, hand-computed by the LLM. On bodies with multi-byte / typographic characters (em-dashes, curly quotes) the offset drifts, and the `source-ingester`'s `body.find(excerpt_quote) == excerpt_position` sanity check then **drops** the mismatched claims. When an ingester does *not* self-correct, a source contributes a fraction of its claims and most of its evidence is lost — silently.

## Evidence (perf study, two cold-cache runs)
- Run 1: the `nemko.com` source kept **1 of 12** extracted claims ("11 dropped due to extractor offset drift").
- Several other ingesters across both runs reported "corrected offset via `body.find()`" — i.e. the drift is routine; only the agents that happen to re-locate avoid the loss.

## Impact
Silent evidence loss. A source we paid to fetch + extract reaches the synthesis with a fraction of its claims, with no warning surfaced. Affects every run.

## Proposed fix
Make the position authoritative *by computation*, not by LLM hand-count:
- Drop `excerpt_position` from the `claim-extractor` output contract; have the ingester (or claim-extractor) re-locate each `excerpt_quote` via `body.find()` and compute the position at write time.
- If a quote genuinely isn't found, surface it as a per-claim warning (not a silent drop), so a real extraction error is visible.

This removes a whole fragile failure class at low cost.

sdh07:
Closes #836.

## Summary
- `source-ingester` no longer **drops** claims when the extractor's hand-counted `excerpt_position` mismatches `body.find(excerpt_quote)`. It now **recomputes** the offset via `body.find()` at write time: on `find() >= 0` it overwrites `excerpt_position` with the recomputed value and keeps the claim — offset drift on multi-byte / typographic characters is no longer a drop condition.
- A genuinely-absent quote (`body.find() == -1`) is now a **per-claim warning** in the Phase 4 batch envelope `notes` field (a real extraction error becomes visible) rather than a silent drop. The Phase 4 success-envelope shape now documents the optional `notes` field so the warning has a home.
- `claim-extractor`'s `excerpt_position` is softened to **advisory** at both contract anchors: a hand-count mismatch on multi-byte chars is expected and harmless because the ingester recomputes. The field stays in the emitted shape + frontmatter description.
- `references/claim-at-ingest.md` and `references/inverted-pipeline.md` updated to reflect "computed authoritatively at write time", reconciling the stale "sanity-check is positional, not semantic" restatement.
- Patch bump cogni-knowledge 1.0.10 → 1.0.11 (plugin.json + marketplace.json mirror).

## Why
The `excerpt_position` was hand-computed by the LLM and drifts on multi-byte / typographic bodies; the old `body.find(excerpt_quote) == excerpt_position` equality check then dropped the mismatched claims silently (perf study: a source kept 1 of 12 claims). Making the offset authoritative-by-computation removes the whole fragile failure class at low cost.

## Scope
`excerpt_position` stays present and is now always correct, so downstream readers are unaffected: `wiki-verifier` does not score on it (context-render only), `source-contradictor` ignores it, `wiki-composer` uses only `text` + `excerpt_quote`. No script reads or recomputes the offset (it is purely an agent-prompt + frontmatter concern), so there is no Python change and no test change.

## Test plan
- [x] `tests/test_ingest_contract.sh` — ALL PASS (asserts `excerpt_position` field presence in claim-extractor — still present).
- [x] `tests/test_skill_contracts.sh` — ALL PASS.
- [x] `tests/test_knowledge_lib.sh`, `tests/test_sub_index.sh`, `tests/test_wiki_source_manifest.sh`, `tests/test_verify_store.sh` — all pass (use `excerpt_position` only as fixture frontmatter).
- [x] Breadcrumb guard clean (0 violations); both manifests valid JSON at 1.0.11.


sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/847

**Summary:** Single perspective (plugin-validator) passed cleanly — version bumped and mirrored to 1.0.11, all four prose files internally consistent, no findings raised.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- plugin.json + marketplace.json both 1.0.11, valid JSON
- notes field coherently threaded across the recompute rule + Phase 4 envelope + claim-extractor-failure path
- claim-extractor still lists excerpt_position (presence test stays green)
- no script/test change, no structural breakage

### Recommendation

---

PR #1121, opened 2026-07-17
cogni-projects: data model + entity-authoring skill (Phase 1)

claude-cogni-work:
Closes #1113.

Phase-1 data-model foundation for epic #1119: defines the three core entity schemas, an authoring skill, and a stdlib validator on top of the #1112 scaffold.

## Summary
- Add `cogni-projects/references/data-model.md` — the YAML-frontmatter schema for all three entity types: **consultant** (skills, seniority, availability window, allocation), **project** (client, open roles, timeline, `strategic_impact` 1–5), and **assignment** (consultant×project join with role + date window). Mirrors the sibling data-model layout (`# … Data Model Reference` H1 → `## Entity Schemas` → per-entity `###` subsections): per-entity fenced example, Required/Optional field lists, enum tables, a kebab-case Naming Conventions table (composite `<consultant>--<project>` assignment slug), the manifest summary-ref shape, and a closing mermaid erDiagram.
- Add the domain-prefixed `projects-entities` skill — locate the portfolio, read siblings, write ONE entity markdown file into `consultants/|projects/|assignments/`, register a summary ref idempotently (dedupe by slug) into the matching `projects-portfolio.json` array + bump `updated`, append an `.metadata/execution-log.json` transition, and gate the write on the validator.
- Add `cogni-projects/scripts/validate-entities.py` — python3 stdlib (no PyYAML): a frontmatter-subset parser plus schema checks (missing required key, invalid enum, non-kebab slug, invalid ISO date, out-of-range integer). Returns the repo `{"success", "data": {"errors", "warnings"}, "error"}` envelope with per-field `{entity, file, field, message}` errors and `success:false` on any error. Accepts a single entity file or a whole portfolio directory.
- Add `cogni-projects/scripts/register-entity.py` — the single writer of manifest refs (see Scope decisions for why it is a separate script).
- Patch-bump `cogni-projects/.claude-plugin/plugin.json` 0.0.1 → 0.0.4 and mirror it in `.claude-plugin/marketplace.json`.

## Scope decisions
- **Three entities, not four.** The #1112 scaffold committed to the noun `assignment`/`assignments` (dir + manifest array + CLAUDE.md + README). The issue's "availability" is a consultant/assignment allocation-window **attribute**, not a fourth entity, so it is modeled as frontmatter fields (`available_from`/`available_until`/`allocation_pct`). This rationale now lives in `data-model.md`'s assignment section, not only in this PR body.
[...]

Issue #1113, opened 2026-07-17
cogni-projects: define self-contained data model + entity-authoring skill


claude-cogni-work:
## Plugin
cogni-projects

## Epic context
Phase 1 (foundation) of "Roadmap: cogni-projects — partner project-portfolio steering (staffing MVP first)". This child defines the consultant/project/availability entities the Phase 2 staffing engine scores over; every later child reads these schemas.

## Use case
The staffing engine and all portfolio reporting need structured, self-contained records of consultants, projects, and availability. Partners author these manually first (no external system dependency), so the plugin needs a documented entity schema and a skill to create and validate entity files.

## Proposed solution
Define three entity schemas as Obsidian-browsable markdown with YAML frontmatter (kebab-case slugs), stored under the portfolio dir: **consultant** (profile: skills, seniority/grade, allocation/availability windows), **project** (client, open staffing roles/needs, timeline, `strategic_impact` score), and **availability** (per-consultant allocation windows). Document them in `references/data-model.md`. Add a domain-prefixed `projects-entities` skill to author/register a consultant/project/availability entity, and a `scripts/validate-entities.py` (python3 stdlib, `{success,data,error}` JSON contract) that checks frontmatter schema conformance. AC: (1) `references/data-model.md` specifies the frontmatter schema for all three entity types; (2) `projects-entities` writes a valid entity file under `cogni-projects/<portfolio-slug>/`; (3) `validate-entities.py` returns `{success:false}` with per-field errors on a malformed entity. ~1 skill + 1 references doc + 1 validation script.

## Alternatives considered
A single monolithic portfolio JSON instead of per-entity markdown — rejected: breaks the Obsidian-browsable, YAML-frontmatter entity convention used across cogni-* plugins. Starting from an external PSA/HR schema — deferred to the import child; the MVP stays self-contained.

_Affected components: cogni-projects/references/data-model.md, cogni-projects/skills/projects-entities/SKILL.md, cogni-projects/scripts/validate-entities.py_

_Downstream hint: plugin-dev:skill-development_

_Complexity hint: medium_

## Depends on
- #1112 cogni-projects: scaffold plugin skeleton + marketplace registration + projects-setup

## Blocks
- #1114 cogni-projects: staffing match engine (availability + profile + strategic impact) — MVP
- #1116 cogni-projects: partner-meeting portfolio dashboard (health + value)
- #1117 cogni-projects: external PSA/HR import of consultants and projects
[...]

PR #1123, opened 2026-07-17
cogni-projects: return the entity-script envelope on every path, resolve symlinks in the containment guard

claude-cogni-work:
Follow-up to the merged #1121. The review gate on that PR ran against its final head and surfaced findings after the merge had already been decided, so they land here instead. The substantive one is a live bug in `main`.

## Why this exists

Issue #1113 AC3 requires the entity scripts to return the `{success, data, error}` envelope on **every** path — "never a bare traceback or non-JSON stderr-only output". Three paths in `main` still break that contract. Each is reproduced below, and each is fixed here.

## Summary

- **Non-UTF-8 entity record crashed the validator** (`validate-entities.py`). `except OSError` does not catch `UnicodeDecodeError` (a `ValueError` subclass), so a record saved as latin-1 escaped the envelope and printed a bare traceback. A DACH name is the usual way in — the repo's CLAUDE.md mandates ä/ö/ü/ß support, so this is a realistic record, not a contrived fixture. Worse, in a whole-portfolio scan the first such record **aborted the run**, hiding every valid entity behind it.
- **Unreadable entity directory crashed the validator.** The directory walk sat outside any guard, so `chmod 000` on `consultants/` raised `PermissionError` past the envelope — the same class of filesystem failure the per-read catch addresses.
- **Corrupt manifest crashed the registrar** (`register-entity.py`). A `projects-portfolio.json` hand-edited to the wrong shape raised `AttributeError` mid-write, past the envelope, from the script that performs the writes.
- **Containment guard was symlink-blind.** The out-of-portfolio check resolved with `os.path.abspath`, which leaves symlinks unresolved, so an entity reached through a symlinked directory cleared the guard while still resolving outside the root — precisely the cross-portfolio ref-escape the guard exists to prevent. Now resolves with `realpath`.
- Cosmetic, from the same review: title-case the five remaining `data-model.md` H2s to match the sibling convention in cogni-portfolio / cogni-trends / cogni-consult, and drop a redundant data-model link from the skill's Step 2 lead sentence.
- Patch-bump `cogni-projects` 0.0.4 -> 0.0.5, mirrored in `marketplace.json`.

## Design note

The fix is deliberately at two levels rather than one. The narrow catch stays in `validate_file()` because `register-entity.py` calls it as a **library function**, bypassing `main()` — and because it yields an actionable message ("re-save it as UTF-8") that a generic top-level catch cannot. The top-level guard in each script's `__main__` dispatch is what actually makes "envelope on every path" true, covering argument handling, directory walking, and manifest I/O. The scan-loop guard is kept for a distinct reason a top-level guard cannot serve: it lets the scan **continue** past a bad record so one file no longer suppresses the rest.

## Verification

Every path re-checked on this branch; all return parseable JSON.

| Case | Before | After |
|---|---|---|
| valid entity | `{"success": true}` | `{"success": true}` rc=0 |
| latin-1 entity | **bare traceback** rc=1 | envelope, per-field `<file>` error rc=1 |
| portfolio scan (1 bad + 1 good) | **traceback, scan aborted** | envelope, `checked: 2` — good entity reported rc=1 |
| unreadable directory | **PermissionError traceback** | envelope rc=2 |
| corrupt manifest | **AttributeError traceback** | envelope rc=2 |
| symlink escape | silently **registered** | refused rc=2 |
| register valid / idempotent re-run | ok | `created` -> `updated`, array stays length 1 rc=0 |

- `python3 -m py_compile` passes on both scripts; stdlib-only preserved (no pyyaml or any pip import).
- `scripts/check-breadcrumbs.py` passes.
- Both manifests parse and mirror 0.0.5; maturity stays `incubating` for the 0.0.x band; the em dash remains genuine UTF-8 (no escapes).

## Scope

README regeneration remains deferred to the cogni-docs pipeline, as declared in #1121.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -270,7 +270,7 @@
     {
       "name": "cogni-projects",
       "source": "./cogni-projects",
-      "version": "0.0.4",
+      "version": "0.0.5",
       "maturity": "incubating",
       "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
       "keywords": [
--- a/cogni-projects/.claude-plugin/plugin.json
+++ b/cogni-projects/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "cogni-projects",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
   "author": {
     "name": "Stephan de Haas",
--- a/cogni-projects/references/data-model.md
+++ b/cogni-projects/references/data-model.md
@@ -10,7 +10,7 @@ the later staffing-match / backfilling / dashboard skills.
 Partners author these records manually (no external system dependency in the
 MVP), so the schema is deliberately flat and human-writable.
 
-## Portfolio layout
+## Portfolio Layout
 
 A portfolio is one `cogni-projects/<portfolio-slug>/` directory rooted by a
 `projects-portfolio.json` manifest (scaffolded by `projects-setup`):
@@ -27,7 +27,7 @@ cogni-projects/<portfolio-slug>/
 Each entity lives in one markdown file under its type's subdirectory. The file's
 subdirectory determines its type, and the frontmatter `type` field must agree.
 
-## Frontmatter conventions
+## Frontmatter Conventions
 
 - Frontmatter is a leading `---` … `---` fence at the top of the file.
 - Values are **flat**: scalars, ISO dates (`YYYY-MM-DD`), integers, and simple
@@ -180,7 +180,7 @@ Valid `status` values:
 
 ---
 
-## Naming conventions
+## Naming Conventions
 
 | Entity | Slug shape | Example |
 |--------|-----------|---------|
@@ -191,7 +191,7 @@ Valid `status` values:
 The `--` double-hyphen separator in assignment slugs makes the join legible and
 keeps one assignment per consultant-project pair addressable by a stable slug.
 
-## Manifest registration
+## Manifest Registration
 
 When `projects-entities` authors an entity, `scripts/register-entity.py
 <portfolio-dir> <entity-file>` upserts a **summary ref** into the matching array
@@ -218,7 +218,7 @@ The entity markdown file is the source of truth for full field values; the
 manifest ref is the index the dashboard and staffing skills scan without opening
 every file.
 
-## Entity relationships
+## Entity Relationships
 
 ```mermaid
 erDiagram
--- a/cogni-projects/scripts/register-entity.py
+++ b/cogni-projects/scripts/register-entity.py
@@ -112,7 +112,10 @@ def main(argv):
     ref = {key: str(fm[key]) for key in REF_FIELDS[entity_type]}
     # The ref points at the entity file relative to the portfolio root, so the
     # dashboard and staffing skills can resolve it without knowing the cwd.
-    rel = os.path.relpath(os.path.abspath(entity_file), os.path.abspath(portfolio_dir))
+    # realpath, not abspath: abspath leaves symlinks unresolved, so an entity
+    # reached through a symlinked directory would clear the check below while
+    # still resolving outside the root.
+    rel = os.path.relpath(os.path.realpath(entity_file), os.path.realpath(portfolio_dir))
     # Refuse an entity that lives outside the target portfolio. Validation cannot
     # catch this — a foreign entity file is itself perfectly valid — but the ref
     # would escape the root, and consumers resolve `file` relative to it, so one
@@ -194,4 +197,11 @@ def main(argv):
 
 
 if __name__ == "__main__":
-    sys.exit(main(sys.argv[1:]))
+    # Same envelope contract as validate-entities.py: this script also writes,
+    # so a hand-edited manifest of the wrong shape must report as a failure the
+    # caller can read rather than a traceback mid-write.
+    try:
+        _code = main(sys.argv[1:])
+    except Exception as _exc:  # noqa: BLE001 — deliberate catch-all
+        _code = _fail("unexpected failure: %s: %s" % (type(_exc).__name__, _exc))
+    sys.exit(_code)
--- a/cogni-projects/scripts/validate-entities.py
+++ b/cogni-projects/scripts/validate-entities.py
@@ -233,6 +233,12 @@ def validate_file(filepath):
     except OSError as exc:
         err("<file>", "cannot read file: %s" % exc)
         return errors, warnings
+    except UnicodeDecodeError as exc:
+        # Entity records are UTF-8 by contract, but a record saved as latin-1
+        # (a DACH name is the usual way in) would otherwise raise past the
+        # envelope every other path returns.
+        err("<file>", "cannot decode file as UTF-8 — re-save it as UTF-8: %s" % exc)
+        return errors, warnings
 
     fm = parse_frontmatter(text)
     if fm is None:
@@ -361,7 +367,20 @@ def main(argv):
 
     errors, warnings = [], []
     for f in all_files:
-        e, w = validate_file(f)
+        # Report an unexpected failure as an ordinary per-field error rather
+        # than raising: one unreadable record must not abort the scan and hide
+        # every valid entity behind it, and callers gate registration on this
+        # envelope.
+        try:
+            e, w = validate_file(f)
+        except Exception as exc:  # noqa: BLE001 — the envelope is the contract
+            e, w = [{
+                "entity": "unknown",
+                "file": f,
+                "field": "<file>",
+                "message": "unexpected failure while validating: %s: %s"
+                           % (type(exc).__name__, exc),
+            }], []
         errors.extend(e)
         warnings.extend(w)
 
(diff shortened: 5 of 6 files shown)


claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1123

**Summary:** All three perspectives passed. This PR fixes an acceptance-criterion violation (issue #1113 AC3: envelope on every path) that shipped to main in the merged #1121. plugin-validator verified the fix empirically against synthetic fixtures: the three traceback paths (latin-1 record, chmod-000 directory, list-shaped manifest) now return the envelope, the portfolio scan continues past a bad record, and the symlink guard is proven load-bearing — abspath registered an escaping ref, realpath refuses it. The three findings that remain are advisory minors under Rule 5b; none is major or critical and none blocks merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (auto-fixable) `cogni-projects/references/data-model.md:30` — The 'Frontmatter Conventions' and 'Validation' sections never state that entity files must be UTF-8, yet this PR makes that a hard, enforced contract with a dedicated error ('cannot decode file as UTF-8 — re-save it as UTF-8'). Given that these portfolios carry European names with genuine non-ASCII characters, a partner who hits this error has no canonical doc stating the requirement — the data model is declared canonical over SKILL.md, so it is the right home for it.
  - *Suggested fix:* Add one bullet under 'Frontmatter Conventions': entity files are UTF-8 encoded; a record saved as latin-1 (the usual way in is a DACH name) is rejected by validate-entities.py and must be re-saved as UTF-8.

- **[minor]** (auto-fixable) `cogni-projects/scripts/register-entity.py:133-136` — AC1(c) is met in its hard clause (no traceback), but only via the top-level catch-all, so the message is not actionable: registering against a list-shaped manifest emits error 'unexpected failure: AttributeError: list object has no attribute setdefault' (exit 2, verified). It leaks a Python implementation detail and never says the manifest root is the wrong shape. The script already has the precedent guard for exactly this class one line later at :137 (if not isinstance(array, list)). Confirmed no partial write occurs.
  - *Suggested fix:* After the json.load at :133, mirror the :137 precedent with an isinstance(manifest, dict) check returning a _fail naming the manifest shape. The top-level guard then stays a true safety net rather than the handler for a named AC path.

- **[minor]** (auto-fixable) `cogni-projects/scripts/validate-entities.py:214` — AC1(b) is met for an unreadable entity FILE, which the narrow except OSError localizes into a per-field error while the scan continues (verified). But an unreadable entity DIRECTORY raises PermissionError out of os.listdir in _entity_files, which sits outside the per-file loop guard, so it is caught only at top level: the whole scan aborts, hiding valid entities in sibling dirs. AC2's letter says 'one bad record', and a directory is not a record, so this is beyond its scope — and the outcome is fail-closed and names the offending dir, so nothing is green-lit.
  - *Suggested fix:* Wrap the os.listdir(d) at :214 in a try/except OSError and surface it as a per-subdir entry the caller can read, letting the walk continue to the remaining subdirs.

### Positive observations

- The three-layer catch design is sound, not redundant — each layer verified independently load-bearing by execution; removing the narrow except UnicodeDecodeError would change register-entity.py's exit-code contract (1 = validation failure vs 2 = environment failure). (plugin-validator)
- The realpath fix is empirically load-bearing, not speculative: abspath resolved the symlinked-dir entity to an escaping ref and would have registered it; realpath refuses it, with no false positives on a legitimately symlinked portfolio root. (plugin-validator)
- UnicodeDecodeError is a ValueError subclass, so the added except is correctly disjoint from the existing except OSError — precisely the subtlety that caused the original miss, and the inline comment names it. (plugin-validator)
- The manifest diff is surgical (two version lines only), 0.0.5 mirrored in both files, maturity correct for the 0.0.x band, and byte-level inspection confirms genuine UTF-8 em dashes across all six changed files. (plugin-validator)
- check-breadcrumbs.py exits 0 with zero violations, and cogni-projects is confirmed absent from the baseline — so the clean run means genuinely no breadcrumbs, not a ratcheted exemption. (plugin-validator)
- Happy paths do not regress: all three entity types validate, idempotency holds exactly (array length 1, action created->updated), and the umlaut round-trips as raw UTF-8. (plugin-validator)
- Failure modes are fail-closed throughout — every new path returns success:false, so a caller gating registration on the envelope can never mistake a handled failure for a clean validation. (plugin-validator)
- Both scripts remain stdlib-only; no pyyaml or pip import. (plugin-validator)
- Only two data-model links remain, down from three; both resolve correctly and use text==href so they render identically in Obsidian and on GitHub. (skill-reviewer)
- The drift guard survived the dedupe intact — the dropped link was the redundant one, not the load-bearing one. (skill-reviewer)
- Field lists verified against references/data-model.md line-by-line for all three entity types — they match exactly. (skill-reviewer, skill-quality-reviewer)
- Progressive disclosure is well-judged: 148-line body against the 500-line soft cap, with detail offloaded to references/data-model.md. (skill-quality-reviewer)
- Voice is imperative and reasons rather than commands — every non-obvious rule explains its rationale. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
### Code review

Second-opinion pass (high effort, 8 angles) on the diff vs `main`. The PR's own change is sound — the three envelope paths and the `realpath` guard were re-verified by execution and behave as claimed. The findings below are about the code the diff *touches* but does not fix; the first is a pre-existing data-loss bug this PR's own framing ("the envelope is the contract") brings into focus, since the envelope now reports "handled" while the data is gone.

**1. `cogni-projects/scripts/register-entity.py:174-180` — manifest write is not atomic; a mid-write failure silently corrupts the portfolio while stdout reports a clean handled failure.** (pre-existing; not introduced here)

`open(manifest_path, "w")` truncates in place before `json.dump` streams into it, so any failure *during* the write leaves a half-written file. The `except OSError` catches it and returns a well-formed envelope — which is precisely what makes it dangerous: the caller reads a tidy `{"success": false}` and has no signal that the manifest is now unparseable.

Reproduced (file-size limit standing in for a full disk):

```
manifest before: 18664 bytes, 200 consultants
$ ( ulimit -f 2; register-entity.py pf pf/consultants/ana-silva.md )
{"success": false, "data": {}, "error": "cannot write portfolio state: [Errno 27] File too large"}
manifest after: 1024 bytes -> *** CORRUPT: Unterminated string starting at line 59 ***
```

The portfolio is destroyed and the only record of it is an error string about writing, not about corruption. `.metadata/execution-log.json` is written *first* (line ~174), so it can also record a transition the manifest never received.

*Suggested fix:* write both files via temp-file + `os.replace()` (atomic rename on POSIX), so a failed write leaves the previous manifest intact. This is the standard fix and matches the "manifest written last / manifest existence is a reliable marker" idempotency posture in `cogni-projects/CLAUDE.md`.

**2. `cogni-projects/scripts/validate-entities.py:214` — an unreadable entity *directory* aborts the whole scan with an empty `errors[]`.**

`_entity_files`'s `os.listdir(d)` sits outside the per-file loop guard this PR adds, so `chmod 000 consultants/` raises `PermissionError` out of the walk and is caught only by the new top-level guard: `data.errors` is `[]` and the valid entities in sibling `projects/`/`assignments/` are never reported. The envelope contract holds and it fails closed, so nothing is green-lit — but the localization the PR adds for a bad *file* doesn't extend to a bad *directory*.

*Suggested fix:* wrap the `os.listdir(d)` in `try/except OSError` and surface it as a per-subdir error entry, letting the walk continue to the remaining subdirs.

**3. `cogni-projects/scripts/register-entity.py:133` — a list-shaped manifest reports a Python implementation detail.**

`manifest.setdefault(...)` on a manifest whose root is a JSON array raises `AttributeError`, now caught by the top-level guard: `error: "unexpected failure: AttributeError: 'list' object has no attribute 'setdefault'"`. No traceback (the contract holds) and no partial write (the raise precedes both writes), but the message never says the manifest root is the wrong shape. The precedent guard for exactly this class already exists one line later at `:137` (`if not isinstance(array, list)`).

*Suggested fix:* after the `json.load` at `:133`, add `if not isinstance(manifest, dict): return _fail(...)` naming the shape — leaving the top-level guard as a true safety net rather than the handler for a named path.

**4. `cogni-projects/scripts/register-entity.py:96` — dead handler.** (pre-existing)

`_ve.validate_file()` at `:85` reads the same file with the same encoding and the `if errors:` gate at `:86` returns first, so the `except OSError` on the `parse_frontmatter` read at `:96` is unreachable outside a TOCTOU rewrite between the two reads. Harmless, but it reads as a live path to the next maintainer. (The same reasoning is why this PR does *not* add a `UnicodeDecodeError` twin next to it — the decode error is already surfaced through the validator gate.)

**5. `cogni-projects/references/data-model.md:30` — the UTF-8 requirement is now enforced but undocumented.**

This PR makes UTF-8 a hard contract with a dedicated error (`cannot decode file as UTF-8 — re-save it as UTF-8`), but neither `## Frontmatter Conventions` nor `## Validation` states it. Since these portfolios carry European names by design and the data model is declared canonical over SKILL.md, that is the right home for the rule.

*Suggested fix:* one bullet under `## Frontmatter Conventions` — entity files are UTF-8; a record saved as latin-1 is rejected and must be re-saved.

---

**Verified working, for the record:** all three envelope paths (latin-1 record, unreadable directory, list-shaped manifest) return parseable JSON; the portfolio scan continues past a bad record (`checked: 2` with the valid entity still reported); the `realpath` guard refuses a symlink-escaping entity that `abspath` would have registered; idempotency holds (array length 1, `created` -> `updated`); both manifests parse and mirror 0.0.5; scripts remain stdlib-only; `check-breadcrumbs.py` passes clean.

Checked for bugs and CLAUDE.md compliance.

claude-cogni-work:
## Follow-up issues

The code-review pass above raised one finding that is out of this PR's scope — it is pre-existing in `main` (shipped with the merged #1121), and fixing it is a distinct concern from this PR's envelope work. Filed rather than left in a comment so it cannot evaporate:

- **#1124** — `register-entity.py`'s manifest write is not atomic, so a mid-write failure corrupts `projects-portfolio.json` while the script reports a clean handled envelope. Reproduced: an 18,664-byte manifest truncated to 1,024 bytes and left unparseable. Filed as S2-major because the manifest is the portfolio's root index and every downstream skill reads it.

The remaining review findings (2-5: the unreadable-directory scan abort, the list-shaped-manifest message, the dead handler at `:96`, and the undocumented UTF-8 requirement) are minor and non-blocking — the automated verdict is `svc:review-pass`. Happy to fold any of them into this PR if you'd prefer them handled here rather than deferred.

PR #1126, opened 2026-07-17
cogni-projects: atomic register-entity manifest + log writes

### LINKED ISSUE #1124: cogni-projects: register-entity.py manifest write is not atomic — a mid-write failure corrupts the portfolio
## Context

Surfaced by the second-opinion code review on PR #1123. It is **pre-existing in `main`** (shipped with the merged #1121), not introduced by that PR, and was left out of its scope because the fix is a distinct concern from the envelope work #1123 does.

## Problem

`register-entity.py` writes `projects-portfolio.json` and `.metadata/execution-log.json` by opening the target path directly with `open(path, "w")`. That truncates the file in place *before* `json.dump` streams the new content into it, so any failure **during** the write leaves a half-written file with no path back to the original.

The `except OSError` around the writes catches the failure and returns a well-formed envelope — which is what makes this dangerous rather than merely unlucky. The caller reads a tidy `{"success": false}` and is told the failure was *handled*; nothing in the output says the manifest is now unparseable.

## Reproduction

A file-size limit standing in for a full disk:

```
manifest before: 18664 bytes, 200 consultants

$ ( ulimit -f 2; python3 cogni-projects/scripts/register-entity.py pf pf/consultants/ana-silva.md )
{"success": false, "data": {}, "error": "cannot write portfolio state: [Errno 27] File too large"}

manifest after: 1024 bytes
*** CORRUPT: json.loads -> Unterminated string starting at line 59 column 7 ***
```

The portfolio is destroyed, and the only record is an error string about *writing*, not about corruption.

## Why it matters

The manifest is the portfolio's root index — `projects-setup`'s idempotency contract keys on it, and `cogni-projects/CLAUDE.md` states the manifest is "written last so its existence is a reliable 'fully initialized' marker". A truncated-but-present manifest breaks that invariant in the worst way: it exists, so it reads as initialized, but it no longer parses. Every downstream skill (staffing match #1114, backfilling #1115, the dashboard #1116) reads this file.

There is a second-order effect: the execution log is written **first**, so a manifest write that fails after the log write leaves the log recording a transition the manifest never received.

## Proposed solution

Write both files via temp-file + `os.replace()` (atomic rename on POSIX) so a failed write leaves the previous file byte-intact:

1. `json.dump` to a temp file in the same directory (same filesystem, so the rename is atomic).
2. `os.replace(tmp, target)` — atomic; either the old file or the complete new file is present, never a truncation.
3. Clean up the temp file on the failure path.
4. Apply to both `projects-portfolio.json` and `.metadata/execution-log.json`; keep the existing `ensure_ascii=False` writer settings so European names are unaffected.

Ordering is worth a thought while in here: writing the manifest before the log would make the log the lossy one rather than the index.

## Acceptance criteria

1. A mid-write failure (reproduce with `ulimit -f`) leaves `projects-portfolio.json` byte-identical to its pre-run content and still parseable.
2. The `{success,data,error}` envelope still reports the failure, and its message distinguishes "nothing was written" from a partial write.
3. The same guarantee holds for `.metadata/execution-log.json`.
4. No temp files are left behind on either the success or failure path.
5. Happy path and idempotency do not regress: registering the same slug twice leaves the array at length 1 with `action` flipping `created` -> `updated`.
6. Stays python3 stdlib-only; scripts keep returning the `{success,data,error}` envelope on every path.

claude-cogni-work:
Closes #1124.

## Summary
- Rewrite `register-entity.py`'s manifest and execution-log writes to `json.dump` into temp files in the target directory, then `os.replace` both — POSIX-atomic, so a mid-write failure (e.g. full disk) leaves each target byte-identical to its pre-run content instead of a truncated, unparseable file.
- Dump both temp files before either `os.replace`, so the common failure touches neither target; the `{success,data,error}` envelope's message now distinguishes "nothing written, existing files are intact" from the rare partial (log replaced, manifest failed) case.
- Clean up temp files on both success and failure paths; preserve `ensure_ascii=False` and the trailing newline so European names and on-disk bytes are unaffected; stay python3 stdlib-only.
- Apply the same temp-file + `os.replace` pattern to `portfolio-init.sh`'s three `.metadata` seed-log writes and its manifest write (identical non-atomic pattern, same plugin/scripts dir), keeping manifest-written-last so its existence stays the completion marker.
- Add `cogni-projects/tests/test_register_entity.sh` (first test in this plugin) covering the failure, happy, and idempotency paths.
- Bump cogni-projects 0.0.5 → 0.0.6 in `plugin.json` and `marketplace.json`.

## Why
`register-entity.py` and `portfolio-init.sh` wrote JSON state with `open(path, "w")` + `json.dump`, which truncates the target in place before the new bytes land. A failure during the write left a half-written, unparseable file with no path back to the original — and the surrounding `except OSError` returned a tidy `{"success": false}` envelope that told the caller the failure was *handled* while `projects-portfolio.json`, the portfolio's root index, was now corrupt. `os.replace` is an atomic rename on the same filesystem: either the old file or the complete new file is present, never a truncation.

Triage found the blast radius wider than the issue stated: `portfolio-init.sh` carried the identical non-atomic pattern for the manifest and the three `.metadata/*.json` seed files. Both are fixed here — they are not genuinely separable (same plugin, same scripts dir, same pattern).

## Scope decisions
- **In scope:** both write sites in `register-entity.py`, the two write sites in `portfolio-init.sh`, a new test, and the version bump.
- **Out of scope:** `validate-entities.py` (no file writes); no inversion of the log-first/manifest-second order — dumping both temps before either replace already leaves both files intact on the common failure, so per-file atomicity makes each write individually safe without changing the documented order or its docstring.
- The atomic-write pattern is vendored (imitating the house `cogni-knowledge/_knowledge_lib.py` / `cogni-consult/resolve-assumptions.py` shape), not imported across the plugin boundary — plugins stay self-contained.

## Test plan
- [x] `bash cogni-projects/tests/test_register_entity.sh` passes — 16 checks (happy, idempotency, encoding, monkeypatched-`os.replace` failure, real `NotADirectoryError`).
- [x] Monkeypatched-`os.replace` failure: envelope reports `success:false` with a message distinguishing nothing-written from partial; the pre-existing `projects-portfolio.json` is byte-identical on re-read and still parses; no `.projects-portfolio.json.*.tmp` / `.execution-log.json.*.tmp` debris remains.
- [x] Real reproduction from the issue: under `ulimit -f` with a manifest larger than one block, the fixed script returns a `{success:false}` envelope (`[Errno 27] File too large (nothing was written; existing files are intact)`), the manifest hash is unchanged, it still parses, and no temp debris remains.
- [x] Happy path + idempotency: registering a fresh entity returns `success:true`, array length 1, action `created`; re-registering the same slug leaves length 1 with action flipping to `updated`; European (ä/ö/ü/ß) names round-trip without ASCII escaping.
- [x] `portfolio-init.sh` still scaffolds correctly, seeds the three `.metadata` logs and the manifest atomically, and re-run returns the "already initialized" envelope; no temp debris.
- [x] `plugin.json` and `marketplace.json` both read 0.0.6 for cogni-projects.

Note: the test simulates the failure by monkeypatching `os.replace`, **not** `ulimit -f` — the latter's `SIGXFSZ` can terminate the process before the handler runs (platform-dependent), which would both skip the cleanup and be non-deterministic across macOS/Linux.


sdh07:

_The merger formed this expectation from issue #1124 before checking out the PR head — the review grades the diff against it._

**Expected approach:** Replace the truncate-in-place writes in `register-entity.py` (the `open(log_path,"w")` then `open(manifest_path,"w")` block inside the single `except OSError`) with a stage-then-rename pattern: `json.dump` each payload into a temp file created *in the same directory as its target* (same filesystem, so `os.replace` is a true atomic rename), then `os.replace(tmp, target)`, unlinking any surviving temp on the failure path. Both writes must be staged before either replace, narrowing the log/manifest divergence to the `os.replace` window. The existing log-first order may be kept or inverted, but the module docstring must stop describing a write scheme the code no longer uses. Writer settings (`indent=2`, `ensure_ascii=False`, trailing newline, `encoding="utf-8"`) must survive verbatim so European names are not ASCII-escaped, and `_fail()` must still fire on every write failure with a message stating the targets are unchanged. The same pattern applies to the four writes in `portfolio-init.sh`'s python heredoc — three `.metadata` seed logs (`json.dump` WITHOUT `ensure_ascii=False`) and the manifest (WITH it) — and the manifest must still be the last thing to appear on disk, since the idempotency guard keys on its existence. Version bumps 0.0.5 → 0.0.6 in both `plugin.json` and `marketplace.json` are required by repo convention, and the new test should follow the harness shape of `cogni-portfolio/tests/test-validate-entities.sh`.

**Acceptance checklist:**
- [ ] `cogni-projects/.claude-plugin/plugin.json` `version` bumped 0.0.5 → 0.0.6, and the `cogni-projects` entry in root `.claude-plugin/marketplace.json` carries the identical 0.0.6 string.
- [ ] The marketplace.json `cogni-projects` entry keeps `"maturity": "incubating"` (0.0.x range) — the diff must not flip it.
- [ ] `cogni-projects/tests/test-register-entity.sh` exists, is invoked as `bash cogni-projects/tests/test-register-entity.sh`, and exits non-zero on assertion failure.
- [ ] The diff introduces no new maintainer breadcrumbs (`#NNNN` refs, version tags, slice codes) — confirm via `scripts/check-breadcrumbs.py` against its baseline, not by eye.
- [ ] In `register-entity.py`, neither `projects-portfolio.json` nor `.metadata/execution-log.json` is opened with mode `"w"` on its final target path any more.
- [ ] Each temp file is created in the SAME directory as its target (`dir=...`) — a bare tempfile default lands in `/tmp` and makes `os.replace` raise `EXDEV`.
- [ ] `os.replace(tmp, target)` (not `os.rename` across devices, not `shutil.move`) is used for both targets.
- [ ] Both payloads are serialized to their temp files BEFORE either `os.replace` runs.
- [ ] The manifest write still uses `indent=2, ensure_ascii=False`, appends a trailing newline, with `encoding="utf-8"` on the temp handle.
- [ ] The execution-log write keeps its base settings (`indent=2`, `ensure_ascii=False`, trailing newline).
- [ ] Every failure path still returns via `_fail()` — single-line envelope on stdout, exit code 2, no traceback escaping.
- [ ] The write-failure message is changed from the flat "cannot write portfolio state: %s" to wording that tells the caller the on-disk files are unchanged, with a distinct message for the rare `os.replace`-stage failure (issue AC 2).
- [ ] Temp files are unlinked on the failure path, and the cleanup is failure-tolerant (unlinking a temp `os.replace` already consumed must not raise and turn a success into a failure).
- [ ] No non-stdlib import is added.
- [ ] The module docstring's "Not transactional: ... log first" is updated to describe the actual new scheme.
- [ ] `portfolio-init.sh`'s heredoc writes all three `.metadata` seed logs via the same temp+`os.replace` pattern, with temps in `.metadata`.
- [ ] Seed-log writes preserve `json.dump(payload, f, indent=2)` WITHOUT `ensure_ascii=False`, while the manifest keeps `ensure_ascii=False` — the diff must not homogenize the two.
- [ ] `portfolio-init.sh` still writes `projects-portfolio.json` LAST (its `os.replace` is the final filesystem mutation before the success envelope).
- [ ] The idempotency guard (existing manifest → "already initialized", `exit 0`) is untouched in behavior.
- [ ] The test covers the corruption regression itself: a constrained-write fixture that asserts the pre-run manifest bytes are unchanged AND still `json.loads`-parseable after the failed run.
- [ ] The test asserts no temp files remain after BOTH a successful and a failed run (issue AC 4).
- [ ] The test asserts idempotency still holds (issue AC 5): same slug twice → array length 1, `action` flips `created` → `updated`.
- [ ] The test invokes `register-entity.py` in a way that works with its hyphenated filename (subprocess or `importlib.util.spec_from_file_location`).
- [ ] No production behavior outside the write mechanics changes: validator gate, outside-portfolio realpath/relpath guard, `REF_FIELDS`, and success-envelope `data` keys all unmodified.

**Risk watchlist:**
- Temp created without `dir=` → lands on `/tmp`, `os.replace` fails with `EXDEV` in the field while passing a same-filesystem test.
- `mkstemp`/`NamedTemporaryFile` create at mode 0600 and `os.replace` preserves it — the manifest silently loses group/other read bits.
- A half-measure fixing only the manifest and leaving `.metadata/execution-log.json` truncating, or fixing only `register-entity.py` and skipping `portfolio-init.sh`.
- Fixing atomicity but forgetting the error-message refinement, so issue AC 2 is unmet while everything else passes.
- Losing `ensure_ascii=False` on either writer (or adding it to the seed logs where base lacks it) — European names become `\uXXXX` escapes.
- Dropping the trailing newline when moving to a temp handle, churning every subsequent diff.
- An unlink-in-`finally` that runs after a successful `os.replace` and raises `FileNotFoundError`, converting a healthy run into an "unexpected failure" envelope.
- Inverting log/manifest order without updating `CLAUDE.md`'s "Manifest written last" convention and the docstring — two documents contradicting the code.
- A `ulimit`-based test that is platform-dependent: exceeding `RLIMIT_FSIZE` can deliver `SIGXFSZ` and kill the interpreter rather than surface `EFBIG`.
- A test that leaks its temp portfolio into the repo tree — `portfolio-init.sh` hardcodes `BASE_DIR="cogni-projects/${SLUG}"` relative to cwd.
- `portfolio-init.sh` runs under `set -euo pipefail`; a raise inside the new heredoc write logic kills bash with no JSON envelope — pre-existing gap the fix could widen.
- `cogni-projects/CLAUDE.md`'s architecture tree does not list a `tests/` directory; adding the first test leaves the developer guide stale (not a hard gate — CLAUDE.md is not in the diff).
- Durability beyond atomicity (`fsync` on the temp fd and the containing dir) is not required by the stated guarantee; absence should not block, presence should not be penalized.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1126

**Summary:** contract-grading-reviewer found two contract items (6, 15) unsatisfied — an unaddressed file-permission regression from mkstemp()'s 0600 default, and a stale module docstring — while plugin-validator independently confirmed the same permission regression (rated minor) plus four further polish items. All 24 other contract items are satisfied and the new 16/16 test suite passes.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[major]** (auto-fixable) `review-contract:item-6` — The dir= requirement is met (register-entity.py:194-198 and portfolio-init.sh:39-43 both pass dir=os.path.dirname(path)), but the same tempfile.mkstemp() calls introduce an unaddressed permission regression the contract's own risk_watchlist flagged in advance as a required explicit check: mkstemp() always creates at mode 0600 regardless of umask, and os.replace(tmp, path) preserves that mode. Verified by direct execution (three times independently, incl. by the merger): a manifest written via the pre-existing open(path,'w') pattern is 0644 under umask 022, but after register-entity.py's os.replace-based write both projects-portfolio.json and .metadata/execution-log.json come out 0600. The identical _atomic_write_json helper in portfolio-init.sh applies the same regression to all three seed logs plus the manifest. No chmod anywhere in the diff restores the prior mode.
  - *Suggested fix:* After each os.replace (or on the temp before it), restore the mode a normal open(path,'w') would have produced — capture the previous file's mode via os.stat when the target exists and re-apply it, else 0o644 & ~umask — in both register-entity.py's write block and portfolio-init.sh's _atomic_write_json helper.
  - *Raised by:* cogni-service:contract-grading-reviewer, also flagged by plugin-dev:plugin-validator

- **[minor]** (auto-fixable) `review-contract:item-15` — register-entity.py's module docstring is untouched by the diff — it still reads 'Not transactional: the execution log and the manifest are two writes, log first.' MERGER ADJUDICATION: the contract-grading-reviewer graded this major on the reading that the docstring describes a scheme the code no longer uses. On the merger's own read the statement remains literally TRUE — the code still performs two writes, log first, and still offers no both-or-neither guarantee (the partial case exists and the envelope names it). So this is a documentation-COMPLETENESS gap (the headline property the PR adds, per-file atomicity, goes undocumented in the file the PR rewrote), not a false claim. Downgraded major -> minor accordingly; it does not independently gate merge.
  - *Suggested fix:* Supplement the docstring paragraph to describe the new per-file atomic write (temp + os.replace) and the two distinct failure envelopes (nothing written vs. log updated but manifest not). Keeping the existing 'two writes, log first' sentence is fine — it is still accurate.

- **[minor]** (waived) (auto-fixable) `cogni-projects/scripts/register-entity.py:200` — The temp file is closed but never fsync'd before os.replace, and the containing directory is not fsync'd after. This gives atomicity but not durability: on a power loss shortly after the rename, some filesystems can surface a zero-length manifest. Same gap at portfolio-init.sh:48. NOTE: the review contract explicitly stated durability beyond atomicity is NOT required by the issue's stated guarantee and that its absence must not block — recorded as advisory only.
  - *Justification:* The committed review contract's risk_watchlist states durability beyond atomicity is not required by the issue's stated guarantee — absence must not block, presence must not be penalized. Waived per the merger's own pre-committed expectation.

- **[minor]** (auto-fixable) `cogni-projects/tests/test-register-entity.sh` — Filename deviates from the established repo-wide test convention — existing test files across cogni-consult, cogni-knowledge, cogni-portfolio and cogni-trends use underscore naming (test_citation_store.sh etc). This PR introduces the first hyphenated one. The file is also mode 0644 while most existing tests are 0755.
  - *Suggested fix:* Rename to cogni-projects/tests/test_register_entity.sh and chmod +x for consistency with the sibling plugins.

- **[minor]** (auto-fixable) `cogni-projects/CLAUDE.md:12` — The Plugin Architecture tree lists .claude-plugin/, skills/, scripts/ and references/ but was not updated to include the newly introduced tests/ directory. README.md's components table likewise lists only the script, so a new contributor will not know tests exist or how to run them.
  - *Suggested fix:* Add a tests/ entry to the CLAUDE.md architecture tree with the invocation, matching how cogni-knowledge documents its tests.

- **[minor]** (auto-fixable) `cogni-projects/scripts/portfolio-init.sh:55` — _atomic_write_json re-raises OSError, so a write failure during init escapes the heredoc as a Python traceback under set -euo pipefail rather than the {success,data,error} envelope the repo convention requires. Unchanged from the base version (pre-existing, not a regression introduced here), but the PR rewrote exactly this code path without closing the gap.
  - *Suggested fix:* Wrap the two _atomic_write_json call sites in try/except OSError and print the failure envelope before exiting non-zero. Reasonable as a follow-up rather than in this PR.

### Positive observations

- Contract items 1-5, 7-14 and 16-24 all satisfied with cited file:line evidence
- New test suite runs green 16/16 (exit 0), bash + python3 only
- The test deliberately avoids ulimit -f and documents why — SIGXFSZ terminates the process before the handler runs, leaving the .tmp debris the contract forbids. A genuinely subtle trap avoided
- Test meaningfully covers the fix: happy path, idempotency upsert, European-character round-trip without \u escaping, monkeypatched os.replace failure asserting byte-identical manifest plus no debris, and a real filesystem OSError
- check-breadcrumbs.py reports zero violations against the 45-entry baseline; check-skill-names.sh OK
- Dumping both temp files before either os.replace is the correct sequencing — the disk-full-prone step happens while both live files are untouched
- ensure_ascii=False preserved on the manifest writer while seed logs correctly keep ensure_ascii=True — the two are not homogenized
- Manifest still written LAST in portfolio-init.sh, preserving the existence-as-completion-marker invariant
- Idempotency guard byte-for-byte unchanged; temp cleanup guarded by os.path.exists plus try/except OSError
- Envelope returned on every path including the __main__ catch-all, with a genuine nothing-written vs partially-written distinction driven by a replaced counter
- Version synced at 0.0.6 across plugin.json and root marketplace.json; zero version drift across all 14 marketplace entries; maturity: incubating correctly retained
- Only new import is stdlib tempfile — no pip dependency introduced

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1126

**Summary:** Re-review after revision 9a13eb7e. All five actionable findings from the prior verdict are genuinely closed. The permission regression is confirmed fixed by independent empirical test — the validator checked out the pre-fix writer at 486691c9 and reproduced 0600 on all four files, then confirmed the revised writer yields 0644 under umask 022 and 0600 under umask 077, so _intended_mode tracks the umask rather than hardcoding a mode. 21/21 tests pass, and the five new mode assertions were verified to FAIL against the unfixed writer, making them a real regression gate rather than a tautology. No regressions in temp cleanup, manifest-last ordering, the deliberate ensure_ascii asymmetry, or the idempotency guard. One minor remains: a commit-message inaccuracy introduced by the revision itself, disclosed below.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** `commit ec95f067` — The rename-only commit ec95f067 carries the subject 'revise(cogni-projects): preserve file mode across atomic writes, document the write contract', but its entire content is a single R100 pure rename. The work it claims lives in the following commit 9a13eb7e, whose subject is near-identical. Cause, disclosed for the record: the revising session's `git add` listed the pre-rename pathspec, which git rejected, aborting the whole add — so only the already-staged rename was committed, and the substantive changes landed in a second commit. Anyone reading `git log --oneline` sees two adjacent commits claiming the same change, one of which does none of it. No functional impact: the rename is recorded as R100 and git-blame follows it.
  - *Suggested fix:* If this PR is squash-merged the issue self-resolves and no action is needed. If merged with history preserved, reword ec95f067 to 'revise(cogni-projects): rename test file to test_register_entity.sh' before merge.

### Positive observations

- Permission fix verified empirically and independently of the test suite's own assertions: fresh portfolio-init under umask 022 yields 0644 on all four seeded JSON files; under umask 077 all four yield 0600 — so _intended_mode tracks the umask rather than hardcoding 0644.
- Regression proof: the validator checked out the pre-fix writer from 486691c9 and reproduced 0600 on all four files, confirming the reported regression was real and that this revision is what closes it.
- Operator-set modes survive the upsert path — projects-portfolio.json set to 0640 and execution-log.json to 0600 both came back unchanged after a re-register, so the os.stat branch correctly takes precedence over the umask branch for existing targets.
- The new mode assertions were confirmed to fail against the unfixed writer (600 != 644, 600 != 640) and pass with it — a genuine regression gate, not a test that would pass either way.
- No regressions on any watch item: zero .tmp debris after success, after a monkeypatched os.replace failure, and after a real OSError; manifest-written-last ordering intact; the deliberate ensure_ascii asymmetry (True on seed logs, False on the manifest) was NOT homogenized, with the umlaut round-trip confirming literal storage; idempotency guard intact.
- os.fchmod is applied to the temp fd inside the with-block rather than to the target after the swap, so the file is never observable at 0600 at its final path — and because it lands after json.dump, the temp stays at the more restrictive 0600 for the duration of the write itself.
- Stdlib-only holds across all three touched files; both Python files pass py_compile and both shell files pass bash -n.
- Test file is 0755 on disk and 100755 in the git index; a repo-wide grep for the old hyphenated filename returns nothing.
- The module docstring now states the temp+os.replace contract, the staging order, both named failure envelopes, and the mode-preservation rationale — closing the documentation-completeness finding.
- CLAUDE.md's architecture tree carries the tests/ entry with its bash invocation.
- Version discipline correct: plugin.json 0.0.6 mirrored in root marketplace.json, still in the Incubating band.
- The waived fsync finding was correctly left untouched, and the deferred envelope finding was filed as tracked follow-up issue #1128 rather than dropped.

### Recommendation


---

claude-cogni-work:
## Revision posted

Commit `9a13eb7e` (plus rename-only `ec95f067`). Fresh verdict: **pass**.

### Findings resolved

- **[major] `review-contract:item-6` — file permission regression.** Added an `_intended_mode(path)` helper to both `register-entity.py` and `portfolio-init.sh`: an existing target keeps its own mode (`os.stat`), a new one gets `0o666 & ~umask`. Applied via `os.fchmod` on the temp fd *before* `os.replace`, so the file is never observable at 0600 at its final path. Verified empirically — pre-fix reproduces 0600, post-fix yields 0644 under umask 022 and 0600 under umask 077. `9a13eb7e`
- **[minor] `review-contract:item-15` — docstring completeness.** Supplemented with the per-file atomicity contract, the both-temps-staged-before-either-replace ordering, and the two named failure envelopes. The existing "two writes, log first" sentence is retained, as you noted it remains accurate. `9a13eb7e`
- **[minor] test filename + mode.** `git mv` to `test_register_entity.sh` (rename tracked as R100), `chmod +x`, and the file's own line-13 usage self-reference updated. It was the only in-repo reference to the old name. `ec95f067` / `9a13eb7e`
- **[minor] `CLAUDE.md` architecture tree.** `tests/` entry added with the bash invocation. `9a13eb7e`

### Not addressed, by agreement

- **[minor] fsync durability** — waived per the review contract's `risk_watchlist`, which pre-committed that durability beyond atomicity is not required and its absence must not block. Left untouched deliberately.
- **[minor] `portfolio-init.sh` OSError escapes the JSON envelope** — pre-existing rather than a regression from this PR, and you scoped it as a follow-up. Filed as #1128 rather than dropped.

### New regression coverage

Five assertions added to the suite (21/21 now pass). They were checked against the *unfixed* writer and genuinely fail there (`600 != 644`, `600 != 640`), so they gate the regression rather than passing either way. They also cover the upsert path preserving an operator-set mode.

### One thing I got wrong

The first commit `ec95f067` claims the mode/docstring work in its subject line but contains only the rename — my `git add` listed the pre-rename pathspec, git rejected it, and that aborted the whole add, so only the already-staged rename landed. The substantive changes are in `9a13eb7e`. Harmless if this is squash-merged; if you preserve history, `ec95f067` is worth a reword.

### Merge note

This PR is at version `0.0.6`; sibling PR #1127 pre-reserved `0.0.7` off the same base. Whichever lands second will conflict in `plugin.json` and `marketplace.json` and needs manual resolution.

claude-cogni-work:
## Revision cycle — no code change warranted

`service-ship` dispatched one `service-revise` cycle against the single un-waived
finding on the latest verdict. `lead-designer` (revise mode) returned an empty
plan with an explicit rationale rather than a no-op commit.

**Finding:** `[minor] commit ec95f067` — the rename-only commit's subject
overclaims work that actually lives in `9a13eb7e`.

**Disposition: waived.** Two independent grounds:

1. The finding's own *Suggested fix* states the issue self-resolves under
   squash-merge and no action is needed. This repo squash-merges every PR
   (recent `main`: `7ed4bb7a`, `b014fec9`, `4751ad64`, all carrying the squashed
   `(#NNNN)` suffix), so the misleading intermediate commit is collapsed into a
   single accurate subject and never lands on `main`.
2. The only alternative remedy — rewording `ec95f067` — requires a history
   rewrite plus force-push. `service-revise` is contractually forbidden from
   both, so it was not an action available to this cycle.

The finding is cosmetic and branch-local, severity `minor`, with no functional
impact (the rename is recorded `R100`; `git-blame` follows it).

### One real fix applied (no commit)

The revise pass did surface a genuine inaccuracy: this PR's **body** still
referenced the pre-rename path `cogni-projects/tests/test-register-entity.sh`
in the Summary and Test plan, while the branch file is
`test_register_entity.sh`. Verified `git grep test-register-entity` over the
branch returns **zero** hits — every in-repo reference (`CLAUDE.md:24`,
`CLAUDE.md:32`, the script's own usage line) was already correct. The stale
path existed only in PR metadata, so it was corrected via `gh pr edit` with no
commit and no branch mutation.

**Branch unchanged. Verdict `pass` and CI green still stand.**

claude-cogni-work:
## Ship-ready — your turn to merge

`service-ship` drove the PR to mergeable state:

- Verdict: ✅ pass
- CI: ✅ green (3/3 checks)
- Review threads: 0 unresolved
- Base: up to date (behind by 0)
- Mergeable: `MERGEABLE` / `CLEAN`
- Draft: flipped to ready

The one remaining `minor` finding (commit `ec95f067`'s subject overclaiming
work that lives in `9a13eb7e`) is **waived** — see the revision-cycle comment
above for the full justification. In short: it self-resolves under this repo's
squash-merge policy, and the only alternative remedy requires a force-push that
`service-revise` is contractually forbidden from performing.

Remaining step: click **Merge** (squash).

> Note: this repo runs single-account — the runner is the PR author, so
> self-approval and self-merge are unavailable. The merge is a human gate by
> construction, not by omission.

Actions taken this invocation: `dispatch_revise` (no-op, waived), `pr_body_fix`
(metadata only, no commit), `flip_ready`.

PR #1125, opened 2026-07-17
cogni-projects: staffing engine — ranked shortlists per open role

claude-cogni-work:
Closes #1114.

Phase 2 (staffing MVP) of cogni-projects: a deterministic staffing engine that consumes the Phase 1 consultant/project/assignment entities to produce ranked candidate shortlists per open project role.

## Summary
- Add `scripts/staffing-score.py` (python3 stdlib, `{success,data,error}` envelope, exit 0/1/2): for each project with `open_roles`, ranks candidate consultants per role on three sub-scores in [0,1] shown separately plus a combined score — availability (consultant `available_from`/`available_until` overlap with the project window + `allocation_pct` headroom), profile fit (`skills`/`seniority` vs the role need), and strategic impact (normalized project `strategic_impact` 1-5). Excludes consultants whose availability window does not overlap the project window (AC2). Deterministic: stable sort with a slug tiebreak and no wall-clock in scores (AC3).
- Add `skills/projects-staff/SKILL.md` (domain-prefixed, satisfies check-skill-names.sh): locates the portfolio, runs `staffing-score.py`, writes a `staffing-recommendations.md` artifact (ranked candidates per open role with the three sub-scores) plus JSON into the portfolio, appends a run record to `.metadata/staffing-log.json`, and summarizes.
- Reuse `validate-entities.py`'s `parse_frontmatter` via `importlib.util.spec_from_file_location` (same pattern as `register-entity.py`) — no duplicated parser, no PyYAML.
- Register the new skill + script in `cogni-projects/CLAUDE.md` (developer-guide tree; drop the staffing engine from the "Planned" section).
- Bump `.claude-plugin/plugin.json` 0.0.4 -> 0.0.5 and mirror the version in the root `.claude-plugin/marketplace.json` cogni-projects entry (stays Incubating).

## Scope decisions
- **In:** the two artifacts triage named (skill + scoring script), the CLAUDE.md registration, and the mandatory version-bump + marketplace mirror. All three acceptance criteria are delivered end-to-end, so the issue is fully closed (`Closes`).
- **Out (separate filed children, not deferred here):** #1115 (backfilling recommender) builds on this engine; the partner-meeting dashboard is a later child.

## Test plan (verified locally against a fixture portfolio)
- [x] `staffing-score.py` with no args exits 2 with the `{success,data,error}` envelope; a bad portfolio dir exits 1.
[...]

PR #1127, opened 2026-07-17
cogni-projects: partner-meeting portfolio dashboard (health + value)

### LINKED ISSUE #1116: cogni-projects: partner-meeting portfolio dashboard (health + value)
## Plugin
cogni-projects

## Epic context
Phase 3 (portfolio steering — deferred) of "Roadmap: cogni-projects — partner project-portfolio steering (staffing MVP first)". Deferred reporting layer over the Phase 1 data model; independent of the staffing engine and must not block the MVP.

## Use case
Partners review the project portfolio in partner meetings and need a single readable snapshot of portfolio health (staffing coverage, at-risk projects) and value (strategic impact, utilization signals) rather than reading raw entity files.

## Proposed solution
Add a domain-prefixed `projects-dashboard` skill and a `scripts/render-dashboard.py` that read the portfolio entities and render a partner-meeting dashboard (markdown and/or HTML) summarizing: staffing coverage per project (roles filled vs open), project health/risk flags, and portfolio value by strategic impact. AC: (1) the dashboard lists every project with fill status and a health flag; (2) an aggregate portfolio value/impact summary is shown; (3) generation degrades gracefully to a partial snapshot when some entities are missing fields, never hard-failing. ~1 skill + 1 render script.

## Alternatives considered
Embedding portfolio value pulled live from cogni-portfolio — deferred to the bridges child; the dashboard reads self-contained impact scores for now. A live web app — out of scope; a static generated artifact fits the ecosystem.

_Affected components: cogni-projects/skills/projects-dashboard/SKILL.md, cogni-projects/scripts/render-dashboard.py_

_Complexity hint: medium_

## Depends on
- #1113 cogni-projects: define self-contained data model + entity-authoring skill

claude-cogni-work:
Closes #1116.

## Summary
- Add a domain-prefixed **`projects-dashboard`** skill that renders a partner-meeting snapshot of a cogni-projects portfolio (find portfolio → run renderer → open the HTML).
- Add stdlib-only **`scripts/render-dashboard.py`** (read-only): enumerates entities via `projects-portfolio.json`, then opens and parses every `projects/`, `consultants/`, `assignments/` markdown frontmatter — reusing `validate-entities.py`'s `parse_frontmatter` and `_entity_files` via the importlib file-path idiom `register-entity.py` already uses, so the dashboard reads exactly the shape the validator enforces instead of re-implementing a parser. Derives per-project staffing coverage (`open_roles` filled vs open) with a health flag, aggregates portfolio value by `strategic_impact` (1–5), and writes a self-contained, `html.escape()`d `output/dashboard.html`. Prints the `{success,data,error}` envelope with `data.path`; a top-level `__main__` try/except keeps a malformed entity from escaping as a traceback.
- **Graceful degradation (AC-3):** a missing field or a role-label mismatch (`erp-lead` vs `ERP Lead`) yields a partial snapshot with a surfaced warning and `success: true` — never a hard failure.
- Add **`tests/test-render-dashboard.sh`** (38 assertions): AC-1 (every project listed with fill status + health flag), AC-2 (value aggregate present), AC-3 (partial snapshot + warnings on a missing field / label mismatch), plus a manifest-missing clean-failure case, a read-only/idempotent re-render check, and an HTML-injection escaping check.
- Update `CLAUDE.md` (dashboard moved from "Planned" to present; skill/script/tests added to the architecture tree) and bump `cogni-projects` 0.0.5 → **0.0.8**.

## Acceptance criteria
- [x] **AC-1** — the dashboard lists every project with fill status and a health flag.
- [x] **AC-2** — an aggregate portfolio value/impact summary (projects grouped by `strategic_impact`) is shown.
- [x] **AC-3** — generation degrades gracefully to a partial snapshot when entities miss fields, never hard-failing.

## Scope decisions
- **HTML-only output** (the issue allows "markdown and/or HTML"), matching the house dashboards. Markdown dual-output, live cogni-portfolio value enrichment, and richer utilization heuristics are deliberately **out of scope** — separate concerns, not partial slices, so this issue closes complete.
- **Read-only by contract:** the renderer never writes `projects-portfolio.json` (`register-entity.py` is its sole writer) and never appends to `.metadata/`.
- **Copied the envelope-conformant `consult-dashboard` pattern**, not `cogni-portfolio/portfolio-dashboard`, which does not follow the `{success,data,error}` contract.
- **Theming is optional:** a `DEFAULT_THEME` is embedded so the MVP needs no `pick-theme` step; `--design-variables <path.json>` is an optional override.

## ⚠️ Version-line conflict expected with #1126
This PR bumps `cogni-projects` to **0.0.8** to sequence past the already-open sibling PR #1126 (which bumps 0.0.5 → 0.0.6). The `version` lines in `cogni-projects/.claude-plugin/plugin.json` and the cogni-projects entry of `.claude-plugin/marketplace.json` will **textually conflict** at merge time — whichever PR merges second should keep this PR's 0.0.8 on top of #1126's 0.0.6.

## Test plan
- [x] `bash cogni-projects/tests/test-render-dashboard.sh` → all 38 assertions pass.
- [x] `python3 cogni-projects/scripts/render-dashboard.py <portfolio-dir>` against a seeded portfolio → valid `{success:true}` envelope with `data.path` at `<portfolio-dir>/output/dashboard.html`; the HTML lists each project with roles-filled-vs-open, a health flag, and the strategic-impact value table.
- [x] AC-3 fixture: a project missing `strategic_impact` plus a role-label mismatch → `success:true`, `data.partial:true`, warnings surfaced in the envelope and rendered on the page.
- [x] `scripts/check-breadcrumbs.py` passes (exit 0); no breadcrumbs in the new skill/script.
- [ ] `cogni-workspace/scripts/check-skill-names.sh` — could not run locally (the script uses `declare -A`, which needs bash ≥ 4; the local shell is macOS bash 3.2). The name `projects-dashboard` clears the guard's exact-match check on the bare word `dashboard`, matching the existing `portfolio-dashboard` / `consult-dashboard` precedent; CI runs it on Linux.

## Note
The test simulates failure paths in-process (missing manifest, malformed fields, injection) rather than depending on host/OS state, so it is deterministic across platforms.

---

## Revision (automated review follow-up)

Addressed all 10 findings from the automated review verdict. Key behavioral changes:

- **Undeclared `open_roles` no longer renders green.** Absence is normalized once at read time (`_normalize_open_roles`) into a single unknown state covering every authoring shape — key absent, `open_roles:` with no value, explicit null — and surfaces a warning plus a distinct `staffing unknown` flag. An explicit empty list stays a real answer. *The bare-scalar shape was an escape the first fix missed and the review pass caught.*
- **One non-UTF-8 entity file no longer aborts the render.** `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so it escaped the per-file guard and discarded every valid project.
- **Usage errors now print the envelope.** `argparse` raises `SystemExit` (a `BaseException`), which the top-level handler never caught, so `python3 render-dashboard.py` printed nothing to stdout. Message derived from the parser, not restated.
- **Theme values are validated** before interpolation into the `<style>` block; entity values already were.
- `avg_allocation` and `fully_allocated` are surfaced in the envelope so the documented utilization figures are reachable.
- SKILL.md documents the health labels the renderer actually emits, the `success:false` branch, and a cross-platform open step.

Test suite grew 23 → 38 assertions. The new fixtures cover genuinely malformed entities (type-invalid values, non-UTF-8 bytes, both undeclared-roles shapes) rather than merely incomplete ones — the prior fixtures were all well-formed, which is why the decode abort shipped against a green suite. Verified the new fixture reproduces the original defect when the fix is reverted.

Deferred: the same unguarded theme interpolation exists in six sibling renderers — filed as #1129 rather than expanded into this PR.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -272,7 +272,7 @@
       "source": "./cogni-projects",
       "version": "0.0.8",
       "maturity": "incubating",
-      "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
+      "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory, authors the validated consultant, project, and assignment records, and renders a read-only partner-meeting dashboard over them, so later skills (staffing match, backfilling) build on the same portfolio.",
       "keywords": [
         "projects",
         "portfolio",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -270,7 +270,7 @@
     {
       "name": "cogni-projects",
       "source": "./cogni-projects",
-      "version": "0.0.7",
+      "version": "0.0.8",
       "maturity": "incubating",
       "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
       "keywords": [
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -270,7 +270,7 @@
     {
       "name": "cogni-projects",
       "source": "./cogni-projects",
-      "version": "0.0.5",
+      "version": "0.0.7",
       "maturity": "incubating",
       "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
       "keywords": [
--- a/cogni-projects/.claude-plugin/plugin.json
+++ b/cogni-projects/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "cogni-projects",
-  "version": "0.0.7",
+  "version": "0.0.8",
   "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
   "author": {
     "name": "Stephan de Haas",
--- a/cogni-projects/.claude-plugin/plugin.json
+++ b/cogni-projects/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "cogni-projects",
-  "version": "0.0.5",
+  "version": "0.0.7",
   "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
   "author": {
     "name": "Stephan de Haas",
--- a/cogni-projects/.claude-plugin/plugin.json
+++ b/cogni-projects/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "cogni-projects",
   "version": "0.0.8",
-  "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory and authors the validated consultant, project, and assignment records that later skills (staffing match, backfilling, partner-meeting dashboard) read from.",
+  "description": "Partner-facing project-portfolio steering for consulting firms — models consultants, projects, and staffing so partners can match people to work by availability, profile fit, and strategic impact. Scaffolds a self-contained portfolio directory, authors the validated consultant, project, and assignment records, and renders a read-only partner-meeting dashboard over them, so later skills (staffing match, backfilling) build on the same portfolio.",
   "author": {
     "name": "Stephan de Haas",
     "email": "stephan@cogni-work.ai"
--- a/cogni-projects/CLAUDE.md
+++ b/cogni-projects/CLAUDE.md
@@ -2,9 +2,9 @@
 
 Partner-facing project-portfolio steering for consulting firms. Models
 consultants, projects, and staffing so partners can match people to work by
-availability, profile fit, and strategic impact. The data model and entity
-authoring have landed; the staffing engine and the skills that read these
-entities arrive in later roadmap children.
+availability, profile fit, and strategic impact. The data model, entity
+authoring, and the read-only portfolio dashboard have landed; the staffing
+engine and backfilling recommender arrive in later roadmap children.
 
 ## Plugin Architecture
 
--- a/cogni-projects/CLAUDE.md
+++ b/cogni-projects/CLAUDE.md
@@ -15,19 +15,22 @@ cogni-projects/
 ├── CLAUDE.md                         This developer guide
 ├── skills/
 │   ├── projects-setup/SKILL.md       Initialize a portfolio directory (entry point)
-│   └── projects-entities/SKILL.md    Author + register one consultant/project/assignment
+│   ├── projects-entities/SKILL.md    Author + register one consultant/project/assignment
+│   └── projects-dashboard/SKILL.md   Render a partner-meeting portfolio dashboard (read-only)
 ├── scripts/
 │   ├── portfolio-init.sh             Idempotent portfolio scaffolder (stdlib-only)
 │   ├── validate-entities.py          Entity frontmatter validator (stdlib-only)
-│   └── register-entity.py            Slug-keyed manifest upsert + execution-log append
+│   ├── register-entity.py            Slug-keyed manifest upsert + execution-log append
+│   └── render-dashboard.py           Portfolio health + value HTML render (read-only, stdlib-only)
+├── tests/
+│   └── test-render-dashboard.sh      Dashboard render regression test (stdlib-only)
 └── references/
     └── data-model.md                 Consultant / project / assignment entity schemas
 ```
 
 Planned (not yet scaffolded — see the roadmap epic):
 
-- `skills/` for the staffing match engine, backfilling recommender, and the
-  partner-meeting dashboard.
+- `skills/` for the staffing match engine and backfilling recommender.
 
 ## Data Model
 
--- a/cogni-projects/README.md
+++ b/cogni-projects/README.md
@@ -2,7 +2,7 @@
 
 > **Incubating** (v0.0.x) — skills, data formats, and workflows may change at any time.
 
-Partner project-portfolio steering for consulting firms on Claude Code. cogni-projects gives partners one place to model consultants, projects, and staffing — so they can match the right people to the right work by availability, profile fit, and strategic impact. This foundation release scaffolds the plugin and its portfolio entry point; the staffing engine and dashboards arrive in later releases.
+Partner project-portfolio steering for consulting firms on Claude Code. cogni-projects gives partners one place to model consultants, projects, and staffing — so they can match the right people to the right work by availability, profile fit, and strategic impact. The data model, entity authoring, and a read-only partner-meeting dashboard have landed; the staffing match engine and backfilling recommender arrive in later releases.
 
 ## Why this exists
 
@@ -18,13 +18,15 @@ Consulting partners steer a portfolio of projects and people with spreadsheets a
 
 cogni-projects is a Claude Code plugin that holds a **self-contained project portfolio**: a directory of consultants, projects, and assignments rooted by a single manifest. It is the 14th plugin in the insight-wave ecosystem and the home for partner-facing portfolio steering.
 
-At this foundation stage it provides the plugin skeleton and the `projects-setup` skill that scaffolds a portfolio directory. Staffing, backfilling, and partner-meeting views build on top of it.
+It provides the `projects-setup` skill that scaffolds a portfolio directory, the `projects-entities` skill that authors consultant/project/assignment records, and the read-only `projects-dashboard` skill that renders a partner-meeting portfolio snapshot. Staffing match and backfilling build on top of these.
 
 ## What it does
 
 - **projects-setup** — initialize a new portfolio directory (`cogni-projects/<portfolio-slug>/`) with a root manifest and metadata logs. Idempotent: re-running never overwrites an existing portfolio.
+- **projects-entities** — author and register consultant, project, and assignment records into the portfolio — the entities later staffing scores over.
+- **projects-dashboard** — render a read-only partner-meeting snapshot of the portfolio: staffing coverage per project, at-risk projects, and portfolio value by strategic impact.
 
-_More skills (staffing match engine, backfilling recommender, partner-meeting dashboard) land in later roadmap releases._
+_More skills (staffing match engine, backfilling recommender) land in later roadmap releases._
 
 ## What it means for you
 
@@ -76,11 +78,16 @@ The manifest holds portfolio identity (`slug`, `name`, `language`, timestamps) p
 | Type | Name | Purpose |
 |------|------|---------|
 | Skill | `projects-setup` | Initialize a portfolio directory |
+| Skill | `projects-entities` | Author and register consultant/project/assignment records |
+| Skill | `projects-dashboard` | Render a read-only partner-meeting portfolio snapshot |
 | Script | `portfolio-init.sh` | Idempotent portfolio scaffolder |
+| Script | `register-entity.py` | Atomic entity register into the portfolio manifest |
+| Script | `validate-entities.py` | Validate entity records against the data model |
+| Script | `render-dashboard.py` | Render the portfolio dashboard HTML from entity data |
 
 ## Architecture
 
-cogni-projects is standalone at this stage. Its portfolio directory is designed as the shared substrate for later skills (staffing match, backfilling, dashboard) and for planned cross-plugin bridges to cogni-consult and cogni-portfolio.
+cogni-projects is standalone at this stage. Its portfolio directory is designed as the shared substrate for later skills (staffing match, backfilling) and for planned cross-plugin bridges to cogni-consult and cogni-portfolio.
 
 ## Dependencies
 
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -1,0 +1,427 @@
+#!/usr/bin/env python3
+"""Render a partner-meeting portfolio dashboard from a cogni-projects portfolio.
+
+Reads projects-portfolio.json to confirm the portfolio, then opens and parses
+every consultant / project / assignment entity's frontmatter to compute:
+  - per-project staffing coverage (open_roles filled vs still open) + a health flag,
+  - an aggregate portfolio value summary grouped by strategic_impact (1..5),
+  - a simple consultant-utilization aggregate.
+Writes a self-contained HTML dashboard to <portfolio-dir>/output/dashboard.html.
+
+Read-only: never writes projects-portfolio.json (register-entity.py is its only
+writer) and never appends to .metadata/. Degrades gracefully — a missing or
+malformed entity field yields a partial snapshot with a surfaced warning, never
+a hard failure (AC-3).
+
+Stdlib-only (no PyYAML). Reuses validate-entities.py's parse_frontmatter and
+_entity_files by file-path load rather than re-implementing a parser or a
+directory walk — the same idiom register-entity.py uses.
+
+Usage:
+  python3 render-dashboard.py <portfolio-dir> [--design-variables <path.json>]
+
+Output: a single JSON line following the repo contract
+  {"success": bool, "data": {...}, "error": str}
+Exit: 0 ok / 2 usage or environment failure.
+"""
+
+import argparse
+import datetime
+import html
+import importlib.util
+import json
+import os
+import sys
+
+# validate-entities.py is not an importable module name (hyphens), so load it by
+# file location and reuse its parser + entity-file discovery — the same idiom
+# register-entity.py uses, so the dashboard reads exactly the shape the
+# validator enforces rather than duplicating (and drifting from) those rules.
+_v_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "validate-entities.py")
+_spec = importlib.util.spec_from_file_location("validate_entities", _v_path)
+if _spec is None or _spec.loader is None:
+    print(json.dumps({
+        "success": False, "data": {},
+        "error": "cannot load validator module: %s" % _v_path,
+    }))
+    sys.exit(2)
+_ve = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_ve)
+
+
+# The built-in palette. The dashboard renders with this whenever no
+# --design-variables file is supplied, so the script never depends on
+# cogni-workspace:pick-theme; a themed override is a purely optional flag.
+DEFAULT_THEME = {
+    "theme_name": "cogni-work",
+    "bg": "#0f1419",
+    "surface": "#1a2028",
+    "surface2": "#232b35",
+    "text": "#e6edf3",
+    "muted": "#9aa7b4",
+    "accent": "#4f9cf9",
+    "ok": "#3fb950",
+    "warn": "#d29922",
+    "risk": "#f85149",
+    "border": "#2d333b",
+    "font": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif",
+}
+
+# The status values that count an assignment as actively covering a role. A
+# completed assignment no longer staffs an open role.
+ACTIVE_ASSIGNMENT_STATES = ("planned", "active")
+
+
+def _load_theme(path):
+    """Return the theme dict. Falls back to DEFAULT_THEME on any read problem.
+
+    Theming is a nicety, never a hard dependency: a missing or malformed
+    --design-variables file must not fail the render, only fall back.
+    """
+    if not path:
+        return dict(DEFAULT_THEME), None
+    try:
+        with open(path, "r", encoding="utf-8") as f:
+            overrides = json.load(f)
+    except (OSError, ValueError) as exc:
+        return dict(DEFAULT_THEME), "design-variables ignored (%s): %s" % (path, exc)
+    theme = dict(DEFAULT_THEME)
+    if isinstance(overrides, dict):
+        # Accept both a flat {key: value} map and a nested {"colors": {...}} one.
+        for src in (overrides, overrides.get("colors", {})):
+            if isinstance(src, dict):
+                for key, value in src.items():
+                    if key in theme and isinstance(value, str):
+                        theme[key] = value
+    return theme, None
+
+
+def _read_entities(portfolio_dir):
+    """Parse every entity file into type-keyed lists, collecting warnings.
+
+    Returns (entities, warnings) where entities is
+    {"consultant": [...], "project": [...], "assignment": [...]} of frontmatter
+    dicts. A file that cannot be read or has no frontmatter is recorded as a
+    warning and skipped — the rest of the portfolio still renders (AC-3).
+    """
+    entities = {"consultant": [], "project": [], "assignment": []}
+    warnings = []
+    for path in _ve._entity_files(portfolio_dir):
+        rel = os.path.relpath(path, portfolio_dir)
+        try:
+            with open(path, "r", encoding="utf-8") as f:
+                fm = _ve.parse_frontmatter(f.read())
+        except OSError as exc:
+            warnings.append("cannot read %s: %s" % (rel, exc))
+            continue
+        if not fm:
+            warnings.append("no frontmatter in %s — skipped" % rel)
+            continue
+        etype = fm.get("type")
+        if etype not in entities:
+            warnings.append("unknown entity type %r in %s — skipped" % (etype, rel))
+            continue
+        fm["_file"] = rel
+        entities[etype].append(fm)
+    return entities, warnings
+
+
+def _coerce_impact(value, project_label, warnings):
+    """Return strategic_impact as an int 1..5, or None with a warning."""
+    if value is None:
+        warnings.append("%s has no strategic_impact — omitted from the value summary" % project_label)
+        return None
+    try:
+        impact = int(str(value).strip())
+    except (TypeError, ValueError):
+        warnings.append("%s has a non-numeric strategic_impact %r — omitted from the value summary" % (project_label, value))
+        return None
+    if not 1 <= impact <= 5:
+        warnings.append("%s strategic_impact %d is outside 1..5 — omitted from the value summary" % (project_label, impact))
+        return None
+    return impact
+
+
+def _project_health(filled, total, status):
+    """Map (roles filled, roles total, project status) to (label, severity)."""
+    if status == "closed":
+        return ("closed", "muted")
+    if total == 0:
+        return ("no open roles", "ok")
+    if filled >= total:
+        return ("fully staffed", "ok")
+    if filled == 0 and status == "active":
+        return ("unstaffed", "risk")
+    return ("%d/%d roles open" % (total - filled, total), "warn")
+
+
+def _compute(entities, warnings):
+    """Derive per-project staffing + the portfolio value/utilization aggregates.
+
+    Fill status is derived, not stored: for a project's open_roles list, a role
+    is "filled" when some planned/active assignment for that project carries a
+    matching role label. Role labels are free strings, so a label an assignment
+    names that no open_roles entry matches is surfaced as a warning rather than
+    silently changing the counts (AC-3).
+    """
+    assignments = entities["assignment"]
+    projects_out = []
+    value_by_impact = {i: 0 for i in range(1, 6)}
+
+    for proj in sorted(entities["project"], key=lambda p: str(p.get("name") or p.get("slug") or "")):
+        slug = proj.get("slug")
+        label = "project %s" % (proj.get("name") or slug or "(unnamed)")
+        open_roles = proj.get("open_roles") or []
+        if not isinstance(open_roles, list):
+            open_roles = [open_roles]
+        status = (proj.get("status") or "").strip()
+
+        covered = set()
+        for a in assignments:
+            if a.get("project") == slug and (a.get("status") or "").strip() in ACTIVE_ASSIGNMENT_STATES:
+                role = a.get("role")
+                if role:
+                    covered.add(role)
+        # An assignment role that matches no listed open role may be a label
+        # mismatch (erp-lead vs "ERP Lead") — flag it, don't hard-fail.
+        for role in sorted(covered):
+            if open_roles and role not in open_roles:
+                warnings.append(
+                    "%s: assignment role %r matches no open_roles label — possible label mismatch" % (label, role)
+                )
+        filled = [r for r in open_roles if r in covered]
+        health_label, health_sev = _project_health(len(filled), len(open_roles), status)
+
+        impact = _coerce_impact(proj.get("strategic_impact"), label, warnings)
+        if impact is not None:
+            value_by_impact[impact] += 1
+
+        projects_out.append({
+            "name": proj.get("name") or slug or "(unnamed)",
+            "client": proj.get("client") or "",
+            "status": status or "unknown",
+            "impact": impact,
+            "roles_total": len(open_roles),
+            "roles_filled": len(filled),
+            "open_roles": [r for r in open_roles if r not in covered],
+            "health_label": health_label,
+            "health_sev": health_sev,
+        })
+
+    # Utilization: a simple average of consultant allocation_pct, plus a count of
+    # consultants at or above 100%. Richer heuristics are out of scope.
+    allocations = []
+    for c in entities["consultant"]:
+        raw = c.get("allocation_pct")
+        if raw is None:
+            continue
+        try:
+            allocations.append(int(str(raw).strip()))
+        except (TypeError, ValueError):
+            warnings.append("consultant %s has a non-numeric allocation_pct %r — omitted from utilization" % (c.get("slug") or "(unknown)", raw))
+    util = {
+        "consultants": len(entities["consultant"]),
+        "with_allocation": len(allocations),
+        "avg_allocation": round(sum(allocations) / len(allocations)) if allocations else None,
+        "fully_allocated": sum(1 for a in allocations if a >= 100),
+    }
+
+    return projects_out, value_by_impact, util
+
+
+def _esc(value):
+    return html.escape(str(value), quote=True)
+
+
+def _render_html(portfolio, projects, value_by_impact, util, warnings, theme):
+    t = theme
+    generated = datetime.date.today().isoformat()
+    sev_color = {"ok": t["ok"], "warn": t["warn"], "risk": t["risk"], "muted": t["muted"]}
+
+    rows = []
+    for p in projects:
+        impact = "—" if p["impact"] is None else ("★" * p["impact"])
+        coverage = "%d/%d" % (p["roles_filled"], p["roles_total"]) if p["roles_total"] else "—"
+        open_roles = ", ".join(_esc(r) for r in p["open_roles"]) if p["open_roles"] else "—"
+        rows.append(
+            "<tr>"
+            "<td><strong>%s</strong><div class='sub'>%s</div></td>"
+            "<td>%s</td>"
+            "<td>%s</td>"
+            "<td>%s</td>"
+            "<td>%s</td>"
+            "<td><span class='flag' style='color:%s'>%s</span></td>"
+            "</tr>" % (
+                _esc(p["name"]), _esc(p["client"]),
+                _esc(p["status"]), impact, coverage, open_roles,
+                sev_color.get(p["health_sev"], t["text"]), _esc(p["health_label"]),
+            )
+        )
+
+    value_rows = []
+    for impact in range(5, 0, -1):
+        count = value_by_impact.get(impact, 0)
+        bar = "█" * count
+        value_rows.append(
+            "<tr><td>%s</td><td>%d</td><td class='bar'>%s</td></tr>"
+            % ("★" * impact, count, bar)
+        )
+
+    warn_block = ""
+    if warnings:
+        items = "".join("<li>%s</li>" % _esc(w) for w in warnings)
+        warn_block = (
+            "<section class='card warnings'><h2>Warnings — partial snapshot "
+            "(%d)</h2><ul>%s</ul></section>" % (len(warnings), items)
+        )
+
+    avg = "—" if util["avg_allocation"] is None else "%d%%" % util["avg_allocation"]
+    total_open = sum(p["roles_total"] - p["roles_filled"] for p in projects)
+
+    return """<!DOCTYPE html>
+<html lang="en"><head><meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>Portfolio dashboard — {name}</title>
+<style>
+  :root {{ color-scheme: dark; }}
+  * {{ box-sizing: border-box; }}
+  body {{ margin: 0; background: {bg}; color: {text}; font-family: {font}; line-height: 1.5; }}
+  .wrap {{ max-width: 1040px; margin: 0 auto; padding: 32px 20px 64px; }}
+  header h1 {{ margin: 0 0 4px; font-size: 1.6rem; }}
+  header .meta {{ color: {muted}; font-size: 0.9rem; }}
+  .tiles {{ display: flex; flex-wrap: wrap; gap: 14px; margin: 22px 0; }}
+  .tile {{ flex: 1 1 150px; background: {surface}; border: 1px solid {border};
+           border-radius: 10px; padding: 16px; }}
+  .tile .n {{ font-size: 1.8rem; font-weight: 700; }}
+  .tile .l {{ color: {muted}; font-size: 0.82rem; text-transform: uppercase; letter-spacing: 0.04em; }}
+  .card {{ background: {surface}; border: 1px solid {border}; border-radius: 10px;
+           padding: 18px 20px; margin: 18px 0; overflow-x: auto; }}
+  .card h2 {{ margin: 0 0 12px; font-size: 1.05rem; }}
+  table {{ width: 100%; border-collapse: collapse; font-size: 0.92rem; }}
+  th, td {{ text-align: left; padding: 9px 10px; border-bottom: 1px solid {border}; vertical-align: top; }}
+  th {{ color: {muted}; font-weight: 600; font-size: 0.78rem; text-transform: uppercase; letter-spacing: 0.03em; }}
+  td .sub {{ color: {muted}; font-size: 0.8rem; }}
+  .flag {{ font-weight: 600; }}
+  .bar {{ color: {accent}; letter-spacing: 1px; }}
+  .warnings {{ border-color: {warn}; }}
+  .warnings h2 {{ color: {warn}; }}
+  .warnings li {{ color: {muted}; font-size: 0.88rem; }}
+  footer {{ color: {muted}; font-size: 0.8rem; margin-top: 26px; }}
+</style></head>
+<body><div class="wrap">
+<header>
+  <h1>{name}</h1>
+  <div class="meta">Partner-meeting portfolio dashboard · generated {generated}</div>
+</header>
+<div class="tiles">
+  <div class="tile"><div class="n">{n_projects}</div><div class="l">Projects</div></div>
+  <div class="tile"><div class="n">{n_consultants}</div><div class="l">Consultants</div></div>
+  <div class="tile"><div class="n">{open_roles}</div><div class="l">Open roles</div></div>
+  <div class="tile"><div class="n">{avg}</div><div class="l">Avg allocation</div></div>
+</div>
+<section class="card">
+  <h2>Projects — staffing &amp; health</h2>
+  <table>
+    <thead><tr><th>Project</th><th>Status</th><th>Impact</th><th>Roles filled</th><th>Still open</th><th>Health</th></tr></thead>
+    <tbody>{rows}</tbody>
+  </table>
+</section>
+<section class="card">
+  <h2>Portfolio value — projects by strategic impact</h2>
+  <table>
+    <thead><tr><th>Impact</th><th>Projects</th><th></th></tr></thead>
+    <tbody>{value_rows}</tbody>
+  </table>
+</section>
+{warn_block}
+<footer>cogni-projects · read-only render · {generated}</footer>
+</div></body></html>
+""".format(
+        name=_esc(portfolio.get("name") or portfolio.get("slug") or "Portfolio"),
+        generated=generated,
+        bg=t["bg"], surface=t["surface"], surface2=t["surface2"], text=t["text"],
+        muted=t["muted"], accent=t["accent"], border=t["border"],
+        ok=t["ok"], warn=t["warn"], risk=t["risk"], font=t["font"],
+        n_projects=len(projects),
+        n_consultants=util["consultants"],
+        open_roles=total_open,
+        avg=avg,
+        rows="".join(rows) or "<tr><td colspan='6'>No projects yet.</td></tr>",
+        value_rows="".join(value_rows),
+        warn_block=warn_block,
+    )
+
+
+def _fail(message, code=2):
+    print(json.dumps({"success": False, "data": {}, "error": message}, ensure_ascii=False))
+    return code
+
+
+def main(argv):
+    ap = argparse.ArgumentParser(
+        description="Render a cogni-projects partner-meeting portfolio dashboard.",
+    )
+    ap.add_argument("portfolio_dir")
+    ap.add_argument("--design-variables", dest="design_variables", default=None)
+    args = ap.parse_args(argv)
+
+    portfolio_dir = os.path.abspath(args.portfolio_dir)
+    if not os.path.isdir(portfolio_dir):
+        return _fail("portfolio directory not found: %s" % portfolio_dir)
+
+    manifest_path = os.path.join(portfolio_dir, "projects-portfolio.json")
+    if not os.path.isfile(manifest_path):
+        return _fail(
+            "portfolio manifest not found: %s (run /cogni-projects:projects-setup first)"
+            % manifest_path
+        )
+
+    warnings = []
+    try:
+        with open(manifest_path, "r", encoding="utf-8") as f:
+            portfolio = json.load(f)
+    except (OSError, ValueError) as exc:
+        return _fail("cannot read portfolio manifest: %s" % exc)
+
+    theme, theme_warning = _load_theme(args.design_variables)
+    if theme_warning:
+        warnings.append(theme_warning)
+
+    entities, read_warnings = _read_entities(portfolio_dir)
+    warnings.extend(read_warnings)
+    projects, value_by_impact, util = _compute(entities, warnings)
+
+    out_dir = os.path.join(portfolio_dir, "output")
+    try:
+        os.makedirs(out_dir, exist_ok=True)
+        out_path = os.path.join(out_dir, "dashboard.html")
+        with open(out_path, "w", encoding="utf-8") as f:
+            f.write(_render_html(portfolio, projects, value_by_impact, util, warnings, theme))
+    except OSError as exc:
+        return _fail("cannot write dashboard: %s" % exc)
+
+    print(json.dumps({
+        "success": True,
+        "data": {
+            "path": out_path,
+            "portfolio": portfolio.get("slug") or "",
+            "projects": len(projects),
+            "consultants": util["consultants"],
+            "open_roles": sum(p["roles_total"] - p["roles_filled"] for p in projects),
+            "warnings": warnings,
+            "partial": bool(warnings),
+        },
+        "error": "",
+    }, ensure_ascii=False))
+    return 0
+
+
+if __name__ == "__main__":
+    # Same envelope contract as the sibling scripts: a hand-edited manifest or
+    # entity of the wrong shape must report as a readable failure, never a
+    # traceback that breaks the {success,data,error} contract mid-render.
+    try:
+        _code = main(sys.argv[1:])
+    except Exception as _exc:  # noqa: BLE001 — deliberate catch-all
+        _code = _fail("unexpected failure: %s: %s" % (type(_exc).__name__, _exc))
+    sys.exit(_code)
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -11,7 +11,8 @@ Writes a self-contained HTML dashboard to <portfolio-dir>/output/dashboard.html.
 Read-only: never writes projects-portfolio.json (register-entity.py is its only
 writer) and never appends to .metadata/. Degrades gracefully — a missing or
 malformed entity field yields a partial snapshot with a surfaced warning, never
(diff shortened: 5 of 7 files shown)


sdh07:

_The merger formed this expectation from issue #1116 before checking out the PR head — the review grades the diff against it._

**Expected approach:** Add `cogni-projects/skills/projects-dashboard/SKILL.md` (domain-prefixed) plus a stdlib-only `cogni-projects/scripts/render-dashboard.py` that reads a portfolio rooted at `projects-portfolio.json` and renders a partner-meeting snapshot. The renderer should reuse `parse_frontmatter` and `_entity_files` from `validate-entities.py` via the importlib file-path idiom already established in `register-entity.py`, rather than re-implementing a second frontmatter parser. It computes per-project fill status by joining `assignments/*.md` (`project`, `role`) against each project's `open_roles`, derives a documented deterministic health flag, and rolls `strategic_impact` (int 1–5) into an aggregate value summary. AC-3 is the load-bearing behavior: like `validate-entities.py`'s `__main__` catch-all, every path — missing manifest, absent frontmatter, missing optional field, unreadable/non-UTF-8 record, non-integer impact — must return the `{success,data,error}` envelope with a partial snapshot and never raise. HTML output must pass every interpolated entity value through `html.escape` (house style: cogni-consult's `generate-dashboard.py`, also explicitly read-only), the renderer must not mutate the manifest or `.metadata/`, and the plugin version must be bumped above base and mirrored byte-identically in `marketplace.json`.

**Acceptance checklist:**
- [ ] `skills/projects-dashboard/SKILL.md` exists with valid YAML frontmatter, `name: projects-dashboard` matching its directory, and a trigger-phrase `description`.
- [ ] The skill name is domain-prefixed (`projects-dashboard`, not bare `dashboard`) so `check-skill-names.sh` passes.
- [ ] `scripts/render-dashboard.py` exists and its import block contains only stdlib modules (no pip dependency).
- [ ] The new SKILL.md contains no maintainer breadcrumbs (`#NNN` refs, version tags, slice codes) — `check-breadcrumbs.py` stays clean.
- [ ] `plugin.json` `version` bumped above base 0.0.5, and the `marketplace.json` `cogni-projects` entry carries the byte-identical string.
- [ ] `CLAUDE.md`'s Plugin Architecture tree lists both new files, and any "Planned (not yet scaffolded)" note no longer claims the dashboard is unscaffolded.
- [ ] `tests/test-render-dashboard.sh` runs standalone, builds fixtures in a `mktemp` dir with a cleanup trap, and exits non-zero on any assertion failure.
- [ ] `render-dashboard.py` prints exactly one `{success,data,error}` JSON line on EVERY path — including no-args usage, a nonexistent path, and a directory lacking `projects-portfolio.json` — via a `_fail`-style helper plus a top-level `try/except Exception` under `__main__`.
- [ ] The renderer does NOT duplicate frontmatter parsing: it loads `validate-entities.py` by file path via `importlib.util.spec_from_file_location` and calls its `parse_frontmatter` / `_entity_files`.
- [ ] The renderer is read-only: no write to `projects-portfolio.json`, none under `.metadata/`, no `os.remove`/`shutil`/rename against the portfolio. The only file written is the documented dashboard artifact.
- [ ] AC-1 structure: the output includes one entry per project under `projects/`, each carrying (a) fill status expressing roles filled vs open and (b) a health flag.
- [ ] AC-1 join correctness: fill status is computed by matching assignment `project` against the project slug AND `role` against the project's `open_roles` — not by counting assignments alone.
- [ ] The health-flag rule is deterministic and documented with its thresholds stated, so two runs on the same portfolio produce the same flag.
- [ ] AC-2: the output contains an aggregate portfolio value/impact summary derived from `strategic_impact`.
- [ ] AC-3 (partial snapshot): a project missing an optional field still appears with an explicit unknown/placeholder marker, and the run returns success with the degradation surfaced (e.g. `data.warnings[]`) rather than silently dropped.
- [ ] AC-3 (never hard-fails): unreadable, non-UTF-8, frontmatter-less, or type-invalid records (e.g. `strategic_impact: high`) produce no traceback — per-file parsing is wrapped so one bad record cannot abort the scan.
- [ ] The test exercises AC-3 directly: at least one fixture with a missing field and one with a malformed entity, asserting no traceback and a parseable envelope.
- [ ] HTML output escapes untrusted entity content: every interpolated entity-derived value (project name, client, role labels, consultant names) passes through `html.escape`.
- [ ] The renderer does not import, exec, or depend on any Phase-2 staffing artifact (e.g. `staffing-score.py`), which does not exist on base.
- [ ] SKILL.md invokes the script with the `${CLAUDE_PLUGIN_ROOT:-...}` resolution idiom, and the output path it documents matches what the script actually writes.
- [ ] SKILL.md stays under the skill length cap and contains no invented cross-plugin dependency.

**Risk watchlist:**
- 771 changed lines against the 400-line reviewable bound: confirm the excess is presentation boilerplate (inline CSS/theme) and not unreviewed logic, and that no second frontmatter parser is hiding in it.
- "Graceful degradation" implemented as a blanket `except: pass` around the whole render, converting real bugs into silently empty dashboards — degradation must be per-record and must surface what degraded.
- The inverse failure: hard-failing with a traceback on a portfolio whose entity dirs are empty or whose manifest is missing, violating AC-3.
- HTML injection via a client or project name containing markup — a real risk, since these are partner-authored free-text fields rendered into a browser artifact.
- A dashboard that mutates portfolio state (bumping manifest `updated`, appending to the execution log) — reporting must never be a writer; `register-entity.py` is the only documented manifest writer.
- **Semantic ambiguity in fill status:** a project with no `open_roles` key could mean "fully staffed" or "unknown". Treating unknown as 100% filled would make an unstaffed project look green — the most consequential silent-wrong-answer failure mode here.
- Health flag defined by an undocumented or LLM-judgment heuristic rather than a deterministic rule, making partner-meeting numbers unreproducible.
- Version collision: a sibling cogni-projects PR may bump concurrently — check `plugin.json` and `marketplace.json` as a pair, not against a hardcoded target.
- Scope creep: pulling live cogni-portfolio value (deferred to the bridges child) or depending on the unmerged staffing engine.
- Judgment-only, not a gate: `cogni-projects/README.md` is already stale (lists only projects-setup) and is not in the diff. Whether the dashboard belongs there too is a cogni-docs follow-up call for a human, not a blocking item.

**Note on issue ambiguity (resolved, not escalated):** the issue says "markdown and/or HTML" and never defines what a "health flag" means or its thresholds. Resolved by requiring whichever format is produced to be HTML-escaped if HTML, and the health rule to be deterministic and documented — either output format satisfies the AC.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1127

**Summary:** The implementation is genuinely good — the renderer is stdlib-only, state-read-only, consistently HTML-escaped, and the 23-assertion suite really passes when executed (verified, not taken on faith). What needs work is the SKILL.md's contract with its own script and the version lineage. Three skill-vs-script drifts are confirmed against the tree: the documented health-flag vocabulary includes a 'partially staffed' label the script never emits, the blanket 'any missing field warns' claim hides two silent-undercount paths, and Step 3 opens the HTML with no success:false gate so a failed render opens a stale dashboard. Separately, the 0.0.5 -> 0.0.7 jump pre-reserves 0.0.6 for an unmerged sibling PR and a simulated merge produces a real conflict in both version files.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (14)

- **[major]** `cogni-projects/skills/projects-dashboard/SKILL.md:32` — Health-flag drift, CONFIRMED against the tree. SKILL.md documents four flags: 'fully staffed / partially staffed / unstaffed / closed'. _project_health (render-dashboard.py:145-155) actually emits five: 'closed', 'no open roles', 'fully staffed', 'unstaffed', and '%d/%d roles open'. There is no 'partially staffed' label anywhere — the partial case renders as e.g. '2/3 roles open'. 'no open roles' is undocumented. Also unstated: 'unstaffed' is reachable only when status == 'active', so a prospective project with zero coverage renders as 'N/M roles open', not unstaffed.
  - *Suggested fix:* Replace the parenthetical with the five real labels and their conditions: closed (status closed) / no open roles (total == 0) / fully staffed (filled >= total) / 'N/M roles open' when partly covered / unstaffed when an *active* project has zero coverage.

- **[major]** `cogni-projects/skills/projects-dashboard/SKILL.md:38-40` — Overbroad degradation claim, CONFIRMED. SKILL.md says 'Any missing or malformed field produces a partial snapshot with a warning, never a hard failure.' Two paths silently undercount with NO warning. (1) render-dashboard.py:181 counts only assignments whose status is in ACTIVE_ASSIGNMENT_STATES via (a.get('status') or '').strip() — status is an optional field per references/data-model.md:168, so an assignment authored without it is excluded from coverage silently and the project reads as less staffed than it is. (2) A project with no open_roles falls to total == 0 -> 'no open roles'/ok, also unwarned. In a partner meeting an under-reported project looks like real staffing risk.
  - *Suggested fix:* Narrow the claim to what the script actually warns on (missing/non-numeric/out-of-range strategic_impact, unmatched role labels, non-numeric allocation_pct, unreadable or frontmatter-less files, unknown type), and add: 'an assignment counts toward coverage only when its status is planned or active — an assignment with no status is not counted and is not warned about, so author status on every assignment.'

- **[major]** `cogni-projects/skills/projects-dashboard/SKILL.md:58-70` — No failure path, CONFIRMED. render-dashboard.py returns success:false / exit 2 for a missing directory (:370), missing manifest (:373-377), unreadable manifest (:384), and unwritable output dir (:401). SKILL.md documents only the success shape and then Step 3 unconditionally runs `open .../output/dashboard.html`, which on any failure path opens a STALE dashboard from a previous run — the worst outcome for a partner meeting, since it looks like a successful render of current data. The sibling projects-entities skill does gate on success in its Step 4, so this is also an inconsistency within the plugin.
  - *Suggested fix:* In Step 2 add: 'If success is false, stop and relay error — a missing manifest means the directory is not an initialized portfolio; run /cogni-projects:projects-setup first. Only proceed to Step 3 when success is true.'

- **[major]** `cogni-projects/.claude-plugin/plugin.json:3` — Version lineage is order-dependent and conflict-guaranteed, CONFIRMED by simulated merge. origin/main is at 0.0.5. This branch forks from 0.0.5 and jumps straight to 0.0.7, pre-reserving 0.0.6 for the unmerged sibling PR #1126. Simulating the merge: after #1126 lands, merging this branch produces CONFLICT (content) in BOTH cogni-projects/.claude-plugin/plugin.json and .claude-plugin/marketplace.json, since both PRs rewrite the same version line from the same base. If #1126 is dropped or reordered, 0.0.6 is skipped and the published version history has a gap.
  - *Suggested fix:* Do not pre-reserve a version for an unmerged sibling. Set this PR to 0.0.6 and rebase whichever PR lands second, bumping to the next free patch at rebase time. Either way resolve manually before merge — do not rely on auto-merge.

- **[major]** `cogni-projects/.claude-plugin/plugin.json:4` — The manifest description this PR edits still reads '...later skills (staffing match, backfilling, partner-meeting dashboard) read from' — describing the partner-meeting dashboard as unshipped in the very PR that ships it. The identical stale sentence is mirrored verbatim into .claude-plugin/marketplace.json:275, so the marketplace listing understates the plugin too.
  - *Suggested fix:* Move the dashboard from the future-work clause into the shipped capability list, and mirror the corrected text into the marketplace.json entry in the same commit (both files are already touched here).

- **[major]** `cogni-projects/README.md:5,21,25,27,78` — README is not in the changed-file set and is now contradicted in three places: line 5 'the staffing engine and dashboards arrive in later releases', line 21 'At this foundation stage it provides the plugin skeleton and the projects-setup skill', line 27 'More skills (... partner-meeting dashboard) land in later roadmap releases'. The 'What it does' (25) and 'Components' (78) tables list only projects-setup, omitting projects-dashboard and the already-merged projects-entities. Per the repo's 16-section convention, sections 4 and 11 are auto-generated from SKILL.md. NOTE: the identical finding was raised on sibling PR #1125 — this is the second skill to land without a README update.
  - *Suggested fix:* Run cogni-docs doc-generate on cogni-projects (or hand-edit sections 4 and 11-13) and delete the three now-false 'later releases' claims. Worth coordinating with #1125 so one of the two closes the gap rather than both conflicting.

- **[minor]** `cogni-projects/tests/test-render-dashboard.sh:130` — Assertion count is 23, not the 24 claimed in the PR body. Verified by execution: 10 + 3 + 6 + 2 + 2 = 23 OK lines, 0 failures, exit 0. Scenario coverage is complete; only the count is overstated.
  - *Suggested fix:* Correct the PR body to 23, or add the missing assertion — a natural gap is asserting data.path points at output/dashboard.html, which no current assertion checks.

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:80-82` — Theming note omits that a bad theme file marks the whole snapshot partial. _load_theme (render-dashboard.py:86-87) turns an unreadable/malformed --design-variables file into a warning, and partial is bool(warnings) (:412) — so a typo'd theme path makes data.partial true, which SKILL.md frames exclusively as 'missing or mismatched records'. The agent would tell the partner the portfolio data is incomplete when only the palette was.
  - *Suggested fix:* Note that a bad --design-variables file falls back to the built-in palette and adds a warning (so the snapshot reports as partial) without affecting the data.

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:25-36` — Progressive-disclosure gap: the skill leans on open_roles, strategic_impact, allocation_pct, and assignment status semantics but never links references/data-model.md, the canonical field contract. The sibling projects-entities skill links it twice. Raised independently by two perspectives.
  - *Suggested fix:* Add under Core concept: 'Field semantics are canonical in [../../references/data-model.md](../../references/data-model.md).'

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:55,69` — Unquoted <portfolio-dir> placeholder in both bash blocks (the sibling skill quotes its path arguments), so a slug or parent path containing a space word-splits and the script receives a truncated path. Step 3 also hardcodes macOS `open`, which fails on Linux and reads to the user as a render failure. Raised independently by two perspectives.
  - *Suggested fix:* Quote the placeholder: ... render-dashboard.py" "<portfolio-dir>". Note `open` is macOS and suggest xdg-open on Linux, or just report the absolute data.path — the file is self-contained, so surfacing the path is a complete result.

- **[minor]** `cogni-projects/scripts/render-dashboard.py:59` — Theme key 'surface2' is defined in DEFAULT_THEME, accepted from --design-variables overrides, and passed into the format() call at line 342, but the CSS template never interpolates {surface2}. Dead configuration surface: overriding it has no effect.
  - *Suggested fix:* Either use surface2 in the template (e.g. table header or tile background) or drop it from DEFAULT_THEME and the format() call.

- **[minor]** `cogni-projects/scripts/render-dashboard.py:109` — The renderer couples to a private, underscore-prefixed function in a sibling script: _ve._entity_files(portfolio_dir). parse_frontmatter is public API, but _entity_files signals internal-only by convention, so a refactor of validate-entities.py can silently break the dashboard with no signal until runtime.
  - *Suggested fix:* Rename _entity_files to entity_files in validate-entities.py and update both call sites, making the reuse contract explicit.

- **[minor]** `cogni-projects/references/data-model.md:218-219` — Cross-doc drift in the same tree: data-model.md says 'the manifest ref is the index the dashboard and staffing skills scan without opening every file'. render-dashboard.py reads the manifest only for portfolio name/slug (:381-384, :340) and otherwise walks and parses every entity file via _ve._entity_files (:109). A portfolio whose entity files exist but were never registered still renders fully.
  - *Suggested fix:* Correct the sentence to reflect that the dashboard parses entity files directly (the manifest is identity + registration index), so the next skill built off this reference does not inherit the wrong assumption.

- **[minor]** (waived) `.github/workflows/lint.yml` — PRE-EXISTING, repo-wide: tests/test-render-dashboard.sh is never executed by CI. The Lint workflow has only two jobs (Maintainer-breadcrumb guard, External-dispatch guard) and no test runner; no workflow references any tests/ path. The suite only protects the code if someone runs it locally. Sibling PR #1126's test-register-entity.sh has the same gap.
  - *Justification:* Pre-existing repo-infrastructure gap not introduced by this PR; annotated for awareness, must not block this merge.

### Positive observations

- Test suite genuinely passes when actually executed — 23 OK, 0 failures, exit 0 — verified by running it rather than trusting the PR body. (plugin-validator, re-confirmed by the orchestrator)
- Stdlib-only verified by AST import extraction: argparse, datetime, html, importlib, json, os, sys — zero third-party imports. (plugin-validator)
- Read-only verified by write-target audit: the only mutating calls are os.makedirs(out_dir) and open(out_path,'w') for output/dashboard.html; projects-portfolio.json is opened exclusively 'r' and .metadata/ is never touched. Test 2c independently asserts the manifest is unchanged after a render. (plugin-validator)
- HTML escaping is applied consistently — _esc() wraps every entity-derived value including nested open_roles items and warning text, using html.escape(quote=True). A <script> payload in a project name provably does not reach the output raw. (plugin-validator)
- AC-3 graceful degradation is implemented with real care: missing strategic_impact, non-numeric allocation_pct, out-of-range impact, unreadable files, missing frontmatter, unknown entity types, and role-label mismatches each produce a surfaced warning and a partial snapshot, with a module-level catch-all guaranteeing the envelope even on an unexpected traceback. (plugin-validator)
- Description is exemplary: third person, concrete trigger phrases ('portfolio health', 'staffing coverage', 'which projects are at risk') plus an explicit catch-all for users who never say 'dashboard'. (skill-reviewer, skill-quality-reviewer)
- The envelope claims that matter most are accurate — data.path is the written file, and data.partial / data.warnings behave exactly as described. (skill-reviewer)
- Read-only framing is enforced end to end: allowed-tools omits Write, the note correctly states register-entity.py is the sole manifest writer, and the test asserts it. (skill-reviewer)
- Body explains the why, not just the how — role labels are free strings so mismatches warn rather than mis-count, and partial snapshots are framed as expected mid-authoring. (skill-reviewer)
- Lean and well-scoped at 83 lines / ~600 words with a clean three-step workflow, and the skill orchestrates rather than reimplements — no rendering, grouping, or HTML logic leaks into the body. (skill-quality-reviewer)
- Domain-prefixed skill name verified against the check-skill-names.sh convention (reimplemented in Python since macOS ships bash 3.2 and that script needs bash 4+); all three skill directory names match their frontmatter, no duplicate bare names across the 14-plugin repo. (plugin-validator)
- CLAUDE.md was updated in the same PR to register both the new skill and the new script in the architecture tree, keeping the developer guide in sync. (plugin-validator)
- Maintainer-breadcrumb guard passes clean — AC references live only in scripts and test comments, never in SKILL.md. (plugin-validator, skill-quality-reviewer)

### Recommendation


---

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1127
**Perspectives consulted:** cogni-service:contract-grading-reviewer, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, plugin-dev:plugin-validator

**Summary:** Well-built feature — read-only by construction, validator reuse, 23-assertion suite green, all repo guards clean, version pair deliberately skipping 0.0.6 to dodge the sibling collision. But grading the diff by EXECUTION rather than reading found five gating defects the shipped tests do not catch: a project missing the optional open_roles field renders green and warning-free (indistinguishable from fully staffed — the silent-wrong-answer risk the committed contract named as highest-value); one non-UTF-8 entity file aborts the entire scan, violating AC-3's never-hard-fail; the no-args path prints no JSON envelope at all; the AC-3 test fixture never exercises a genuinely malformed entity, which is why that regression shipped undetected; and the documented health-flag vocabulary does not match the labels the code emits.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (10)

- **[major]** (auto-fixable) `review-contract:item-15` — HIGHEST-VALUE FINDING — the exact ambiguity the committed contract's risk watchlist named in advance as the most consequential silent-wrong-answer failure mode. A project entity that legitimately omits the optional open_roles field (open_roles IS optional for type=project per validate-entities.py's schema) is surfaced with NO unknown/placeholder marker and NO warning: `open_roles = proj.get("open_roles") or []` collapses 'field absent' into the same total==0 branch as 'explicitly zero roles needed', yielding health_label='no open roles', health_sev='ok' (#3fb950 — the SAME GREEN as fully staffed), warnings=[] and partial=False. Reproduced by direct execution against a fixture project with strategic_impact:4, status:active and no open_roles key: it rendered fully green with zero warnings. On a partner-meeting decision surface, an unstaffed project is displayed as healthy.
  - *Suggested fix:* Distinguish `'open_roles' not in proj` from an explicit empty list. When the key is absent, append a warning ('<project> has no open_roles — fill status unknown') and render a distinct 'unknown' health label/severity rather than reusing the fully-staffed-equivalent 'ok' branch.

- **[major]** (auto-fixable) `review-contract:item-16` — AC-3 VIOLATION (never hard-failing / one bad record cannot abort the scan). A single non-UTF-8 entity file aborts the ENTIRE render rather than degrading per-file: _read_entities' per-file guard catches only OSError, but opening a non-UTF-8 file with encoding='utf-8' raises UnicodeDecodeError — a ValueError subclass, not an OSError — which propagates past main() to the top-level except Exception, producing {success:false, data:{}} and discarding every other valid project in the portfolio. Reproduced with a portfolio of two valid projects plus one non-UTF-8 file: the whole render failed instead of returning a partial snapshot with one warning.
  - *Suggested fix:* Widen the per-file except in _read_entities to `except (OSError, UnicodeDecodeError, ValueError)` (or read binary and decode with errors='replace'), so one unreadable/non-UTF-8 file is recorded as a warning and skipped — matching the handling OSError already gets.

- **[major]** (auto-fixable) `review-contract:item-8` — The {success,data,error} envelope is NOT printed on every path. argparse.parse_args() raises SystemExit(2) on a missing required argument, and SystemExit inherits BaseException — not Exception — so the top-level `except Exception` in __main__ never catches it. Verified by direct execution: `python3 render-dashboard.py` with no args produces EMPTY stdout, argparse usage text on stderr, exit 2. The repo convention requires the envelope on every path, and a caller parsing stdout gets nothing to parse.
  - *Suggested fix:* Wrap args.parse_args(argv) in try/except SystemExit inside main() (or override ArgumentParser.error) so a usage error also emits the {success:false,...} envelope before exiting.

- **[major]** (auto-fixable) `review-contract:item-17` — The AC-3 test fixture does not exercise a genuinely malformed entity. Fixture 3 in test-render-dashboard.sh covers only a missing optional field (strategic_impact) and a role-label mismatch — both well-typed, well-formed data. There is no fixture for a type-invalid value (strategic_impact: high), non-UTF-8 bytes, or frontmatter-less content mixed among valid records. This gap is precisely why the item-16 non-UTF-8 hard-fail shipped with a green 23-assertion suite: the suite tests the degradation paths the author anticipated, not the ones that break.
  - *Suggested fix:* Add a fixture mixing a valid project with a type-invalid strategic_impact and/or a non-UTF-8 entity file, asserting the envelope still returns success:true, partial:true, and no traceback on stderr.

- **[major]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:32` — CONTRACT ITEM 13 — the health-flag rule must be deterministic AND documented with thresholds stated. Determinism holds (_project_health is a pure function of filled/total/status). But the documented vocabulary does not match what the code emits: SKILL.md lists 'fully staffed / partially staffed / unstaffed / closed', while render-dashboard.py:145-155 emits 'closed', 'no open roles', 'fully staffed', 'unstaffed', and the literal '<n>/<m> roles open'. So 'partially staffed' is never produced; the 'no open roles' branch is undocumented; and 'unstaffed' fires only when filled==0 AND status=='active', so a PLANNED project with zero coverage shows '<n>/<m> roles open', not 'unstaffed'. MERGER NOTE ON A PERSPECTIVE SPLIT: contract-grading-reviewer graded item 13 satisfied, but both skill perspectives independently cited the exact emitted-vs-documented label strings, and the item-15 finding above independently confirms 'no open roles' is a real, undocumented, and consequential label. The merger sides with the two skill perspectives here.
  - *Suggested fix:* State the real rule and labels: 'a health flag derived deterministically from (roles filled, roles listed, project status): closed when status is closed; no open roles when the project lists none; fully staffed when every listed role is covered; unstaffed when an active project has zero coverage; otherwise <n>/<m> roles open. A role counts as filled only when a planned or active assignment for that project names a matching label.' Fold in the item-15 'unknown' state once that is fixed.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:50-62` — Step 2 documents only the success path. The script returns success:false with exit 2 on a missing portfolio directory, a missing projects-portfolio.json, or an unwritable output dir — the one case that is not a graceful partial — and the skill gives no instruction for that branch, so the model will improvise. The sibling projects-setup SKILL.md documents both envelope branches. (The hard-fail boundary itself is CORRECT and not an AC-3 violation: AC-3 governs entity-level degradation; a missing portfolio is environment-level.)
  - *Suggested fix:* Add: 'When success is false the render did not run at all — read error. A missing projects-portfolio.json means the directory is not an initialized portfolio; point the user at projects-setup rather than retrying.'

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:66-70` — Step 3 uses the macOS-only `open` command with no fallback, while the rest of the plugin is platform-neutral. On Linux the command does not exist and the step fails opaquely at the end of an otherwise successful run.
  - *Suggested fix:* Report data.path and offer opening as a convenience — `open` on macOS, `xdg-open` on Linux — noting a failure here is cosmetic since the dashboard is already written.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:35` — 'Utilization: a simple average of consultant allocation_pct' under-specifies a number partners read off the dashboard. Consultants lacking allocation_pct are silently excluded from the average rather than counted as zero — a sensible choice, but invisible. The tile row also surfaces a fully_allocated count the body never mentions. Separately, data.partial is set by ANY warning including a merely-unreadable --design-variables file, so a cosmetic theming miss marks the whole snapshot partial.
  - *Suggested fix:* Note that consultants without allocation_pct are excluded rather than zero-counted (so a thinly authored portfolio is not made to look under-allocated), and that partial can be set by a non-substantive warning.

- **[minor]** (auto-fixable) `cogni-projects/scripts/render-dashboard.py:341-344` — Theme values from a --design-variables JSON are interpolated raw into the <style> block, while every entity-derived value correctly goes through _esc(). _load_theme allowlists the KEY and requires isinstance(value, str) but does not constrain the VALUE, so a crafted file with e.g. "bg": "#000</style><script>...</script>" breaks out of the stylesheet into the self-contained HTML. Low reachability — the file is local and operator-supplied, not portfolio-derived — hence minor.
  - *Suggested fix:* Route theme values through _esc() at format time, or validate in _load_theme against a conservative pattern and fall back to DEFAULT_THEME on mismatch.

- **[minor]** (auto-fixable) `cogni-projects/scripts/render-dashboard.py:14,105,166` — The script docstring and comments carry literal '(AC-3)' acceptance-criterion codes — a finding-code breadcrumb of the family the repo's maintainer-breadcrumb guard exists to prevent. check-breadcrumbs.py scopes to SKILL.md and agent files so it does NOT trip (verified: 0 violations), but the convention's rationale — state the reason semantically, leave provenance to git history — applies to scripts too.
  - *Suggested fix:* Replace the '(AC-3)' markers with the semantic rationale, e.g. 'a partial snapshot is preferable to a hard failure mid-authoring'.

### Positive observations

- Read-only guarantee enforced in three independent layers: source (a single write to output/dashboard.html, zero destructive calls, .metadata never referenced), documentation (SKILL.md states it twice), and a behavioral regression test (2c manifest untouched after re-render)
- The importlib file-location load of validate-entities.py reuses the validator's own parser and entity discovery instead of duplicating frontmatter rules, so the dashboard cannot drift from the shape the validator enforces — exactly what the contract asked for
- Version pair byte-identical at 0.0.7 across plugin.json and marketplace.json, deliberately skipping 0.0.6 to avoid collision with the concurrent sibling PR — the pair-check the contract asked for, with no drift; maturity: incubating correctly retained
- AC-1 join correctness confirmed: fill status matches assignment.project against the project slug AND assignment.role against open_roles — not a bare assignment count (the shortcut the contract warned about was avoided)
- HTML escaping of entity-derived content is genuinely correct — every interpolated entity value passes through _esc(), confirmed by both code read and a live <script>-in-project-name fixture
- 23-assertion regression suite ships with the feature and passes green, with mktemp fixtures and a cleanup trap; above the bar for an incubating plugin even though its AC-3 coverage has the gap noted above
- allowed-tools omits Write and Edit, making the read-only contract structurally enforced rather than merely asserted
- Description carries eight concrete trigger phrases plus an explicit anti-under-trigger clause; skill name correctly domain-prefixed and matching its directory
- All repo guards pass clean: check-breadcrumbs.py 0 violations, check-skill-names.sh OK
- Body is 69 lines against a 500-line cap, with detail correctly delegated to the script rather than inlined
- CLAUDE.md architecture tree updated to list all three new files, and the 'Planned (not yet scaffolded)' note correctly edited to drop the dashboard clause
- Theming is optional-by-construction with a built-in palette, so the script carries no runtime dependency on cogni-workspace:pick-theme
- No dependency on the unmerged Phase-2 staffing engine — the scope-creep risk the contract flagged was avoided

### Recommendation

⚠️ **Revise before merge.** 10 finding(s) need attention.

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1127

**Summary:** The revision genuinely closed all 10 prior findings — independently re-verified by the validator's adversarial probes and by the quality reviewer against the diff — and two of three perspectives now pass cleanly. The merged verdict is needs_revision because plugin-validator's own verdict is needs_revision, driven by one newly-surfaced major finding: the README still describes the partner-meeting dashboard as a future roadmap item and omits it from the Components table, contradicting the deliverable this PR ships. The remaining seven findings are minor, newly surfaced (not re-opened), and every finding across all three perspectives is marked auto-fixable. Nothing in the feature logic is flagged.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (8)

- **[major]** (auto-fixable) `cogni-projects/README.md:27` — README line 27 states 'More skills (staffing match engine, backfilling recommender, partner-meeting dashboard) land in later roadmap releases.' and line 83 repeats 'later skills (staffing match, backfilling, dashboard)' — both are falsified by this PR, which ships projects-dashboard. The 'What it does' section (lines 23-27) and the Components table (lines 74-79) also list only projects-setup / portfolio-init.sh, omitting projects-dashboard, render-dashboard.py, and tests/. Repo CLAUDE.md designates README sections 4 and 11-13 as auto-generated from SKILL.md, so this is a convention violation.
  - *Suggested fix:* Run cogni-docs doc-generate against cogni-projects to regenerate sections 4 and 11-13, or hand-edit: add a projects-dashboard bullet under 'What it does', add Skill projects-dashboard + Script render-dashboard.py rows to the Components table, and drop 'partner-meeting dashboard'/'dashboard' from the forward-looking lists on lines 27 and 83. Fold in the missing projects-entities entries while there.

- **[minor]** (auto-fixable) `cogni-projects/CLAUDE.md:5` — The intro paragraph still reads 'the staffing engine and the skills that read these entities arrive in later roadmap children' — projects-dashboard is precisely a skill that reads these entities and it landed in this PR. The file's architecture tree and script list were correctly updated in this commit, so the stale sentence contradicts the tree three lines below it.
  - *Suggested fix:* Reword to: 'the data model, entity authoring, and the read-only portfolio dashboard have landed; the staffing engine and backfilling recommender arrive in later roadmap children.'

- **[minor]** (auto-fixable) `cogni-projects/scripts/render-dashboard.py:59` — The surface2 theme key is defined in DEFAULT_THEME, accepted and validated by _load_theme, and passed to _render_html's .format() call, but no {surface2} placeholder exists anywhere in the CSS block. It is dead configuration surface an operator could set with no effect.
  - *Suggested fix:* Either use surface2 in the stylesheet or remove it from DEFAULT_THEME and the format kwargs.

- **[minor]** (auto-fixable) `cogni-projects/scripts/render-dashboard.py:83` — The theme denylist blocks markup and property-injection breakouts but permits parentheses. Verified by probe that a design-variables value of url(https://evil.example/track.png) is interpolated verbatim into background: url(...), turning the self-contained HTML dashboard into one that phones home on open. The threat model is documented as operator-supplied, so this is not a breakout — but a partner-meeting artifact that silently fetches a remote beacon is a different failure class than the one the guard reasons about.
  - *Suggested fix:* Add ( and ) to _THEME_VALUE_FORBIDDEN, or validate color keys against a token pattern and only allow the freer character set for the font key. Update the docstring rationale to name external-resource fetches alongside markup injection.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:42` — The '<n>/<m> roles open' row does not say what n and m are. The script emits (total - filled, total) — i.e. n is the count of roles still open, not roles filled. A reader or downstream agent can plausibly read it the other way, which inverts the meaning on a partner decision surface.
  - *Suggested fix:* Change the Flag cell to '<open>/<total> roles open' and note the leading number is the count still open, not the count filled.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:46-50` — The utilization bullet covers only the 'no allocation_pct' exclusion. The script also excludes a non-numeric allocation_pct (with its own warning), and sets avg_allocation to null when no consultant has a usable value — so the field the skill names can legitimately be null, which the SKILL.md does not say. An agent that formats data.avg_allocation unconditionally will print 'None%'.
  - *Suggested fix:* Extend the bullet to cover the non-numeric exclusion and state that data.avg_allocation is null when no consultant has a usable value; report it as unknown rather than 0%.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:52-53` — 'a label an assignment names that no open_roles entry matches is surfaced as a warning' is unconditional, but the script guards the warning with 'if open_roles and ...' — a project whose open_roles is absent or empty produces no mismatch warning at all. Documented behavior is slightly broader than the code.
  - *Suggested fix:* Qualify as: 'when a project lists open_roles, a label an assignment names that no entry matches is surfaced as a warning...'

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:70` — '<portfolio-dir>' is the only unquoted script argument across the plugin's four invocation snippets; a portfolio path containing a space would word-split.
  - *Suggested fix:* Quote the placeholder.

### Positive observations

- All 10 findings from the prior verdict are closed and independently re-verified — not merely claimed. The validator re-probed each degradation path adversarially rather than trusting the diff.
- AC-3 verified across a genuinely adversarial fixture — missing/non-numeric strategic_impact, absent/bare open_roles, a role-label mismatch, a non-UTF-8 entity file, and a malformed design-variables file each degrade to a surfaced warning with success staying true and no traceback on stderr. (plugin-validator)
- Envelope contract holds on every path probed — unknown flag, excess positional args, nonexistent directory, file-as-directory — all returning well-formed JSON with exit 2; the SystemExit-vs-Exception distinction is the subtle bug most implementations miss, and the --help carve-out correctly avoids dressing a success path as a failure. (plugin-validator)
- Version pair is byte-identical at 0.0.8 across plugin.json and the marketplace entry, and the description strings hash-match. (plugin-validator)
- The revision found and closed an escape the first fix missed: a bare 'open_roles:' with no value also collapsed into the healthy-green branch. Absence is now normalized once at read time so every authoring shape reaches the same unknown state. (plugin-validator, locked by test 6i)
- Read-only guarantee is asserted, not just claimed: test 2c re-reads the manifest after a double render. (plugin-validator)
- The health-flag table is an exact, ordered mirror of _project_health(), including the non-obvious distinction that an undeclared open_roles is 'staffing unknown' while an explicitly empty one is 'no open roles'. (skill-reviewer)
- Every envelope field the SKILL.md names — data.path, data.partial, data.warnings, data.avg_allocation, data.fully_allocated — is actually printed by the script. (skill-reviewer)
- Description is third-person, trigger-rich, and explicitly covers the no-'dashboard'-keyword case; body is lean, imperative, and correctly scoped to read-only with the write-ownership boundary stated. (skill-reviewer)
- Test suite grew 23 to 38 assertions, and the new malformed-entity fixture was confirmed to reproduce the original decode abort when the fix is reverted — a real regression test, not a decorative one.
- Body is 115 lines against the 500-line soft cap, with schema detail correctly offloaded to references/data-model.md; no maintainer breadcrumbs in body or description. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the 10 findings from the [previous verdict](https://github.com/cogni-work/insight-wave/pull/1127#issuecomment-5015227460) in `ec9c7557`. All 10 are resolved — none waived.

### Findings resolved

| # | Severity | Finding | How it was resolved |
|---|---|---|---|
| item-15 | major | Undeclared `open_roles` rendered green with no warning | Absence normalized once at read time (`_normalize_open_roles`) into a single unknown state, surfacing a warning and a distinct `staffing unknown` flag. An explicit empty list stays a real answer. |
| item-16 | major | Non-UTF-8 entity file aborted the whole render | Per-file guard widened to `(OSError, UnicodeDecodeError)` — named explicitly rather than widening to `ValueError`, so a future parser fault still surfaces instead of being mislabeled an unreadable file. |
| item-8 | major | No envelope on the argparse usage path | `SystemExit` caught inside `main()`; message derived from `ap.format_usage()` rather than restated. `--help` carved out so a success path is not dressed as a failure. |
| item-17 | major | AC-3 fixture never exercised a genuinely malformed entity | Fixture 6 adds a type-invalid `strategic_impact`, both undeclared-roles shapes, and a real non-UTF-8 file; fixtures 7 and 8 cover the usage envelope and theme rejection. |
| SKILL.md:32 | major | Documented health vocabulary did not match emitted labels | Replaced with a table mirroring `_project_health` label-for-label and in precedence order, including the new `staffing unknown` state. |
| SKILL.md:50-62 | minor | Only the success path documented | Both branches documented and distinguished by kind — environment-level failure vs per-entity degradation — each with a concrete next action. |
| SKILL.md:66-70 | minor | macOS-only `open` | Reports `data.path` as the deliverable, then offers `open` / `xdg-open`, noting the step is cosmetic. |
| SKILL.md:35 | minor | Utilization under-specified | Names `data.avg_allocation` and `data.fully_allocated` (both now surfaced in the envelope) and states the exclusion rule and the `partial` caveat. |
| render-dashboard.py:341-344 | minor | Theme values interpolated raw into `<style>` | `_valid_theme_value` denylist with per-key fallback to the built-in palette; fixture 8 asserts a `</style><script>` payload does not reach the output. |
| render-dashboard.py:14,105,166 | minor | `(AC-3)` breadcrumbs | Replaced with semantic rationale. |

### Worth flagging

The review pass caught an **escape the first fix missed**: a bare `open_roles:` with no value also collapsed into the healthy-green branch, since `or []` normalizes an empty scalar the same way. That is why absence is now normalized once at read time rather than tested as key-presence. Both shapes are asserted in fixture 6e.

The new fixture was verified to reproduce the original decode abort when the fix is reverted — it is a real regression test, not a decorative one.

Test suite: 23 → 38 assertions, all green. Version 0.0.8 in both manifests.

### Follow-up issues

- #1129 — the same unguarded theme interpolation exists in six sibling renderers. Filed rather than expanded into this PR.

### Not done

The fresh verdict flags README drift (the README still lists the dashboard as future work). `service-revise` revises at most once per dispatch and has used its pass here, so that finding is queued for the next `service-resume` rather than applied now.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1127

**Summary:** All four perspectives passed; the sole un-waived finding is an advisory minor (oversized diff) surfaced per Rule 5b — non-blocking.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** `PR diff` — Diff is 1075 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- Prior major finding resolved: plugin.json:4 description no longer lists the partner-meeting dashboard among 'later skills' — now credits it as shipped, with only staffing-match/backfilling forthcoming. (plugin-validator)
- plugin.json and marketplace.json descriptions mirror byte-for-byte; versions mirror at 0.0.8. (plugin-validator)
- SKILL.md utilization wording now faithful to render-dashboard.py (missing allocation_pct silently excluded; non-numeric excluded and warned). (plugin-validator)
- None of the 8 findings closed in the prior revision regressed; Issue #1116 AC behaviors intact. (plugin-validator)
- Prior major finding resolved and verified against render-dashboard.py lines 289-298: the reworded utilization bullet correctly splits the silently-excluded and excluded-and-warned cases. (skill-reviewer)
- Reword introduced no clarity/triggering regression; body ~107 lines, well under the 500 cap. (skill-reviewer)
- Grades clean against Issue #1116 AC: per-project fill status + health flag, aggregate strategic-impact value summary, and graceful partial-snapshot degradation. (skill-reviewer)
- projects-dashboard is a lean (~107-line) read-only skill; description triggers well and body explains the why behind each rule. (skill-quality-reviewer)
- All four F5-F8 accuracy fixes correctly integrated; all three issue #1116 acceptance criteria addressed. (skill-quality-reviewer)
- Body and description are breadcrumb-free. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the automated review verdict. Fresh verdict: **✅ PASS** (head `4f77fc13`).

**Verdict findings resolved (8):**
- ✅ **[major]** `README.md` / `CLAUDE.md` listed the now-shipped dashboard as forthcoming and the Components table omitted it — added `projects-entities` / `projects-dashboard` bullets + the missing skill/script rows, dropped `dashboard` from the later-skills lists. (`41b6712c`)
- ✅ **[minor]** `CLAUDE.md:5` intro contradicted the shipped dashboard — reworded. (`41b6712c`)
- ✅ **[minor]** `render-dashboard.py:59` dead `surface2` theme key — removed from `DEFAULT_THEME` and the `.format()` kwargs. (`41b6712c`)
- ✅ **[minor]** `render-dashboard.py:83` theme denylist permitted `url(...)` (external-fetch beacon) — added `(` / `)` to `_THEME_VALUE_FORBIDDEN`, with a regression test. (`41b6712c`)
- ✅ **[minor]** `SKILL.md:42` `<n>/<m>` roles-open ambiguity — relabeled `<open>/<total>` with semantics. (`41b6712c`)
- ✅ **[minor]** `SKILL.md:46-50` utilization null/non-numeric cases undocumented — documented. (`41b6712c`)
- ✅ **[minor]** `SKILL.md:52-53` role-label mismatch warning was unconditional — qualified to `open_roles`-declaring projects. (`41b6712c`)
- ✅ **[minor]** `SKILL.md:70` unquoted `<portfolio-dir>` — quoted. (`41b6712c`)

**Also fixed (surfaced re-reviewing the revision):**
- ✅ `plugin.json` + `marketplace.json` descriptions still listed the shipped dashboard among "later skills" — corrected in both (kept mirrored). (`4f77fc13`)
- ✅ `SKILL.md` utilization bullet over-claimed a warning on the *missing*-`allocation_pct` case — separated the silent (missing, optional field) vs warned (non-numeric) paths to match `render-dashboard.py`. (`4f77fc13`)

Full test suite passes (26 assertions, incl. the `url()`-beacon regression). Advisory: PR diff is 1075 lines (>400 reviewable bound) — non-blocking.

sdh07:

_The merger formed this expectation from issue #1116 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds a read-only, stdlib-only `cogni-projects/scripts/render-dashboard.py` plus a domain-prefixed `cogni-projects/skills/projects-dashboard/SKILL.md`, modeled directly on the sibling analytics script `cogni-projects/scripts/staffing-score.py` (which at :100-116 loads `parse_frontmatter` from `validate-entities.py` by file location, at :279-306 loads entities preferring manifest refs with a directory-scan fallback, at :92-97 emits the `_fail` envelope, and at :411-444 validates argv/portfolio-dir and reserves exit 2 for usage vs 1 for domain failure). The dashboard must read project entities (`strategic_impact` 1..5 and `open_roles[]` per the schema in `validate-entities.py:76-88`) and assignment entities (`role`, `project`, `status` in planned/active/completed per :89-104) to derive per-project roles-filled-vs-open (AC-1), a portfolio-level strategic-impact/value aggregate (AC-2), and must never hard-fail on missing fields — `portfolio-init.sh:23` scaffolds only `consultants/ projects/ assignments/ .metadata/`, so any output directory has to be created by the script itself, and entities with absent optional fields must degrade to a partial snapshot rather than raising (AC-3). The change must also register the new skill in `cogni-projects/CLAUDE.md` (architecture tree at :11-29, Planned bullet at :37-39 which still names the 'partner-meeting dashboard') and in `cogni-projects/README.md` (What-it-does list ~:27 and Components table ~:76-82), add a house-style bash test alongside `cogni-projects/tests/test_register_entity.sh`, and bump `cogni-projects/.claude-plugin/plugin.json` (base 0.0.7) with a mirrored bump in `.claude-plugin/marketplace.json:273`.

**Acceptance checklist:**
- [ ] `cogni-projects/skills/projects-dashboard/SKILL.md` exists and its YAML frontmatter is valid — `name: projects-dashboard` matching the directory name, plus a `description` in the third-person "This skill should be used when..." trigger style used by `projects-staff/SKILL.md:2-13`.
- [ ] The skill name carries the required domain prefix: `projects-dashboard`, not the bare generic `dashboard` (CLAUDE.md flags `dashboard` as a generic name `check-skill-names.sh` rejects unprefixed).
- [ ] `render-dashboard.py` imports only Python stdlib — no PyYAML, no pip dependency (house rule, cogni-projects/CLAUDE.md:59-60).
- [ ] `render-dashboard.py` reuses the canonical frontmatter parser by loading `validate-entities.py` via `importlib.util.spec_from_file_location` (the idiom at `staffing-score.py:100-116` / `register-entity.py:48-57`) rather than defining its own frontmatter regex — a second parser in the diff is a fail.
- [ ] Every exit path prints exactly one JSON line of the repo envelope `{"success","data","error"}` — including the no-args/wrong-arity and not-a-portfolio-directory paths (cf. `staffing-score.py:92-97`, `:411-421`).
- [ ] A top-level `__main__` guard catches `Exception` and converts it into a `success: false` envelope, so no traceback escapes (pattern at `validate-entities.py:400-413`).
- [ ] Read-only w.r.t. entity records and the manifest: no write targeting `projects-portfolio.json`, `consultants/`, `projects/`, `assignments/`; any artifact written lands under a directory the script creates itself with `os.makedirs(..., exist_ok=True)` (`portfolio-init.sh:23` scaffolds no `output/`).
- [ ] **AC-1:** the artifact enumerates EVERY non-excluded project record, each carrying both a fill status (roles filled vs open, from `open_roles[]` against assignment records) and a discrete health/risk flag computed from data, not a hardcoded constant.
- [ ] **AC-1:** fill-status derivation reads assignment `status` against the schema's actual enum `planned|active|completed` (`validate-entities.py:96-98`); keying off a value outside that enum is a fail.
- [ ] **AC-2:** a portfolio-level aggregate computed over `strategic_impact` (1..5 int, `validate-entities.py:83`) — a sum/mean/weighted rollup rendered under an identifiable aggregate heading, not merely per-project repetition.
- [ ] **AC-3:** every optional field touched (`open_roles`, `start_date`, `end_date`, `status`, `strategic_impact`, `allocation_pct`) is accessed with a guarded/defaulting read so a missing field renders a placeholder rather than raising (cf. `staffing-score.py:261-272`, `:342-344`).
- [ ] **AC-3:** a project with NO `strategic_impact` is distinguishable in output from one with `strategic_impact: 1` (em dash / "unknown", not a silent 0) — the null-vs-lowest distinction `projects-staff/SKILL.md:106-116` makes.
- [ ] If HTML is emitted, every interpolated entity-derived value (project name, client, role labels, portfolio name) passes through `html.escape()` at the interpolation site — an unescaped f-string/`%`/`.format()` of entity text into markup is a fail (the defect class issue #1129 was filed for).
- [ ] `cogni-projects/tests/` gains a renderer test in the house shape of `test_register_entity.sh`: `set -u`, `mktemp -d` fixture with `trap ... EXIT`, pass/fail counter, non-zero exit on any assertion failure — no pytest, no pip.
- [ ] The test asserts all three ACs against a real invocation: builds a fixture portfolio (manifest + ≥1 project + ≥1 assignment), runs the renderer, and asserts (a) envelope parses `success: true`, (b) artifact contains fill status + health flag per project, (c) a fixture with fields deliberately omitted still returns `success: true` with a partial artifact rather than a traceback.
- [ ] `cogni-projects/CLAUDE.md` updated on both sides: architecture tree (:11-29) gains the skill, script and tests entry, AND the Planned bullet (:37-39) plus intro (:3-7) no longer list the partner-meeting dashboard as unbuilt — while still listing the backfilling recommender as planned.
- [ ] `cogni-projects/README.md` gains the dashboard in both auto-generated spots: the "What it does" bullet list (~:27) and the Components table (~:76-82); no other section silently rewritten.
- [ ] `plugin.json` `version` bumped above base 0.0.7 and the `cogni-projects` entry in `.claude-plugin/marketplace.json` (:271-275) set to the EXACT same string; marketplace.json stays valid JSON and the bump stays inside 0.0.x so `maturity: incubating` remains correct.
- [ ] No maintainer breadcrumbs in the new SKILL.md or any agent/skill file — no `#NNNN` refs, version tags, or milestone/slice codes (breadcrumb guard ratchets on newly introduced ones).
- [ ] The new SKILL.md respects the length cap and stays advisory-not-duplicative: it does not restate the renderer's aggregation formulas as a second source of truth but points at the script as authoritative (the discipline in `projects-staff/SKILL.md:163-167`).

**Risk watchlist:**
- A renderer that silently SKIPS projects missing a field, so AC-1's "lists EVERY project" quietly becomes "lists every well-formed project" — partial-snapshot must degrade the ROW, not drop it.
- Bare `except:` or a blanket try/except around the whole render that swallows real bugs and always reports success — degradation must be per-field, not global suppression that makes AC-3 untestable.
- Unescaped interpolation of entity text into HTML (the #1129 defect class in six sibling renderers); a name containing `<` or `&` corrupts or injects into the artifact.
- Duplicating the frontmatter parser or entity-discovery walk instead of reusing `validate-entities.py` — a second copy drifts from the schema the validator enforces.
- Reading `.metadata/staffing-recommendations.json` as a REQUIRED input: CLAUDE.md:50-55 describes it as a last-run snapshot that may not exist, so hard-failing without it violates AC-3.
- Mutating entity records or the manifest — the dashboard is a reporting layer, not a second writer of the data model.
- A health flag that merely restates fill status (e.g. "at risk" iff `open_roles` non-empty) with no independent signal — satisfies the letter of AC-1 while delivering nothing actionable. Human judgment call; not a hard gate.
- Non-deterministic output (wall-clock timestamps, dict-iteration ordering) making the artifact churn every run and the test flaky — `staffing-score.py` deliberately excludes wall-clock values.
- Version bump in plugin.json not mirrored in marketplace.json, or mirrored with a different string — the recurring drift this repo calls out.
- README/CLAUDE.md left claiming the dashboard is still "planned" while the skill ships — a later doc-audit drift finding.
- Scope creep into cogni-portfolio live value lookups, which the issue explicitly defers to the bridges child.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1127
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:contract-grading-reviewer

**Summary:** Three independent perspectives (plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:contract-grading-reviewer) converged on the same contract-item-20 defect — SKILL.md restates render-dashboard.py's health-flag and utilization derivation logic as a second source of truth without naming the script as authoritative — so the merged verdict is needs_revision; every other finding (oversized-PR diff, stale README line, missing data-model.md link, a ragged line wrap) is minor and non-blocking. 19 of the 20 committed contract items were confirmed satisfied by execution.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (5)

- **[major]** `cogni-projects/skills/projects-dashboard/SKILL.md:36-54` — Contract item 20 violated: the health-flag table (lines 36-42) restates render-dashboard.py's _project_health() decision logic (lines 204-222) condition-for-condition including precedence order, and the Utilization bullet (lines 46-54) restates the avg_allocation aggregation formula and null-handling rationale (script lines 286-300) near-verbatim. This is a second source of truth the agent never actually needs — it runs the script and relays data.path plus warnings — and it will silently drift when the renderer changes. Unlike sibling projects-staff/SKILL.md, this skill's Notes section (lines 108-121) carries no pointer naming render-dashboard.py as authoritative for the logic it just restated. Corroborated independently by cogni-service:skill-quality-reviewer and cogni-service:contract-grading-reviewer.
  - *Suggested fix:* Replace the health-flag table with one or two conceptual sentences and trim the Utilization bullet to only what the agent needs to read the envelope correctly (data.avg_allocation is null, not 0%, when no consultant has a usable allocation_pct). Then add a Notes bullet mirroring projects-staff: 'The renderer is the source of truth for the flags and aggregates. scripts/render-dashboard.py computes every health flag, the strategic-impact grouping, and the utilization figures; this section explains what the output means so it can be narrated to a partner, not how it is computed. To change a threshold or a flag rule, change the script, not this skill.'

- **[minor]** (auto-fixable) `cogni-projects/README.md:34` — Stale forward-looking sentence contradicts the rest of the README now that the dashboard has shipped: 'later backfilling and dashboard views plug into the same directory as they ship' still frames the dashboard as unshipped, while lines 5, 21, 28 and the components table at line 84 all correctly present it as delivered.
  - *Suggested fix:* Reword line 34 to scope the forward-looking clause to backfilling alone.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:25-61` — Unlike both siblings, this skill never links references/data-model.md, yet it names field contracts inline (open_roles, strategic_impact, allocation_pct, status).
  - *Suggested fix:* Add a link to ../../references/data-model.md in Core concept as the field contract. This also absorbs most of the prose the item-20 finding above asks to cut.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:59` — Ragged mid-sentence line wrap ('...no such warning). Any' followed by a short line) left over from an edit; cosmetic inconsistency with the otherwise uniform ~78-column wrapping.
  - *Suggested fix:* Re-wrap the closing paragraph of Core concept.

- **[minor]** `PR diff` — Diff is 1070 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- Contract items 1, 3, 4, 5 and 19 satisfied and independently verified by EXECUTION: version bump exact-match (0.0.7 -> 0.0.8 mirrored in marketplace.json), stdlib-only imports, importlib reuse of validate-entities.py rather than a second parser, hostile-fixture handling with zero tracebacks, and zero maintainer breadcrumbs across 50 scanned occurrences.
- AC-3 handled with unusual care: success:true-but-partial is separated from success:false environment failure, and the skill warns that data.partial can be set by a non-substantive warning so the agent does not mis-describe portfolio data as incomplete.
- Security posture sound: script/img-onerror/onmouseover payloads verified HTML-escaped by execution, and _valid_theme_value rejects CSS-structural characters so a url()/@import beacon cannot turn a self-contained partner artifact into a network-fetching page.
- Frontmatter, domain-prefixed naming and trigger-phrase quality all clean; body is 107 lines against the 500-line cap, leanest of the three sibling skills.
- Read-only contract enforced and tested: allowed-tools omits Write/Edit, and test 2c asserts the manifest is untouched across a re-render.
- Author's 38-assertion test suite passes in full, and independent hostile fixtures (non-dict manifest, out-of-enum assignment status, impact-1-vs-missing) confirmed the behavior rather than only re-running the author's own cases.

### Recommendation

---

sdh07:
## Correction: the verdict above was posted after this PR had already merged

**This PR merged at 18:14:23Z** (`00cd40dd`). The automated review verdict immediately above was posted at **18:24:45Z** — about ten minutes *later* — by a second, concurrent merger instance that had read this PR as `OPEN` at ~18:08Z and was still running its review when the merge landed.

The `svc:review-needs-revision` label that verdict applied has been removed and replaced with `svc:merged`, since a needs-revision label on a merged PR misrepresents the state to the author drain.

**What this means for the verdict itself:**

- It is **not** a request to revise this PR — there is nothing left to revise here.
- The findings were graded against commit `286161da`, which is what merged, so they remain **factually valid against `main`**.
- 19 of the 20 committed contract items were confirmed satisfied *by execution* (test suite plus independent hostile fixtures for out-of-enum assignment status, HTML injection, non-dict manifest, and impact-1-vs-missing). That part of the verdict stands as a positive record.
- The one substantive finding — committed contract **item 20**, the `projects-dashboard` SKILL.md restating `render-dashboard.py`'s `_project_health()` decision table and utilization aggregation as a second source of truth with no pointer naming the script authoritative — is now **on `main`** and is drift risk rather than a live defect (the restated logic is currently accurate). It is worth a small follow-up rather than anything urgent.

**Process note for maintainers:** two merger instances were draining the `svc:needs-review` queue against this repository at the same time. `service-gatekeeper` checks PR state once at Step 1 and does not re-check before posting its verdict, so a merge landing mid-review produces exactly this artifact. A state re-check immediately before `post-verdict` (and before the Step 5 label strip) would turn this into a clean `skipped: pr_not_open`.

sdh07:
Follow-up filed: #1134 — carries the committed contract item 20 finding (SKILL.md duplicating `render-dashboard.py`'s derivation logic) plus the three minor nits from the same review. Nothing further is needed on this PR.

Issue #1155, opened 2026-07-27
cogni-projects: a non-string project status crashes the whole dashboard render instead of warning


claude-cogni-work:
## Plugin
cogni-projects

## Problem
`render-dashboard.py:245` computes a project's status as:

```python
status = (proj.get("status") or "").strip()
```

`.strip()` is a `str` method, so a **non-string** `status:` frontmatter value raises `AttributeError`. This is reachable: `validate-entities.py`'s `_coerce` turns a bare numeric like `status: 2026` into an `int`. That is not a valid status per the schema enum, but the renderer crashes on it *before* any validation warning can be attached.

The consequence is a contract violation. `render-dashboard.py` is written to degrade gracefully — a malformed entity is supposed to become a per-entity warning in `data.warnings[]` with `data.partial: true`, and the rest of the portfolio still renders (Fixture 6 in `tests/test-render-dashboard.sh` exercises exactly that). Instead, one hand-edited project with a numeric status collapses the whole render into the top-level "unexpected failure" envelope, and the partner gets no dashboard at all.

## Proposed solution
Coerce defensively at the read site — `str(proj.get("status") or "").strip()` — and, if the value was not already a string, append a per-entity warning so the bad record is visible rather than silently normalized. `staffing-score.py:331` already wraps in `str(...)` for the same reason, so this also brings the two scripts closer together.

## Acceptance criteria
1. A portfolio containing a project with a non-string `status` (e.g. `status: 2026`) renders successfully, with the offending project surfaced in `data.warnings[]` and `data.partial: true`.
2. Regression coverage in `cogni-projects/tests/test-render-dashboard.sh`, alongside the existing malformed-entity fixture.

_Affected components: cogni-projects/scripts/render-dashboard.py, cogni-projects/tests/test-render-dashboard.sh_

claude-cogni-work:
## Triage Report

- **Severity:** S3-minor
- **Component:** cogni-projects / render-dashboard.py (`_compute`)
[...]

Issue #1152, opened 2026-07-27
cogni-projects: dashboard open-role total counts roles on closed projects


sdh07:
## Plugin
cogni-projects

## Problem
The dashboard's headline open-role count includes roles declared on **closed** projects.

`render-dashboard.py:498` computes `data.open_roles` as `sum(p["roles_total"] - p["roles_filled"] for p in projects)` across every project, and the "Open roles" tile does the same. A project with `status: closed` still gets its declared `open_roles` counted toward that total — it only receives a `closed` health flag.

The sibling `projects-staff` explicitly skips closed projects when building shortlists, so the two surfaces disagree: a partner reading the dashboard sees live hiring demand that `projects-staff` will not act on. On a portfolio carrying finished engagements the headline overstates demand, with nothing indicating why.

## Proposed solution
Exclude closed projects from the open-role total (and the corresponding tile), matching the behaviour `projects-staff` already implements. If the total is intentionally all-projects, say so explicitly in the tile label and in `projects-dashboard/SKILL.md` so the figure is not narrated as live demand.

## Acceptance criteria
1. `data.open_roles` and the dashboard tile agree with `projects-staff` on which projects count, or the divergence is labelled where a reader sees it.
2. Regression coverage in `cogni-projects/tests/test-render-dashboard.sh` for a portfolio containing a closed project with declared `open_roles`.

_Affected components: cogni-projects/scripts/render-dashboard.py, cogni-projects/tests/test-render-dashboard.sh_

_Found during review of #1136; renderer behaviour, out of scope for that documentation-anchoring PR._

claude-cogni-work:
## Triage Report

- **Severity:** S3-minor
- **Component:** cogni-projects / projects-dashboard (`render-dashboard.py`)
- **Root cause hypothesis:** The renderer already knows about closed status — `_project_health` (`render-dashboard.py:211`) special-cases `status == "closed"`, and `_compute` (`:246`) suppresses the "no open_roles" warning for closed projects — but the two open-role **aggregations** were never wired to that check. A closed project's declared-but-never-cleared roles therefore leak into the headline total. `staffing-score.py:326-332` filters them out before scoring (`str(p.get("status", "")).strip().lower() != "closed"`, commented "A closed project is delivered or lost — its roles are not open work to staff"), so the two surfaces genuinely disagree. Confirmed, not refuted.
[...]

PR #1136, opened 2026-07-19
cogni-projects: anchor render-dashboard.py as dashboard skill's source of truth

claude-cogni-work:
Closes #1134.

## Summary
- Add a **source-of-truth Notes bullet** to `projects-dashboard/SKILL.md` naming `scripts/render-dashboard.py` authoritative for health-flag and aggregate derivation, mirroring the sibling `projects-staff` anchor pattern.
- Trim the health-flag table so it conveys what each flag (`closed`, `staffing unknown`, `no open roles`, `fully staffed`, `unstaffed`, `<open>/<total> roles open`) **means to a partner reader**, without restating `_project_health()`'s branch conditions or precedence order.
- Trim the Utilization bullet to retain the **`null`-vs-`0%` envelope-reading guidance** while dropping the aggregation formula and exclusion rationale (owned by the script).
- Link `references/data-model.md` as the field contract (relative `../../references/data-model.md`, matching the sibling skills).
- Repair the ragged line wrap and rescope the README `## What it means for you` clause so it no longer describes the already-shipped dashboard as unshipped.
- Patch-bump `cogni-projects` `0.0.9 → 0.0.10` in `plugin.json` and `marketplace.json` (base was `0.0.9` on `main` after #1135; marketplace edited surgically, version line only).

## Scope
- **In scope:** all six acceptance criteria in one coherent doc/skill PR — a single drift-removal concern touching the same skill file, its README framing, and the mandatory paired version bump.
- **No renderer logic change.** `render-dashboard.py` is untouched; every flag's interpretive meaning is retained, only the duplicated *derivation* prose is removed.

## Test plan
- [x] `bash cogni-projects/tests/test-render-dashboard.sh` passes (no renderer logic changed).
- [x] Notes names `scripts/render-dashboard.py` authoritative; health-flag table no longer restates `_project_health()` conditions/precedence; utilization keeps the `null`-vs-`0%` guidance only.
- [x] `references/data-model.md` linked; line-wrap repaired; README no longer frames the dashboard as unshipped.
- [x] `plugin.json` and `marketplace.json` both read `0.0.10`.
- [x] `check-breadcrumbs.py` reports no new breadcrumbs.

PR #1156, opened 2026-07-27
cogni-projects: exclude closed projects from open-role total

claude-cogni-work:
Refs #1152.

## Summary
- `render-dashboard.py` counted unfilled roles on **every** project, including ones with `status: closed`, in both the JSON envelope (`data.open_roles`) and the "Open roles" HTML tile. A partner reading the dashboard saw hiring demand that `projects-staff` would never act on.
- Both summation sites now call one `_open_role_demand` helper, so they cannot drift apart again, and the closed check is named as `_is_closed` and matches case-insensitively the way `staffing-score.py` does.
- Closed projects still appear in the table with their `closed` flag and still count toward the `Projects` tile — only the demand figure drops them.
- Added regression Fixture 9 and a scope caveat in `projects-dashboard/SKILL.md`.

## Scope decisions
- **Route (a), not (b).** Acceptance criterion 1 allowed either excluding closed projects or labelling the divergence. Excluded, because every other closed-project signal in the plugin already points that way: `staffing-score.py` drops them before scoring, `references/data-model.md` says their roles are not open work to staff, and `_project_health` already renders them muted. Labelling would have entrenched a number no other surface agrees with. The SKILL.md caveat is added anyway, so the narrower figure is not mis-narrated.
- **Other aggregates untouched.** `value_by_impact` and the `Projects` tile still count all projects — lifetime-value and portfolio-size views, not demand figures, and outside the acceptance criteria.
- **No cross-script refactor.** A rejected alternative plan extracted a shared `is_closed_project()` into `validate-entities.py` for both this script and `staffing-score.py`. Not taken: `staffing-score.py` has no test suite, and no CI runs any of these tests, so the broader change's only guard would have been manual.

## Review notes
**A pre-existing case-sensitivity split, now visible.** `_project_health` and `_compute`'s warning gate both compare `status == "closed"` **case-sensitively**; the new `_is_closed` is case-**insensitive**, because that is what parity with `staffing-score.py` requires. So on a non-conforming `status: Closed`, the tile treats the project as closed while the row does not — the tile and row would disagree.

This is deliberate but worth a reviewer's eye. Making the new predicate case-sensitive instead would keep the file internally consistent but break the parity the issue asks for; unifying all three would change what a row renders, which is a behaviour change beyond this fix's scope. `Closed` is not a schema-valid status, so reaching the split requires a record that already fails validation. `_is_closed`'s docstring records the trade-off and names the unification as the eventual fix. Raised by the pre-commit quality pass, not by the original plan.

## Verification
- `bash cogni-projects/tests/test-render-dashboard.sh` — all fixtures pass, including new Fixture 9. There is no CI job for this suite (`lint.yml` has only the breadcrumb, external-dispatch, and version-bump gates), so this was run by hand.
- The regression test was confirmed to **fail against the unfixed renderer**: with `render-dashboard.py` restored from `origin/main`, assertions `9c` (envelope reports `open_roles: 3`) and `9f` (tile renders 3) both fail. A test that passed either way would prove nothing.
- `bash cogni-projects/tests/test_register_entity.sh` and `test_portfolio_init.sh` — both still pass.
- No `plugin.json` / `marketplace.json` version line is touched; the post-merge workflow owns the bump.

## Follow-up issues
- #1155 — cogni-projects: a non-string project status crashes the whole dashboard render instead of warning

#1155 is an **adjacent pre-existing defect** found while planning this fix, not deferred scope of #1152. Because the plan carries a non-empty `deferred[]`, this PR links with `Refs` rather than `Closes`, so merging will not auto-Refs #1152 — it can be closed by hand at merge, since its acceptance criteria are fully delivered here.

### DIFF
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -222,6 +222,43 @@ def _project_health(filled, open_roles, status):
     return ("%d/%d roles open" % (total - filled, total), "warn")
 
 
+def _is_closed(status):
+    """Whether a project status means the work is over — delivered or lost.
+
+    Matches case-insensitively, the way staffing-score.py does, so the dashboard
+    and the staffing shortlist agree on which projects are still live.
+
+    Two older checks in this file — _project_health's `status == "closed"` and
+    the warning gate in _compute — compare the literal case-sensitively and are
+    deliberately left alone here, since switching them changes what a row
+    renders. That means a non-conforming `status: Closed` is read as closed by
+    this predicate but not by those two, so the tile and the row would disagree.
+    The schema only admits lowercase `closed`, so that split needs a record that
+    already failed validation; unify the three when one is willing to take the
+    rendering change.
+    """
+    return str(status or "").strip().lower() == "closed"
+
+
+def _open_role_demand(projects):
+    """Sum unfilled roles across the projects that still represent live demand.
+
+    A closed project is delivered or lost, so its declared roles are not work
+    anyone will staff — staffing-score.py drops the same set before scoring, and
+    the data model says as much. Counting them here made the headline number
+    disagree with the shortlist a partner acts on.
+
+    Only `closed` is excluded, never an `active` allowlist: `prospective` is a
+    valid status, and filtering to active alone would silently drop pipeline
+    demand — the same divergence in the other direction.
+    """
+    return sum(
+        p["roles_total"] - p["roles_filled"]
+        for p in projects
+        if not _is_closed(p["status"])
+    )
+
+
 def _compute(entities, warnings):
     """Derive per-project staffing + the portfolio value/utilization aggregates.
 
@@ -349,7 +386,7 @@ def _render_html(portfolio, projects, value_by_impact, util, warnings, theme):
         )
 
     avg = "—" if util["avg_allocation"] is None else "%d%%" % util["avg_allocation"]
-    total_open = sum(p["roles_total"] - p["roles_filled"] for p in projects)
+    total_open = _open_role_demand(projects)
 
     return """<!DOCTYPE html>
 <html lang="en"><head><meta charset="utf-8">
@@ -495,7 +532,7 @@ def main(argv):
             "consultants": util["consultants"],
             "avg_allocation": util["avg_allocation"],
             "fully_allocated": util["fully_allocated"],
-            "open_roles": sum(p["roles_total"] - p["roles_filled"] for p in projects),
+            "open_roles": _open_role_demand(projects),
             "warnings": warnings,
             "partial": bool(warnings),
         },
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -33,6 +33,13 @@ records (their field contract lives in
   Two flags read alike but are not: `no open roles` means the project needs
   nobody right now, while `staffing unknown` means it never declared
   `open_roles`, so coverage cannot be read — never narrate it as covered.
+  The headline figure (`data.open_roles` and the `Open roles` tile) counts
+  unfilled roles on non-closed projects only — closed work is delivered or
+  lost, not live demand — so it matches the set `projects-staff` shortlists. A
+  `prospective` project still counts, so read it as live-and-pipeline demand,
+  not active-only. The closed project stays in the table with its `closed`
+  flag and still counts toward the `Projects` tile, so a shrinking open-role
+  total is not a missing project.
 - **Portfolio value**: projects grouped by `strategic_impact` (1–5), so the
   high-impact work is visible at a glance.
 - **Utilization**: the envelope carries an average-allocation figure and a
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -391,6 +391,50 @@ assert_html_lacks "8e url() beacon theme value rejected" \
 assert_html "8f safe accent still applied alongside rejected url()" \
   "#00ff88" "$PF8/output/dashboard.html"
 
+# Fixture 9 — a closed project's declared roles are not open demand. The closed
+# project stays visible in the table with its flag (9g/9h); only the headline
+# figure drops it, so a shrinking total is not a vanished project. 9c and 9f are
+# a pair because the envelope and the HTML tile are separate call sites in the
+# renderer — fixing one alone would ship a dashboard that disagrees with itself,
+# so asserting only one of them would let that regression through.
+PF9="$TMPROOT/closed"
+seed_portfolio "$PF9"
+write_entity "$PF9/projects/live-crm.md" <<'EOF'
+---
+type: project
+slug: live-crm
+name: Live CRM Rollout
+client: Northwind
+strategic_impact: 4
+status: active
+open_roles: [crm-lead]
+---
+# Live CRM Rollout
+EOF
+write_entity "$PF9/projects/legacy-migration.md" <<'EOF'
+---
+type: project
+slug: legacy-migration
+name: Legacy Migration
+client: Northwind
+strategic_impact: 3
+status: closed
+open_roles: [migration-lead, data-engineer]
+---
+# Legacy Migration
+EOF
+run "$PF9"
+HTML9="$PF9/output/dashboard.html"
+assert_json "9a closed-portfolio render succeeds" "d['success'] is True"
+assert_json "9b closed project still counted as a project" "d['data']['projects'] == 2"
+assert_json "9c open roles exclude the closed project" "d['data']['open_roles'] == 1"
+assert_json "9d closed project emits no warning" "d['data']['warnings'] == []"
+assert_json "9e clean snapshot" "d['data']['partial'] is False"
+assert_html "9f tile agrees with the envelope" \
+  '<div class="n">1</div><div class="l">Open roles</div>' "$HTML9"
+assert_html "9g closed project still listed" "Legacy Migration" "$HTML9"
+assert_html "9h closed health flag still rendered" ">closed</span>" "$HTML9"
+
 echo
 if [ "$failures" -eq 0 ]; then
   echo "All render-dashboard tests passed."


claude-cogni-work:
## Ship-ready

Author instance finished this change and is handing it to the merger for an independent review. This PR is deliberately left as a **draft** with `svc:needs-review` — the author drain does not flip draft→ready and does not run its own review gate, so the first substantive review of this diff is the merger's.

**What was done**
- Both open-role summation sites in `render-dashboard.py` now go through one `_open_role_demand` helper that skips closed projects, with the closed check named as `_is_closed`.
- Regression Fixture 9 added, and confirmed to fail against the unfixed renderer (assertions 9c and 9f) — it is a real guard, not a test that passes either way.
- A scope caveat added to `projects-dashboard/SKILL.md` so the narrower figure is not narrated as all-projects demand.
- Full `test-render-dashboard.sh` suite passes, as do the two sibling suites. No CI covers these, so they were run by hand.

**Worth the reviewer's attention**
- The `## Review notes` section in the PR body: this change introduces a case-sensitivity split between the new case-insensitive `_is_closed` and the two pre-existing case-sensitive checks in the same file. Deliberate — parity with `staffing-score.py` requires it — but it is the judgment call most worth a second opinion.
- The PR links `Refs #1152`, not `Closes`, because the plan carries a non-empty `deferred[]`. The deferred item (#1155) is an adjacent pre-existing defect rather than unfinished scope, so #1152's acceptance criteria are fully met and it can be closed by hand at merge.

sdh07:

_The merger formed this expectation from issue #1152 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix makes the dashboard's headline open-role figure agree with the sibling staffing engine on which projects count. Today `cogni-projects/scripts/render-dashboard.py` sums demand across every project at two independent sites — `total_open` at :352 (which feeds the 'Open roles' tile through the `open_roles=` format key at :420 into the tile markup at :392) and the envelope's `data.open_roles` at :498 — while `cogni-projects/scripts/staffing-score.py:329-332` drops any project whose `str(p.get('status','')).strip().lower() == 'closed'` before scoring. The fix should introduce one shared exclusion helper in render-dashboard.py (naturally placed near `_project_health` at :204-222 / `_compute` at :225) and point BOTH summation sites at it, using the same strip+lowercase denylist as staffing-score.py so parity is real rather than incidental — and specifically a `!= 'closed'` denylist, never an `active` allowlist, which would silently erase prospective demand. It must be a headline-only change: closed projects must still appear in the projects table with their real coverage cell and their existing `closed` health flag (`_project_health`:211), still be counted by `data.projects` / the `n_projects` tile (:418), and still land in `value_by_impact` (:265-267) and `util` (:295-300). `cogni-projects/tests/test-render-dashboard.sh` gains a new fixture (existing fixtures run 1-8, all-active, with Fixture 1's `1e open roles = 3` at :146 as the parity anchor) pairing an active and a closed project so both the envelope value and the tile markup are asserted, with a negative guard against the pre-fix number; `cogni-projects/skills/projects-dashboard/SKILL.md` should state the figure's scope in its **Staffing coverage** bullet (:30-35) so a partner is not narrated a demand number whose scope is unstated.

**Acceptance checklist:**
- [ ] Scope: the diff touches only cogni-projects/scripts/render-dashboard.py, cogni-projects/skills/projects-dashboard/SKILL.md, and cogni-projects/tests/test-render-dashboard.sh — no version bump in cogni-projects/.claude-plugin/plugin.json or the repo-root .claude-plugin/marketplace.json (confirm both version lines are absent from the diff).
- [ ] Repo structure and plugin manifest health are intact after the change (structure/manifest gate reports clean).
- [ ] The projects-dashboard SKILL.md frontmatter is still valid with its required keys (`name`, `description`, `allowed-tools`) — frontmatter gate reports clean.
- [ ] The projects-dashboard skill still satisfies the skill length / body cap.
- [ ] No maintainer breadcrumbs are introduced — no `#1152`, `#1136`, version tags, or slice/finding codes appear in SKILL.md or the script; the breadcrumb gate reports clean.
- [ ] render-dashboard.py gains exactly one closed-project exclusion predicate (a named module-level helper, e.g. `_open_role_demand`/`_is_closed`), and grepping the file for `roles_total"] - p["roles_filled` returns zero occurrences that are NOT routed through that helper — i.e. neither the `total_open` site (was :352) nor the envelope site (was :498) still carries an independent unfiltered sum.
- [ ] The tile number and `data.open_roles` are provably the same value: either the envelope reuses the same helper call or the value computed for the `open_roles=` format key is the value placed in `data.open_roles` — a reader can trace one expression to both sinks.
- [ ] The predicate matches staffing-score.py:331 exactly in normalization: it strips whitespace and lowercases before comparing to `closed` (grep the new helper for `.strip()` and `.lower()`), so `status: Closed` and `status: " closed "` are excluded on both surfaces.
- [ ] The predicate is a `!= "closed"` denylist, not an allowlist — grep shows no `in ("active"` / `== "active"` membership test gating the sum, so a project with `status: prospective`, an empty status, or a missing `status` key still contributes its demand.
- [ ] The envelope key is still named `data.open_roles` (not renamed or removed) and the tile label element `<div class="l">Open roles</div>` still exists, so existing readers and assertions do not break.
- [ ] Closed projects are NOT dropped from `projects_out`: `data.projects` and the `{n_projects}` tile still count them, the closed project still renders a `<tr>` in the projects table, and `_project_health`'s `("closed", "muted")` branch at :211 still fires so the row shows the `closed` flag.
- [ ] The closed project's per-project data is not doctored to make the total work: its `roles_total` / `roles_filled` / `open_roles` list in `projects_out` and the table's coverage cell still report its real declared figures rather than being zeroed or blanked.
- [ ] `value_by_impact` (:265-267), the `util` aggregate (:295-300), and the `declared_roles is None and status != "closed"` warning guard (:246) are unchanged — the diff shows no edit to those blocks (scope creep guard).
- [ ] test-render-dashboard.sh gains a new fixture seeding at least one `status: active` project with declared `open_roles` and one `status: closed` project with declared `open_roles`, using the existing `seed_portfolio` / `write_entity` / `run` helpers.
- [ ] The new fixture asserts the envelope with `assert_json` on `d['data']['open_roles']` equal to the active-only count, AND asserts the tile with `assert_html` on a needle that binds the number to its label (e.g. the `<div class="n">N</div><div class="l">Open roles</div>` markup) rather than a bare digit that could match anywhere else in the page.
- [ ] The new fixture carries a mutation guard: an `assert_html_lacks` (or negative `assert_json`) against the pre-fix total — the sum that includes the closed project's roles — so reverting the fix fails the suite rather than passing vacuously.
- [ ] The new fixture also asserts the non-regression half: the closed project's name still appears in the rendered HTML and its `closed` health flag is still present (grep-able needle from the `<span class='flag'...>closed</span>` markup).
- [ ] Fixture 1's existing assertion `1e open roles = 3` is unchanged and still passes — an all-active portfolio's headline is unaffected by the fix.
- [ ] Running `bash cogni-projects/tests/test-render-dashboard.sh` against the PR head exits 0 and prints `All render-dashboard tests passed.` with zero FAIL lines.
- [ ] render-dashboard.py still prints a single-line `{success, data, error}` envelope on every path and remains stdlib-only — the diff adds no new import beyond the existing argparse/datetime/html/importlib/json/os/sys set and no pip dependency.
- [ ] projects-dashboard/SKILL.md states the open-role figure's scope in prose — its **Staffing coverage** (or tile-narration) bullet names that the headline excludes closed projects (grep for `closed` in that bullet), and that prose matches the code shape actually shipped in render-dashboard.py. This mechanism is prose-carried, so a code-shape-anchored grep plus the documented behavior match is the achievable bar; no behavioral test of the SKILL.md is expected.
- [ ] The SKILL.md edit is additive: the existing caveats survive verbatim — the `no open roles` vs `staffing unknown` distinction, the `avg_allocation` null-is-unknown caveat, the 'partial is set by any warning' caveat, and the 'renderer is the source of truth' note are all still present (diff shows no deleted qualifier lines).

**Risk watchlist:**
- Over-filtering: filtering `entities["project"]` at the top of `_compute` instead of only at the summation sites — this fixes the headline but silently drops closed projects from the table, from `data.projects`, and from the strategic-impact value summary, turning a headline bug into data loss on the partner's decision surface.
- Half-fix: patching only the envelope at :498 or only `total_open` at :352, leaving the tile and the JSON disagreeing with each other — strictly worse than the current consistent-but-wrong state.
- Allowlist inversion: implementing the exclusion as `status == "active"` rather than `status != "closed"`, which silently erases demand from `prospective`, blank, or missing-status projects without any warning.
- Casing parity theatre: an exact `== "closed"` comparison (copying `_project_health`:211) instead of staffing-score.py's `.strip().lower()`, so `status: Closed` still diverges from projects-staff — the exact divergence AC-1 exists to close.
- Residual internal inconsistency (judgment, not gateable): if the new helper normalizes case but `_project_health`:211 and the `:246` warning guard stay case-sensitive, a `status: Closed` project is excluded from the headline yet renders without the `closed` flag and may emit a spurious 'staffing status unknown' warning. This is strictly less wrong than today and arguably a separate pre-existing defect, but a human should decide whether to normalize all three sites now or file a follow-up. Do not bounce the PR on this alone.
- Vacuous test harness (this repo has form here): a tile assertion whose needle is a bare digit, or an assertion of a value the pre-fix code also produced, would pass against an unfixed renderer. Confirm the new fixture actually distinguishes fixed from unfixed — the negative guard is the load-bearing assertion.
- Fixture arithmetic error: choosing an active count and a closed count that happen to coincide with some other number on the page (e.g. the projects count or avg allocation), making the tile grep ambiguous.
- Silent envelope contract drift: renaming `data.open_roles`, adding a competing `data.open_roles_all`, or emitting a new warning per closed project that flips `data.partial` to true on a healthy portfolio — the SKILL.md narrates `partial` to the user, so a spurious flip is user-visible.
- Doc-side scope creep or caveat deletion while editing SKILL.md — this repo has a recurring pattern of doc edits quietly removing the qualifiers that document deliberately-silent cases.
- Whether the projects table's own coverage/'Still open' columns should also be visually muted for closed projects is a presentation judgment the issue does not settle; a human may want it, but it is not a gate and its absence is not a defect.
- A version bump landing in the branch would reintroduce the repo-wide marketplace.json conflict class the post-merge auto-bump eliminated.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1156

**Summary:** All four perspectives passed and the committed review contract's dedicated grading pass confirmed all 22 acceptance-checklist items satisfied with zero findings; the only un-waived finding is an advisory minor on fixture coverage, and the two case-sensitivity findings are explicitly waived under the contract's own risk-watchlist instruction not to bounce on them alone.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (waived) `cogni-projects/scripts/render-dashboard.py:225-240` — Residual internal case-sensitivity split between `_is_closed` (normalized) and `_project_health`'s comparison (case-sensitive): for a non-schema-conformant `status: Closed` record, the headline correctly excludes it but the table row renders without the `closed` flag, so tile and row disagree for that one record. The corresponding SKILL.md prose at skills/projects-dashboard/SKILL.md:40-42 ('The closed project stays in the table with its `closed` flag') holds unconditionally only for the schema-conformant lowercase case, and inherits the same divergence. The author's `_is_closed` docstring and the PR's Review notes already disclose this; it is pre-existing, not introduced by this diff, and unreachable without a schema-invalid record.
  - *Justification:* The committed review contract's risk_watchlist item 5 names this exact condition and instructs 'Do not bounce the PR on this alone.' Pre-existing divergence, not introduced by this change, and unreachable without a schema-invalid record.
  - *Raised by:* plugin-dev:plugin-validator, also flagged by plugin-dev:skill-reviewer

- **[minor]** (auto-fixable) `cogni-projects/tests/test-render-dashboard.sh:394-436` — Fixture 9 pairs only an `active` and a `closed` project, so it does not lock in the denylist-vs-allowlist property. Mutation-tested: rewriting the predicate to an `== "active"` allowlist (the inversion named in the contract's risk watchlist) leaves the whole suite green. The shipped code is correct (verified live: a `status: prospective` project with 2 declared roles does contribute), but nothing in the suite prevents a future refactor from silently erasing pipeline demand.
  - *Suggested fix:* Add a third project to Fixture 9 with `status: prospective` and declared `open_roles`, raise the 9c/9f expected total accordingly, and keep the closed project's roles out of it. That one extra entity converts the allowlist inversion from a passing mutation into a failing one.

### Positive observations

- Denylist-not-allowlist verified in code and at runtime: no `== "active"` membership test exists anywhere in the file, and a `status: prospective` project with 2 declared roles still contributes its demand (plugin-validator, contract-grading-reviewer item 9).
- Tile/envelope agreement is provable by construction — both sinks call the identical `_open_role_demand(projects)` expression over the identical list, and tests 9c/9f each fail independently under a tile-only or envelope-only half-fix (plugin-validator, contract-grading-reviewer item 7).
- Item 16 satisfied by direct execution, not inspection: the pre-fix renderer copied from origin/main into an isolated scratch worktree fails both 9c and 9f, proving the new fixture is non-vacuous rather than trusted on inspection alone (contract-grading-reviewer).
- SKILL.md edit is genuinely additive (7 added / 0 deleted) and all four pre-existing load-bearing caveats survive verbatim - this repo has a recurring pattern of doc edits quietly dropping qualifiers for deliberately-silent cases, and this one does not (skill-reviewer, skill-quality-reviewer).
- Prose-to-code fidelity confirmed on every clause of the new caveat, including that `_is_closed`'s normalization is byte-identical to staffing-score.py:331 (skill-quality-reviewer, plugin-validator).
- Manifest and structural gates all clean: no version bump in plugin.json or the marketplace.json mirror, zero new breadcrumbs, zero new imports (stdlib-only preserved), and the suite exits 0 with 63/63 passing at head (plugin-validator, contract-grading-reviewer).

### Recommendation

---

Issue #1162, opened 2026-07-27
cogni-projects: a non-string assignment role still crashes the whole dashboard render


claude-cogni-work:
## Plugin
cogni-projects

## Problem
`render-dashboard.py` still loses the entire dashboard to a single hand-edited record — the same contract inversion #1155 fixed for `status`, reached through `role` instead.

`sorted(covered)` in `_compute` raises `TypeError` when an assignment's `role` is a non-string. `validate-entities.py`'s `_coerce` turns any all-digit scalar into an `int` at parse time, and the renderer reuses that parser without running the validator, so `covered` can hold a mix of `int` and `str`.

### Reproduction

One project with `open_roles: [lead]`, and two active assignments on it — one with `role: 2`, one with `role: lead`:

```
{"success": false, "data": {}, "error": "unexpected failure: TypeError: '<' not supported between instances of 'str' and 'int'"}
```

Verified against `main` with #1155's fix applied. Every accumulated warning is discarded and the partner gets no dashboard, exactly as before that fix.

## Proposed solution
Two candidates, and the choice is the real question here:

1. **Read-site defense**, mirroring #1155 — coerce `role` where `covered` is built. Minimal and consistent with what just shipped, but it hardens one more field and leaves the next one exposed.
2. **Normalize at the parse boundary** — one pass in `_read_entities` (which already owns the per-entity warning surface, already has the relative path as a label, and already mutates the parsed dict to set `_file`) that coerces a known list of string-typed fields per entity type, never touching numeric fields like `strategic_impact` / `allocation_pct`. This closes the class rather than the instance.

Option 2 is the stronger fix and would make #1155's `_status_text` redundant. It is filed separately rather than folded into #1155 because it changes normalization for every field on every entity type — a materially wider blast radius than that issue's acceptance criteria, and it needs its own decision about which fields count as string-typed.

Do **not** change `_coerce` itself: its int coercion is wanted for numeric fields, `validate-entities.py` is shared with other scripts, and the validator already reports `status: 2026` correctly as an invalid enum value. The defect is entirely that the renderer reuses the parser but skips the validator.

## Acceptance criteria

_Ratified 2026-07-28 by maintainer ruling clearing the `svc:needs-human-review` hold on the linked PR. The original two criteria under-specified the bar for an issue that deliberately leaves a design choice open; the merger's independent contract derived the items below from the base code, the delivered change satisfies them, and they are recorded here as the accepted bar._

[...]

Issue #1161, opened 2026-07-27
cogni-projects: warn on an assignment whose project matches no project in the portfolio


claude-cogni-work:
## Plugin
cogni-projects

## Current behavior
In `render-dashboard.py`'s `_compute`, the assignment loop reads an assignment's `status` only after the project filter matches:

```python
if a.get("project") == slug and (a.get("status") or "").strip() in ACTIVE_ASSIGNMENT_STATES:
```

An assignment whose `project` value matches no project in the portfolio — a typo, or a project file that was deleted — is therefore never status-read at all. It contributes nothing to role coverage and produces no warning: it is silently invisible.

## Desired behavior
Surface an orphan assignment as a per-entity warning, so a mistyped `project:` reference is visible in `data.warnings[]` rather than quietly reducing coverage.

## Motivation
Silent invisibility is the same failure shape the parent issue fixed for a non-string status: a malformed record that the renderer skips without telling anyone. A partner reading the dashboard sees an under-staffed project with no indication that an assignment meant to cover it was dropped on a typo.

## Why this is separate
The parent issue's acceptance criteria are about a non-string `status` crashing the render. Orphan-assignment detection is a distinct coverage gap that needed its own decision about whether every unmatched assignment warrants a warning (a portfolio mid-authoring may legitimately carry assignments for projects not yet filed).

**That decision was made during grooming — it is no longer open:** the warning is **status-independent**. Every assignment whose `project` matches no project in the portfolio warns, including one with `status: completed` — a completed assignment naming a non-existent project is still a broken reference, and scoping the check to `ACTIVE_ASSIGNMENT_STATES` would leave a typo in a historical record invisible.

The mid-authoring consequence is **accepted, not designed around**: a portfolio with no project files yet will orphan every assignment and flip `partial` true. Suppressing the warning wholesale in that case would restore exactly the silent invisibility this issue is about. Gated in the `## Acceptance criteria` below.

## Acceptance criteria

_Ratified via `/cogni-service:service-groom` — a base-grounded derivation against `origin/main`, with one maintainer decision folded in: the orphan warning is **status-independent** (a completed assignment naming a non-existent project is still a broken reference)._

**Scope and gates**

[...]

Issue #1160, opened 2026-07-27
cogni-projects: unify the three case-sensitivity treatments of project status in render-dashboard.py


claude-cogni-work:
## Plugin
cogni-projects

## Current behavior
`render-dashboard.py` treats a project's `status` three different ways in the same file:

- `_project_health` compares against the case-sensitive literals `== "closed"` / `== "active"`
- the `_compute` open-roles warning gate compares `status != "closed"`, also case-sensitive
- `_is_closed` compares case-**in**sensitively: `str(status or "").strip().lower() == "closed"`

So a project written `status: Closed` is treated as closed by the open-role demand total but as *not* closed by the health flag and the missing-`open_roles` warning gate. `_is_closed`'s own docstring already flags this split as something worth unifying "when one is willing to take the rendering change".

## Desired behavior
One consistent treatment of `status` across all three sites.

## Motivation
The split is latent today because entities are usually machine-written in lower case, but it is a real correctness gap for hand-edited portfolios — the same record can be simultaneously counted as closed and rendered as active.

## Why this is separate
Unifying them changes what a row renders for any portfolio that currently has mixed-case statuses, which is a product decision with its own review question. It is deliberately out of scope for the crash fix in the parent issue, which only makes a non-string status degrade to a warning.

**The review questions were answered during grooming — they are no longer open:**

1. **Direction: case-insensitive.** This was never actually a free choice. `staffing-score.py:331` already filters with `str(...).strip().lower() != "closed"`, and `_is_closed`'s own docstring says it matches case-insensitively *so the dashboard and the staffing shortlist agree*. Unifying case-**sensitively** would re-diverge them, which is the defect class this issue exists to close.
2. **Rendering: normalized.** A project authored `status: Closed` renders `closed`, so the column matches the value the logic acted on.
3. **Normalization is surfaced.** A status that required normalizing emits a warning, matching this file's existing convention for schema-deviating records (`_text_field`, `_coerce_impact`). A correctly-lowercase status emits none.

All three are gated in the `## Acceptance criteria` below.

## Acceptance criteria

_Ratified via `/cogni-service:service-groom` — a base-grounded derivation against `origin/main`, with three maintainer decisions folded in: unify **case-insensitively** (the direction `staffing-score.py:331` and `_is_closed`'s own docstring already point to), render the **normalized** value in the Status column, and **warn** when a status had to be normalized._

**Scope and gates**

[...]

PR #1163, opened 2026-07-27
cogni-projects: non-string status warns instead of crashing the render

### LINKED ISSUE #1155: cogni-projects: a non-string project status crashes the whole dashboard render instead of warning

claude-cogni-work:
Closes #1155.

## Summary

`render-dashboard.py` called `.strip()` straight on a frontmatter `status` at two sites in `_compute` — the project status, and the assignment status inside the role-coverage loop. The frontmatter parser it borrows from `validate-entities.py` coerces any all-digit scalar to an `int` before the schema enum is ever checked, and the renderer never runs that validator, so a hand-edited `status: 2026` arrived as an `int`. The int is truthy, so the `or ""` fallback never fired and `.strip()` raised.

Nothing local catches that, so it reached the module-level catch-all, whose `_fail` hard-codes empty `data` — discarding every warning collected so far. One bad record cost the partner the whole dashboard, inverting the script's contract that a malformed entity becomes a per-entity warning and nothing else.

Both sites now read through `_status_text`, which coerces to text and appends a warning naming the record.

## Two things worth a reviewer's attention

**1. The guard keys on `raw is not None`, not on truthiness.** `status: 0` coerces to a *falsy* int. A truthiness guard (`str(raw or "")`) would silently normalize it to `"unknown"` and warn about nothing at all — a quieter version of the same defect. Fixture 10's `10c`/`10d` are a pair specifically to pin both halves: `2026` used to crash, `0` used to vanish, and a regression to either shape fails exactly one of them.

**2. The issue documents one crash site; there are two.** The assignment-status read at the same nesting level has the identical signature. Fixing only the documented one would leave the contract violation reachable, so both are covered. The loop's compound condition became an early-`continue` purely so the coercion can sit in statement position — the project filter already ran first via `and` short-circuit, so this is not a behavior change and there was never a duplicate-warning bug to fix.

## Scope decisions

- **No parse-boundary normalization.** The strongest argument against this change's shape is that defending at read sites is per-field and the field list is unbounded — see the known adjacent defect below. Normalizing known string fields once in `_read_entities` would close the class. It is deliberately not done here: it changes normalization for every field on every entity type, which is a materially wider blast radius than this issue's acceptance criteria and needs its own decision about which fields count as string-typed. Filed as #1162.
- **`validate-entities.py` is untouched.** Its int coercion is genuinely wanted for `strategic_impact` / `allocation_pct`, the file is shared with other scripts, and the validator already reports `status: 2026` correctly as an invalid enum value. The defect is entirely that the renderer reuses the parser but skips the validator.
- **Fixture 10 is a sibling of Fixture 6, not a case inside it.** Fixture 6 pins a hardcoded project count and shares one stderr capture; folding a fourth malformed shape in would rewrite passing assertions and conflate a decode failure with this crash. No existing fixture, count, or comment is touched.
- **The assignment loop was not hoisted into a pre-pass.** Two reviewers suggested it (fewer nesting levels, `O(A)` instead of `O(P×A)`). Declined: hoisting makes orphan assignments newly status-read and warned, which is a behavior change beyond the acceptance criteria and is tracked separately as #1161.

## Known adjacent defect — not fixed here

The same whole-render loss is still reachable through `role` rather than `status`. `sorted(covered)` raises `TypeError` on a mixed int/str role set:

```
{"success": false, "data": {}, "error": "unexpected failure: TypeError: '<' not supported between instances of 'str' and 'int'"}
```

Reproduced against this branch (one project with `open_roles: [lead]`, two active assignments on it, one `role: 2` and one `role: lead`). Filed as #1162 with the repro and an exposure audit of every other frontmatter field this script reads.

This is surfaced rather than fixed because it is a different field with a different failure mode and is outside this issue's acceptance criteria — but a reviewer who thinks the parse-boundary fix should land instead of this one has everything needed to make that call, and this PR should then be closed in favor of #1162.

## Test plan

- [x] `bash cogni-projects/tests/test-render-dashboard.sh` — all 10 fixtures pass (46 assertions).
- [x] **The test actually pins the bug.** Reverting only `render-dashboard.py` to `origin/main` and re-running fails 7 of Fixture 10's 8 assertions with exactly `AttributeError: 'int' object has no attribute 'strip'`. (`10g` passes in both states — the catch-all suppresses the traceback — so it guards the same shape `6g` does rather than being the discriminating assertion.)
- [x] Warning cardinality: a portfolio with 3 projects and 1 bad-status assignment emits exactly **one** assignment warning, not one per project.
- [x] `bash cogni-projects/tests/test_register_entity.sh` and `test_portfolio_init.sh` — unaffected siblings still pass.
- [x] `check-breadcrumbs.py` 0 findings · `check-version-bump.py` `[ok]` (14 plugins scanned, 0 touched, 0 violations) · `check-external-dispatch.py` 0 findings · `check-skill-names.sh` OK.
- [x] `git diff origin/main -- '*/.claude-plugin/*.json'` is empty — no version line in this branch; the post-merge workflow owns the bump.

## Follow-up issues

- #1160 — unify the three case-sensitivity treatments of project `status`
- #1161 — warn on an assignment whose `project` matches no project
- #1162 — a non-string assignment `role` still crashes the whole render

### DIFF
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -240,6 +240,31 @@ def _is_closed(status):
     return str(status or "").strip().lower() == "closed"
 
 
+def _status_text(raw, label, warnings):
+    """Read an entity's `status` as text, warning when it was not text already.
+
+    The frontmatter parser this script borrows from validate-entities.py coerces
+    any all-digit scalar to an int before the schema enum is ever checked — and
+    the renderer never runs that validator. So a hand-edited `status: 2026`
+    arrives here as an int, and calling a str method on it straight raises
+    inside _compute, which no local handler catches: the module-level catch-all
+    then discards every warning collected so far and fails the whole render.
+    That inverts this script's contract, where one bad record costs a warning
+    and nothing else.
+
+    The guard keys on `raw is not None`, deliberately not on truthiness —
+    `status: 0` coerces to a falsy int, which a truthy check would silently
+    normalize to "unknown" with no warning at all.
+    """
+    if raw is None:
+        return ""
+    if not isinstance(raw, str):
+        warnings.append(
+            "%s has a non-string status %r — coerced to text; fix the record" % (label, raw)
+        )
+    return str(raw).strip()
+
+
 def _open_role_demand(projects):
     """Sum unfilled roles across the projects that still represent live demand.
 
@@ -279,13 +304,18 @@ def _compute(entities, warnings):
         # An undeclared open_roles is not an empty one — see _normalize_open_roles.
         declared_roles = _normalize_open_roles(proj)
         open_roles = declared_roles or []
-        status = (proj.get("status") or "").strip()
+        status = _status_text(proj.get("status"), label, warnings)
         if declared_roles is None and status != "closed":
             warnings.append("%s has no open_roles — staffing status unknown" % label)
 
         covered = set()
         for a in assignments:
-            if a.get("project") == slug and (a.get("status") or "").strip() in ACTIVE_ASSIGNMENT_STATES:
+            # Filter first — an assignment belonging to another project must not
+            # have its status read here, or it would warn once per project.
+            if a.get("project") != slug:
+                continue
+            a_label = "assignment %s" % (a.get("_file") or a.get("slug") or "(unknown)")
+            if _status_text(a.get("status"), a_label, warnings) in ACTIVE_ASSIGNMENT_STATES:
                 role = a.get("role")
                 if role:
                     covered.add(role)
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -111,9 +111,11 @@ A failure here is cosmetic — the dashboard is already written to `data.path`.
   `projects-entities` does, via `register-entity.py`) and never touches
   `.metadata/`. Re-running only rewrites `output/dashboard.html`.
 - **Partial snapshots are expected mid-authoring.** A project without a
-  `strategic_impact`, a project that omits `open_roles`, or an entity file
-  that cannot be read or decoded is reported in the warnings list, not
-  treated as an error. One bad record never costs the rest of the portfolio.
+  `strategic_impact`, a project that omits `open_roles`, an entity whose
+  `status` is not text (an all-digit value such as `status: 2026` is read as a
+  number, then coerced back to text), or an entity file that cannot be read or
+  decoded is reported in the warnings list, not treated as an error. One bad
+  record never costs the rest of the portfolio.
 - **Theming is optional.** The dashboard renders with a built-in palette;
   pass `--design-variables <path.json>` to override colors when a themed look is
   wanted.
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -63,6 +63,17 @@ assert_html_lacks() {
   fi
 }
 
+# Every "the renderer degraded instead of crashing" fixture needs this, and the
+# crash signature belongs in one place — widening it later should be one edit.
+assert_no_traceback() {
+  local label="$1" file="$2"
+  if grep -qF 'Traceback' "$file"; then
+    fail "$label" "traceback present: $(head -3 "$file")"
+  else
+    pass "$label"
+  fi
+}
+
 seed_portfolio() {
   # $1 = portfolio dir. Seeds a manifest + entity subdirs.
   local pf="$1"
@@ -327,11 +338,7 @@ assert_json "6e warns on unknown staffing (both shapes)" \
   "sum('staffing status unknown' in w for w in d['data']['warnings']) == 2"
 assert_json "6f warns on undecodable file" \
   "any('cannot read' in w and 'latin1' in w for w in d['data']['warnings'])"
-if grep -qF 'Traceback' "$STDERR6"; then
-  fail "6g no traceback on stderr" "traceback present: $(head -3 "$STDERR6")"
-else
-  pass "6g no traceback on stderr"
-fi
+assert_no_traceback "6g no traceback on stderr" "$STDERR6"
 assert_html "6h unknown staffing flagged in HTML" \
   "staffing unknown" "$PF6/output/dashboard.html"
 
@@ -434,6 +441,92 @@ assert_html "9f tile agrees with the envelope" \
   '<div class="n">1</div><div class="l">Open roles</div>' "$HTML9"
 assert_html "9g closed project still listed" "Legacy Migration" "$HTML9"
 assert_html "9h closed health flag still rendered" ">closed</span>" "$HTML9"
+
+# ---------------------------------------------------------------------------
+# Fixture 10 — a status that is not text. The frontmatter parser coerces an
+# all-digit scalar to an int before any enum check runs, so `status: 2026`
+# reached a str method and raised, and the module-level catch-all then threw
+# away every warning collected so far — one hand-edited record cost the whole
+# dashboard. 10c and 10d are a pair: `2026` used to crash while `0` is *falsy*
+# and used to normalize silently to "unknown", so a guard keyed on truthiness
+# rather than on None would pass one and fail the other.
+# ---------------------------------------------------------------------------
+PF10="$TMPROOT/nonstring-status"
+seed_portfolio "$PF10"
+write_entity "$PF10/projects/numeric-status.md" <<'EOF'
+---
+type: project
+slug: numeric-status
+name: Numeric Status
+client: NumCo
+strategic_impact: 3
+status: 2026
+open_roles: [lead]
+---
+# Numeric Status
+EOF
+write_entity "$PF10/projects/zero-status.md" <<'EOF'
+---
+type: project
+slug: zero-status
+name: Zero Status
+client: ZeroCo
+strategic_impact: 2
+status: 0
+open_roles: [lead]
+---
+# Zero Status
+EOF
+write_entity "$PF10/projects/sound.md" <<'EOF'
+---
+type: project
+slug: sound
+name: Sound Delivery
+client: SoundCo
+strategic_impact: 4
+status: active
+open_roles: [architect]
+---
+# Sound Delivery
+EOF
+# `project: sound` must name a real project — _compute filters on the project
+# before it reads an assignment's status, so an orphan assignment would never
+# reach the coercion and the fixture would pass against the unfixed script.
+write_entity "$PF10/assignments/numeric-assignment.md" <<'EOF'
+---
+type: assignment
+slug: rena--sound
+consultant: rena
+project: sound
+role: architect
+start_date: 2026-02-01
+end_date: 2026-08-01
+status: 1
+---
+# assignment
+EOF
+
+# The shared run() helper discards stderr, so invoke directly the way Fixture 6
+# does — 10g asserts on the absence of a traceback.
+STDERR10="$TMPROOT/stderr10.txt"
+LAST_JSON="$(python3 "$SCRIPT" "$PF10" 2>"$STDERR10" | tail -n 1)"
+
+assert_json "10a non-string status still succeeds" "d['success'] is True"
+assert_json "10b marked partial"                   "d['data']['partial'] is True"
+assert_json "10c falsy int status still warns" \
+  "any('non-string status' in w and 'Zero Status' in w for w in d['data']['warnings'])"
+assert_json "10d project surfaced by name" \
+  "any('non-string status' in w and 'Numeric Status' in w for w in d['data']['warnings'])"
+# The assignment is named by its relative path — _read_entities sets _file on
+# every kept entity, while slug is optional frontmatter.
+assert_json "10e assignment surfaced by file" \
+  "any('non-string status' in w and 'numeric-assignment.md' in w for w in d['data']['warnings'])"
+assert_json "10f every project still counted"      "d['data']['projects'] == 3"
+assert_no_traceback "10g no traceback on stderr" "$STDERR10"
+# A needle without quote characters: warnings reach the HTML through _esc, which
+# rewrites the quotes %r puts around a value.
+assert_html "10h healthy project still rendered" \
+  "Sound Delivery" "$PF10/output/dashboard.html"
 
 echo
 if [ "$failures" -eq 0 ]; then


claude-cogni-work:
## Ship-ready

Authored under the two-instance drain's author role and handed off for independent review. No in-session review verdict exists by design — the merger instance forms its own review contract from the issue and grades this diff against it.

**State:** draft, `svc:needs-review`, linked `Refs #1155` (three follow-ups deferred, so the parent stays open until the merger's merge and the reaping sweep close it).

**Where a reviewer should push hardest:**

1. **Is the altitude right?** This defends at two read sites. A reviewer could reasonably hold that the fix belongs at the parse boundary in `_read_entities`, which would close the whole class instead of one field — and the evidence for that view is in this PR: `role` still crashes the render the same way (#1162, with a repro). The case for shipping this anyway is that parse-boundary normalization touches every field on every entity type, well beyond this issue's acceptance criteria. If the merger disagrees, closing this in favor of #1162 is a legitimate outcome.
2. **The falsy-int guard.** `raw is not None` rather than truthiness is load-bearing; `status: 0` would otherwise normalize silently to `"unknown"`. Assertions `10c` and `10d` exist to pin both halves — check they actually would fail independently.
3. **The scope of the loop restructure.** The compound condition became an early-`continue`. Confirm this is behavior-preserving (the project filter already ran first via `and` short-circuit) and that the assignment warning still fires exactly once, not once per project.

**Verified before handoff:** all 10 fixtures pass; reverting only the script fails 7 of Fixture 10's 8 assertions with the exact reported `AttributeError`, so the test genuinely pins the bug; warning cardinality checked against a 3-project portfolio; both sibling suites pass; breadcrumb, version-bump, external-dispatch, and skill-name gates all clean.

claude-cogni-work:

_The merger formed this expectation from issue #1155 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The defect is entirely inside `_compute` in cogni-projects/scripts/render-dashboard.py: line 282 reads `status = (proj.get("status") or "").strip()` and line 288 reads `(a.get("status") or "").strip()` for assignments. validate-entities.py's `_coerce` (lines 120-124, applied at parse_frontmatter lines 159/171/176) turns a bare numeric scalar like `status: 2026` into an `int`, so `.strip()` raises AttributeError; that propagates out of `_compute` (called from main at line 515) into the `__main__` catch-all at lines 548-552 and collapses the whole render into the `unexpected failure:` envelope with exit 2 — violating the graceful-degradation contract stated in the module docstring (lines 11-15) and exercised by Fixture 6 of cogni-projects/tests/test-render-dashboard.sh (lines 250-342). A correct fix coerces at both read sites the way `_is_closed` (line 240) and staffing-score.py:331 already do (`str(...) .strip()`), and appends a per-entity warning to the `warnings` list `_compute` already threads (so it lands in `data.warnings[]` and flips `data.partial` via line 537) whenever the raw value was present but not a `str` — keying the guard on `is not None`/`isinstance` rather than truthiness, because `status: 0` is a falsy int a truthiness guard would silently normalize. Regression coverage belongs as a new fixture appended after Fixture 9 (test file lines 394-436) and before the summary block at lines 438-444, following Fixture 6's stderr-capture idiom (line 317) so the traceback class itself is asserted away. The projects-dashboard SKILL.md 'Partial snapshots are expected mid-authoring' bullet (lines 113-116) should name the new degradation class without restating renderer logic (its own note at lines 105-109 makes the script the source of truth).

**Acceptance checklist:**
- [ ] cogni-projects/scripts/render-dashboard.py no longer contains the pattern `(proj.get("status") or "").strip()` (base line 282): grep the changed file for `.strip()` applied to an un-`str()`-wrapped `.get("status")` and get zero hits.
- [ ] The assignment-side read site inside `_compute`'s `for a in assignments:` loop (base line 288, `(a.get("status") or "").strip()`) is coerced the same way — the identical crash is reachable from an assignment with `status: 1`; a diff that fixes only the project site leaves the bug live.
- [ ] The non-string status warning is appended to the same `warnings` list parameter `_compute` receives (so it reaches `data.warnings[]` and flips `data.partial` via line 537), not to a new list, a print, or a logging call.
- [ ] The warning message identifies the offending record: for a project it uses the existing `label` built at base line 278 (name/slug fallback); for an assignment it names the record via `_file` (set in `_read_entities` at base line 164) or its slug — grep the new format string for that interpolation.
- [ ] The presence/absence guard keys on `is not None` and/or `isinstance(raw, str)`, NOT on truthiness — grep the new lines to confirm no `if raw and ...` form, so a falsy `status: 0` still warns instead of being silently normalized to the empty string.
- [ ] render-dashboard.py stays stdlib-only: the import block (base lines 29-35, argparse/datetime/html/importlib.util/json/os/sys) gains no new third-party or heavyweight import, and `_coerce`/`parse_frontmatter` in validate-entities.py are NOT modified (that would change validator behavior repo-wide).
- [ ] Version-bump gate: no version line is touched — neither cogni-projects/.claude-plugin/plugin.json nor the repo-root .claude-plugin/marketplace.json appears among the changed files, per the post-merge bump convention.
- [ ] Maintainer-breadcrumb guard: the SKILL.md and any new comments introduce no leaked breadcrumbs — no `#NNN` issue/PR refs, version tags, or slice/finding codes (scripts/check-breadcrumbs.py passes on the changed files).
- [ ] SKILL.md frontmatter validity: cogni-projects/skills/projects-dashboard/SKILL.md keeps its `name: projects-dashboard`, `description`, and `allowed-tools: Read, Bash, Glob, Grep` frontmatter block (base lines 1-13) unchanged and still parses.
- [ ] cogni-projects/tests/test-render-dashboard.sh gains a NEW fixture block appended after the Fixture 9 block (base lines 394-436) and before the summary at base lines 438-444; no existing fixture body, seeded entity, assertion expression, or assertion label is modified — in particular Fixture 6's `6c valid projects still counted` count of 4 and `6e ... == 2` are untouched.
- [ ] The new fixture seeds a project whose frontmatter carries a bare numeric `status` (e.g. `status: 2026`) alongside at least one healthy project, and asserts `d['success'] is True` and `d['data']['partial'] is True` (issue AC-1).
- [ ] The new fixture asserts the offending project is *named* in the warnings — an `assert_json` whose expression matches both a status-related substring and the bad project's name or slug (e.g. `any('status' in w and '<slug>' in w for w in d['data']['warnings'])`) — not merely `len(warnings) >= 1`.
- [ ] The new fixture proves one bad record does not cost the snapshot: `d['data']['projects']` equals the number of seeded projects (bad one included) and the healthy project's name is asserted present in the written output/dashboard.html via `assert_html`.
- [ ] The new fixture captures stderr the way Fixture 6 does (base line 317: `python3 "$SCRIPT" "$PF" 2>"$STDERR" | tail -n 1`) and asserts no `Traceback` appears — the regression being guarded is precisely an uncaught AttributeError.
- [ ] The coerced status reaches the render path as text: the numeric status appears as `2026` in the project's Status cell of output/dashboard.html rather than the `unknown` fallback from base line 309, so the bad record is visible on the partner-facing surface (assert_html, or an equivalent envelope assertion).
- [ ] Assignment-side coverage: the fixture seeds an assignment with a numeric `status` (e.g. `status: 1`) whose `project` matches a seeded project slug, and asserts the render still succeeds, surfaces a warning for that assignment, and does not crash — a non-string assignment status must not silently pass the `ACTIVE_ASSIGNMENT_STATES` membership test at base line 288/72.
- [ ] The falsy-int edge is covered, not just the truthy one: a `status: 0` project (or an equivalent explicit assertion) still produces a warning rather than being normalized to the empty string.
- [ ] Absent/empty status still warns NOT at all: existing Fixture 1 assertions `1c not partial (clean)` and `1d no warnings`, and Fixture 9's `9d closed project emits no warning` / `9e clean snapshot`, are unchanged and still pass — a fix that warns on a missing `status:` key or an empty `status:` value makes every thin portfolio look broken.
- [ ] Closed-project semantics are unchanged: `_is_closed` (base line 240) and `_project_health` (base lines 204-222) are not rewritten, `status: closed` still yields the `closed` flag, and Fixture 9's `9c open roles exclude the closed project` and `9f tile agrees with the envelope` still pass — the deferred case-sensitivity unification stays out of scope.
- [ ] The SKILL.md edit is confined to the 'Partial snapshots are expected mid-authoring' bullet (base lines 113-116) and names the new degradation class (a type-invalid / non-string `status`); because this mechanism is prose-carried in a SKILL.md, the achievable bar is a grep for the added clause plus confirmation that the 'renderer is the source of truth for the flags and aggregates' note (base lines 105-109) is not contradicted — no behavioral test is demanded of the prose.
- [ ] The new fixture's assertion labels are unique and sequentially numbered in the file's existing convention (e.g. `10a`, `10b`, ...), so a failing assertion message points at exactly one check rather than colliding with an existing label.

**Risk watchlist:**
- A half-fix that coerces only the project read site (base line 282) and leaves the assignment site (base line 288) — the identical AttributeError is still reachable from `status: 1` on an assignment produced by the same `_coerce` path.
- A truthiness guard (`if raw and not isinstance(raw, str)`) that silently swallows `status: 0` / `status: -1`, normalizing a bad record to the empty string with no warning — exactly the 'silently normalized' failure the issue's proposed solution warns against.
- Warning on an absent or empty `status`, which turns every legitimately status-less project partial and breaks the clean-snapshot assertions in Fixtures 1 and 9.
- Wrapping the `_compute` body (or the whole per-project loop) in a try/except instead of fixing the read sites — the render stops crashing but the warning loses per-entity attribution and future type faults are swallowed rather than surfaced.
- Fixing upstream in validate-entities.py `_coerce` or adding a blanket 'stringify every frontmatter value' shim — that would defeat `_coerce_impact`'s non-numeric warning path and change validator behavior for register-entity.py and staffing-score.py too.
- A test that asserts only `len(d['data']['warnings']) >= 1`, which passes even when the warning names the wrong record or fires for an unrelated reason.
- A test that does not capture stderr, letting a traceback still print while the tail-of-stdout envelope looks healthy.
- Mutating an existing fixture's seeded entities or counts (e.g. Fixture 6's `projects == 4`) to make room for the new case, which misattributes any future failure.
- Rendering the coerced status as the `unknown` fallback so the malformed record disappears from the partner-facing table even though a warning exists.
- Scope creep into the deferred case-sensitivity unification of `_is_closed` vs `_project_health` vs the `_compute` warning gate — that silently changes what a `status: Closed` row renders and is explicitly deferred.
- SKILL.md drift: restating renderer logic (thresholds, exact warning strings) in prose, which the file's own 'renderer is the source of truth' note forbids and which rots on the next renderer change.
- Judgment-only (no gate can decide it): whether the chosen warning wording reads clearly to a partner relaying it verbally — a human should eyeball the exact phrasing against the existing warning voice in `_coerce_impact` and `_read_entities`.

- The plan-time acceptance criteria list 2 criteria (project-status render + a regression fixture), but a base-grounded regeneration derives ~21 gating items — including the second, identical crash site for assignment status at render-dashboard.py base line 288, the falsy `status: 0` case that a truthiness guard would silently normalize, and the must-not-regress clean-snapshot assertions in Fixtures 1 and 9. The acceptance bar was materially under-specified relative to what the base code requires.

sdh07:
## Human verdict: merge

The hold was `insufficient_coverage`. That escalation grades the **issue's** acceptance bar — #1155 stated 2 criteria where a base-grounded contract derives ~21 — not this diff. Graded the diff against the committed checklist by execution, not by reading:

- Full suite on the branch: 10 fixtures, 46 assertions, all pass.
- **The tests pin the bug.** Reverting only `render-dashboard.py` to base fails 7 of Fixture 10's 8 assertions with exactly `AttributeError: 'int' object has no attribute 'strip'`. `10g` passes in both states, as the PR body already discloses.
- **The falsy half is independently pinned.** Swapping the guard from `raw is None` to a truthiness form fails `10c` alone — `status: 0` cannot silently regress to `unknown`.
- **Warning cardinality holds.** 3 projects + 1 bad-status assignment emits exactly one assignment warning; the early-`continue` restructure is behaviour-preserving.
- **The coerced value reaches the partner-facing surface.** `status: 2026` renders `<td>2026</td>`, `status: 0` renders `<td>0</td>` — not the `unknown` fallback.
- **The disclosed adjacent defect is real and pre-existing.** A mixed `role: 2` / `role: lead` assignment set crashes the render with `TypeError` on base *and* on this branch. Not introduced here.
- Warning wording matches the file's existing register (`has a non-numeric strategic_impact %r — omitted from …`), which closes the contract's judgment-only item.

Two deviations from the letter of the contract, both accepted:

1. Fixture 6's `6g` inline traceback check was extracted into a shared `assert_no_traceback` helper — an existing fixture body was touched, but behaviour-preservingly, and the invariants the contract named (`6c` count of 4, `6e == 2`) are untouched.
2. "Coerced status appears in the HTML Status cell" is satisfied in behaviour but not asserted. Verified by hand above; not worth a CI cycle to add.

On altitude: the parse-boundary fix in `_read_entities` would close the whole class, and this PR argues its own case for that alternative honestly. It does not supersede this one — it is a materially wider blast radius across every field of every entity type, and it stays tracked separately. This lands; the class-level fix gets its own decision.

Follow-ups stay open and unclaimed by this merge.

Issue #1147, opened 2026-07-26
Roadmap: cogni-consult — a user-facing output register that governs the default path


sdh07:
PR #1139 added a lexicon layer and a German sister register to cogni-consult's output styles. Quality improved, but two live transcripts show the fix does not reach far enough. The doctrine exists and is good — `output-styles/strategy-advisor-de.md:35-36` ("Executive-Register: präzise, knapp, kein Füllmaterial") and `:60-68` ("Systemvokabular bleibt im System", with `complete` as a named negative example) — but it is **unreachable on the default path**: output styles are opt-in via `/config`, fixed at session start, and per `cogni-consult/CLAUDE.md` they do not load `references/`. Meanwhile every existing rule is scoped to prose (`strategy-advisor-de.md:52`, `:58`, `strategy-advisor.md:40` "in prose"), so tables are outside the rules by construction — which is why a clean-prose session still leaks slugs and engine values in its columns.

This roadmap moves the register from an opt-in voice file to a plugin-owned contract the skills themselves load, pursuing three goals: (1) stop leaking process terminology into user-facing text, (2) make output markedly shorter, and (3) describe steps in user terms — the skills currently undersell themselves, and consultants trust the system more when they understand what a step gives them. Goal (3) is a positive requirement currently covered by nothing, not a side effect of (1). The audience is the consultant (Berater), never the operator of the system.

The single criterion separating this epic from PR #1139: re-run both transcripts **with no output style active**.

## Tracks
- [ ] #1140 cogni-consult: rewrite the three seed table templates in consult-resume and consult-action-fields
- [ ] #1141 cogni-consult: move the interaction-language rule into a Step 0 gate and make Bash descriptions user copy
- [ ] #1142 cogni-consult: add references/user-facing-output.md as the register's single source of truth, loaded from all nine skills
- [ ] #1143 cogni-consult: cap the closing report at six lines and emit one ranked offer instead of several
- [ ] #1144 cogni-consult: plugin-owned anglicism lexicon, closed disposition vocabulary, Action Field → Handlungsfeld
- [ ] #1145 cogni-consult: split the slug contradiction between interaction-language.md (how) and the register (whether)
- [ ] #1146 cogni-consult: collapse the two output styles into one language-neutral stance file over the register reference

## Close gate
- [ ] #1140 cogni-consult: rewrite the three seed table templates in consult-resume and consult-action-fields
[...]

PR #1164, opened 2026-07-27
cogni-workspace: identifier clause governs spelling, not surfacing

### LINKED ISSUE #1148: cogni-workspace: the DE/EN CLAUDE.md template teaches identifiers may be surfaced, contradicting the consult register
## Plugin
cogni-workspace

## Current behavior
`cogni-workspace/assets/claude-templates/CLAUDE.de.md:8` states:

> Fachbegriffe, Code-Bezeichner und Dateinamen bleiben auf Englisch

`cogni-workspace/skills/manage-workspace/SKILL.md:117-126` copies that file to the workspace root as `CLAUDE.md` at init. It is therefore the **only** register-adjacent instruction that reliably loads in every session on the default path.

The rule is about orthography — do not translate an identifier — but it reads as permission to *surface* identifiers. That is the opposite of what a consultant-facing register wants, and it is why a German session cheerfully prints a `Slug` column.

## Desired behavior
Narrow the clause so it governs spelling, not existence:

> Bezeichner, Dateinamen und CLI-Befehle bleiben in ihrer englischen Form, **wenn sie genannt werden müssen**.

Apply the equivalent narrowing to the EN twin `CLAUDE.en.md`.

## Motivation
Filed as a standalone issue because epic #1147 was deliberately scoped to cogni-consult only. But this line materially limits how far that epic's children can reach: they install a register in cogni-consult while the workspace root keeps teaching the opposite. The two are worth landing near each other.

This is a wording change to a template, not a behavior change to any skill.

## Acceptance criteria
- [ ] `CLAUDE.de.md:8` carries the "wenn sie genannt werden müssen" qualifier.
- [ ] `CLAUDE.en.md` carries the equivalent English qualifier.
- [ ] Neither file's other behavioral anchors change.
- [ ] A workspace re-init (or `manage-workspace` update) propagates the corrected template, since `SKILL.md:251-253` overwrites it as plugin-managed.

## Related
- Epic #1147 — cogni-consult user-facing output register (the doctrine this template contradicts).

claude-cogni-work:
Closes #1148.

## Summary

- **`assets/claude-templates/CLAUDE.de.md:8`** — appended the qualifier: `- Fachbegriffe, Code-Bezeichner und Dateinamen bleiben in ihrer englischen Form, wenn sie genannt werden müssen`. The clause now governs orthography (do not translate an identifier) instead of reading as standing permission to print one.
- **`assets/claude-templates/CLAUDE.en.md:7`** — equivalent English narrowing: `- Technical terms, code identifiers, and filenames remain in their English form when they must be named`.
- **`assets/output-styles/workspace-de.md:11`** — carried the identical defective sentence verbatim; narrowed the same way.
- **`skills/manage-workspace/SKILL.md`** Update Mode Step 4 — added a CLAUDE-template refresh so existing workspaces actually receive the correction.

## Two corrections to the issue as filed

**1. Acceptance criterion 4 was false, and its anchor pointed at the wrong thing.** The criterion reads "a `manage-workspace` update propagates the corrected template, since `SKILL.md:251-253` overwrites it as plugin-managed". That anchor is Update Mode Step 4 "Update Output Styles and Theme Template" — the "plugin-managed, not user-customized" language governs **output styles**, and Update Mode referenced `CLAUDE.md` nowhere at all. Every claude-template reference lived in Init Mode (`SKILL.md:113-134`), which runs only when `.workspace-config.json` is **absent**. Any workspace old enough to need this fix already has that file, so it goes through Update Mode and would never have received the correction. A prose-only edit could not have satisfied the criterion, so this PR adds the missing step.

**2. The issue's proposed replacement sentence would have loosened the rule, not narrowed it.** It drops `Fachbegriffe`, narrows `Code-Bezeichner` to `Bezeichner`, and adds `CLI-Befehle`. Dropping `Fachbegriffe` removes technical terms from the rule's coverage — a change in what the rule governs, not in how tightly it governs it, and in tension with acceptance criterion 3 ("neither file's other behavioral anchors change"). The original three nouns are kept in both languages; only the qualifier is appended.

## Scope decisions

**Included a file the criteria do not name.** `assets/output-styles/workspace-de.md:11` is the same sentence character-for-character, in the same plugin — and it is the surface Update Mode *already* refreshed on every update. Fixing the template while leaving the output style teaching the opposite would have reproduced, at plugin scale, exactly the contradiction this issue is about. Its EN twin `workspace-en.md` has no identifier bullet at all, so there is nothing to narrow there; that pre-existing DE/EN asymmetry is left alone.

**The propagation step is deliberately non-destructive.** Update Mode now unconditionally refreshes `.claude/templates/CLAUDE.{en,de}.md` — an unambiguously plugin-managed cache nobody hand-edits, which `cogni-workspace/templates/obsidian/plugins/terminal/workspace-orchestrator.sh:84-95` copies to the workspace root on every per-session language switch. The workspace-root `CLAUDE.md` is compared rather than clobbered, defaulting to keep. See the open question.

## Test plan

- [x] `git diff --numstat` shows exactly **+1/-1** on each of the three asset files — no other bullet touched.
- [x] Real `ä/ö/ü/ß` throughout both DE files (`müssen` included); grep for `muessen` / `\uXXXX` escapes returns nothing.
- [x] `check-breadcrumbs.py` 0 violations · `check-version-bump.py` `[ok]` (14 plugins scanned, 0 touched, 0 violations) · `check-external-dispatch.py` 0 violations · `check-skill-names.sh` OK.
- [x] `git diff origin/main -- '*/.claude-plugin/*.json'` is empty — no version line; the post-merge workflow owns the bump.
- [x] `manage-workspace/SKILL.md` body is 294 lines against the 500-line soft cap after +11 net.
- [ ] **Manual, not run here:** against a scratch workspace with a stale `.claude/templates/CLAUDE.de.md` and a hand-edited root `CLAUDE.md`, run `manage-workspace` update and confirm the templates are refreshed while the root file stays byte-identical unless the user opts in.
- [ ] **Manual, not run here:** in that workspace, run the Obsidian launcher language switch and confirm the root `CLAUDE.md` picks up the corrected sentence.

## Open questions

**Is the workspace-root `CLAUDE.md` plugin-managed or user-owned?** Nothing in cogni-workspace states an ownership contract for it, and no checksum, version stamp, or drift detection covers it — all such machinery is theme-file-only. It is also the natural place for a user to add project instructions on top of the language rules. This PR therefore takes the safe path: refresh the plugin-managed template cache unconditionally, diff-and-ask before touching the root file, defaulting to keep.

Evidence pulling the other way, found while validating the plan: `copy_claude_template` in the Obsidian launcher (`workspace-orchestrator.sh:84-95`) **already** overwrites the workspace-root `CLAUDE.md` unconditionally on every language switch — no diff, no prompt. So the root file is arguably already treated as plugin-managed in practice, and this PR's prompt is *stricter* than shipped behavior. If that is the intent, the prompt collapses to an unconditional `cp` and acceptance criterion 4 holds with no user interaction.

This is a product-ownership go/no-go that cannot be answered from the repository, which is why this PR is a draft.

### DIFF
--- a/cogni-workspace/assets/claude-templates/CLAUDE.de.md
+++ b/cogni-workspace/assets/claude-templates/CLAUDE.de.md
@@ -5,4 +5,4 @@
 - Auf Deutsch antworten, es sei denn, der Benutzer wechselt die Sprache
 - Korrekte deutsche Rechtschreibung verwenden, einschließlich aller Umlaute (ä, ö, ü, Ä, Ö, Ü) und Eszett (ß)
 - Keine Umschreibungen wie "ae" für "ä", "oe" für "ö", "ue" für "ü" oder "ss" für "ß" verwenden
-- Fachbegriffe, Code-Bezeichner und Dateinamen bleiben auf Englisch
+- Fachbegriffe, Code-Bezeichner und Dateinamen bleiben in ihrer englischen Form, wenn sie genannt werden müssen
--- a/cogni-workspace/assets/claude-templates/CLAUDE.en.md
+++ b/cogni-workspace/assets/claude-templates/CLAUDE.en.md
@@ -4,4 +4,4 @@
 
 - Respond in English unless the user switches language
 - Use concise, professional language
-- Technical terms, code identifiers, and filenames remain in English
+- Technical terms, code identifiers, and filenames remain in their English form when they must be named
--- a/cogni-workspace/assets/output-styles/workspace-de.md
+++ b/cogni-workspace/assets/output-styles/workspace-de.md
@@ -8,7 +8,7 @@
 - Bei Workspace-Pfaden Umgebungsvariablen verwenden (z.B. `$COGNI_RESEARCH_ROOT`) statt absolute Pfade
 - Bei Dateioperationen relative Pfade vom Workspace-Root anzeigen
 - Bei Multi-Plugin-Operationen angeben, welches Plugin welches Artefakt besitzt
-- Fachbegriffe, Code-Bezeichner und Dateinamen bleiben auf Englisch
+- Fachbegriffe, Code-Bezeichner und Dateinamen bleiben in ihrer englischen Form, wenn sie genannt werden müssen
 
 ## Intent Router
 
--- a/cogni-workspace/skills/manage-workspace/SKILL.md
+++ b/cogni-workspace/skills/manage-workspace/SKILL.md
@@ -248,9 +248,20 @@ bash ${CLAUDE_PLUGIN_ROOT}/scripts/install-workspace-deps.sh --force
 Same fail-soft contract as Init Mode Step 3.5: warn and continue if `python3` /
 the `venv` module / the network is unavailable — never block the update.
 
-### 4. Update Output Styles and Theme Template
+### 4. Update Output Styles, CLAUDE.md Templates, and Theme Template
 
 Copy latest output-style files from `${CLAUDE_PLUGIN_ROOT}/assets/output-styles/` to `.claude/output-styles/`, overwriting existing ones (these are plugin-managed, not user-customized).
+
+Refresh both language templates in `.claude/templates/`. Like the output styles these are plugin-managed — the Obsidian Terminal launcher copies from them to the workspace root on every per-session language switch, so a correction to a template only reaches an existing workspace if this step runs. Init Mode installs them; without the refresh here they stay at whatever version was current when the workspace was created:
+
+```bash
+mkdir -p "${WORKSPACE_DIR}/.claude/templates"
+cp "${CLAUDE_PLUGIN_ROOT}/assets/claude-templates/CLAUDE.en.md" \
+   "${CLAUDE_PLUGIN_ROOT}/assets/claude-templates/CLAUDE.de.md" \
+   "${WORKSPACE_DIR}/.claude/templates/"
+```
+
+Do **not** overwrite the workspace-root `CLAUDE.md` here. Nothing declares it plugin-managed the way the output styles are declared above, and it is the natural place for a user to add project-specific instructions on top of the language rules. Compare it against `.claude/templates/CLAUDE.${LANGUAGE}.md` instead; if they differ, show the difference and ask whether to adopt the updated template, keep the current file, or merge by hand — defaulting to keeping the current file.
 
 Refresh `_template/theme.md` from `${CLAUDE_PLUGIN_ROOT}/themes/_template/`. Preserve all user-created themes.
 


claude-cogni-work:
## Open question resolved — the propagation step is the right direction

The PR body asks whether the workspace-root `CLAUDE.md` is plugin-managed or user-owned, and raises the possibility that this whole copy mechanism is vestigial and should be removed rather than extended. That was investigated. It should **not** be removed, and this PR's Update Mode step stands.

**Why.** The template is redundant in *content* — the output style `workspace-de.md` is a strict superset of `CLAUDE.de.md` — but not in *activation*, and activation is the point:

- `CLAUDE.md` is auto-loaded as project context every session with no user action.
- Output styles are opt-in and fixed at session start, and **nothing in this repository ever activates one**. `generate-settings.sh` writes only an `env` object into `.claude/settings.local.json`; no `outputStyle` key exists anywhere in the monorepo. The sole mention of activation is `templates/obsidian/plugins/terminal/workspace-orchestrator.sh:134`, which *echoes* `Activate with: /output-style workspace-${LANGUAGE}` and never invokes it.

So `CLAUDE.md` is currently the only surface that reliably carries the language and orthography rules into a session. Removing the copy step would mean a fresh session in a newly-created or newly-language-switched workspace silently falls back to default behavior — a regression, not a cleanup. Extending Update Mode so existing workspaces actually receive template corrections, which is what this PR does, is repairing that backstop.

**This also settles the ownership question in the direction the PR chose cautiously.** Because `CLAUDE.md` is load-bearing for language behavior, the plugin does have a legitimate claim to its content — which is consistent with `copy_claude_template` already overwriting it unconditionally on every language switch. A reviewer who wants acceptance criterion 4 to hold with no user interaction can collapse this PR's diff-and-ask into an unconditional `cp`; that is now a defensible simplification rather than a risk.

**One defect surfaced by the investigation, filed separately as #1165:** the two copies of the language rules have already drifted. The output style has gained env-var path rules, relative-path display, ownership attribution, and an Intent Router table that never reached the template — and the wording defect this PR fixes had to be corrected in both files independently, because nothing connects them. #1165 proposes either collapsing the template to a pointer, or the sequenced work that would make removal safe (write `outputStyle` directly, verify a fresh session honors it, then retire the copy).

Repo-wide grep confirms nothing else depends on the workspace-root `CLAUDE.md` or `.claude/templates/` — no other plugin, and no `workspace-status` / `workspace-dashboard` foundation check.

claude-cogni-work:
## Ship-ready

Authored under the two-instance drain's author role and handed off for independent review. No in-session review verdict exists by design — the merger instance forms its own review contract from the issue and grades this diff against it.

**State:** draft, `svc:needs-review`, `Closes #1148`. It stays draft because of a genuine escalation, which the comment above now resolves in favor of the direction taken.

**Where a reviewer should push hardest:**

1. **Two acceptance criteria were wrong as filed, and the PR departs from both.** AC 4's anchor pointed at the output-style step, and the criterion was false — Update Mode never touched `CLAUDE.md` at all, so no existing workspace could ever have received the fix. And the issue's own proposed replacement sentence drops `Fachbegriffe` and adds `CLI-Befehle`, which loosens the rule's coverage rather than narrowing its scope; the original nouns were kept instead. A reviewer should confirm both departures are right rather than assume the issue text was authoritative.
2. **A fourth file was changed that the criteria do not name.** `assets/output-styles/workspace-de.md:11` carried the identical sentence and is the surface an update already refreshes. Confirm including it is in scope.
3. **The diff-and-ask on the workspace-root `CLAUDE.md` is stricter than shipped behavior.** The Obsidian launcher already clobbers that file unconditionally on every language switch. Collapsing the prompt to an unconditional `cp` is a defensible simplification — see the comment above.

**Verified before handoff:** exactly +1/-1 on each of the three asset files; real `ä/ö/ü/ß` throughout with no ASCII substitutes or unicode escapes; breadcrumb, version-bump, external-dispatch, and skill-name gates all clean; no version line in the diff; `manage-workspace/SKILL.md` at 294 body lines against the 500-line cap after +11 net.

**Not verified — needs a human with a real workspace:** the two manual checks in the test plan (that an update refreshes `.claude/templates/` while leaving a hand-edited root `CLAUDE.md` byte-identical, and that a launcher language switch then picks up the corrected sentence). Both are behavioral claims about the Obsidian launcher that cannot be exercised from a diff.

claude-cogni-work:

_The merger formed this expectation from issue #1148 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix is a prose/asset edit confined to cogni-workspace, with no executable behaviour change. In the base clone the defective sentence exists in exactly three places (verified by grep): `cogni-workspace/assets/claude-templates/CLAUDE.de.md:8` (`- Fachbegriffe, Code-Bezeichner und Dateinamen bleiben auf Englisch`), its EN twin `cogni-workspace/assets/claude-templates/CLAUDE.en.md:7` (`- Technical terms, code identifiers, and filenames remain in English`), and a character-for-character duplicate at `cogni-workspace/assets/output-styles/workspace-de.md:11`. The fix should rewrite those bullets so the rule governs *form* under a condition ('… bleiben in ihrer englischen Form, wenn sie genannt werden müssen' / '… remain in their English form when they must be named') while leaving every other line byte-identical — CLAUDE.de.md lines 1-7, CLAUDE.en.md lines 1-6, workspace-de.md lines 1-10 and 12-26. The DE twin `workspace-en.md` carries no such bullet (its Behavioral Anchors are lines 5-9), so it correctly stays untouched, and `templates/obsidian/plugins/terminal/workspace-orchestrator.sh:84-95` (`copy_claude_template`) already re-copies `.claude/templates/CLAUDE.<lang>.md` onto the workspace root on every language switch, so no script change is needed. The issue's AC 4 is false against base: `skills/manage-workspace/SKILL.md:251-255` is Update Mode Step 4 'Update Output Styles and Theme Template' and its 'plugin-managed, not user-customized' clause governs output styles only; every claude-template reference lives in Init Mode Step 4 (SKILL.md:113-134), which runs only when `.workspace-config.json` is absent. So a correct fix must additionally extend Update Mode Step 4 with an explicit CLAUDE.md-template refresh that uses Update Mode's own `${WORKSPACE_DIR}` (SKILL.md:230, 268) rather than Init Mode's `${TARGET_DIR}`, and must state an explicit disposition for the workspace-root `CLAUDE.md` — noting that Update Mode's backup set (SKILL.md:202-207) and the Rollback block (SKILL.md:286-294) cover only `.workspace-config.json`, `.claude/`, and `.workspace-env.sh`, so an unconditional root overwrite would be unrecoverable. The mechanism is prose-carried (asset templates plus a SKILL.md body), so no behavioural test is achievable or demanded here; the gradable bar is exact-string grep anchors plus scope containment.

**Acceptance checklist:**
- [ ] Scope containment: the diff touches only `cogni-workspace/assets/claude-templates/CLAUDE.de.md`, `cogni-workspace/assets/claude-templates/CLAUDE.en.md`, `cogni-workspace/assets/output-styles/workspace-de.md`, and `cogni-workspace/skills/manage-workspace/SKILL.md` — confirm no other path appears in the changed-file list.
- [ ] `cogni-workspace/assets/output-styles/workspace-en.md` is NOT modified: its Behavioral Anchors block (lines 5-9 in base) carries no identifier bullet, so adding one would be scope creep rather than a narrowing.
- [ ] `cogni-workspace/templates/obsidian/plugins/terminal/workspace-orchestrator.sh` is NOT modified — `copy_claude_template()` (base lines 84-95) already copies `.claude/templates/CLAUDE.${lang}.md` to `$WORKPLACE_ROOT/CLAUDE.md`, so template propagation to the root needs no script edit.
- [ ] No plugin version was touched: neither `cogni-workspace/.claude-plugin/plugin.json` nor the repo-root `.claude-plugin/marketplace.json` appears in the diff. Adjudicate by running the version-bump gate (`python3 scripts/check-version-bump.py`) and confirming status is not `version-touched` / `version-mirror-desync`.
- [ ] No maintainer breadcrumbs were introduced into the changed SKILL.md: the new Update Mode prose contains no `#1147`/`#1148`-style issue refs, no `vX.Y.Z` tag, no `M<n>`/`Slice <n>`/`F<nn>` code. Adjudicate by running the maintainer-breadcrumb guard (`python3 scripts/check-breadcrumbs.py`), which globs `*/skills/*/SKILL.md`, and confirming exit 0.
- [ ] `manage-workspace` SKILL.md frontmatter is intact and unchanged in its identity fields: `name: manage-workspace`, the `description` block, and `allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion`. Adjudicate the naming half by running the skill-name check (`bash cogni-workspace/scripts/check-skill-names.sh`) and confirming exit 0; confirm `description` is byte-identical so no README section-4 regeneration is owed.
- [ ] `CLAUDE.de.md` still opens with `# Workspace-Anweisungen` and `## Sprache` and still contains exactly four `- ` bullets — the structural shape of base lines 1-8 is preserved.
- [ ] `CLAUDE.en.md` still opens with `# Workspace Instructions` and `## Language` and still contains exactly three `- ` bullets — the structural shape of base lines 1-7 is preserved.
- [ ] `workspace-de.md` still contains its three `##` sections (`## Verhaltensanker`, `## Intent Router`, `## Sprachpräferenz`), exactly seven `- ` bullets under Verhaltensanker, and the three-row Intent-Router table (manage-workspace / manage-themes / workspace-status).
- [ ] `CLAUDE.de.md` line 8 carries the conditional qualifier verbatim: grep the file for the literal substring `wenn sie genannt werden müssen` and confirm one hit on the identifier bullet. (Prose-carried mechanism — no behavioural test is possible; the documented mutation check is that deleting this substring must make this item fail.)
- [ ] `CLAUDE.en.md` carries the equivalent English conditional qualifier on its identifier bullet — a clause of the form `when they must be named` (or a semantically identical conditional such as `when they need to be named`). Confirm the bullet now scopes the rule to the *form* of an identifier that is named, not to whether identifiers may be surfaced. A bare restatement without a condition fails this item.
- [ ] DE/EN twin parity: the reworked `CLAUDE.de.md` and `CLAUDE.en.md` bullets cover the same noun set as each other (DE `Fachbegriffe/Code-Bezeichner/Dateinamen` ↔ EN `technical terms/code identifiers/filenames`, or an identically-shaped pair on both sides). If either file's noun set was narrowed or widened relative to base, the same change must appear on both sides — a one-sided noun-list change is a divergence defect.
- [ ] `cogni-workspace/assets/output-styles/workspace-de.md` line 11 receives the same narrowing: after the diff, its identifier bullet and `CLAUDE.de.md`'s identifier bullet are character-for-character identical (they are identical in base). A stale duplicate here reinstates the exact contradiction the issue is closing, because Update Mode Step 4 refreshes output styles on every update.
- [ ] German orthography convention holds in every changed DE string: grep the three DE-touching files for `muessen`, `ae`/`oe`/`ue` transliterations of the changed bullet, and any `\u00` escape sequence — all must be absent, and `müssen` must appear with a real `ü`.
- [ ] Non-regression of the other behavioral anchors (issue AC 3): `CLAUDE.de.md` lines 1-7 and `CLAUDE.en.md` lines 1-6 are byte-identical to base, and `workspace-de.md` lines 1-10 plus 12-26 are byte-identical to base. Exactly one line changed per asset file.
- [ ] `manage-workspace` SKILL.md Update Mode Step 4 (base lines 251-255) gains an explicit CLAUDE.md-template propagation step: the new text names the source `${CLAUDE_PLUGIN_ROOT}/assets/claude-templates/` and the destination `.claude/templates/`, so a workspace that already has `.workspace-config.json` (and therefore never re-runs Init Mode Step 4) actually receives the corrected template. Both existing Step-4 sentences (output styles; `_template/theme.md` refresh preserving user themes) survive verbatim.
- [ ] Any bash added to Update Mode uses Update Mode's own workspace variable `${WORKSPACE_DIR}` (as at base SKILL.md:230 and :268) and `${CLAUDE_PLUGIN_ROOT}` — NOT Init Mode's `${TARGET_DIR}`, which is undefined in the Update Mode flow. A copy-paste of the Init Mode Step 4 block carrying `${TARGET_DIR}` fails this item.
- [ ] The new Update Mode text states an explicit, unambiguous disposition for the **workspace-root** `CLAUDE.md` — either 'overwritten unconditionally (plugin-managed)' or 'compared and the user asked, defaulting to keep'. Silence, or wording that leaves a reader unable to tell whether the root file is clobbered, fails this item.
- [ ] Backup/rollback consistency for whichever disposition was chosen: if the root `CLAUDE.md` is overwritten unconditionally, Update Mode Step 1's backup commands (base SKILL.md:202-207, which copy only `.workspace-config.json`, `.claude/`, `.workspace-env.sh`) AND the `## Rollback` block (base SKILL.md:286-294) must be extended to include `CLAUDE.md`; if instead the user is asked before touching it, no backup change is required and the already-present `AskUserQuestion` in `allowed-tools` must be the stated mechanism.
- [ ] Init Mode is untouched: SKILL.md lines 113-143 (Step 4 'Install Output Styles, CLAUDE.md Templates, and Theme Template', including both `cp` blocks and the `.claude/templates` mkdir) are byte-identical to base — Init already copies the corrected asset, so no Init-side change is warranted.
- [ ] Cross-reference integrity after the edit: Update Mode's step headings still run `1, 2, 3, 3.5, 4, 5, 6, 7`, and the three in-body pointers still resolve — 'Update Mode step 5' (base line 179), 'Init Mode step 5' (base line 259), and 'Init Mode Step 3.5' (base line 248). Any renumbering must be accompanied by matching updates to these three references.
- [ ] Style parity in the rewritten bullets: no trailing period is added (no bullet in `CLAUDE.de.md`, `CLAUDE.en.md`, or `workspace-de.md`'s anchor list ends with one in base), and each file keeps a single trailing newline.

**Risk watchlist:**
- The DE and EN twins drift: the author applies the issue's literal replacement sentence to DE (which silently drops `Fachbegriffe`, narrows `Code-Bezeichner` to `Bezeichner`, and adds `CLI-Befehle`) but only appends a qualifier to EN — leaving the two templates covering different noun sets. Note the issue's own proposed DE sentence *loosens* coverage by dropping `Fachbegriffe`, which is in tension with its own AC 3; a one-sided application is worse than either choice made consistently.
- The `workspace-de.md:11` duplicate is left stale because the issue's acceptance criteria never mention it. That is the single highest-value miss here: the DE output style is the surface Update Mode already refreshes on every update, so a stale copy re-teaches the defect to every existing workspace even after the templates are fixed.
- Umlaut mangling: the changed bullets are rewritten through a tool that emits `\u00fc` escapes or ASCII transliteration (`muessen`), violating the repo-wide encoding convention — a known failure mode when files are round-tripped through a JSON serializer rather than edited surgically.
- A prose-only edit is shipped with no Update Mode change, so the fix reaches only brand-new workspaces. Every workspace old enough to carry the defective template has `.workspace-config.json` and therefore goes through Update Mode, which references CLAUDE.md nowhere in base — the issue's AC 4 would be nominally 'checked' while being factually unsatisfiable.
- The Update Mode addition unconditionally `cp`s over the workspace-root `CLAUDE.md`, destroying user-added workspace instructions with no backup entry and no rollback path (Step 1 and the Rollback block cover only `.workspace-config.json`, `.claude/`, `.workspace-env.sh`).
- Collateral edits smuggled into `manage-workspace` SKILL.md under cover of a one-step addition — rewording of Init Mode, renumbering that silently breaks the three in-body step cross-references, or an `allowed-tools`/`description` frontmatter change that would drag README section 4 out of sync.
- Over-reach into `workspace-en.md` or `workspace-orchestrator.sh`: adding a new EN output-style bullet or changing the launcher's copy semantics turns a wording fix into a behaviour change the issue explicitly disclaims ('a wording change to a template, not a behavior change to any skill').
- Judgment-only, not gated: whether the new German sentence actually *reads* as an orthography rule to a native speaker rather than as a subtly reworded permission slip. No grep can decide this — a human should read the final DE bullet aloud against the consultant-facing register the epic is installing.
- Judgment-only, not gated: whether narrowing the workspace-root template is sufficient to stop a German session printing a `Slug` column, or whether the real fix lives in the consult register the related epic owns. This PR should not be graded on that downstream outcome.

- Is the workspace-root `CLAUDE.md` plugin-managed or user-owned? Nothing in cogni-workspace states an ownership contract for it, and no drift/checksum machinery covers it (`check-theme-drift.py` is theme-file-only). Base evidence pulls both ways: `templates/obsidian/plugins/terminal/workspace-orchestrator.sh:84-95` already overwrites it unconditionally on every per-session language switch with no diff and no prompt, while Update Mode's backup set (`skills/manage-workspace/SKILL.md:202-207`) and the Rollback block (lines 286-294) do not cover it, so an unconditional Update Mode overwrite would be unrecoverable. This is a product decision the issue leaves open; the checklist gates that a disposition is stated and made backup-consistent, not which disposition is chosen.
- Acceptance criterion 4's anchor is false against base. It asserts `SKILL.md:251-253` 'overwrites it as plugin-managed', but those lines are Update Mode Step 4 'Update Output Styles and Theme Template' and the plugin-managed clause governs output styles; Update Mode references CLAUDE.md nowhere, and every claude-template reference lives in Init Mode Step 4 (lines 113-134), which runs only when `.workspace-config.json` is absent. As written, AC 4 is unsatisfiable by a prose-only template edit — a human should confirm the intended reading (add propagation to Update Mode) before this criterion is treated as met.
- The plan-time acceptance criteria list 4 items, but a base-grounded regeneration derives 22 gating items — including the character-for-character duplicate at `assets/output-styles/workspace-de.md:11` (never named in the AC, yet the surface Update Mode already refreshes), the repo-wide real-umlaut encoding constraint, the `${WORKSPACE_DIR}` vs `${TARGET_DIR}` variable-scope hazard in Update Mode, the backup/rollback consequence of touching the workspace-root `CLAUDE.md`, and the CI gates (breadcrumb guard, version-bump gate) that a SKILL.md edit puts in scope. The acceptance bar was materially under-specified.

sdh07:
## Human verdict: merge

Three escalations. Two are settled by base evidence; the third was mine to answer.

**AC 4's anchor was false, and the PR is right to depart from it.** Confirmed against base: Update Mode referenced `CLAUDE.md` nowhere at all — every claude-template reference lived in Init Mode Step 4, which runs only when `.workspace-config.json` is absent. Any workspace old enough to carry the defective sentence goes through Update Mode, so a prose-only edit could never have satisfied the criterion. The added Step 4 propagation is what makes AC 4 satisfiable, not scope creep.

**`insufficient_coverage` grades the issue's bar, not this diff.** Graded all 22 contract items by inspection of the trees rather than by reading the diff:

- Scope containment holds — exactly the four named files; `workspace-en.md` and `workspace-orchestrator.sh` untouched; exactly +1/-1 per asset file.
- **The duplicate is kept in lockstep** — `CLAUDE.de.md:8` and `workspace-de.md:11` are character-for-character identical after the diff. That was the highest-value item on the risk watchlist and it passes.
- Structural shape preserved on all three assets (headings, bullet counts, the three `##` sections and the 3-row Intent Router). No trailing period added; single trailing newline each.
- Encoding clean — real `ü` in `müssen`, no transliteration, no `\u00` escapes anywhere under `assets/`.
- SKILL.md: frontmatter byte-identical, Init Mode byte-identical, step numbering `1,2,3,3.5,4,5,6,7` intact with all three in-body cross-references still resolving, and the new bash correctly uses Update Mode's `${WORKSPACE_DIR}` rather than Init Mode's `${TARGET_DIR}`. `${LANGUAGE}` is already in scope there too.
- **Backup consistency comes for free under the chosen disposition** — Step 1 backs up `.claude/` wholesale, so the new `.claude/templates/` refresh is already covered by the existing backup and Rollback blocks. Nothing owed.
- The Update Mode heading rename mirrors Init Mode's; repo-wide grep finds no reference to the old heading text.

**The workspace-root `CLAUDE.md` is adjudicated user-owned.** The PR's compare-and-ask stands as written. The argument for collapsing it to an unconditional `cp` — that the launcher already clobbers the file — is what decided it the other way: `copy_claude_template` is called from `launch_claude`, so it fires on **every** launcher start, not merely on a language switch, and it overwrites the root file with no backup. Any project instructions a user added there are already being destroyed each session. That is a defect in the launcher, not a model for `manage-workspace` to copy.

Two things filed out of this review:

- A new issue to rethink, and possibly delete, the launcher's per-session copy and the `.claude/templates/` cache — including that clobber. `.claude/templates/` has exactly one reader, so if the copy goes, the cache goes, and this PR's Update Mode step goes with it.
- A correction on #1165: its option B (write `outputStyle` into `settings.local.json`, then retire the template) is not viable. Output styles cannot be activated by a plugin — a non-default style replaces the system prompt, and the key is project-global and resolved at session start. `CLAUDE.md` is the surface that reliably loads; the output style is the vestigial one.

Known-unverified and carried forward, as the PR itself flags: both manual checks need a real workspace and cannot be exercised from a diff.

PR #1167, opened 2026-07-27
cogni-projects: a non-string role no longer fails the whole render

### LINKED ISSUE #1162: cogni-projects: a non-string assignment role still crashes the whole dashboard render

claude-cogni-work:
Closes #1162.

## Summary

A hand-edited `role: 2` cost the entire dashboard. The frontmatter parser this renderer borrows from `validate-entities.py` int-coerces any all-digit scalar before the schema enum is ever checked, and the renderer never runs that validator — so `_compute`'s `sorted(covered)` was asked to order an `int` against a `str`, raised, and the module-level catch-all discarded every warning collected so far. This is the same contract inversion already fixed for `status`, reached through `role` instead.

- Generalize the existing read-site guard into `_text_field(raw, field, label, warnings)` and route the assignment `role` through it. `None` returns `""` silently; any other non-string warns and coerces. Like the original it keys on `raw is not None`, not truthiness, so a falsy `role: 0` is surfaced rather than dropped without a word.
- **Normalize `open_roles` entries the same way.** This is the load-bearing half. They pass through the same int coercion, and the fill checks are membership tests that never raise — they just answer wrongly. A one-sided fix would have converted a today-correct `2026 in [2026]` match into `"2026" in [2026]`, silently reporting a filled role as open on a partner-facing staffing number.
- Regression coverage as Fixture 11, following the Fixture 10 pattern.
- One clause in the skill's Notes bullet so the warn-and-degrade contract documents a non-text role label alongside a non-text status.

## Why both sides, not just the one that crashed

This is the finding that shaped the change. An exploration pass patched a throwaway copy to stringify `role` alone and ran the symmetric fixture:

```
"open_roles": 1,
"warnings": [
  "assignment assignments/symmetry-role.md has a non-string role 2026 — coerced to text; fix the record",
  "project Symmetry Check: assignment role '2026' matches no open_roles label — possible label mismatch"
]
```

`roles_filled` flipped from 1 to 0 and a spurious label-mismatch warning appeared. That trades a loud crash for a quiet wrong number, which is worse. Fixture 11 carries a second project specifically to pin it: assertions 11f/11g/11k fail if either side is left uncoerced.

## Scope

Takes the issue's **option 1** (read-site defense), symmetrically. `_coerce` in `validate-entities.py` is untouched, per the issue's explicit constraint — its int coercion is wanted for numeric fields and the parser is shared.

**Option 2 (normalize every text field at the `_read_entities` parse boundary) stays filed as its own change**, as the issue itself scoped it: it decides the string-typed field set for every entity type. The helper's docstring carries that signpost, and now also names `staffing-score.py` — it reads the same parser and, on a numeric role label, scores every consultant at zero profile fit and echoes the raw int into the shortlist rather than raising. Whoever picks that change up needs to know there is a second consumer.

A quality pass folded `_status_text` into a delegate of the new helper. It turned out to be the same function with the field name hard-coded, so keeping both would have meant two copies of one body and one operator-facing warning template. Warning strings are byte-identical; fixtures 10c/10d/10e pass unchanged.

## Accepted behaviour change

A numeric `open_roles` entry now emits a warning, so an all-numeric portfolio that rendered clean before becomes `partial: true`. Kept rather than coerced silently — it mirrors the existing "fix the record" contract, and only fires on records already off-schema, where a numeric role label is pathological rather than idiomatic.

## Verification

- `bash cogni-projects/tests/test-render-dashboard.sh` — green, 59 assertions (48 existing + 11 new).
- `test_register_entity.sh` and `test_portfolio_init.sh` — green.
- **Mutation-tested, both directions.** Reverting both call sites fails 10 of Fixture 11's assertions with the reported `TypeError`. Reverting *only* the `open_roles` normalization fails 4 — including 11f reading `open_roles: 1` instead of `0` — so the symmetry guard genuinely bites and is not decoration.
- The issue's exact repro (`open_roles: [lead]`, assignments `role: 2` and `role: lead`) now returns `success: true`, `partial: true`, names the offending record in `warnings[]`, and writes `dashboard.html`.
- Breadcrumb guard: 0 violations. Version-bump gate: 0 plugins touched. Skill-name check: OK.
- Performance measured on a 300-project x 6000-assignment portfolio: +1.0 ms (+1.1%). The pre-existing project filter fires before the coercion, so nothing is added to the quadratic path — 8,400 helper calls against 1.8M inner-loop iterations.

## Reviewer notes

- No version bump in this branch; the post-merge workflow owns it.
- `sorted(covered)` is deliberately left without `key=str`. The coercion at the read sites is the contract; a local sort key would let a future gap degrade to a silently wrong count instead of a visible failure.
- Enumerated every other parsed-entity value in this file that is compared, sorted, or joined — `type` (dict-key membership, already degrades), `assignment.project` vs `project.slug` (both coerced identically, symmetric), the display fields (all through `_esc`), and the numeric fields. No third exposed read site in this file.

### DIFF
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -240,29 +240,50 @@ def _is_closed(status):
     return str(status or "").strip().lower() == "closed"
 
 
-def _status_text(raw, label, warnings):
-    """Read an entity's `status` as text, warning when it was not text already.
+def _text_field(raw, field, label, warnings):
+    """Read a parsed entity field as text, warning when it was not text already.
 
     The frontmatter parser this script borrows from validate-entities.py coerces
     any all-digit scalar to an int before the schema enum is ever checked — and
-    the renderer never runs that validator. So a hand-edited `status: 2026`
-    arrives here as an int, and calling a str method on it straight raises
-    inside _compute, which no local handler catches: the module-level catch-all
-    then discards every warning collected so far and fails the whole render.
-    That inverts this script's contract, where one bad record costs a warning
-    and nothing else.
+    the renderer never runs that validator. So a hand-edited `status: 2026` or
+    `role: 2` arrives here as an int. Comparing or calling a str method on it
+    raises inside _compute, which no local handler catches: the module-level
+    catch-all then discards every warning collected so far and fails the whole
+    render. That inverts this script's contract, where one bad record costs a
+    warning and nothing else.
 
-    The guard keys on `raw is not None`, deliberately not on truthiness —
-    `status: 0` coerces to a falsy int, which a truthy check would silently
-    normalize to "unknown" with no warning at all.
+    `role` fails harder than `status`, because _compute sorts the covered-role
+    set and Python refuses to order a str against an int — one numeric role
+    beside any text role was enough.
+
+    Both sides of a comparison must come through here, not just the side that
+    raised. `open_roles` entries are coerced by the same parser, so normalizing
+    only the assignment role would trade the crash for something quieter and
+    worse: a numeric role that matches its numeric open_roles entry today would
+    stop matching, silently reporting a filled role as open.
+
+    The guard keys on `raw is not None`, deliberately not on truthiness — a
+    falsy `status: 0` or `role: 0` would otherwise normalize away with no
+    warning at all.
+
+    Normalizing every text field once, where the entities are parsed, would
+    close this class rather than its instances, and retire this helper with it.
+    That is tracked as its own change: it decides the string-typed field set for
+    every entity type, and staffing-score.py reads the same parser, where a
+    numeric role label scores every consultant at zero fit instead of raising.
     """
     if raw is None:
         return ""
     if not isinstance(raw, str):
         warnings.append(
-            "%s has a non-string status %r — coerced to text; fix the record" % (label, raw)
+            "%s has a non-string %s %r — coerced to text; fix the record" % (label, field, raw)
         )
     return str(raw).strip()
+
+
+def _status_text(raw, label, warnings):
+    """Read an entity's `status` as text. See _text_field for the why."""
+    return _text_field(raw, "status", label, warnings)
 
 
 def _open_role_demand(projects):
@@ -303,7 +324,10 @@ def _compute(entities, warnings):
         label = "project %s" % (proj.get("name") or slug or "(unnamed)")
         # An undeclared open_roles is not an empty one — see _normalize_open_roles.
         declared_roles = _normalize_open_roles(proj)
-        open_roles = declared_roles or []
+        # Coerced so both sides of the covered/open_roles comparison are text.
+        # Length is preserved, so roles_total and _project_health still see the
+        # declared count.
+        open_roles = [_text_field(r, "open_roles entry", label, warnings) for r in (declared_roles or [])]
         status = _status_text(proj.get("status"), label, warnings)
         if declared_roles is None and status != "closed":
             warnings.append("%s has no open_roles — staffing status unknown" % label)
@@ -316,7 +340,7 @@ def _compute(entities, warnings):
                 continue
             a_label = "assignment %s" % (a.get("_file") or a.get("slug") or "(unknown)")
             if _status_text(a.get("status"), a_label, warnings) in ACTIVE_ASSIGNMENT_STATES:
-                role = a.get("role")
+                role = _text_field(a.get("role"), "role", a_label, warnings)
                 if role:
                     covered.add(role)
         # An assignment role that matches no listed open role may be a label
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -112,10 +112,10 @@ A failure here is cosmetic — the dashboard is already written to `data.path`.
   `.metadata/`. Re-running only rewrites `output/dashboard.html`.
 - **Partial snapshots are expected mid-authoring.** A project without a
   `strategic_impact`, a project that omits `open_roles`, an entity whose
-  `status` is not text (an all-digit value such as `status: 2026` is read as a
-  number, then coerced back to text), or an entity file that cannot be read or
-  decoded is reported in the warnings list, not treated as an error. One bad
-  record never costs the rest of the portfolio.
+  `status` or role label is not text (an all-digit value such as `status: 2026`
+  or `role: 2` is read as a number, then coerced back to text), or an entity
+  file that cannot be read or decoded is reported in the warnings list, not
+  treated as an error. One bad record never costs the rest of the portfolio.
 - **Theming is optional.** The dashboard renders with a built-in palette;
   pass `--design-variables <path.json>` to override colors when a themed look is
   wanted.
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -110,12 +110,16 @@ A failure here is cosmetic — the dashboard is already written to `data.path`.
 - **Read-only.** The renderer never writes `projects-portfolio.json` (only
   `projects-entities` does, via `register-entity.py`) and never touches
   `.metadata/`. Re-running only rewrites `output/dashboard.html`.
-- **Partial snapshots are expected mid-authoring.** A project without a
+- **Partial snapshots are expected mid-authoring.** Any of these is reported in
+  the warnings list, not treated as an error: a project without a
   `strategic_impact`, a project that omits `open_roles`, an entity whose
-  `status` or role label is not text (an all-digit value such as `status: 2026`
-  or `role: 2` is read as a number, then coerced back to text), or an entity
-  file that cannot be read or decoded is reported in the warnings list, not
-  treated as an error. One bad record never costs the rest of the portfolio.
+  `status`, role label, or `open_roles` entry is not text, or an entity file
+  that cannot be read or decoded. A non-text value — `status: 2026`, `role: 2`,
+  or `open_roles: [2]` — is read as a number, then coerced back to text. Both
+  sides of the role comparison are coerced, so a numeric role still matches its
+  numeric `open_roles` entry and still counts as filled: such a warning says to
+  fix the record, not that the coverage figure beside it is wrong. One bad
+  record never costs the rest of the portfolio.
 - **Theming is optional.** The dashboard renders with a built-in palette;
   pass `--design-variables <path.json>` to override colors when a themed look is
   wanted.
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -518,9 +518,11 @@ assert_json "10c falsy int status still warns" \
 assert_json "10d project surfaced by name" \
   "any('non-string status' in w and 'Numeric Status' in w for w in d['data']['warnings'])"
 # The assignment is named by its relative path — _read_entities sets _file on
-# every kept entity, while slug is optional frontmatter.
-assert_json "10e assignment surfaced by file" \
-  "any('non-string status' in w and 'numeric-assignment.md' in w for w in d['data']['warnings'])"
+# every kept entity, while slug is optional frontmatter. Counted for the same
+# reason 11c/11d are: this read sits inside _compute's per-project slug filter,
+# and hoisting it out would warn once per project while `any(...)` stayed green.
+assert_json "10e assignment surfaced by file, exactly once" \
+  "sum('non-string status' in w and 'numeric-assignment.md' in w for w in d['data']['warnings']) == 1"
 assert_json "10f every project still counted"      "d['data']['projects'] == 3"
 assert_no_traceback "10g no traceback on stderr" "$STDERR10"
 # A needle without quote characters: warnings reach the HTML through _esc, which
@@ -543,6 +545,12 @@ assert_html "10h healthy project still rendered" \
 # something quieter: `symmetry`'s numeric role matches its numeric open_roles
 # entry today, and would stop matching, reporting a filled role as open. With
 # only one side coerced, 11f reads 1 instead of 0.
+#
+# The third project is clean throughout, and 11h/11l are its pair: they assert a
+# bad record costs neither the project count nor the render of an untouched
+# project — the same claim Fixture 10 makes with Sound Delivery at 10f/10h.
+# Both of this fixture's other projects are bad-record demonstrations, so
+# without it nothing here checks that resilience.
 # ---------------------------------------------------------------------------
 PF11="$TMPROOT/nonstring-role"
 seed_portfolio "$PF11"
@@ -569,6 +577,18 @@ status: active
 open_roles: [2026]
 ---
 # Symmetry Check
+EOF
+write_entity "$PF11/projects/clean.md" <<'EOF'
+---
+type: project
+slug: clean
+name: Clean Delivery
+client: CleanCo
+strategic_impact: 4
+status: active
+open_roles: [architect]
+---
+# Clean Delivery
 EOF
 # Every assignment names a project that exists — _compute filters on the
 # project before it reads the role, so an orphan would never reach the
@@ -626,6 +646,22 @@ status: active
 ---
 # assignment
 EOF
+# Text role matching the declared open_roles entry, so `clean` contributes no
+# open role and 11f/11k stay at 0 — the clean project has to be genuinely
+# clean, or it would perturb the very counts the symmetry guard pins.
+write_entity "$PF11/assignments/clean-role.md" <<'EOF'
+---
+type: assignment
+slug: eli--clean
+consultant: eli
+project: clean
+role: architect
+start_date: 2026-02-01
+end_date: 2026-08-01
+status: active
+---
+# assignment
+EOF
 
 # Direct invoke for the same reason Fixture 10 does it — run() discards stderr
 # and 11i asserts a traceback never reached it.
@@ -635,10 +671,14 @@ HTML11="$PF11/output/dashboard.html"
 
 assert_json "11a non-string role still succeeds" "d['success'] is True"
 assert_json "11b marked partial"                 "d['data']['partial'] is True"
-assert_json "11c non-string role surfaced by file" \
-  "any('non-string role' in w and 'numeric-role.md' in w for w in d['data']['warnings'])"
-assert_json "11d falsy zero role still warns" \
-  "any('non-string role' in w and 'zero-role.md' in w for w in d['data']['warnings'])"
+# Counted, not merely present: the role read sits inside _compute's per-project
+# slug filter, so each offending assignment must warn exactly once. Moving that
+# read outside the filter would warn once per project iteration instead — a
+# regression `any(...)` cannot see, because the warning is still there.
+assert_json "11c non-string role surfaced by file, exactly once" \
+  "sum('non-string role' in w and 'numeric-role.md' in w for w in d['data']['warnings']) == 1"
+assert_json "11d falsy zero role still warns, exactly once" \
+  "sum('non-string role' in w and 'zero-role.md' in w for w in d['data']['warnings']) == 1"
 assert_json "11e non-string open_roles entry surfaced by project" \
   "any('non-string open_roles entry' in w and 'Symmetry Check' in w for w in d['data']['warnings'])"
 assert_json "11f numeric role still fills its numeric open_role" \
@@ -647,11 +687,12 @@ assert_json "11f numeric role still fills its numeric open_role" \
 # coerced '0' and '2' roles as matching no open_roles label.
 assert_json "11g symmetric match raises no mismatch warning" \
   "not any('matches no open_roles label' in w and 'Symmetry Check' in w for w in d['data']['warnings'])"
-assert_json "11h every project still counted"    "d['data']['projects'] == 2"
+assert_json "11h every project still counted"    "d['data']['projects'] == 3"
 assert_no_traceback "11i no traceback on stderr" "$STDERR11"
 assert_html "11j symmetry project still rendered" "Symmetry Check" "$HTML11"
 assert_html "11k open-roles tile agrees with the envelope" \
   '<div class="n">0</div><div class="l">Open roles</div>' "$HTML11"
+assert_html "11l clean sibling project still rendered" "Clean Delivery" "$HTML11"
 
 echo
 if [ "$failures" -eq 0 ]; then
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -528,6 +528,131 @@ assert_no_traceback "10g no traceback on stderr" "$STDERR10"
 assert_html "10h healthy project still rendered" \
   "Sound Delivery" "$PF10/output/dashboard.html"
 
+# ---------------------------------------------------------------------------
+# Fixture 11 — a role label that is not text. The same parse-time int coercion
+# that hit `status` reaches `role`, and lands harder: _compute sorts the
+# covered-role set, and Python refuses to order a str against an int, so one
+# `role: 2` next to any text role raised and cost the whole render.
+#
+# 11c and 11d are the truthy/falsy pair, mirroring 10c/10d — a guard keyed on
+# truthiness rather than on None would let `role: 0` through unremarked.
+#
+# 11f, 11g and 11k are the symmetry guard, and they are the reason this fixture
+# carries a second project. `open_roles` entries are coerced by the same parser,
+# so stringifying only the assignment side would fix the crash and introduce
+# something quieter: `symmetry`'s numeric role matches its numeric open_roles
+# entry today, and would stop matching, reporting a filled role as open. With
+# only one side coerced, 11f reads 1 instead of 0.
+# ---------------------------------------------------------------------------
+PF11="$TMPROOT/nonstring-role"
+seed_portfolio "$PF11"
+write_entity "$PF11/projects/mixed.md" <<'EOF'
+---
+type: project
+slug: mixed
+name: Mixed Roles
+client: MixCo
+strategic_impact: 3
+status: active
+open_roles: [lead]
+---
+# Mixed Roles
+EOF
+write_entity "$PF11/projects/symmetry.md" <<'EOF'
+---
+type: project
+slug: symmetry
+name: Symmetry Check
+client: SymCo
+strategic_impact: 2
+status: active
+open_roles: [2026]
+---
+# Symmetry Check
+EOF
+# Every assignment names a project that exists — _compute filters on the
+# project before it reads the role, so an orphan would never reach the
+# coercion and the fixture would pass against the unfixed script. `active` is
+# in ACTIVE_ASSIGNMENT_STATES, so these genuinely reach the role read.
+write_entity "$PF11/assignments/numeric-role.md" <<'EOF'
+---
+type: assignment
+slug: ada--mixed
+consultant: ada
+project: mixed
+role: 2
+start_date: 2026-02-01
+end_date: 2026-08-01
+status: active
+---
+# assignment
+EOF
+write_entity "$PF11/assignments/text-role.md" <<'EOF'
+---
+type: assignment
+slug: bo--mixed
+consultant: bo
+project: mixed
+role: lead
+start_date: 2026-02-01
+end_date: 2026-08-01
+status: active
+---
+# assignment
+EOF
+write_entity "$PF11/assignments/zero-role.md" <<'EOF'
+---
+type: assignment
+slug: cy--mixed
+consultant: cy
+project: mixed
+role: 0
+start_date: 2026-02-01
+end_date: 2026-08-01
+status: active
+---
+# assignment
+EOF
+write_entity "$PF11/assignments/symmetry-role.md" <<'EOF'
+---
+type: assignment
+slug: di--symmetry
+consultant: di
+project: symmetry
+role: 2026
+start_date: 2026-02-01
+end_date: 2026-08-01
+status: active
+---
+# assignment
+EOF
+
+# Direct invoke for the same reason Fixture 10 does it — run() discards stderr
+# and 11i asserts a traceback never reached it.
+STDERR11="$TMPROOT/stderr11.txt"
+LAST_JSON="$(python3 "$SCRIPT" "$PF11" 2>"$STDERR11" | tail -n 1)"
+HTML11="$PF11/output/dashboard.html"
+
+assert_json "11a non-string role still succeeds" "d['success'] is True"
+assert_json "11b marked partial"                 "d['data']['partial'] is True"
+assert_json "11c non-string role surfaced by file" \
+  "any('non-string role' in w and 'numeric-role.md' in w for w in d['data']['warnings'])"
+assert_json "11d falsy zero role still warns" \
+  "any('non-string role' in w and 'zero-role.md' in w for w in d['data']['warnings'])"
+assert_json "11e non-string open_roles entry surfaced by project" \
+  "any('non-string open_roles entry' in w and 'Symmetry Check' in w for w in d['data']['warnings'])"
+assert_json "11f numeric role still fills its numeric open_role" \
+  "d['data']['open_roles'] == 0"
+# Scoped to Symmetry Check on purpose — Mixed Roles legitimately reports its
+# coerced '0' and '2' roles as matching no open_roles label.
+assert_json "11g symmetric match raises no mismatch warning" \
+  "not any('matches no open_roles label' in w and 'Symmetry Check' in w for w in d['data']['warnings'])"
+assert_json "11h every project still counted"    "d['data']['projects'] == 2"
+assert_no_traceback "11i no traceback on stderr" "$STDERR11"
+assert_html "11j symmetry project still rendered" "Symmetry Check" "$HTML11"
+assert_html "11k open-roles tile agrees with the envelope" \
+  '<div class="n">0</div><div class="l">Open roles</div>' "$HTML11"
+
 echo
 if [ "$failures" -eq 0 ]; then
   echo "All render-dashboard tests passed."


claude-cogni-work:
## Ship-ready

Draft PR handed off for independent review. The author instance deliberately did **not** run the in-session review gate and did **not** flip this PR ready — under the two-instance posture the review and the draft→ready decision belong to the merger.

**What landed:** `_text_field` at the read site coerces a non-string `role`, and `open_roles` entries are normalized through the same helper so both sides of the coverage comparison are text. `_status_text` became a delegate of the generalized helper rather than a second copy of the same body.

**Where to aim review effort:**

1. **The symmetric coercion is the substantive judgment**, not the crash fix. Coercing only the assignment role passes the issue's stated acceptance criteria while silently reporting a filled role as open — verified empirically, and reproduced again as mutation 2 below. If you disagree that both sides belong in this change, that is the thing to push on.
2. **`_status_text` was folded into a delegate.** The plan originally said to leave it untouched; three independent quality reviewers flagged it as an exact clone of the new helper, so it was collapsed. Warning strings are byte-identical and fixtures 10c/10d/10e pass unchanged — but this does touch already-shipped code, so it is a legitimate place to disagree.
3. **A numeric `open_roles` entry now warns**, flipping a previously-clean all-numeric portfolio to `partial: true`. Accepted deliberately; the reasoning is in the PR body under "Accepted behaviour change."

**Evidence the tests bite** (not just that they pass):

| Mutation | Result |
|---|---|
| Revert both call sites | 10 of Fixture 11's assertions fail with the reported `TypeError` |
| Revert **only** the `open_roles` normalization | 4 fail — 11f reads `open_roles: 1` instead of `0` |
| Neither reverted | 59 assertions green |

**Known-unverified:** nothing material. The issue's exact repro was executed and produces `success: true` / `partial: true` with the offending record named. Sibling suites (`test_register_entity.sh`, `test_portfolio_init.sh`) are green. Breadcrumb guard, version-bump gate, and skill-name check all pass.

**Out of scope, tracked, and worth knowing:** `staffing-score.py` reads the same parser and has the same exposure with a different failure mode — a numeric role label scores every consultant at zero profile fit and echoes the raw int into the shortlist, with no warnings channel to surface it. That is not regressed by this PR and is not fixed by it; it is named in the new helper's docstring so whoever picks up the parse-boundary change knows there is a second consumer.

sdh07:

_The merger formed this expectation from issue #1162 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The single crash site is `sorted(covered)` at cogni-projects/scripts/render-dashboard.py:324, fed by `covered.add(role)` at :311-321 where `a.get("role")` is added raw — and validate-entities.py `_coerce` (:120-124) has already turned an all-digit scalar into an int, so a project with one `role: 2` and one `role: lead` assignment yields a mixed-type set. A correct fix normalizes `role` to text before the set is built (a `_role_text`-style sibling to `_status_text` at render-dashboard.py:243-265, keyed on `raw is not None` rather than truthiness) or, per the issue's option 2, in `_read_entities` (:132-166) over an explicit per-type list of string-typed fields that excludes `strategic_impact`/`allocation_pct` — either way emitting a per-record warning that names the assignment by its `_file` relative path, leaving `partial` (already `bool(warnings)` at :567) to follow. Because `open_roles` items pass through the same `_coerce`, the fix must keep both sides of the membership tests at :326/:329/:343 comparable — coercing `role` alone would make `role: 2` stop matching `open_roles: [2]`, converting a previously-correct 1/1 count into 0/1 plus a spurious label-mismatch warning. Regression coverage belongs in cogni-projects/tests/test-render-dashboard.sh as a Fixture 11 mirroring Fixture 10 (:445-529), whose own comments record the vacuity trap this fixture must avoid: the assignments must name a real project slug and there must be *two* of them on that project with different value types, since `sorted({2})` does not raise. `_coerce` in validate-entities.py stays untouched per the issue, and the projects-dashboard SKILL.md partial-snapshot note (:113-118) currently names only `status` and needs to cover a non-text `role`.

**Acceptance checklist:**
- [ ] Scope: the diff touches only cogni-projects/scripts/render-dashboard.py, cogni-projects/skills/projects-dashboard/SKILL.md, and cogni-projects/tests/test-render-dashboard.sh — confirm cogni-projects/scripts/validate-entities.py is absent from the changed-file list, so `_coerce` (validate-entities.py:120-124) is provably unmodified as the issue requires.
- [ ] No version bump in the branch: neither cogni-projects/.claude-plugin/plugin.json nor the repo-root .claude-plugin/marketplace.json appears in the changed files (the post-merge workflow owns the bump).
- [ ] SKILL.md frontmatter stays valid — name, description, allowed-tools present and parseable (adjudicate by running the frontmatter check against cogni-projects/skills/projects-dashboard/SKILL.md and reading data.clean).
- [ ] No maintainer breadcrumb is introduced: neither the script, the SKILL.md, nor the test carries an issue/PR ref, a version tag, or a slice/finding code — rationale must be stated semantically (adjudicate by running the breadcrumb guard).
- [ ] Plugin structure/manifest health is unaffected — cogni-projects still validates (adjudicate by running the structure/manifest check).
- [ ] render-dashboard.py stays stdlib-only: no import added beyond the existing argparse/datetime/html/importlib.util/json/os/sys set (grep the import block at :29-35).
- [ ] A new fixture block exists in cogni-projects/tests/test-render-dashboard.sh with a banner comment and labeled assertions in the house `assert_json`/`assert_html` style (e.g. `11a`…), following the Fixture 10 pattern at :445-529 — AC-2.
- [ ] The whole suite passes: `bash cogni-projects/tests/test-render-dashboard.sh` exits 0 and prints "All render-dashboard tests passed."
- [ ] No pre-existing assertion is deleted or weakened: all 57 existing labels (1a-1j, 2a-2c, 3a-3f, 4a-4b, 5a-5b, 6a-6i, 7a-7c, 8a-8f, 9a-9h, 10a-10h) are still present with unchanged expressions — diff the assertion lines, do not just trust a green run.
- [ ] The new fixture is vacuity-proof, not merely present: it seeds at least TWO assignments whose `project:` names a real project slug in the same fixture and whose `status:` is planned/active, one with an all-digit `role:` and one with an alphabetic `role:` — a single numeric role cannot reproduce the crash because `sorted({2})` succeeds. Confirm by reverting the script guard and re-running the suite: the new assertions must fail.
- [ ] The mixed-type set can no longer reach `sorted(covered)`: the coercion happens where `covered` is built (render-dashboard.py:311-321) or earlier in `_read_entities` (:132-166), never as a try/except around the sort — grep that `covered.add(...)` receives an already-normalized value.
- [ ] AC-1 behavior: for that fixture the envelope has `d['success'] is True`, `d['data']['partial'] is True`, and the offending record is surfaced in `d['data']['warnings']` — asserted in the test, not merely observed.
- [ ] The role warning names the offending assignment by its relative file path (the `_file` set at render-dashboard.py:164), mirroring assertion 10e's `numeric-assignment.md` needle — a warning that names only the project does not identify the record to fix.
- [ ] The role warning text is distinguishable from the status one: it contains a role-specific token (e.g. 'non-string role') so a grep/assertion cannot pass on a `_status_text` warning by accident.
- [ ] The warning fires exactly once per offending assignment, not once per project iteration — assert a count (`sum(... ) == 1`) rather than `any(...)`, as `_compute` iterates all assignments inside the project loop (:312).
- [ ] Symmetry preserved: a project with `open_roles: [2]` and an active assignment with `role: 2` still counts as filled (coverage 1/1) and emits NO 'label mismatch' warning — the membership tests at :326, :329 and :343 must stay comparable on both sides. Asserted in the test, since a role-only coercion silently regresses this previously-correct case.
- [ ] No traceback escapes: the new fixture invokes the script directly capturing stderr (the Fixture 6/10 idiom at :328/:512, since the shared `run()` helper discards stderr) and calls `assert_no_traceback`.
- [ ] One bad record does not cost the rest of the snapshot: the fixture includes a sound sibling project and asserts both the project count and that the healthy project's name appears in the rendered HTML (mirroring 10f/10h).
- [ ] Falsy-value edge case covered, mirroring Fixture 10's deliberate 10c/10d pair: the fixture includes a falsy non-string role (`role: 0`) alongside the crashing one, and the chosen behavior (warn-and-count vs. silently skip at the `if role:` guard on :320) is stated in a script comment or test comment rather than left incidental.
- [ ] Absent/empty `role` does not start warning: Fixture 1's `d['data']['warnings'] == []` / `partial is False` and Fixture 9's `warnings == []` still hold, so a clean portfolio is not newly marked partial.
- [ ] If the parse-boundary option is taken, its blast radius is bounded and visible: the coerced field list is an explicit named constant or per-type mapping in render-dashboard.py that excludes `strategic_impact` and `allocation_pct`, and the existing 'non-string status' warning wording survives so assertions 10c/10d still pass unchanged.
- [ ] The new guard carries a why-comment or docstring in the established style of `_status_text` (render-dashboard.py:243-265) explaining the parser/validator split that causes the int — this file documents every guard's rationale.
- [ ] cogni-projects/skills/projects-dashboard/SKILL.md is updated so its partial-snapshot note (:113-118), which today names only a non-text `status`, also covers a non-text `role` — and the 'Role labels are free strings' paragraph (:52-56) is still accurate after the change. This mechanism is prose-carried, so the achievable bar is a grep for the role wording anchored to the changed script behavior plus the mutation check that the described warning text matches what the script actually emits; no behavioral test of SKILL.md is demandable.

**Risk watchlist:**
- The likeliest wrong fix: coerce `role` only where `covered` is built, leaving `open_roles` items as ints — `role: 2` then stops matching `open_roles: [2]`, turning a previously-correct 1/1 into 0/1 plus a bogus label-mismatch warning. The issue's own exposure table calls `open_roles` items 'cosmetic' precisely because both sides pass through `_coerce`; a one-sided fix invalidates that premise.
- A vacuous fixture: one assignment with a numeric role, or an assignment whose `project:` names no real project. Neither reaches `sorted(covered)` with mixed types, so the test passes against the unfixed script — the exact trap Fixture 10's own comment at test-render-dashboard.sh:492-494 documents.
- Wrapping `sorted(covered)` in try/except or switching to `sorted(covered, key=str)` — that stops the traceback without surfacing the offending record, so AC-1's warning requirement goes unmet and the next read site (`role not in open_roles`) still compares across types silently.
- Placing the guard after the `if role:` truthiness test at :320, so `role: 0` is dropped with no warning — the same falsy-int hole `_status_text`'s docstring explicitly warns against and Fixture 10's 10c exists to catch.
- Option 2 over-reach: normalizing every scalar at the parse boundary can newly warn on `name`/`client`/`slug` values that are all-digit, flipping previously clean portfolios to `partial: true` and breaking Fixture 1's `warnings == []`; or it can drop/reword the `non-string status` warning and silently break 10c/10d.
- Touching `_coerce` in validate-entities.py despite the issue's explicit prohibition — it is shared with register-entity.py and staffing-score.py, and the validator correctly reports `status: 2026` as an invalid enum today.
- Warning emitted per project rather than per assignment, producing N duplicate lines on a portfolio with several projects — an `any(...)`-style assertion would not catch it.
- A green suite that green-lit itself: assertions rewritten or deleted so an unchanged behavior still 'passes'. Diff the assertion expressions, not just the exit code.
- SKILL.md drift left behind — the script starts warning on `role` while the skill still tells the user only `status` can be non-text, so a partner sees a warning the skill cannot explain.
- Judgment-only (no gate can decide it): the issue frames the read-site-vs-parse-boundary choice as 'the real question' while also declaring option 2 stronger. Both satisfy the AC, so this checklist is deliberately option-agnostic — a human should confirm they are content with whichever the author chose, and, if option 2, with the specific set of fields declared string-typed.

- The plan-time acceptance criteria list 2 items (render successfully with a warning + partial; a Fixture 10-style regression test), but a base-grounded regeneration derives ~22 gating items — including the open_roles/role comparison symmetry that a naive read-site fix silently regresses, the fixture-vacuity requirement that Fixture 10's own comments already document as a known trap, a bound on which fields the parse-boundary option may normalize, and the projects-dashboard SKILL.md update that the AC never asks for even though SKILL.md is in the changed-file scope. The acceptance bar was materially under-specified.

sdh07:
## Hold cleared by maintainer — proceeding to review

The `insufficient_coverage` escalation on the committed review contract was
adjudicated live by the maintainer: **waived, review proceeds.**

The ruling is that the contract's derived acceptance checklist — not issue #1162's
two-item `## Acceptance criteria` — is the grading rubric for this PR. The
under-specified plan-time bar is therefore no longer load-bearing: the ~22 derived
items above stand in for it, including the `open_roles`/`role` comparison symmetry
and the fixture-vacuity requirement the AC never asked for.

No acceptance-criteria writeback was made to #1162, because this ruling does not
overrule or supersede any criterion the issue states — it supplements a bar that was
merely incomplete. `svc:needs-human-review` removed; `service-review` runs next
against the committed contract.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1167

**Summary:** Two contract-grading majors (items 15 and 18) are both mutation-confirmed and independently trigger revision. Separately, three reviewers converged on a SKILL.md documentation gap - the partial-snapshot note omits the `open_roles entry` warning variant the script also emits - deduped into one minor finding; plugin-validator offered a stylistic waiver on this point but skill-reviewer and skill-quality-reviewer held it open, so the merged finding is treated as un-waived. The core fix itself is sound: both mutation checks the contract demanded (vacuity-proof fixture, load-bearing open_roles symmetry) passed.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[major]** (auto-fixable) `review-contract:item-15` — The contract requires asserting a count (`sum(...) == 1`) rather than `any(...)` so a per-project-iteration duplicate cannot slip through, but Fixture 11's role-warning assertions (test-render-dashboard.sh:638, 640) both use `any(...)`. Mutation-confirmed: duplicating the `_text_field` role-warning call in render-dashboard.py (simulating the exact per-project-duplication risk the contract names) still leaves the full suite green — all 11a-11k pass with the warning emitted twice per offending assignment.
  - *Suggested fix:* Add a count-based assertion, e.g. `sum('non-string role' in w and 'numeric-role.md' in w for w in d['data']['warnings']) == 1`, so a regression that fires the warning once per project iteration (instead of once per assignment) is caught.

- **[major]** (auto-fixable) `review-contract:item-18` — Fixture 11 lacks a sound sibling project mirroring Fixture 10's clean 'Sound Delivery' project (test-render-dashboard.sh:479-484, asserted clean via 10f/10h). Both Fixture 11 projects carry warnings: 'Mixed Roles' via numeric-role.md and zero-role.md (non-string role), and 'Symmetry Check' via its own non-string open_roles entry (open_roles: [2026]) and its non-string role: 2026 assignment. 11j asserts only 'Symmetry Check' appears in the rendered HTML, and Symmetry Check is not a healthy project — it is itself a bad-record demonstration; 'Mixed Roles' is never asserted in the HTML at all.
  - *Suggested fix:* Add a third, entirely clean project (no non-string status/role/open_roles anywhere in its own or its assignments' records) to Fixture 11, and assert its name appears in the rendered HTML alongside the project count, mirroring 10f/10h's resilience check.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:113-118` — The change made `_text_field` normalize three fields, and it warns with the field name in the text: `status`, `role`, and `open_roles entry` (render-dashboard.py:278-280, called at :330). The note covers only `status` and an assignment's `role`, and the parenthetical exemplifies only `status: 2026` / `role: 2`, so a user reading a literal warning such as `project acme has a non-string open_roles entry 2 — coerced to text; fix the record` finds no vocabulary for it in the skill. The load-bearing consequence is also unstated: because both sides are coerced, a numeric `role: 2` still matches `open_roles: [2]` and still counts as filled — the reader sees two warnings alongside an unchanged 1/1 coverage figure and needs to know that is correct, not a miscount. Three perspectives independently converged on this same gap (plugin-validator, skill-reviewer, skill-quality-reviewer); plugin-validator waived it as stylistic, the other two held it open, so it is treated as un-waived.
  - *Suggested fix:* In the partial-snapshot bullet, replace the clause with: 'an entity whose `status` or role label is not text (an all-digit value such as `status: 2026`, `role: 2`, or an `open_roles` entry like `[2]` is read as a number, then coerced back to text — both sides of the role comparison are coerced, so a numeric role still matches its numeric `open_roles` entry and still counts as filled; the warnings say to fix the record, not that coverage is wrong)'. (ref: skill-creator §Writing Style (explain the why — state what the output means so it can be narrated) / §Principle of Lack of Surprise)

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:52-56` — Not an inaccuracy, a completeness note. After this change there are two distinct role-related warning classes — the label mismatch this paragraph describes, and the new non-text-label one described 60 lines later under Notes. The paragraph now silently omits that a non-text label is coerced and both sides normalized, which is exactly why a numeric label still counts as filled rather than flipping to a spurious mismatch.
  - *Justification:* The Notes entry already carries the behavior; adding it here risks duplicating the same fact in two places, which this skill otherwise avoids.

### Positive observations

- item 10 satisfied (mutation-verified): reverting the role coercion makes 10 of Fixture 11's assertions genuinely fail with a TypeError, proving the fixture is not vacuous. (contract-grading-reviewer)
- item 16 satisfied (mutation-verified): reverting only the open_roles normalization makes 11e, 11f, 11g and 11k genuinely fail - the symmetry guard is load-bearing, not decorative. (contract-grading-reviewer)
- item 9 satisfied: git diff shows zero removed lines in the test file; all 58 pre-existing assertion labels remain and pass. (contract-grading-reviewer, plugin-validator)
- No version bump: check-version-bump.py returns clean, both manifests still 0.0.12 and in sync. (plugin-validator)
- No breadcrumbs: check-breadcrumbs.py returns 0 violations across 50 scanned occurrences. (plugin-validator)
- stdlib-only import block unchanged; no second uncoerced read site survives - _open_role_demand reads the already-computed counts. (plugin-validator)
- The SKILL.md contract item is satisfied on both halves - the partial-snapshot note covers a non-text role, and the 'Role labels are free strings' paragraph is still accurate against the post-change script. (skill-reviewer)
- No drift between prose and emission: the skill says 'coerced back to text', the script says 'coerced to text; fix the record' and returns the coerced value rather than skipping the record. (skill-reviewer)
- Body length 108 lines against the 500-line soft cap, field contract deferred to references/data-model.md - progressive disclosure intact. (skill-reviewer, skill-quality-reviewer)
- The edit is proportionate: four lines changed inside an existing bullet, no new section, no restructuring. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1167

**Summary:** All three perspectives (plugin-validator, skill-reviewer, skill-quality-reviewer) passed the fix and its skill-doc/test coverage; the two remaining findings are un-waived minors that surface as advisory notes under Rule 5b rather than gating the merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:115-116` — The three-field list mixes literal field names with prose: "an entity whose `status`, role label, or `open_roles` entry is not text". Two of the three are code-ticked matching the emitted warning verbatim; `role` is softened to un-ticked prose, slightly undercutting the goal of letting a reader map a raw warning straight to the field name.
  - *Suggested fix:* Change "role label" to "`role`" so all three list items are code-ticked and match the emitted warning text exactly.

- **[minor]** (auto-fixable) `cogni-projects/tests/test-render-dashboard.sh — assertion 11e` — 11e is the only positive presence assertion in Fixture 11 still on `any(...)` after 10e/11c/11d were tightened to `sum(...) == 1`. The `open_roles` coercion sits inside the per-project loop but outside the assignment loop, so a mirror-image regression (hoisting that read into the assignment loop, warning once per assignment instead of once per project) is invisible to `any(...)` because the warning is still present.
  - *Suggested fix:* Change to sum('non-string open_roles entry' in w and 'Symmetry Check' in w for w in d['data']['warnings']) == 1

### Positive observations

- Symmetric coercion is the right call and the part a narrower read-site fix would have gotten wrong; reverting only the open_roles comprehension reproduces the exact wrong-number regression that 11f/11g/11k catch. (plugin-validator)
- The guard keys on `raw is not None` rather than truthiness, so `role: 0` warns instead of silently normalizing away, and 11d pins that falsy case. (plugin-validator)
- `_status_text` is retained as a thin wrapper over `_text_field`, so existing call sites and Fixture 10 semantics are untouched — the diff is additive at the behavior level. (plugin-validator)
- Fixture 11's third clean project (11h/11l) pins the plugin's resilience claim: one bad record costs neither the project count nor an untouched project's render. (plugin-validator)
- Scope discipline: `_coerce` in the shared validate-entities.py is untouched, and the parse-boundary alternative is referenced semantically in the docstring without a maintainer breadcrumb. (plugin-validator)
- The added sentence stays on the right side of the skill's own stated boundary — it frames coercion as a narration fact rather than restating the algorithm. (skill-reviewer)
- Every behavioral claim verified against source: both-sides coercion holds, the `status: 2026` / `role: 2` / `open_roles: [2]` examples are all genuinely int-coerced, and `data.partial` is `bool(warnings)`. (skill-reviewer)
- Description is third-person, carries eight concrete trigger phrases plus an explicit no-keyword fallback, and the skill name is domain-prefixed per repo convention. (skill-reviewer)
- Progressive disclosure is right-sized for a 112-line skill: the entity field contract is delegated to references/data-model.md rather than inlined. (skill-reviewer)
- The reworked Notes bullet now names all three fields `_text_field` warns on, matching the literal emitted warning text so a reader can map a warning straight to the bullet. (skill-quality-reviewer)
- Body measures 112 lines against the 500-line soft cap with detail correctly offloaded to references/data-model.md — textbook progressive disclosure. (skill-quality-reviewer)
- No CAPS-heavy MUSTs anywhere; every rule is paired with its reason. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressing the [previous verdict](https://github.com/cogni-work/insight-wave/pull/1167#issuecomment-5096637746). Commit `8c358a4d`.

### Findings resolved

- **[major] `review-contract:item-15`** — 11c/11d asserted `any(...)`, so the per-project-iteration duplicate the contract named was invisible. Both now assert `sum(...) == 1`. Mutation-confirmed both ways: duplicating the `_text_field` role call now fails 11c and 11d, where before the suite stayed green. (`8c358a4d`)
- **[major] `review-contract:item-18`** — Fixture 11 gains **Clean Delivery**, a third project with no non-string value anywhere in its own or its assignment's record. Its single open role is filled by a text-role active assignment, so it perturbs neither 11f nor the 11k tile; 11h rises to 3 and a new **11l** asserts it reaches the rendered HTML — the same count-plus-healthy-project pair Fixture 10 makes with Sound Delivery at 10f/10h. Verified non-vacuous: under a reverted-coercion mutation, 11h and 11l both fail. (`8c358a4d`)
- **[minor] `cogni-projects/skills/projects-dashboard/SKILL.md`** — the partial-snapshot note now names the `open_roles` entry alongside `status` and `role`, and states the consequence the reader needs: both sides of the role comparison are coerced, so a numeric role still matches its numeric `open_roles` entry and still counts as filled — the warning asks for the record to be fixed, not for the coverage figure beside it to be doubted. The example was corrected to `open_roles: [2]` (the emitted warning names the *element*, `open_roles entry 2`, not the list). (`8c358a4d`)

The fourth finding (`SKILL.md:52-56`) was waived by the reviewer and left untouched, so the coercion fact still lives in exactly one place.

### Also in this commit

A quality pass converted **10e** to the same `sum(...) == 1` form. It guards the assignment `status` read behind the very same per-project slug filter, and a hoist mutation confirmed it was blind to the regression 11c/11d now catch — Fixture 10 stayed fully green while the warning fired three times.

### Verification

- `bash cogni-projects/tests/test-render-dashboard.sh` — green, 60 assertions.
- `test_register_entity.sh`, `test_portfolio_init.sh` — green.
- Breadcrumb guard 0 violations; version-bump gate clean (`cogni-projects` touched, 0 violations); skill-name check OK.

### Fresh verdict

✅ **PASS** — [new verdict comment](https://github.com/cogni-work/insight-wave/pull/1167#issuecomment-5100020197). Two advisory minors remain (11e could take the same count form; `role` unticked in the SKILL.md field list); neither gates merge.

sdh07:

_The merger formed this expectation from issue #1162 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The crash is at cogni-projects/scripts/render-dashboard.py:324 (`for role in sorted(covered)`), fed by :311-321 where `_compute` adds each active assignment's raw `role` to `covered`; because validate-entities.py's `_coerce` (validate-entities.py:120-124) turns any all-digit scalar into an int at parse time and the renderer never runs the validator, `covered` can hold both int and str and `sorted` raises TypeError, which the module-level catch-all at render-dashboard.py:578-582 converts into a total render failure that discards every accumulated warning. A correct fix normalizes role labels to text before they reach `sorted`/membership — preferably one pass in `_read_entities` (render-dashboard.py:132-166, which already owns the per-entity warning surface and already mutates the dict at :164) over an explicit allowlist of string-typed fields per entity type, leaving `strategic_impact` and `allocation_pct` alone and leaving validate-entities.py `_coerce` untouched — and it must warn per offending record following the documented `_status_text` pattern at :243-265, whose stated rationale is that the guard keys on `is not None`, never truthiness (the `status: 0` trap Fixture 10c pins). The fix must keep both sides of the fill comparison on the same type: `open_roles` items also pass through `_coerce` (`_normalize_open_roles`, :185-201) and are compared against `covered` at :324-329 and :343, so coercing only `role` would newly break a numerically-labelled role that currently matches (a regression the issue's own exposure audit implicitly assumes away). Regression coverage belongs in cogni-projects/tests/test-render-dashboard.sh as a new fixture following the Fixture 10 pattern (test-render-dashboard.sh:445-529): seeded portfolio, direct `python3 "$SCRIPT"` invocation capturing stderr, `assert_json` + `assert_no_traceback` + `assert_html` — and it must place at least two mixed-type active roles on one real project, since a single non-string role never reaches a comparison inside `sorted()` and would pass against the unfixed script.

**Acceptance checklist:**
- [ ] Scope: the diff touches only cogni-projects/scripts/render-dashboard.py, cogni-projects/skills/projects-dashboard/SKILL.md, and cogni-projects/tests/test-render-dashboard.sh — cogni-projects/scripts/validate-entities.py is NOT in the changed-file list, so `_coerce` (validate-entities.py:120-124) is unchanged as the issue directs.
- [ ] No version bump in the branch: neither cogni-projects/.claude-plugin/plugin.json nor the repo-root .claude-plugin/marketplace.json appears in the diff (the post-merge workflow owns the patch bump; the PR-time Version-bump gate flags `version-touched` otherwise).
- [ ] cogni-projects/skills/projects-dashboard/SKILL.md still has valid frontmatter with `name`, `description`, and `allowed-tools` intact — adjudicate by running the frontmatter gate on the PR head and reading `data.clean`, not by eyeballing the diff.
- [ ] No new maintainer breadcrumb is introduced: no `#NNNN` issue/PR ref, version tag, or slice code appears in the SKILL.md prose or in the new render-dashboard.py docstrings/comments — adjudicate with the breadcrumb gate; rationale must be stated semantically the way the existing `_status_text` docstring (render-dashboard.py:243-265) states it without naming an issue.
- [ ] render-dashboard.py remains stdlib-only (no import added beyond the existing argparse/datetime/html/importlib.util/json/os/sys set) and still prints exactly one `{success,data,error}` JSON line, keeping the module-level catch-all at :578-582 as the last resort rather than the primary handler.
- [ ] Every value that reaches `sorted(covered)` (render-dashboard.py:324) is a `str`: the diff shows the normalization either in `_read_entities` (:132-166) over the parsed frontmatter or at the `covered.add(role)` site (:319-321), so a mixed int/str set is structurally unreachable.
- [ ] Reproduction from the issue renders instead of failing: a project with `open_roles: [lead]` plus two active assignments on it, one `role: 2` and one `role: lead`, yields `d['success'] is True`, `d['data']['partial'] is True`, and no `Traceback` on stderr (assert_no_traceback, the Fixture 6/10 pattern).
- [ ] The offending assignment is surfaced in `data.warnings[]` and named by its relative file path (the `_file` label set at render-dashboard.py:164), not by `slug` — mirroring assertion 10e, since `slug` is optional frontmatter on an assignment.
- [ ] Type symmetry preserved: a project with `open_roles: [2]` and an active assignment with `role: 2` still counts as filled — `roles_filled` includes it, the health flag reads `fully staffed`, and NO `matches no open_roles label` warning is emitted. Both sides of the :324-329 and :343 comparisons are normalized the same way; coercing `role` alone silently converts a currently-matching label into a false mismatch.
- [ ] The guard keys on type/None, not truthiness: an active assignment with `role: 0` surfaces a non-string-role warning naming the record rather than being silently dropped by the `if role:` test at :320 — the same failure mode the `status: 0` case (Fixture 10c) was written to pin.
- [ ] Absent is not malformed: an assignment with no `role` key contributes nothing to `covered` and emits no non-string warning, and a project with no `status` key still renders `unknown` in the envelope/table (never the literal string `None`) — normalization must not stringify a missing key (:259-260, :337).
- [ ] Numeric fields are untouched: the diff's set of normalized fields demonstrably excludes `strategic_impact` and `allocation_pct` (name the allowlist/constant in the diff), so `_coerce_impact` (:169-182) and the allocation `int()` path (:353-361) keep receiving ints.
- [ ] Exactly one warning per offending record: in the new fixture's envelope, `sum(<offending file> in w for w in d['data']['warnings']) == 1` — a fix that warns inside the per-project loop must not re-warn the same assignment once per project.
- [ ] cogni-projects/tests/test-render-dashboard.sh gains a new numbered fixture following the Fixture 10 pattern (test-render-dashboard.sh:445-529): `seed_portfolio`, `write_entity` heredocs, a direct `LAST_JSON="$(python3 "$SCRIPT" "$PFn" 2>"$STDERRn" | tail -n 1)"` invocation rather than the stderr-discarding `run()` helper, and `assert_json` / `assert_no_traceback` / `assert_html` assertions.
- [ ] The new fixture is non-vacuous: it places at least TWO planned/active assignments with mixed-type roles (e.g. `role: 2` and `role: lead`) on the SAME project whose `slug` actually exists as a project entity — a single non-string role never reaches a comparison inside `sorted()`, and an orphan `project:` value is filtered out at :313 before the role is ever read, so either shape would pass against the unfixed script.
- [ ] The new fixture also asserts the blast-radius property the issue names: a healthy sibling project still appears in the rendered HTML and `d['data']['projects']` counts every parseable project, proving one bad record no longer costs the whole dashboard.
- [ ] The whole suite is green — `bash cogni-projects/tests/test-render-dashboard.sh` exits 0 with Fixtures 1-10 passing unchanged, in particular 3e (`label mismatch`), 6e (`staffing status unknown` count == 2), 9c/9f (open-role envelope/tile agreement), and 10c/10d/10e (non-string status warnings).
- [ ] If `_status_text` (:243-265) is removed or absorbed into the new normalization (the issue notes option 2 makes it redundant), the `non-string status` warning text and the falsy `status: 0` warning behavior are preserved, or the Fixture 10 assertions are updated deliberately to equivalent semantics — never weakened or deleted to make the suite pass.
- [ ] SKILL.md doc-sync: the `Partial snapshots are expected mid-authoring` bullet (skills/projects-dashboard/SKILL.md:113-118), which today names `status` only, is generalized to cover a non-text `role`. Because this mechanism is prose-carried, the achievable bar is a grep for `role` inside that Notes bullet plus a documented statement of the degraded behavior — not a behavioral test of the SKILL.
- [ ] SKILL.md keeps the renderer as the source of truth (its own convention at :105-109): no health-flag, coverage, or coercion computation rule is duplicated into the prose, and the skill body stays within the skill length cap — adjudicate the cap by running the skill length gate on the PR head.
- [ ] HTML output is still escaped: coerced role/status values reach the page through `_esc` (:372-373, :385, :412), and Fixture 5's `assert_html_lacks` on `<script>alert(1)</script>` still passes.

**Risk watchlist:**
- Read-site-only fix (issue option 1): coerces `role` at `covered.add` but leaves `open_roles` items as parsed ints, so a numerically-labelled role that matches today (`open_roles: [2]` + `role: 2`) becomes a false `matches no open_roles label` warning and a falsely unstaffed project — a silent behavioral regression no existing fixture covers.
- A vacuous regression fixture: one non-string role, or a `sorted()` set of size 1, or an assignment whose `project:` names no real project — each passes against the unfixed script, so the test proves nothing. Ideally the grader confirms the new fixture fails against base HEAD's render-dashboard.py.
- Truthiness guard reintroduced for `role`, repeating exactly the trap `_status_text` documents: `role: 0` silently dropped with no warning, so the partner sees an unfilled role and no explanation.
- Over-broad normalization that sweeps in `strategic_impact` / `allocation_pct`: `_coerce_impact` does `int(str(value).strip())` and would still pass every existing assertion, so this defect is invisible to the test suite and must be caught by reading the field allowlist.
- Stringifying a missing key: `str(None)` leaking `None` into the status column or into `covered`, turning an absent field into a fake value.
- Warning-noise regression: normalizing inside `_compute`'s per-project loop rather than once at the parse boundary, so one bad assignment emits N warnings on an N-project portfolio and inflates the warnings panel.
- Modifying `_coerce` in validate-entities.py despite the issue's explicit prohibition — it is shared with register-entity.py and staffing-score.py, and the validator legitimately reports `status: 2026` as an invalid enum.
- Deleting or loosening Fixture 10 assertions (or `_status_text`'s warning wording) to make a refactor green, quietly retiring #1155-era coverage while claiming to supersede it.
- Judgment-only, no gate can decide it: whether the chosen allowlist of string-typed fields is the *right* set for every entity type (dates, `skills[]`, `seniority`, `consultant`/`project` back-references) is a data-model call a human should eyeball — the checklist only bounds it by requiring numeric fields be excluded and the renderer's behavior be preserved.
- Judgment-only: whether the SKILL.md wording still reads as partner-facing narration rather than a restatement of the coercion algorithm.

- The plan-time acceptance criteria list 2 items (renders with a warning + a Fixture 10-style regression test), but a base-grounded regeneration derives ~21 gating items, several of them substantive and absent from the AC: the open_roles/role type-symmetry regression that the read-site variant (the issue's own option 1) would introduce, the falsy `role: 0` truthiness trap, the non-vacuity requirement that the new fixture carry two mixed-type roles on one real project, the prohibition on normalizing numeric fields, and the SKILL.md doc-sync that changed_files[] shows is in scope but the AC never asks for. The acceptance bar was materially under-specified for an issue that explicitly leaves a design choice open.

sdh07:
## Hold cleared — maintainer ruling

The `svc:needs-human-review` hold on this PR was a **Trigger-1 `insufficient_coverage`** escalation: the merger's independent contract derived ~21 gating items from the base code, against the 2 items issue #1162's acceptance criteria listed. That escalation graded the **issue's acceptance bar**, not this change.

The maintainer has ruled that the delivered work meets the bar. Every substantive item the contract derived beyond the plan-time AC is already satisfied here:

- both sides of the fill comparison normalized (`role` **and** `open_roles` entries), so a numeric role still matches its numeric label rather than being reported falsely open;
- the guard keys on `raw is not None`, so a falsy `role: 0` warns rather than vanishing;
- Fixture 11 is non-vacuous — two mixed-type roles on one real project, plus a symmetry project and a clean sibling;
- numeric fields (`strategic_impact`, `allocation_pct`) excluded from normalization;
- `_status_text` preserved as a delegate with byte-identical warning text, Fixture 10 unweakened;
- SKILL.md's warn-and-degrade bullet generalized to the role label;
- `validate-entities.py` untouched, no version bump in the branch.

Per the **cleared-hold protocol**, the ratified bar has been written back into issue #1162's `## Acceptance criteria` section, so a future gatekeeper run derives its contract from the corrected criteria instead of re-raising this same escalation.

Proceeding to ready + squash-merge.

Issue #1166, opened 2026-07-27
cogni-workspace: rethink (and possibly delete) the per-session CLAUDE.md copy mechanism


sdh07:
## Plugin
cogni-workspace

### The launcher clobbers the workspace-root `CLAUDE.md` every session

The Obsidian Terminal launcher copies a language template over the workspace-root `CLAUDE.md`:

- `templates/obsidian/plugins/terminal/workspace-orchestrator.sh:84-95` — `copy_claude_template()` does `cp "$WORKPLACE_ROOT/.claude/templates/CLAUDE.${lang}.md" "$WORKPLACE_ROOT/CLAUDE.md"`.
- It is called from `launch_claude()` (line 114), immediately after `select_language()` — so it runs on **every launcher start**, not only when the language actually changes.
- There is no diff, no prompt, and no backup. Anything a user added to the root `CLAUDE.md` is silently destroyed each session.

`.claude/templates/` exists solely to feed this function. Repo-wide grep finds no other reader — no other plugin, no `workspace-status` / `workspace-dashboard` foundation check.

### The rules also exist in a second hand-maintained copy that has already drifted

The language/orthography rules live in two places with nothing keeping them in sync:

- `assets/claude-templates/CLAUDE.{de,en}.md` — copied to the workspace root as `CLAUDE.md` at init, and to `.claude/templates/` for the launcher's per-session language switch.
- `assets/output-styles/workspace-{de,en}.md` — installed to `.claude/output-styles/` and refreshed on every update.

The output style is a strict superset. Every instruction in `CLAUDE.de.md` already appears in `workspace-de.md`:

| `CLAUDE.de.md` | `workspace-de.md` |
|---|---|
| `:5` (respond in German) | `:5` — byte-identical |
| `:6` + `:7` (umlauts/ß; no `ae`/`oe`/`ue`/`ss` substitutes) | `:6` — merged into one line |
| `:8` (identifiers stay English) | identifier clause — byte-identical |

**The copies have already drifted.** The output style has since gained env-var path rules, relative-path display, plugin-ownership attribution, and an Intent Router table; none of that reached the template. And a wording defect in the shared identifier sentence had to be fixed in both files independently, because nothing connects them.

## Why this is worth rethinking rather than extending

Two facts pull in opposite directions and nothing in the plugin states which one is the contract:

1. The launcher treats the root `CLAUDE.md` as **plugin-managed** — it clobbers it unconditionally.
[...]

Issue #1165, opened 2026-07-27
cogni-workspace: the CLAUDE.md template and the output style duplicate the language rules and have already drifted


claude-cogni-work:
## Plugin
cogni-workspace

## Current behavior
The language/orthography rules exist in two hand-maintained copies with nothing keeping them in sync:

- `assets/claude-templates/CLAUDE.{de,en}.md` — copied to the workspace root as `CLAUDE.md` at init, and to `.claude/templates/` for the Obsidian launcher's per-session language switch.
- `assets/output-styles/workspace-{de,en}.md` — installed to `.claude/output-styles/` and refreshed on every update.

The output style is a strict superset. Every instruction in `CLAUDE.de.md` already appears in `workspace-de.md` — `:5` is byte-identical to `workspace-de.md:5`, `:6`+`:7` are merged into `workspace-de.md:7`, and `:8` is byte-identical to `workspace-de.md:11`.

**The copies have already drifted.** The output style has since gained env-var path rules, relative-path display, plugin-ownership attribution, and an Intent Router table; none of that reached the template. And a wording defect in the shared identifier sentence had to be fixed in both files independently, because nothing connects them.

## Why this is not simply "delete the template"
Investigated and rejected. The template is redundant in *content* but not in *activation*:

- `CLAUDE.md` is auto-loaded as project context every session, no user action.
- Output styles are opt-in and fixed at session start. **Nothing in this repository ever activates one.** `generate-settings.sh` writes only an `env` object into `.claude/settings.local.json` — no `outputStyle` key exists anywhere in the monorepo. The only mention of activation is `templates/obsidian/plugins/terminal/workspace-orchestrator.sh:134`, which *echoes* `Activate with: /output-style workspace-${LANGUAGE}` and never invokes it.

Deleting the template today would mean a fresh session in a newly-created or newly-language-switched workspace silently falls back to default behavior until the user runs a slash command the onboarding flow never surfaces. That is a regression, not a cleanup.

## Proposed solution
Two options, in preference order.

**A — Kill the duplication, keep the mechanism (smaller, safe now).** Reduce `CLAUDE.{de,en}.md` to a pointer at the output style, or generate them from the output-style source at copy time, so the two cannot drift again.

**B — Make removal safe, then remove (larger, sequenced).**
1. Make `generate-settings.sh` and `workspace-orchestrator.sh` write `outputStyle` into `.claude/settings.local.json` directly, replacing the echoed suggestion with an actual write.
[...]

Issue #1150, opened 2026-07-26
cogni-consult: SKILL.md column-header drift guard (conditional)


sdh07:
## Plugin
cogni-consult

## Use case
`cogni-consult/CLAUDE.md` warns that the two output-style files carry no shared reference and "must be kept in step deliberately". Today that means ~140 lines of hand-maintained parallel doctrine, which will drift.

#1146 addresses this by **collapsing** what can drift — the two files become one language-neutral stance file over `references/user-facing-output.md`. This issue is the fallback if that proves insufficient.

## Proposed solution
`scripts/check-register-drift.py` plus a third job in `.github/workflows/lint.yml`, modelled on the existing ratchet (`scripts/check-breadcrumbs.py` + `scripts/baselines/` + `tests/`).

**One assertion:** a regex over `cogni-consult/skills/*/SKILL.md` table lines flagging a backticked field identifier used as a column header — the defect class that produced the `Scope` column.

Reuse `check-breadcrumbs.py`'s baseline-ratchet and `ALLOW_MARKER` escape hatch so the guard can land without first cleaning every historical occurrence.

### Narrowed: the heading-parity assertion is superseded

This issue originally carried a second assertion — identical `##` heading sets across `strategy-advisor.md` and `strategy-advisor-de.md`, plus a line ceiling. #1146 now collapses those two files into **one**, and one file cannot drift against itself. The failure mode that assertion caught (a section added to one file only) stops existing by construction, so a checker for it would be dead code from the day it lands.

The remaining assertion is unaffected: it reads `skills/*/SKILL.md`, never the style files, and its defect class is orthogonal to how many style files exist. That is why this issue **narrows rather than closes**.

## Alternatives considered
- **Content parity across the two languages.** Moot under #1146's collapse, and was never machine-checkable anyway — the DE file legitimately carried DE-only normative content (orthography, the English-term allowlist, the before/after mapping table) with no EN counterpart.
- **Generate the DE file from the EN one.** Rejected in #1146 with its rationale: reviewing a generated system prompt is worse than reviewing a hand-written one.
- **A guard on the language rules themselves.** Not needed. The workspace language now travels in the `language` settings key, which Claude Code renders into a `# Language` system-prompt section — there is no second hand-maintained copy of "respond in X" left to drift against.

## Acceptance criteria
- [ ] Ship **only if** post-#1146 drift is actually observed — this issue is conditional by design, and closing it as "not needed" is a valid outcome.
[...]

PR #1168, opened 2026-07-28
cogni-workspace + cogni-consult: language settings key as SSOT, and carry language into subagents

sdh07:
## Description

The workspace language lived in four hand-maintained files and reached a session only through a `CLAUDE.md` the Obsidian launcher clobbered on every start. Claude Code's own `language` settings key builds a `# Language` system-prompt section, which outranks all of it — so the key becomes the source of truth and the copy mechanism goes away.

Separately, the four cogni-consult agents received no language at all. A subagent's system prompt is its own body plus a trust note, a notes block, the environment info, and an attachments block — no `CLAUDE.md`, no settings language, no active output style. That is why the persona-challenge transcript could not meet its German-output bar: nothing reached the agent authoring the graded text.

> ### Correction — the first two commits verified against the wrong binary
>
> Commits `9d7ac29` and `ffe2cc4` claimed to be traced against 2.1.220. They were not. `claude` resolves to `/opt/claude-code/bin/claude`, a compiled ELF; what was actually read was a **separate npm install of the package at version 2.1.42** (`/opt/node22/lib/node_modules/@anthropic-ai/claude-code/`, `package.json` → `2.1.42`). Three claims were false, and two had been written into `cogni-consult/CLAUDE.md`, where the doc guards would have preserved them. `4e13219` corrects the code and the docs. In all three cases the premise this branch started from was right and my "correction" was wrong.

## Verified mechanics (`/opt/claude-code/bin/claude`, 2.1.220)

| Mechanism | Finding |
|---|---|
| `language` settings key | Real. Builds `# Language` with **two** paragraphs: respond-in-X plus identifier handling, *and* "Maintain full orthographic correctness… Never substitute accented characters with their ASCII equivalents (e.g., never write … `"fur"` for `"für"`, or `"loeschen"` for `"löschen"`)". |
| Does it reach subagents? | **No.** The builder has exactly one caller, the main-loop assembler (beside `memory`, `env_info_simple`, `output_style`). The subagent assembler returns `[…agentBody, trust-note, Notes block, env info, attachments]`. |
| `keep-coding-instructions` on a **plugin** style | **Honored.** `source:"plugin",forceForPlugin:qde(i["force-for-plugin"]),keepCodingInstructions:qde(i["keep-coding-instructions"])`, carried through the merged style map into `c.keepCodingInstructions===!0?uO_():null`. |
| `force-for-plugin` | Real, session-global, first-loaded-wins with a warning — which is the durable reason not to use it. |
| `SubagentStart` | Carries `agent_id` + `agent_type`; `additionalContext` becomes a `hook_additional_context` message. Matcher `^(a|b|c)$` takes the regex branch, since `^ ( ) $` fall outside the literal-list character class. |

## Changes

**cogni-workspace**

- `scripts/generate-settings.sh` writes the `language` key, mapping the ISO code to the natural-language name the setting expects. The map covers the European languages the market registry names (`en, de, fr, it, nl, pl, es`); an unmapped code is reported as `data.language_key_skipped` rather than silently producing a workspace with no `# Language` section at all.
- Same script: `--update` now read-modify-writes the whole settings document. It previously rebuilt the file from `env` alone, silently destroying any other top-level key the user had added (`permissions`, `hooks`, `outputStyle`). **This is an independent pre-existing bug** and the fix stands on its own.
- New `hooks/on-session-start-language.sh` carries only the residue the built-in section leaves behind: German **ß/ss** discipline (choosing `ß` after a long vowel vs `ss` after a short one is not an ASCII substitution, so the built-in rule says nothing about it) and the mid-conversation language-switch override (the built-in says "Always respond in X" with no escape clause). It emits a `hookSpecificOutput` envelope, because plain stdout is not injected as context. German-only by design: other languages have no residue once diacritics are covered upstream.
- The Obsidian launcher's per-session language switch writes the settings key instead of copying a template over the workspace-root `CLAUDE.md`. That file is now the user's unconditionally.
- `assets/claude-templates/` and the `.claude/templates/` cache are retired; Init Mode no longer writes a root `CLAUDE.md`, and Update Mode gains a migration step.
- The workspace output styles keep their env-var, relative-path, plugin-ownership and Intent Router rules, drop what the settings key now delivers, and declare `keep-coding-instructions: true`.

**cogni-consult**

- New `hooks/hooks.json` + `hooks/on-subagent-start.sh`: a `SubagentStart` hook matched on the four `consult-*` agent types. It emits the resolved language together with `references/subagent-output-contract.md`, read from its first `##` heading onward. Its German block is deliberately **complete** — respond-in-German, diacritics, ß/ss, identifiers-stay-English — because subagents get no `# Language`. The two hooks emitting different rule sets is correct, not drift, and both files say so.
- A hook cannot see the user's message, so the message-detection override travels as an `interaction_language` dispatch input that wins over the hook's default. Documented in both orchestration contracts, the three dispatch sites, and the three agent input contracts.
- Both output styles declare `keep-coding-instructions: true`, which is **behavioral**: it keeps `# Doing tasks` (prefer-editing-existing-files, the OWASP warning, the over-engineering rules) in place, which matters for skills that run `deliverable-graph.py` and do idempotent `Edit` round-trips on `field.json`.

## Testing

- `generate-settings.sh`: `--language de` → `"language": "german"`; `--language fr` → `"french"`; `--language xx` → no key written and `language_key_skipped: "xx"` in the envelope.
- `--update` after hand-adding `permissions` and a custom `env` var: both survive. This fails on `main`.
- `on-session-start-language.sh`: emits **only** the ß/ss and switch rules for German, silent for English, silent with no workspace, and resolves via `.workspace-config.json` when the settings key is absent.
- `on-subagent-start.sh`: German workspace emits respond-in-German + diacritics + ß/ss + identifiers-stay-English + the register block; falls back to English with no workspace. Asserted field by field, so the intended asymmetry against the session hook is tested rather than eyeballed.
- The launcher's `write_settings_language` extracted and run directly: sets the key, preserves `permissions` and `env`, and warns without failing under `set -e` on an unknown code.
- `python3 scripts/check-breadcrumbs.py` — 0. `python3 scripts/check-version-bump.py` — 0 versions touched. `bash cogni-workspace/scripts/check-skill-names.sh` — OK. `bash -n` clean on all four touched scripts; both `hooks.json` parse.

**Still un-run**, because they need a live German workspace and a real session — not claimed as passing:

- End-to-end "German from the first token with no output style and no root `CLAUDE.md`".
- `/cogni-consult:consult-personas` through a full challenge round.
- **Activate `cogni-consult:strategy-advisor` and confirm `# Doing tasks` survives** — the evidence the reviewer asked for, which the earlier "inert" framing meant nobody collected.

## Checklist

- [x] I have read and agree to the [Contributor License Agreement](../CLA.md)
- [x] My changes follow the existing code style
- [x] I have updated documentation where applicable
- [x] I have tested my changes


sdh07:
## Review — two mechanical claims are falsified, one ships into the repo

Verified against the same Claude Code 2.1.220 binary this PR says it traced.

### 1. `keep-coding-instructions` IS read for plugin-shipped styles

The findings table says it is *"**not read at all** for plugin-shipped ones, which therefore always drop the `# Doing tasks` block"*, and `cogni-consult/CLAUDE.md` now states it is *"honored for styles loaded from a directory but silently ignored for plugin-shipped ones."*

The plugin loader reads it:

```js
{ …, source:"plugin",
  forceForPlugin:         Kde(i["force-for-plugin"]),
  keepCodingInstructions: Kde(i["keep-coding-instructions"]) }
```

`Kde` is boolean coercion, and the merged style map carries `keepCodingInstructions` through to the resolver. The CLAUDE.md paragraph is also self-undermining — it says the declaration *"starts working the moment the plugin loader reads it"*, which it already does.

**Impact:** this is a false mechanic promoted from PR text into guarded repo documentation, where doc guards protect it and later sessions cite it.

**Second-order:** because the field *is* honored, adding `keep-coding-instructions: true` to the two cogni-consult styles is **behavioral, not inert** — activating `strategy-advisor` now retains the engineering defaults. That is the correct change, but the PR documents it as a no-op, so the acceptance evidence does not test it. Please add: activate the style, confirm `# Doing tasks` survives.

### 2. `# Language` does contain an orthography rule

The findings table says *"No orthography rule"*. That premise justifies `on-session-start-language.sh` (82 lines + a second `SessionStart` registration) and keeps the orthography block in the workspace output styles. Verbatim sentence 2 of `fMy()`:

> Maintain full orthographic correctness for ${lang}, including all required diacritical marks, accents, and special characters. Never substitute accented characters with their ASCII equivalents (e.g., never write "nao" for "não", "fur" for "für", or "loeschen" for "löschen").

A genuine residue **does** exist and the hook is not pointless: ß-after-long-vowel vs ss-after-short is not an ASCII substitution, so `dass` / `muss` / `Maßnahme` is uncovered, as is the mid-conversation language-switch rule. But the hook is over-scoped and mis-justified on the current premise, and the workspace output styles re-introduce some of the duplication this PR exists to delete. Please re-scope to the actual residue and restate the rationale.

### 3. Verify, don't assume — the subagent-prompt claim

*"A subagent's system prompt is its own body plus a notes block and the environment info"* is the load-bearing justification for the entire `SubagentStart` hook. It was not independently confirmed in this review. Given two of the three sibling claims in the same table are wrong, please show the trace for this one before merge.

### Good, and worth keeping

- The `--update` fix in `generate-settings.sh` is a real independent bug: `json.dump({"env": env})` destroyed `permissions`, `hooks`, and every other top-level key on update. Read-modify-writing the whole document is the right fix.
- Routing rung 2 of the interaction-language ladder through an `interaction_language` dispatch input, because a hook cannot see the user's message — good catch.
- `SubagentStart` matcher as a regex over agent-type names is consistent with the `matchQuery: agentType` binding.
- Guards clean: breadcrumbs 0, no version bump in branch.

### Minor

`LANGUAGE_NAMES` is triplicated across `generate-settings.sh` and both hook scripts. The comment acknowledges it, but that is the drift class this PR removes elsewhere. It also covers only `en`/`de`: any other workspace language silently gets no `language` key and therefore no `# Language` section. If workspace language is deliberately en/de-only, state it; otherwise it is a silent gap.

sdh07:
You're right on both, and the root cause is worse than the two findings: **I traced the wrong binary.**

`claude` resolves to `/opt/claude-code/bin/claude`, a 275 MB compiled ELF at 2.1.220. What I actually read was `/opt/node22/lib/node_modules/@anthropic-ai/claude-code/cli.js` — a separate npm install whose `package.json` says **2.1.42**. So every claim I labelled "verified against the installed 2.1.220" was verified against 2.1.42. That also explains the shape of the errors: the draft this branch started from had the mechanics right, and I "corrected" three of them wrongly.

Fixed in `4e13219`.

### 1. `keep-coding-instructions` — confirmed, you're right

```
source:"plugin",forceForPlugin:qde(i["force-for-plugin"]),keepCodingInstructions:qde(i["keep-coding-instructions"])
```

and it survives the merge (`keepCodingInstructions:u.keepCodingInstructions`) into `c===null||c.keepCodingInstructions===!0?uO_():null`. You also caught that the CLAUDE.md paragraph was self-undermining — "starts working the moment the plugin loader reads it" about a loader that already reads it. That paragraph is rewritten to say the field is honored and load-bearing, the upstream-filing sentence is gone, and the acceptance evidence you asked for is now listed in the PR body as un-run rather than quietly absent.

### 2. `# Language` orthography — confirmed, you're right

Verbatim from the real binary:

> Maintain full orthographic correctness for `${e}`, including all required diacritical marks, accents, and special characters. Never substitute accented characters with their ASCII equivalents (e.g., never write "nao" for "não", "fur" for "für", or "loeschen" for "löschen").

Re-scoped exactly to the residue you identified. `on-session-start-language.sh` now carries only the ß-after-long-vowel vs ss-after-short rule and the mid-conversation switch override; the umlaut/ASCII-substitution bullet is gone. Its header no longer rests on the false premise, and the German-only rule table is now documented as deliberate rather than accidental — once diacritics are covered upstream, other languages have no residue.

One thing I did **not** re-scope, deliberately: the subagent hook keeps the full set including diacritics, because subagents get no `# Language` at all. Both hook headers now say so, so nobody later "aligns" them and silently removes rules from the agent path.

### 3. The subagent-prompt claim — trace, as requested

Fair to demand it given the above. The language builder has **exactly one caller** in the whole binary, the main-loop assembler:

```js
oR("language",()=>iO_(u.language)), oR("output_style",()=>sO_(c)), …
```

The subagent path is a different function:

```js
async function zon(e,t,r){ let i=`Notes:…`, s=await EO_(t,r), a=Zep(t);
  return [...e, "Messages from the agent that launched you …", i, s, ...a!==null?[a]:[]] }
```

`EO_` is the env-info block, `Zep` is attachments/countdown. No language section, no memory. So this one holds — it is now the only justification the hook rests on, and it's the one I checked twice.

### 4. The en/de gap — agreed, it was silent

Not a deliberate restriction worth defending. The map now covers the European languages the market registry actually names (`en, de, fr, it, nl, pl, es`), and an unmapped code is reported as `data.language_key_skipped` instead of producing a workspace with no `# Language` section at all. The triplication stands — the registry carries ISO codes only, no display names, and the three sites span a plugin boundary and a standalone-script boundary — but each now names the others, and the gap is no longer silent.

I'm also correcting the five issue bodies I edited earlier in this branch, since my "corrections" injected these falsehoods into #1146, #1147, #1142, #1166 and #1150 — which is precisely the failure mode #1146 exists to prevent.


---

sdh07:
## Follow-up — code fixes verified, two issue bodies still carry the falsehood

Re-verified against `/Users/stephandehaas/.local/share/claude/versions/2.1.220` (my local install is the real 2.1.220, which is why our traces diverged). The wrong-binary diagnosis is the right root cause and explains all three errors at once.

### Landed correctly

- `cogni-consult/CLAUDE.md` — now states the plugin loader reads the field and treats the declaration as load-bearing behaviour. The `force-for-plugin` rationale is correctly framed as version-independent, which is the part that will survive the next upgrade.
- `on-session-start-language.sh` — re-scoped to the actual residue; the DE block carries only ß/ss discipline and the header no longer rests on the false premise. Keeping the full set in the subagent hook, and documenting *why* the two differ, is the right call — that asymmetry would otherwise get "fixed" later and silently strip rules from the agent path.
- `LANGUAGE_NAMES` — en/de/fr/it/nl/pl/es with `data.language_key_skipped` for unmapped codes. Gap is no longer silent.

### Still outstanding

**#1142** was edited, but the correction embeds the second-order falsehood:

> (That opt-out is honored for directory-loaded styles but not read for plugin-shipped ones on 2.1.220 — a Claude Code limitation, filed upstream.)

#1147 explicitly retracted that exact wording and recorded that **there is no upstream bug**. #1142 never got the follow-up, so it currently asserts both a false mechanic and a filing that does not exist. The parenthetical should go; the surrounding bullet is otherwise correct.

**#1166** is untouched — all three reasons under *"The output style cannot be the single source"* still stand:

1. *"Output styles are opt-in … `generate-settings.sh` writes only an `env` object"* — going stale, since this branch changes exactly that.
2. *"A plugin cannot select one programmatically"* — false (`force-for-plugin: true`).
3. *"An output style replaces the system prompt"* — false as stated; it replaces one block (`# Doing tasks`), with `keep-coding-instructions: true` as the opt-out, and `# Executing actions with care` is unconditional.

Reasons 2 and 3 are the two premises that made #1165 closable as not-planned, so correcting them re-opens the option space #1166 is actually asking about — worth doing before anyone answers its four questions.

#1147, #1146 and #1150 are clean. #1146's new acceptance criterion naming all three forbidden statements is the right guard to have added.

sdh07:
Both bodies are now corrected. You were right about what matters: I posted correction *comments* on #1142 and #1166 but left the falsehoods in the bodies, and the body is what gets read.

**#1142** — the parenthetical is gone from the *Alternatives considered* bullet. Since that bullet has now been wrong twice, I replaced it with a short correction history recording both withdrawn versions, the fact that **no upstream bug was filed**, and the wrong-artifact root cause. Cheaper for the next reader than finding two separate retractions.

**#1166** — all three reasons corrected in the body, struck through rather than deleted, since they are the premises that made #1165 closable:

1. Marked stale in its evidence — `generate-settings.sh` no longer writes only an `env` object. The "nothing activates one" half is still true and still a good reason.
2. and 3. Marked false, with the actual mechanics and the version-durable session-global reason in their place.

Two consequences of that I flagged rather than left implicit:

- The acceptance criterion *"the single source … is the artifact that actually loads (`CLAUDE.md`)"* is falsified by this branch — the single source is the settings key, and `CLAUDE.md` is not involved at all. Struck and corrected in place, since implementing it as written would rebuild the mechanism the issue asks to retire.
- **#1165 was closed as not-planned partly because "its two proposed options both required making the output style authoritative, which the three reasons above rule out."** Two of those three reasons are now corrected, so that closure rests on a premise that no longer holds. I left #1165 closed — its options are superseded by the settings key rather than merely unblocked — but noted in #1166's Related section that they should be re-read on their merits if this ever gets revisited, rather than treated as ruled out. Say the word if you'd rather I reopen it.

Also added the "trace against `readlink -f "$(command -v claude)"`, not a `node_modules` copy" criterion to #1142, matching the one already on #1146.

Worth noting for anyone else on this machine class: the divergence was not a version disagreement between us. Your local install is the real 2.1.220; this container has a stale **2.1.42** npm package sitting beside the real binary, with `/opt/node22/bin/claude` symlinking past it. Anything resolving Claude Code through `node_modules` here reads the wrong version.

No code changed in this round — `4e13219` is still the head, and Lint is green on it.


---

PR #1169, opened 2026-08-02
cogni-projects: dashboard narrate step + envelope health/value fields

### LINKED ISSUE #1151: cogni-projects: projects-dashboard has no narrate step, and the envelope cannot supply one
## Plugin
cogni-projects

## Problem
`projects-dashboard` tells an agent to run the renderer, report `data.path`, and offer to open the file — then stops. There is no step that narrates the snapshot, even though the skill's own Notes bullet states its purpose is to explain the output "so it can be narrated to a partner", and its description promises "staffing coverage per project, at-risk projects, and portfolio value by strategic impact".

The gap is structural, not stylistic. `render-dashboard.py`'s JSON envelope (lines 489-501) carries only `path`, `portfolio`, `projects`, `consultants`, `avg_allocation`, `fully_allocated`, `open_roles`, `warnings`, `partial`. The per-project health flags and the `value_by_impact` grouping exist **only** inside the rendered `output/dashboard.html`. So an agent that follows Step 2 and reads stdout — exactly what the skill instructs — cannot name which projects are at risk or where the strategic-impact weight sits, and the skill correctly forbids re-deriving them from entity records.

The sibling `projects-staff` has an explicit summarize step for the same reason.

## Proposed solution
Add a short Step 4 to `cogni-projects/skills/projects-dashboard/SKILL.md` instructing the agent to read `output/dashboard.html` back for the per-project flags and the impact grouping rather than re-deriving them, and to give the partner a few lines: project and consultant counts, roles still open, the projects carrying a warn/risk flag by name, where the impact weight sits, plus any warnings.

An alternative worth weighing first: emit the per-project flags and `value_by_impact` in the JSON envelope, so the narrating agent never has to parse HTML. That is the cleaner fix if the envelope is allowed to grow.

## Acceptance criteria
1. An agent following the skill end-to-end can name at-risk projects and the impact grouping without re-deriving either from entity records.
2. Whichever route is chosen, the skill does not restate any flag trigger condition or threshold — `render-dashboard.py` stays the source of truth.

_Affected components: cogni-projects/skills/projects-dashboard/SKILL.md, cogni-projects/scripts/render-dashboard.py_

_Found during review of #1136; deliberately out of scope there, which was a documentation-trim PR._

claude-cogni-work:
Closes #1151.

## Summary
The `projects-dashboard` skill's stated purpose is to narrate the snapshot to a partner, but it stopped after reporting `data.path` — and `render-dashboard.py`'s JSON envelope carried none of the per-project health flags or the strategic-impact grouping needed to do so (they lived only in the rendered `output/dashboard.html`). This takes the reporter-preferred **envelope route**:

- **`render-dashboard.py`** — add two additive keys to the success envelope's `data` dict: `projects_detail` (the per-project list `_compute` already builds — name/client/status/impact/roles_total/roles_filled/open_roles/health_label/health_sev) and `value_by_impact` (the `{1..5: count}` grouping). Both are values already in scope at the print site; no new derivation, so the script stays the single source of truth for flags and thresholds. Purely additive — the HTML output and every existing envelope key are untouched.
- **`projects-dashboard/SKILL.md`** — add `### Step 4: Narrate the snapshot` before `## Notes`, telling the agent to read `data.projects_detail` and `data.value_by_impact` from stdout and narrate a few lines (counts, open roles, at-risk projects by name via `health_sev` risk/warn, where impact weight concentrates, warnings) — from the envelope only, never re-deriving from entity records and never restating a flag threshold. One sentence added to Step 2's success paragraph flags the two new fields.
- **`tests/test-render-dashboard.sh`** — three additive Fixture 1 assertions covering `projects_detail` (length + per-project health keys) and `value_by_impact` (string-keyed counts).

## Acceptance criteria
1. ✅ An agent following the skill end-to-end can name at-risk projects and the impact grouping from stdout alone — no re-derivation from entity records.
2. ✅ Step 4 reads `health_sev`/`health_label`/impact tiers as reported and restates no flag trigger condition or threshold; `render-dashboard.py` stays the SSOT.

## Testing
All three `cogni-projects` test suites pass (render-dashboard incl. new 1k/1l/1m, portfolio-init, register-entity). No `plugin.json` version bump — the post-merge workflow owns it.

## Review notes

- **One out-of-delta hunk, taken deliberately.** `c574b714` edits the Core concept's `staffing unknown` gloss, which is pre-existing on `main` and outside this PR's delta. All three review perspectives independently flagged it as restating `_project_health`'s trigger condition; two waived it as out-of-scope, one did not. Fixed here rather than deferred because it is a two-line prose swap with no behaviour change, and leaving it would keep re-firing on every future pass over this file. The renderer still names the cause at runtime, through the `has no open_roles — staffing status unknown` warning Step 4 already relays.
- **One finding's supporting claim was not acted on, because it is wrong.** The last verdict argued the worked example's warnings line "points the agent at the wrong envelope field", on the premise that a missing `open_roles` produces no `data.warnings` entry. `render-dashboard.py:332-333` does the opposite — `if declared_roles is None and status != "closed": warnings.append("%s has no open_roles — staffing status unknown" % label)`. The warning is real, so the example now quotes it verbatim instead of dropping it. The rest of that finding was valid and is fixed: `staffing unknown` carries `warn`, so the example flags three projects, not two.

### DIFF
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -587,6 +587,8 @@ def main(argv):
             "avg_allocation": util["avg_allocation"],
             "fully_allocated": util["fully_allocated"],
             "open_roles": _open_role_demand(projects),
+            "projects_detail": projects,
+            "value_by_impact": value_by_impact,
             "warnings": warnings,
             "partial": bool(warnings),
         },
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -81,9 +81,11 @@ incomplete and which records to fix. Note that `partial` is set by *any*
 warning, including a non-substantive one such as an unreadable
 `--design-variables` file, so check what the warnings actually say before
 describing the portfolio data itself as incomplete. The `data` envelope also
-carries `projects_detail` (a per-project list with `health_label` / `health_sev`
-/ `impact`) and `value_by_impact` (project count per strategic-impact tier) —
-Step 4 reads these to narrate the snapshot.
+carries `projects_detail` (per project: `name`, `client`, `status`, `impact`,
+`roles_total`, `roles_filled`, `health_label`, `health_sev`, plus an
+`open_roles` **list** of that project's still-unfilled role labels — not the
+portfolio integer `data.open_roles`) and `value_by_impact` (project count per
+strategic-impact tier) — Step 4 reads these to narrate the snapshot.
 
 **When `success` is `false`:** the render did not run at all — read `error`. This
 is the environment-level failure branch (missing portfolio directory, missing
@@ -113,11 +115,17 @@ re-derive anything from the entity records:
 - **Counts** — `data.projects` projects and `data.consultants` consultants,
   with `data.open_roles` roles still open across the portfolio.
 - **At-risk projects, by name** — from `data.projects_detail`, name each project
-  whose `health_sev` is `risk` or `warn` (quote its `health_label`). If none are
-  flagged, say the portfolio is fully staffed.
+  whose `health_sev` is `risk` or `warn` (quote its `health_label`). If no
+  project carries `risk` or `warn`, say no project is currently flagged — and
+  name any `closed` or `no open roles` projects rather than folding them into a
+  staffing claim, since those read as `muted`/`ok`, not `fully staffed`.
 - **Where the strategic-impact weight sits** — from `data.value_by_impact` (a
-  map of impact tier `1`–`5` to project count), point to the tier(s) carrying
-  the most projects.
+  map of impact tier `"1"`–`"5"`, string keys in the JSON envelope, to project
+  count), point to the tier(s) carrying the most projects. Projects whose
+  `strategic_impact` is missing or outside 1–5 are omitted (they surface as a
+  warning and carry `impact: null` in `projects_detail`), so these counts need
+  not sum to `data.projects`; when no project declares a valid impact the map is
+  all zeros — a meaningless five-way tie, not a real distribution to point at.
 - **Warnings** — if `data.partial` is `true`, relay `data.warnings` so the
   partner knows which records to fix.
 
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -31,8 +31,8 @@ records (their field contract lives in
   health flag. Only a live commitment covers a role — a finished assignment
   releases it, so a role can reappear as open with no edit to the project.
   Two flags read alike but are not: `no open roles` means the project needs
-  nobody right now, while `staffing unknown` means it never declared
-  `open_roles`, so coverage cannot be read — never narrate it as covered.
+  nobody right now, while `staffing unknown` means its coverage cannot be read
+  at all — never narrate it as covered.
   The headline figure (`data.open_roles` and the `Open roles` tile) counts
   unfilled roles on non-closed projects only — closed work is delivered or
   lost, not live demand — so it matches the set `projects-staff` shortlists. A
@@ -135,12 +135,12 @@ stays the source of truth for how those are computed (see Notes below).
 Illustrative shape only — take every number, name, and label from the envelope,
 never from this example:
 
-> Six projects, nine consultants, eleven roles still open. Two projects are
+> Six projects, nine consultants, eleven roles still open. Three projects are
 > flagged: **Harbour Replatform** is `unstaffed`, **Meridian Rollout** is
-> `2/3 roles open`. **Southbank Archive** carries `closed`. The strategic
-> weight sits in tier 5 (three projects), with nothing at tier 1. One warning:
-> **Tallow Freight** never declared `open_roles`, so its coverage cannot be
-> read.
+> `2/3 roles open`, **Tallow Freight** is `staffing unknown`. **Southbank
+> Archive** carries `closed`. The strategic weight sits in tier 5 (three
+> projects), with nothing at tier 1. One warning, relayed as reported:
+> *project Tallow Freight has no open_roles — staffing status unknown*.
 
 ## Notes
 
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -18,7 +18,7 @@ Render a partner-meeting snapshot of a cogni-projects portfolio: every project
 with its staffing coverage and a health flag, plus an aggregate view of
 portfolio value by strategic impact. The dashboard is a **read-only** view over
 the consultant, project, and assignment records that `projects-setup` and
-`projects-entities` produce — it never changes portfolio state.
+`projects-entities` produce.
 
 ## Core concept
 
@@ -76,16 +76,15 @@ directory and prints a `{"success", "data", "error"}` envelope whose `data.path`
 is the written file.
 
 **When `success` is `true`:** if `data.partial` is `true`, `data.warnings` lists
-what was missing or mismatched — relay those so the user knows the snapshot is
-incomplete and which records to fix. Note that `partial` is set by *any*
+what was missing or mismatched — relay those (Step 4). `partial` is set by *any*
 warning, including a non-substantive one such as an unreadable
 `--design-variables` file, so check what the warnings actually say before
 describing the portfolio data itself as incomplete. The `data` envelope also
 carries `projects_detail` (per project: `name`, `client`, `status`, `impact`,
 `roles_total`, `roles_filled`, `health_label`, `health_sev`, plus an
 `open_roles` **list** of that project's still-unfilled role labels — not the
-portfolio integer `data.open_roles`) and `value_by_impact` (project count per
-strategic-impact tier) — Step 4 reads these to narrate the snapshot.
+portfolio integer `data.open_roles`) and `value_by_impact` — Step 4 reads these
+to narrate the snapshot.
 
 **When `success` is `false`:** the render did not run at all — read `error`. This
 is the environment-level failure branch (missing portfolio directory, missing
@@ -116,15 +115,15 @@ re-derive anything from the entity records:
   with `data.open_roles` roles still open across the portfolio.
 - **At-risk projects, by name** — from `data.projects_detail`, name each project
   whose `health_sev` is `risk` or `warn` (quote its `health_label`). If no
-  project carries `risk` or `warn`, say no project is currently flagged — and
-  name any `closed` or `no open roles` projects rather than folding them into a
-  staffing claim, since those read as `muted`/`ok`, not `fully staffed`.
+  project carries `risk` or `warn`, say no project is currently flagged. Either
+  way, name any project whose flag is `closed` or `no open roles` rather than
+  folding it into a staffing claim — neither means the project is fully staffed.
 - **Where the strategic-impact weight sits** — from `data.value_by_impact` (a
   map of impact tier `"1"`–`"5"`, string keys in the JSON envelope, to project
   count), point to the tier(s) carrying the most projects. Projects whose
   `strategic_impact` is missing or outside 1–5 are omitted (they surface as a
   warning and carry `impact: null` in `projects_detail`), so these counts need
-  not sum to `data.projects`; when no project declares a valid impact the map is
+  not sum to `data.projects`. When no project declares a valid impact the map is
   all zeros — a meaningless five-way tie, not a real distribution to point at.
 - **Warnings** — if `data.partial` is `true`, relay `data.warnings` so the
   partner knows which records to fix.
@@ -133,13 +132,23 @@ Read `health_sev`, `health_label`, and the impact tiers as the renderer reports
 them — never restate a flag's trigger condition or threshold. `render-dashboard.py`
 stays the source of truth for how those are computed (see Notes below).
 
+Illustrative shape only — take every number, name, and label from the envelope,
+never from this example:
+
+> Six projects, nine consultants, eleven roles still open. Two projects are
+> flagged: **Harbour Replatform** is `unstaffed`, **Meridian Rollout** is
+> `2/3 roles open`. **Southbank Archive** carries `closed`. The strategic
+> weight sits in tier 5 (three projects), with nothing at tier 1. One warning:
+> **Tallow Freight** never declared `open_roles`, so its coverage cannot be
+> read.
+
 ## Notes
 
 - **The renderer is the source of truth for the flags and aggregates.**
   `scripts/render-dashboard.py` computes every health flag, the strategic-impact
-  grouping, and the utilization figures; this section explains what the output
-  *means* so it can be narrated to a partner, not how it is computed. To
-  change a threshold or a flag rule, change the script, not this skill.
+  grouping, and the utilization figures; this skill explains what the output
+  *means*, not how it is computed. To change a threshold or a flag rule, change
+  the script, not this skill.
 - **Read-only.** The renderer never writes `projects-portfolio.json` (only
   `projects-entities` does, via `register-entity.py`) and never touches
   `.metadata/`. Re-running only rewrites `output/dashboard.html`.
--- a/cogni-projects/skills/projects-dashboard/SKILL.md
+++ b/cogni-projects/skills/projects-dashboard/SKILL.md
@@ -80,7 +80,10 @@ what was missing or mismatched — relay those so the user knows the snapshot is
 incomplete and which records to fix. Note that `partial` is set by *any*
 warning, including a non-substantive one such as an unreadable
 `--design-variables` file, so check what the warnings actually say before
-describing the portfolio data itself as incomplete.
+describing the portfolio data itself as incomplete. The `data` envelope also
+carries `projects_detail` (a per-project list with `health_label` / `health_sev`
+/ `impact`) and `value_by_impact` (project count per strategic-impact tier) —
+Step 4 reads these to narrate the snapshot.
 
 **When `success` is `false`:** the render did not run at all — read `error`. This
 is the environment-level failure branch (missing portfolio directory, missing
@@ -99,6 +102,28 @@ xdg-open "<data.path>"  # Linux
 ```
 
 A failure here is cosmetic — the dashboard is already written to `data.path`.
+
+### Step 4: Narrate the snapshot
+
+The dashboard's purpose is to be explained to a partner, so close with a few
+plain lines — do not stop at reporting the path. Read these from the renderer's
+stdout envelope only; do not re-open `output/dashboard.html` and do not
+re-derive anything from the entity records:
+
+- **Counts** — `data.projects` projects and `data.consultants` consultants,
+  with `data.open_roles` roles still open across the portfolio.
+- **At-risk projects, by name** — from `data.projects_detail`, name each project
+  whose `health_sev` is `risk` or `warn` (quote its `health_label`). If none are
+  flagged, say the portfolio is fully staffed.
+- **Where the strategic-impact weight sits** — from `data.value_by_impact` (a
+  map of impact tier `1`–`5` to project count), point to the tier(s) carrying
+  the most projects.
+- **Warnings** — if `data.partial` is `true`, relay `data.warnings` so the
+  partner knows which records to fix.
+
+Read `health_sev`, `health_label`, and the impact tiers as the renderer reports
+them — never restate a flag's trigger condition or threshold. `render-dashboard.py`
+stays the source of truth for how those are computed (see Notes below).
 
 ## Notes
 
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -160,6 +160,12 @@ assert_html "1g AC-1 lists other project" "Compliance Audit" "$HTML1"
 assert_html "1h AC-1 fill status shown"  "1/3" "$HTML1"
 assert_html "1i AC-1 health flag shown"  "unstaffed" "$HTML1"
 assert_html "1j AC-2 value section"      "strategic impact" "$HTML1"
+# Narrate-step envelope fields: projects_detail + value_by_impact are surfaced
+# on stdout so the skill can name at-risk projects and the impact grouping
+# without parsing the HTML. json.dumps stringifies the int impact keys.
+assert_json "1k envelope carries projects_detail"        "len(d['data']['projects_detail']) == 2"
+assert_json "1l projects_detail carries per-project health" "all('health_label' in p and 'health_sev' in p for p in d['data']['projects_detail'])"
+assert_json "1m envelope carries value_by_impact"        "d['data']['value_by_impact']['5'] == 1 and d['data']['value_by_impact']['2'] == 1"
 
 # ---------------------------------------------------------------------------
 # Fixture 2 — idempotent re-render: running twice rewrites the same file and
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -165,8 +165,8 @@ assert_html "1j AC-2 value section"      "strategic impact" "$HTML1"
 # without parsing the HTML. json.dumps stringifies the int impact keys.
 assert_json "1k envelope carries projects_detail"        "len(d['data']['projects_detail']) == 2"
 assert_json "1l projects_detail carries per-project health" "all('health_label' in p and 'health_sev' in p for p in d['data']['projects_detail'])"
-assert_json "1n named at-risk project pinned with its flag" "any(p['name'] == 'Compliance Audit' and p['health_label'] == 'unstaffed' and p['health_sev'] == 'risk' for p in d['data']['projects_detail'])"
 assert_json "1m envelope carries value_by_impact"        "d['data']['value_by_impact']['5'] == 1 and d['data']['value_by_impact']['2'] == 1"
+assert_json "1n named at-risk project pinned with its flag" "any(p['name'] == 'Compliance Audit' and p['health_label'] == 'unstaffed' and p['health_sev'] == 'risk' for p in d['data']['projects_detail'])"
 
 # ---------------------------------------------------------------------------
 # Fixture 2 — idempotent re-render: running twice rewrites the same file and
@@ -264,6 +264,9 @@ run "$PF5"
 assert_json "5a injection render succeeds" "d['success'] is True"
 assert_html_lacks "5b entity markup escaped" \
   '<script>alert(1)</script>' "$PF5/output/dashboard.html"
+# _esc is confined to _render_html — the envelope must hand the narrating agent
+# the raw name, not HTML entities.
+assert_json "5c injection name reaches the envelope unescaped" "any(p['name'] == '<script>alert(1)</script>' for p in d['data']['projects_detail'])"
 
 # ---------------------------------------------------------------------------
 # Fixture 6 — genuinely malformed entities, not merely incomplete ones: a
@@ -448,6 +451,10 @@ assert_html "9f tile agrees with the envelope" \
   '<div class="n">1</div><div class="l">Open roles</div>' "$HTML9"
 assert_html "9g closed project still listed" "Legacy Migration" "$HTML9"
 assert_html "9h closed health flag still rendered" ">closed</span>" "$HTML9"
+# _open_role_demand drops closed projects from the headline figure only — the
+# closed row must still reach projects_detail with its flag, or a widened
+# exclusion would pass the suite while the HTML table still showed the row.
+assert_json "9i closed project reaches the envelope with its flag" "any(p['name'] == 'Legacy Migration' and p['health_label'] == 'closed' and p['health_sev'] == 'muted' for p in d['data']['projects_detail'])"
 
 # ---------------------------------------------------------------------------
 # Fixture 10 — a status that is not text. The frontmatter parser coerces an
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -165,6 +165,7 @@ assert_html "1j AC-2 value section"      "strategic impact" "$HTML1"
 # without parsing the HTML. json.dumps stringifies the int impact keys.
 assert_json "1k envelope carries projects_detail"        "len(d['data']['projects_detail']) == 2"
 assert_json "1l projects_detail carries per-project health" "all('health_label' in p and 'health_sev' in p for p in d['data']['projects_detail'])"
+assert_json "1n named at-risk project pinned with its flag" "any(p['name'] == 'Compliance Audit' and p['health_label'] == 'unstaffed' and p['health_sev'] == 'risk' for p in d['data']['projects_detail'])"
 assert_json "1m envelope carries value_by_impact"        "d['data']['value_by_impact']['5'] == 1 and d['data']['value_by_impact']['2'] == 1"
 
 # ---------------------------------------------------------------------------


sdh07:

_The merger formed this expectation from issue #1151 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The gap is real and structural: `cogni-projects/scripts/render-dashboard.py` `_compute` (L308-393) already builds a per-project list carrying `name`, `status`, `impact`, `roles_total`, `roles_filled`, `open_roles`, `health_label`, `health_sev` (L360-370) and a `value_by_impact` dict (L320, L358), but the success print site (L580-594) drops both, emitting only nine scalar/aggregate keys — so an agent following `cogni-projects/skills/projects-dashboard/SKILL.md` Steps 1-3 (L60-101) genuinely cannot name at-risk projects or the impact grouping, and the Notes bullet at L105-109 correctly forbids re-deriving them. The correct fix is the envelope route: additively emit the already-computed per-project detail and `value_by_impact` at the print site (no recomputation, no change to the nine existing keys, `ensure_ascii=False` retained, `_fail` at L519-521 untouched), then add a `### Step 4` narrate step to SKILL.md between `### Step 3: Open it` (L92) and `## Notes` (L103) that reads those named `data.*` keys and gives the partner counts, open roles, the flagged projects by name, the impact weight, and any warnings — modelled on the sibling `cogni-projects/skills/projects-staff/SKILL.md` `### Step 5: Summarize` (L155-161). AC-2 constrains that step to reading flags as values, never restating a trigger or threshold, since `_project_health` (L204-222) and `_coerce_impact` (L169-182) stay the source of truth. `cogni-projects/tests/test-render-dashboard.sh` must gain non-vacuous content assertions on both new keys while all eleven existing fixtures keep passing.

**Acceptance checklist:**
- [ ] SKILL.md gains exactly one new workflow step headed `### Step 4:` placed after `### Step 3: Open it` and before `## Notes`; the Step 1/2/3 headings are unchanged verbatim and nothing is renumbered.
- [ ] SKILL.md frontmatter `name`, `description` and `allowed-tools` are unchanged — `allowed-tools` is still `Read, Bash, Glob, Grep` with no write-capable tool added (the skill stays read-only). If `description` did change, `cogni-projects/README.md` (skill bullet, L28) must also be in the diff; it is not in changed_files, so the description must be untouched.
- [ ] The success print site in render-dashboard.py grows purely additively: all nine existing `data` keys — `path`, `portfolio`, `projects`, `consultants`, `avg_allocation`, `fully_allocated`, `open_roles`, `warnings`, `partial` — are still emitted with identical spelling and semantics.
- [ ] `data.projects` is still the integer project count, not repurposed into the per-project list — the existing assertions `d['data']['projects'] == N` (test lines 153, 170, 333, 436, 526, 690) must still hold.
- [ ] The `_fail` helper (base L519-521) is unmodified: a failure path still prints `{"success": false, "data": {}, "error": ...}` with no new keys, so fixtures 4a/4b and 7a-7c are unaffected.
- [ ] The success `json.dumps` call still passes `ensure_ascii=False`, so non-ASCII names reaching the new keys (`Ana Ström`, `Café Rollout`) are not \u-escaped.
- [ ] The new per-project detail key emits, per project, at minimum `name` and `health_label` (ideally also `health_sev`, `status`, `roles_total`, `roles_filled`) — reviewer confirms those field names in the emitted dicts, since without them an agent still cannot name at-risk projects.
- [ ] The impact grouping is emitted from `_compute`'s existing `value_by_impact` (base L320/358), not recomputed by a second traversal at the print site.
- [ ] Verify-by-acting: run `python3 scripts/check-breadcrumbs.py` — it passes, i.e. the new SKILL.md prose introduces no issue/PR ref (`#NNN`), version tag, or milestone/slice code.
- [ ] No plugin version is touched: neither `cogni-projects/.claude-plugin/plugin.json` nor the repo-root `.claude-plugin/marketplace.json` appears in the diff (version-bump gate).
- [ ] Step 4 instructs the agent to narrate from the Step 2 envelope, naming the concrete `data.*` keys it reads, and explicitly says not to re-derive the flags or grouping from entity records; every field name Step 4 cites exists verbatim in the print-site `data` dict — reviewer cross-checks each cited key against the script.
- [ ] Step 4 covers all the narration elements the issue names: project count, consultant count, roles still open, the warn/risk-flagged projects **by name**, where the strategic-impact weight sits, plus relaying `data.warnings` when `partial` is true.
- [ ] AC-2 gate: the new step restates no flag trigger condition and no threshold — grep it for numerals and for conditional flag language; it must contain no `when/if X then flag Y` rule, no numeric threshold (e.g. `100`, `>= 100%`), and no re-definition of when a project is unstaffed/fully staffed/staffing-unknown.
- [ ] The at-risk set Step 4 asks for is keyed on the renderer's own flag/severity value, not on an agent-side comparison such as `roles_filled < roles_total` — otherwise a `closed` project (severity `muted`, roles still declared; test Fixture 9 `Legacy Migration`) gets narrated as at risk.
- [ ] The per-project list in the envelope is the same list `_render_html` receives (base L569/576), not a re-sorted copy, so the narration order matches the HTML table's name-sorted order.
- [ ] `cogni-projects/tests/test-render-dashboard.sh` gains assertions for **both** new keys that pin content, not mere presence: at least one naming a specific fixture project together with its expected `health_label` (e.g. Fixture 1 `Nordic Retail ERP` / `unstaffed`, or Fixture 9 `closed`) and at least one asserting a specific `value_by_impact` count.
- [ ] The `value_by_impact` assertion uses the key form that survives `json.dumps` — int dict keys serialize to strings, so it must index `['5']` not `[5]`; an int-keyed assertion would KeyError, and getting this right is evidence the key type was considered.
- [ ] Verify-by-acting: run `bash cogni-projects/tests/test-render-dashboard.sh` — it exits 0 and prints `All render-dashboard tests passed.`, with every pre-existing fixture 1-11 assertion still passing alongside the new ones.
- [ ] Mutation check (non-vacuity): delete the new key(s) from the print site, re-run `bash cogni-projects/tests/test-render-dashboard.sh`, and confirm it exits non-zero with at least one NEW assertion failing. A still-green suite means the added assertions are vacuous.
- [ ] Because `value_by_impact` counts only projects with a valid 1-5 `strategic_impact` (`_coerce_impact` returns None and warns otherwise, base L169-182), the skill states that its counts need not sum to `data.projects` — matching the existing `avg_allocation` caveat convention at SKILL.md L46-50. Without it an agent narrating Fixture 3/6 data silently under-reports the portfolio.
- [ ] Route fallback: if the diff instead took the HTML-read route and left the success envelope unchanged, items 3-8, 15-17 and 19 are waived and replaced by — Step 4 must tell the agent to `Read` `output/dashboard.html` and name the specific section/table it extracts the flags and the impact grouping from; items 11-14, 18 and 20 still gate unchanged.

**Risk watchlist:**
- Renaming or repurposing the existing `data.projects` int into the new per-project list — a silent breaking change to the envelope contract that six existing assertions depend on.
- Emitting the per-project detail without `health_label`/`health_sev`, so AC-1 is still unmet: the agent can count roles but cannot name which projects are at risk.
- Dropping `ensure_ascii=False` (or re-serializing through a helper that omits it), re-escaping the non-ASCII names Fixtures 6 and 3 deliberately carry.
- Step 4 explaining *how* a flag is computed (a threshold, a `filled == 0 and status == active` rule) — a direct AC-2 violation that also creates a second source of truth that will drift from `_project_health`.
- Narrating at-risk by an agent-side `roles_filled < roles_total` comparison, which mislabels `closed` projects as at risk and re-litigates the closed-project/open-role split Fixture 9 was written to protect.
- Test assertions that only check key presence (`'value_by_impact' in d['data']`) — technically red against base but blind to a wrong shape, wrong counts, or a detail list missing the health flag.
- Indexing `value_by_impact` with int keys in the test, or telling the agent in prose to read integer keys, when JSON serialization has already turned them into strings.
- Adding the new keys to the failure envelope too, or restructuring `_fail`, breaking the clean-failure fixtures 4 and 7.
- Silently widening `allowed-tools` (e.g. adding Write/Edit) while adding the step, eroding the skill's read-only guarantee.
- A Step 4 so long it re-explains the Core-concept material already at SKILL.md L30-56, duplicating the flag semantics in a second place rather than pointing at them.
- Judgment-only, not gradable: whether the Step 4 prose actually reads as a *partner-meeting* narration rather than a JSON field dump — a human should eyeball the tone against `projects-staff` Step 5.

- The plan-time acceptance criteria list 2 criteria, but a base-grounded regeneration derives 21 gating items — including envelope back-compatibility (`data.projects` staying an int count), `_fail`/failure-path invariance, `ensure_ascii=False` retention, the int-vs-string `value_by_impact` key serialization, test non-vacuity under mutation, and the caveat that the impact grouping omits projects with an invalid `strategic_impact`. None of these were asked for, and several are silent-breakage risks; the acceptance bar was materially under-specified. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1169

**Summary:** All four perspectives independently converge on the same two committed-contract gaps in the new Step 4 narration — item 20 (SKILL.md never states value_by_impact undercounts data.projects when strategic_impact is missing/invalid) and item 16 (the new test assertions check count/schema-presence but never pin a named fixture project to its expected health_label) — both mechanical prose/test additions rather than design problems, plus three minor documentation gaps. 18 of 21 contract items are satisfied, verified by execution (suite green; mutation check non-vacuous; breadcrumb, version-bump and skill-name gates clean). No perspective raised a critical or block-level issue and none flagged an INTENT-UNCLEAR contract divergence.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (5)

- **[major]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:118` — Contract item 20 violated (raised by plugin-validator; also flagged by skill-reviewer, skill-quality-reviewer, contract-grading-reviewer): neither Step 4 nor the Core-concept bullet states that value_by_impact counts only projects with a valid 1-5 strategic_impact, so its tiers need not sum to data.projects. _coerce_impact (render-dashboard.py:169-182) returns None and warns for a missing, non-numeric, or out-of-range value, while Step 4 describes the key flatly as 'a map of impact tier 1-5 to project count' and tells the agent to 'point to the tier(s) carrying the most projects'. An agent narrating Fixture 3 or Fixture 6 data will silently under-report the portfolio; projects_detail[].impact arriving as null for those same projects is likewise unexplained. Secondary consequence: value_by_impact is seeded {1..5: 0} (L320), so when no project declares a valid impact the map is all zeros and 'the tier carrying the most projects' is a meaningless five-way tie the agent has no instruction to recognize. The sibling caveat convention already exists at SKILL.md:45-50 for avg_allocation.
  - *Suggested fix:* Add one sentence to the Step 4 impact bullet (or the Core-concept portfolio-value bullet at L43-44): 'Projects whose strategic_impact is missing or outside 1-5 are omitted (they surface as a warning and carry impact: null in projects_detail), so these counts need not sum to data.projects.'

- **[major]** (auto-fixable) `cogni-projects/tests/test-render-dashboard.sh:166` — Contract item 16 violated (raised by contract-grading-reviewer; also flagged by plugin-validator): assertions 1k (len(...) == 2) and 1l (all('health_label' in p and 'health_sev' in p for p in ...)) check only count and universal key-presence for projects_detail — neither names a specific fixture project (e.g. Nordic Retail ERP) together with its expected health_label value, as the contract explicitly requires. Only the value_by_impact assertion (1m) pins content. Impact is limited because projects_detail is the same object handed to _render_html, so wrong names/labels are still caught indirectly via the HTML assertions (a name-to-slug mutation did go red), but the envelope's own contract is not directly pinned.
  - *Suggested fix:* Add an assertion such as: assert_json "1n at-risk project named with its flag" "any(p['name'] == 'Nordic Retail ERP' and p['health_label'] == 'unstaffed' and p['health_sev'] == 'risk' for p in d['data']['projects_detail'])" — replacing or supplementing 1l's schema-only check. Optionally add a Fixture 9 counterpart pinning closed/muted.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:116` — Flagged by plugin-validator, skill-reviewer and skill-quality-reviewer: 'If none are flagged, say the portfolio is fully staffed' reuses a specific renderer health_label value as an aggregate claim covering unflagged states it does not mean. _project_health (L204-222) returns severity ok for two different labels — 'no open roles' and 'fully staffed' — and muted for 'closed'. So a portfolio of only closed projects, or only projects needing nobody, would be narrated as fully staffed — against the Core-concept warning at L33-35 that two flags 'read alike but are not'. Presentational rather than a correctness break.
  - *Suggested fix:* Soften to state-neutral phrasing: 'If no project carries risk or warn, say no project is currently flagged' — and name any closed or no-open-roles projects rather than folding them into a staffing claim.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:119` — Flagged by skill-reviewer and skill-quality-reviewer: Step 4 describes data.value_by_impact as 'a map of impact tier 1-5 to project count'. _compute builds it with int keys (L320), but json.dumps serializes them as strings, so the agent actually reads {"1": 0, "2": 1, ...}. The prose implies integer keys — the prose half of the int-vs-string key hazard the review contract's risk watchlist names as a live failure mode. (The test itself got this right; only the prose is misleading.)
  - *Suggested fix:* Write the tiers as they appear in the envelope: 'a map of impact tier ("1"-"5", string keys in the JSON envelope) to project count'.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:83` — Flagged by plugin-validator: a key-name collision is now live in the envelope and undocumented. data.open_roles is an integer total of unfilled roles on non-closed projects, while each data.projects_detail[].open_roles is a LIST of that project's unfilled role labels (render-dashboard.py:367). The Step 2 addition describes projects_detail as carrying only health_label / health_sev / impact, and Step 4 sends the agent to data.open_roles for the count — an agent that reaches into the per-project record for the same-named key gets a list where it expects a number, or reports the closed-project roles the headline deliberately excludes.
  - *Suggested fix:* In the Step 2 sentence, name the per-project field set explicitly and disambiguate: 'projects_detail (per project: name, status, impact, roles_total, roles_filled, health_label, health_sev, plus an open_roles LIST of that project's unfilled labels — not the portfolio integer data.open_roles)'.

### Positive observations

- Envelope stays exactly backward compatible: all nine pre-existing data keys keep their spelling and semantics, data.projects stays the int count len(projects), ensure_ascii=False is retained at L594, and the failure path (_fail, L519-521) is untouched — so the clean-failure fixtures 4a/4b and 7a-7c are unaffected. (contract items 3, 4, 5, 6)
- No recomputation: value_by_impact and projects_detail are the identical objects _compute already built and _render_html already receives (L569/L576-577) — no second traversal, and narration order matches the HTML table's name-sorted order. (items 8, 15)
- At-risk narration is keyed on the renderer's own health_sev (never an agent-side roles_filled < roles_total comparison), so Fixture 9's closed Legacy Migration cannot be mislabelled at risk — the exact regression the review contract's risk watchlist called out. (item 14)
- AC-2 is respected structurally, not just verbally: Step 4 carries no numeric threshold, no if-X-then-flag-Y rule, and no redefinition of unstaffed/fully-staffed/staffing-unknown; it closes by pointing back at render-dashboard.py as the source of truth. (items 12, 13)
- Every data.* field Step 4 and the Step 2 extension cite exists verbatim in the print-site dict (L582-594) and the per-project record (L360-370). No invented keys. (items 7, 11)
- Verified by execution rather than reading: the full suite exits 0 (72 OK, 'All render-dashboard tests passed.'), and the item-19 mutation deleting both new print-site keys exits 1 with exactly 1k/1l/1m failing and nothing else — the added assertions are non-vacuous. A second mutation swapping the detail name for slug also went red. The file was restored and re-confirmed green. (items 18, 19)
- Skill hygiene intact: frontmatter name/description and the read-only allowed-tools (Read, Bash, Glob, Grep) are unchanged — so README.md correctly stays out of the diff — and exactly one new '### Step 4:' heading was inserted between Step 3 and '## Notes' with nothing renumbered. (items 1, 2)
- Repo gates all green on the head commit: check-breadcrumbs.py clean (0 violations), check-version-bump.py clean in PR mode (plugin.json and marketplace.json both 0.0.13, mirror in sync — no branch version touch), and check-skill-names.sh OK. (items 9, 10)
- Length and tone discipline hold: body is 7210 chars against a 16500 cap (complexity 7), and Step 4's register matches the sibling projects-staff '### Step 5: Summarize' — verb-first heading, envelope-sourced figures, partner-facing framing per bullet rather than a bare field dump.
- The value_by_impact test assertion (1m) correctly indexes string keys ['5'] and ['2'] as they actually appear after json.dumps serializes an int-keyed dict — the int-key trap was considered and avoided. (item 17)
- Step 4 explicitly forbids both re-derivation routes — no re-opening output/dashboard.html and no re-deriving from entity records — keeping the skill's read-only single-source posture intact, and introduces no maintainer breadcrumbs.

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1169

**Summary:** All three perspectives (plugin-validator, skill-reviewer, skill-quality-reviewer) returned pass on the revision; the one substantive minor finding — a pre-existing Core-concept gloss on staffing flags edging toward describing AC-2's trigger condition — was raised independently by plugin-validator and skill-reviewer, deduped into a single entry, and explicitly waived by both with written justification, so the PR is clean to ship.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (waived) `cogni-projects/skills/projects-dashboard/SKILL.md:34-35 (Core concept, pre-existing — not changed by this PR)` — Pre-existing Core-concept gloss 'staffing unknown means it never declared open_roles' edges toward describing a flag trigger condition (AC-2). Definitional narration intent, not a numeric threshold; no numeric health threshold restated anywhere.
  - *Justification:* Pre-existing text untouched by PR #1169; the newly-added Step 4 fully honors AC-2. Grading scoped to the PR's changes.

### Positive observations

- Envelope contract exact — every field Step 2/Step 4 names is present and correctly typed in render-dashboard.py, no doc/code drift.
- All five prior findings verifiably resolved (string-key note, not-summing caveat, all-zeros tie guard, open_roles list-vs-integer disambiguation, closed/no-open-roles vs fully-staffed).
- AC-1 met: Step 4 names at-risk projects from data.projects_detail and locates impact weight from data.value_by_impact, stdout-only, no entity re-derivation.
- AC-2 met: forbids restating any flag trigger/threshold, render-dashboard.py named source of truth.
- Genuinely additive: only 3 declared files touched, version manifests untouched, mirror consistent at 0.0.13, suite passes.
- Body 7927 chars vs 16500 cap; reasoned imperative voice; no maintainer-breadcrumb tokens.

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed all five findings from the previous automated verdict in `69e08825` (fresh verdict: ✅ **PASS**).

| Finding | Severity | Resolution |
|---|---|---|
| SKILL.md Step 4 `value_by_impact` — tiers need not sum to `data.projects` | major | Added a caveat: projects with missing/out-of-range `strategic_impact` are omitted (surface as a warning, carry `impact: null` in `projects_detail`); an all-invalid portfolio reads as an all-zeros five-way tie, not a real distribution. |
| test — no assertion pins a named fixture project's health flag | major | Added assertion `1n` pinning **Compliance Audit** = `health_label: unstaffed` / `health_sev: risk` in `projects_detail`. *(Corrected from the verdict's suggested "Nordic Retail ERP" — `_project_health` maps Nordic, 1/3 roles filled, to `2/3 roles open`/`warn`; Compliance Audit, 0/1 active, is the genuine unstaffed/risk project.)* |
| SKILL.md Step 4 — "fully staffed" reused as an aggregate claim | minor | Softened to state-neutral "If no project carries `risk` or `warn`, say no project is currently flagged", and name any `closed`/`no open roles` projects separately (they read `muted`/`ok`, not `fully staffed`). |
| SKILL.md Step 4 — `value_by_impact` keys described as ints | minor | Corrected to JSON **string** keys `"1"`–`"5"` as they appear after `json.dumps`. |
| SKILL.md Step 2 — `open_roles` name collision undocumented | minor | Named the full per-project `projects_detail` field set and disambiguated its `open_roles` **list** (unfilled labels) from the portfolio integer `data.open_roles`. |

No `render-dashboard.py` change was needed — the script behavior is already correct; only the narrating prose and one test assertion were tightened. `test-render-dashboard.sh` passes (73 OK), including the new `1n`.

sdh07:

_The merger formed this expectation from issue #1151 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The gap is that `render-dashboard.py`'s `_compute()` (base lines 308-393) already derives per-project `health_label`/`health_sev` and the `value_by_impact` 1..5 grouping, but `main()`'s envelope literal (base lines 580-594) drops both — it emits only counts (`projects` as an int, `consultants`, `avg_allocation`, `fully_allocated`, `open_roles`, `warnings`, `partial`), so the flags and the impact weight survive only inside the rendered HTML. The clean fix is the issue's own alternative: grow the envelope additively — pass the `projects` list and `value_by_impact` dict that `_compute()` already returns at base line 569 into the `json.dumps` payload under new keys (never repurposing the existing integer `data.projects`), so the narrating agent reads structured JSON instead of parsing HTML. `cogni-projects/skills/projects-dashboard/SKILL.md` then gains a short narration step after Step 3 (base lines 92-101), modeled on the sibling `projects-staff/SKILL.md` Step 5 (base lines 156-161), that relays project/consultant counts, open roles, the warn/risk-flagged projects by name, where the impact weight sits, and the warnings — sourcing all of it from the renderer's output and never re-deriving it, consistent with the existing SSOT Notes bullet (SKILL.md base lines 103-109) which forbids restating any flag rule or threshold. `cogni-projects/tests/test-render-dashboard.sh` must grow assertions pinning the new fields against known fixtures (Fixture 1's `unstaffed` Nordic Retail ERP at lines 151-162, Fixture 9's `closed` Legacy Migration at lines 433-443) while every pre-existing assertion still passes.

**Acceptance checklist:**
- [ ] `cogni-projects/skills/projects-dashboard/SKILL.md` gains a narration/summarize step: grep `^### Step ` shows a new heading beyond the base three, and the step numbers are contiguous with no duplicates.
- [ ] The skill's YAML frontmatter stays valid (name, description, allowed-tools present and well-formed) and `allowed-tools` still admits every tool the new step tells the agent to use.
- [ ] No new maintainer breadcrumb (issue/PR `#NNN`, version tag, slice/finding code) is introduced in `cogni-projects/skills/projects-dashboard/SKILL.md` or `cogni-projects/scripts/render-dashboard.py`.
- [ ] Plugin structure/manifests stay intact: no version change in `cogni-projects/.claude-plugin/plugin.json` or the repo-root `.claude-plugin/marketplace.json` (the post-merge workflow owns the bump; neither file appears in the changed set).
- [ ] The skill length stays within the body cap — the added step is a few lines, not a second Core-concept section.
- [ ] `cogni-projects/scripts/render-dashboard.py`'s success envelope (base lines 580-594) gains, under new key(s), per-project entries carrying at least the project name and its `health_label`, plus the strategic-impact grouping (counts for impacts 1..5).
- [ ] The new envelope values are the `projects` and `value_by_impact` objects `_compute()` already returns at the base line-569 call site — grep confirms no second health-flag or impact-grouping derivation was added anywhere in the script or the skill.
- [ ] Backward compatibility: every existing `data` key keeps its name and type — in particular `data.projects` remains an **integer count**, which `test-render-dashboard.sh` asserts six times (1b, 2b, 6c, 9b, 10f, 11h). If the diff repurposes `projects` into a list, all six assertions plus SKILL.md's envelope description (base lines 74-76) must be updated in the same diff; a silent type change is a fail.
- [ ] The new key(s) are emitted unconditionally inside the single `json.dumps` payload literal, not behind a conditional — so a zero-project or zero-consultant portfolio still carries them and the skill's new step never reads a missing key.
- [ ] `value_by_impact` has **int** keys in `_compute` (base line 320) and `json.dumps` serializes them as the strings `"1"`..`"5"`. The diff either converts to explicit string keys / a list, or a new test assertion pins the shape actually emitted (e.g. `d['data']['value_by_impact']['5']`). An int-keyed dict shipped with no assertion pinning the serialized key type is a fail.
- [ ] The envelope carries entries for **every** project the HTML table lists, including closed ones — Fixture 9's `Legacy Migration` (test lines 421-443) must appear with its `closed` flag, so the narrating agent sees the same rows the partner sees.
- [ ] Envelope values are the raw entity values, not HTML-escaped: Fixture 5's project named `<script>alert(1)</script>` (test lines 244-259) must reach the JSON unescaped — `_esc()` output leaking into the envelope is a fail.
- [ ] SKILL.md's new step names each thing the issue's problem statement enumerates: project count, consultant count, roles still open, the warn/risk-flagged projects **by name**, where the impact weight sits, and the warnings.
- [ ] AC-2 (SSOT): the newly added SKILL.md prose restates **no** flag trigger condition and **no** threshold — no `filled == 0`, no "unstaffed when…", no percentage or count rule. It may name flag values as strings to relay, and must point at the renderer/envelope as the source of the flags.
- [ ] The existing Notes bullet "**The renderer is the source of truth for the flags and aggregates.**" (SKILL.md base lines 105-109) survives — deleted or weakened is a fail.
- [ ] AC-1 (no re-derivation): the new step instructs the agent to read the flags and impact grouping from the renderer's output (envelope fields, or `output/dashboard.html` if the HTML route was taken) and explicitly not from the entity records.
- [ ] `bash cogni-projects/tests/test-render-dashboard.sh` exits 0 with every assertion passing, including all pre-existing ones (a behavioral test is demanded here because the mechanism lives in an executable file).
- [ ] New test coverage names specifics, not shapes: at least one assertion tying a named fixture project to its exact flag in the new envelope field (e.g. Nordic Retail ERP → `unstaffed`, test lines 102-161), and at least one on the impact grouping counts for Fixture 1 (impact 5 → 1 project, impact 2 → 1 project).
- [ ] Envelope/HTML agreement is asserted the way the base suite pairs 9f and 11k against the tile: a fixture project's new-envelope health label matches the flag rendered into `dashboard.html`, so the two call sites cannot drift apart.
- [ ] No behavior regression in the existing computations: `_project_health`, `_open_role_demand`, and `_compute`'s counting are unchanged (or, if touched, assertions 1c/1d/1e, 9c/9d/9e and 11f still hold — a clean portfolio must not become `partial` because fields were added).
- [ ] The `{success, data, error}` contract is intact on every path: exactly one JSON line to stdout, `ensure_ascii=False` retained on both the success and `_fail` prints, exit codes 0/2 unchanged, and Fixture 7's no-args usage-envelope assertions still pass.
- [ ] Stdlib-only: no new import beyond the Python standard library in `render-dashboard.py`.

**Risk watchlist:**
- The fix repurposes the existing `data.projects` integer into a list of project dicts — satisfying both AC criteria while silently breaking the envelope contract six existing assertions and the skill's own Step 2 prose depend on.
- `value_by_impact` is dumped with its int keys, so an agent (or a later consumer) that indexes `value_by_impact[5]` gets nothing — the int→str coercion happens invisibly at serialization and no assertion pins it.
- The skill's new step re-derives the flags in prose — e.g. telling the agent to call a project at risk when no roles are filled — which restates a trigger condition `_project_health` (base lines 204-222) owns, violating AC-2 while looking helpful.
- A half-measure: the envelope grows but the skill never gains a narration step (or vice versa), so the structural gap the issue names — an agent following the skill end-to-end still cannot name at-risk projects — persists.
- The HTML read-back route is chosen for the skill *and* the script is changed anyway, leaving two sources for the same facts that can disagree; or the skill instructs HTML parsing with brittle string matching against the `<td class='flag'>` markup.
- Closed projects are filtered out of the new envelope field to mirror `_open_role_demand`'s exclusion, so the narrated snapshot silently loses projects the HTML table still shows — the exact 'shrinking total is not a vanished project' confusion the base skill warns about.
- New tests assert only shape (`'value_by_impact' in d['data']`) rather than values, passing against an implementation that emits the right keys with wrong numbers.
- Adding fields inside the warnings path makes a clean portfolio report `partial: true`, inverting the meaning of the flag the skill's Step 2 branches on.
- Judgment-only, not gated: whether the narration wording actually reads well to a partner, and whether the new envelope key name (`project_health` vs `projects_detail` vs nesting under `projects`) is the one later consumers — the backfilling recommender that CLAUDE.md says reads this directory — will want. A human should eyeball the key naming, since an envelope key is a de-facto public contract.
- Judgment-only: `cogni-projects/README.md` line 89 and `CLAUDE.md` describe the renderer without listing envelope keys, so no doc drift is mechanically forced — but if the skill description changed, `cogni-docs:doc-audit` would want a README section-4 refresh.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1169

**Summary:** Three of four perspectives passed and the merger's committed review contract is fully satisfied (22/22 items, zero INTENT-UNCLEAR), but cogni-service:skill-quality-reviewer returned needs_revision on two un-waived minor SKILL.md polish items (duplicated renderer severity-label mapping; missing worked example for Step 4's narration output) — Rule 3 promotes the merged verdict regardless of the other three passes. Four further minor, non-blocking test/skill findings are deduped across perspectives below. No major or critical finding anywhere.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:108-134 (Step 4)` — Step 4 defines the narration by a four-bullet field checklist but gives no output template or worked example, and the only shape guidance is 'a few plain lines'. skill-creator's Writing Patterns section recommends an explicit output template or an input/output example for structured deliverables like this, because a field list alone leaves register and length underspecified — one agent will emit four terse bullets, another a paragraph of prose, for the same envelope.
  - *Suggested fix:* Append a short illustrative worked example after the bullets showing counts, two flagged projects by name, the impact tier, and a warning — marked as illustrative, with numbers and labels always coming from the envelope, never from the example. (ref: skill-creator §Writing Patterns — defining output formats / examples pattern)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:120-121` — The new step states the label-to-severity mapping the renderer owns: 'name any closed or no open roles projects ... since those read as muted/ok'. _project_health (render-dashboard.py:211-222) is the sole owner of which label carries which severity; repeating the pairing here is a second copy that can drift. Not a trigger condition or threshold, so AC-2 is not broken — but the skill does not need to name the severities to make its point.
  - *Suggested fix:* Drop the severity names and keep the instruction behavioral, e.g. '... and name any project whose flag is closed or no open roles rather than folding it into a staffing claim — neither means the project is fully staffed.'

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:118-128` — The 'name any closed / no open roles projects' clause is grammatically attached to the no-flag branch only, though it is equally needed in the mixed case where some projects are flagged and others are closed. The impact-tier bullet likewise stacks three caveats into one sentence chain.
  - *Suggested fix:* Move the closed / no-open-roles clause out of the conditional into its own sentence applying to both branches, and split the impact bullet's all-zero caveat into a trailing sentence.

- **[minor]** (auto-fixable) `cogni-projects/tests/test-render-dashboard.sh:443` — The closed-project case is correct today but ungated: a closed project does reach projects_detail with its closed/muted flag, yet Fixture 9 gained no envelope assertion (coverage is still HTML-only, 9g/9h). Since _open_role_demand deliberately filters closed projects and sits three lines above the new envelope keys, a later refactor extending that exclusion to projects_detail would pass the whole suite while silently dropping rows the HTML table still shows.
  - *Suggested fix:* Add one assertion after 9h, e.g. assert_json "9i closed project reaches the envelope with its flag" "any(p['name'] == 'Legacy Migration' and p['health_label'] == 'closed' and p['health_sev'] == 'muted' for p in d['data']['projects_detail'])"
  - *Raised by:* plugin-dev:plugin-validator, also flagged by plugin-dev:skill-reviewer

- **[minor]** (auto-fixable) `cogni-projects/tests/test-render-dashboard.sh:264-266` — Fixture 5's injection project name <script>alert(1)</script> is asserted escaped in the HTML but nothing asserts it reaches projects_detail raw. The code is correct today (_esc is confined to _render_html), but an _esc leak into the envelope would surface to the narrating agent as escaped entities with no test catching it.
  - *Suggested fix:* Add an assert_json on Fixture 5 pinning the unescaped name in projects_detail.

- **[minor]** (auto-fixable) `cogni-projects/tests/test-render-dashboard.sh:166-168` — The four new assertion labels run 1k, 1l, 1n, 1m — 1n is emitted before 1m, so pass output and any failure report list them out of sequence, breaking the strict lexical ordering every other fixture in this suite holds to.
  - *Suggested fix:* Swap the last two lines so the value_by_impact assertion keeps label 1m and the named-at-risk-project assertion keeps 1n.
  - *Raised by:* plugin-dev:plugin-validator, also flagged by plugin-dev:skill-reviewer

### Positive observations

- All 22 committed acceptance-checklist items satisfied, confirmed by both source reading and direct execution: the full test suite exits 0 with 79 assertions, plus synthetic fixtures run against the renderer confirming closed-project and injection-name handling. Zero INTENT-UNCLEAR items. (cogni-service:contract-grading-reviewer)
- Backward compatibility is airtight: data.projects stays an int (len(projects)) and new data lands under a distinct projects_detail key rather than repurposing it. (plugin-dev:plugin-validator)
- projects_detail and value_by_impact are the literal returns of the single _compute() call, and the same projects list object feeds _render_html — envelope and HTML are structurally incapable of drifting apart. (plugin-dev:plugin-validator)
- The value_by_impact int-key trap is closed by value, not shape: assertion 1m pins ['5']==1 and ['2']==1, confirming the serialized string-key type and actual Fixture 1 counts. (plugin-dev:plugin-validator, plugin-dev:skill-reviewer)
- AC-1 (no re-derivation) and AC-2 (relay flag values, never restate trigger/threshold) are both satisfied explicitly in the new Step 4 prose, and the pre-existing SSOT Notes bullet survives intact and reinforced. (all three file-routed perspectives)
- Step 4 is contiguous with the base three steps, modeled tightly on the sibling projects-staff Step 5, and introduces no new tool beyond the existing allowed-tools. (plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer)
- Repo conventions clean: no version bump in either manifest, check-breadcrumbs.py 0 violations, check-skill-names.sh passes, imports stdlib-only. (plugin-dev:plugin-validator)
- Body measures 7927/16500 chars, comfortably within budget; voice throughout is imperative and reasoned rather than stacked CAPS MUSTs. (cogni-service:skill-quality-reviewer)

### Recommendation

⚠️ **Revise before merge.** 6 finding(s) need attention.

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1169

**Summary:** All three perspectives (plugin-validator, skill-reviewer, skill-quality-reviewer) passed cleanly with both acceptance criteria verified against render-dashboard.py; every finding is an un-waived minor, which Rule 5b treats as advisory rather than merge-blocking.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (5)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:119-120` — The at-risk bullet switches field vocabulary mid-sentence: the first clause keys on health_sev ('is risk or warn'), the 'Either way' clause keys on 'flag' ('is closed or no open roles'). Those two are health_label values, and their severities are muted and ok respectively - an agent that carries health_sev forward from the previous clause will look for a severity 'closed' that the renderer never emits and silently name nothing.
  - *Suggested fix:* Name the field in the second clause: "Either way, name any project whose `health_label` is `closed` or `no open roles` rather than folding it into a staffing claim".
  - *Raised by:* plugin-dev:skill-reviewer, also flagged by plugin-dev:plugin-validator

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:123-125` — 'Projects whose strategic_impact is missing or outside 1-5 are omitted' restates a renderer rule (_coerce_impact's range check) inside the step that otherwise disclaims restating renderer logic. Borderline against AC-2 - a range filter is not a flag trigger or a threshold, and the sentence earns its place by stopping the agent claiming the tier counts sum to data.projects - so not read as a violation. Flagged only so the choice is deliberate.
  - *Suggested fix:* Optional: lean on the envelope instead of the rule, e.g. 'projects the renderer omitted from the value summary carry impact: null in projects_detail and say so in warnings'. The omission condition is also already spelled out in the Notes 'Partial snapshots' bullet.

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:30-42` — Pre-existing, not introduced by this PR: the staffing-coverage bullet packs five independent caveats into one 13-line bullet. That bullet also holds the text closest to an AC-2 tension ('staffing unknown means it never declared open_roles' mirrors _project_health's branch condition), which the PR inherited rather than added.
  - *Suggested fix:* Out of scope for this PR. If touched later, split into sub-bullets and phrase the look-alike flags in terms of what they mean for narration rather than what triggers them.

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:34` — Pre-existing Core-concept text (unchanged by this PR) sits in mild tension with AC-2: 'staffing unknown means it never declared open_roles' is verbatim the trigger in _project_health(), and the surrounding lines restate _open_role_demand()'s exclusion rule. Both are silent-drift surfaces if the renderer's rules change. AC-2 reads as scoped to the narration route this PR adds - which complies - so this is an observation, not a defect introduced here.
  - *Suggested fix:* Optional follow-up, not for this PR: reduce the Core-concept restatements to what the label means for the reader and drop the condition clause, leaving the renderer as sole authority.

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:45-50` — Core concept invests four lines teaching how to caveat the utilization figure, but Step 4 - now the skill's single narration surface - has no utilization bullet and the worked example omits it. The 'Avg allocation' tile is one of the four tiles the partner sees in the HTML, so the guidance now has no step that consumes it.
  - *Suggested fix:* Add a fifth Step 4 bullet for utilization (data.avg_allocation across data.fully_allocated, caveated per Core concept; null means unknown, not 0%), or trim the Core concept guidance if utilization is deliberately out of narration scope.

### Positive observations

- AC-1 met through a clean two-step contract: Step 2 declares the envelope fields and their shapes, Step 4 consumes exactly those fields, with an explicit prohibition on re-opening the HTML or re-deriving from entity records. (plugin-validator, skill-reviewer, skill-quality-reviewer)
- AC-2 met: every instruction reads a reported value; no numeric threshold and no label-to-severity mapping remain, with an explicit pointer to render-dashboard.py as source of truth. (skill-reviewer, skill-quality-reviewer)
- Every envelope claim was verified against render-dashboard.py - the projects_detail 9-field list, the risk/warn severity vocabulary, impact: null for out-of-range values, and the all-zeros value_by_impact case. (skill-reviewer)
- The envelope change is minimally invasive - two additive keys at the existing print site from values already in scope, no signature change, no restructuring of _compute() or _render_html(). (plugin-validator)
- The data.open_roles integer vs per-project open_roles list collision is called out explicitly at the point of use, pre-empting the most likely misread. (plugin-validator, skill-reviewer)
- Test 5c pins that _esc() stays confined to _render_html() and the narrating agent receives the raw unescaped project name from the envelope - exactly the bug the envelope route could have introduced. (plugin-validator)
- Test 9i guards a widening-exclusion regression, and Test 1n directly pins AC-1 by asserting Compliance Audit arrives with health_label 'unstaffed' and health_sev 'risk'. (plugin-validator)
- Both zero-state branches are handled - no project flagged, and no valid impact declared - which is where a narration step usually degrades into a confident wrong claim. (skill-reviewer)
- The worked example is correctly fenced as illustrative-only, guarding against an agent lifting fixture-shaped numbers. (plugin-validator, skill-reviewer, skill-quality-reviewer)
- Voice is reasoned rather than commanded - every lowercase 'never' is paired with its why, avoiding the CAPS-MUST anti-pattern. (skill-quality-reviewer)
- Progressive disclosure is sound at 8,244 chars against a 16,500 cap, with the entity field contract offloaded to references/data-model.md. (skill-reviewer, skill-quality-reviewer)
- No documentation drift and no maintainer breadcrumbs; whole test suite green and sibling gates (breadcrumb guard, check-skill-names.sh, version-bump gate) pass across the repo. (plugin-validator, skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

**Commit:** ad95addd6827032483fe8b7c73977e7f92c3382c
**Addressed:** Answers the NEEDS_REVISION verdict, all six minor auto-fixable findings. SKILL.md Step 4 gains the illustrative worked example the skill-quality-reviewer asked for, drops the renderer-owned severity pairing flagged by skill-reviewer, and un-nests the closed / no-open-roles clause plus splits the impact bullet's stacked caveat. test-render-dashboard.sh gains assertion 9i pinning the closed project in `projects_detail`, assertion 5c pinning the injection name unescaped in the envelope, and a swap of the Fixture 1 1m/1n lines restoring strict lexical label ordering. `render-dashboard.py` is unchanged — every finding was about the skill's prose or the tests' coverage of an envelope the script already emits correctly.
**New verdict:** PASS — see https://github.com/cogni-work/insight-wave/pull/1169#issuecomment-5160245771

### Findings resolved (6)

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:108-134 (Step 4)` — Step 4 defined the narration by a four-bullet field checklist with no output template or worked example, leaving register and length underspecified.
  - *Resolution:* Appended a short illustrative worked example at the end of Step 4 showing counts, two flagged projects named with their quoted `health_label`, a closed project called out, where the impact weight sits, and one warning — prefixed with an explicit illustrative-only / take-yours-from-the-envelope disclaimer. The example quotes labels as reported and restates no threshold, so the renderer stays SSOT. Its project names are deliberately invented and do not collide with any test-fixture name.
  - *Commit:* ad95addd

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:120-121` — The at-risk bullet restated the label-to-severity mapping owned by `_project_health` ("since those read as muted/ok"), a second copy that can drift.
  - *Resolution:* Rewrote the clause behaviorally as suggested: "... name any project whose flag is `closed` or `no open roles` rather than folding it into a staffing claim — neither means the project is fully staffed." No severity names remain in that bullet.
  - *Commit:* ad95addd

- **[minor]** `cogni-projects/skills/projects-dashboard/SKILL.md:118-128` — The closed / no-open-roles clause was attached to the no-flag branch only though it is equally needed in the mixed case, and the impact bullet stacked three caveats into one sentence chain.
  - *Resolution:* Moved the clause into its own sentence opening "Either way, …" so it applies to both branches, and split the impact bullet's semicolon chain so the all-zero five-way-tie caveat is a separate trailing sentence.
  - *Commit:* ad95addd

- **[minor]** `cogni-projects/tests/test-render-dashboard.sh:443` — Fixture 9's closed-project coverage was HTML-only (9g/9h); no envelope assertion guarded the closed row reaching `projects_detail`, so widening `_open_role_demand`'s exclusion to `projects_detail` would have passed the suite.
  - *Resolution:* Added assertion 9i after 9h pinning that `Legacy Migration` reaches `data.projects_detail` with `health_label` `closed` and `health_sev` `muted`, exactly as suggested, with a comment naming the regression it guards.
  - *Commit:* ad95addd

- **[minor]** `cogni-projects/tests/test-render-dashboard.sh:264-266` — Fixture 5 asserted the injection project name escaped in the HTML but nothing asserted it reaches `projects_detail` raw, so an `_esc` leak into the envelope would have gone uncaught.
  - *Resolution:* Added assertion 5c after 5b pinning the unescaped name in `data.projects_detail`, with a comment noting `_esc` is confined to `_render_html`.
  - *Commit:* ad95addd

- **[minor]** `cogni-projects/tests/test-render-dashboard.sh:166-168` — The four new Fixture 1 assertion labels were emitted 1k, 1l, 1n, 1m, breaking the strict lexical ordering every other fixture holds to.
  - *Resolution:* Swapped the last two lines so emission order is now 1k, 1l, 1m, 1n; each label stays attached to its original expression.
  - *Commit:* ad95addd

Auto-revise triggered by the review verdict's findings — no human feedback since the last automated review. See the linked verdict comment above for the original findings.

## Review notes

Two things worth a maintainer's eye, neither blocking and neither part of the six findings:

- **Prose-simplify pass touched four spots outside Step 4.** The pipeline's prose pass deduped four statements elsewhere in the skill: the intro's trailing read-only clause (the `**Read-only.**` Notes bullet states it more precisely), Step 2's warnings-relay rationale (now a pointer to Step 4, which carries the full rule), Step 2's second `value_by_impact` gloss (Step 4's definition is the fuller one), and the Notes restatement of the source-of-truth framing (the operative "change the script, not this skill" instruction survives). Flagging it so the out-of-delta hunks do not read as unexplained scope creep.
- **Declined a suggested generalization, deliberately.** A quality pass proposed replacing the three per-fixture envelope assertions with one suite-wide `assert_detail_rows` invariant invoked after every `run` — it verified the invariant holds at all 12 probe points, and noted `projects_detail` is currently asserted only in fixtures 1, 5 and 9 (fixtures 3, 6, 8, 10, 11 make no claims about it). That is a real coverage gap and a reasonable follow-up, but it rewrites every fixture and sits well outside this revise's scope, so it was left alone rather than folded in here.

sdh07:

_The merger formed this expectation from issue #1151 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The gap is real and structural: `_compute` (cogni-projects/scripts/render-dashboard.py:308-393) already builds both the per-project payload (`name`, `status`, `impact`, `roles_total`, `roles_filled`, `open_roles`, `health_label`, `health_sev`) and `value_by_impact` (1..5 counts), but the emitted envelope at lines 580-594 carries only nine scalar/aggregate keys, so the flags and the impact grouping exist solely inside `_render_html` (lines 400-516). The correct fix is the issue's own 'alternative worth weighing first': add the per-project health payload and `value_by_impact` to the envelope as NEW keys sourced from `_compute`'s existing return values (no second derivation, so the HTML table and the envelope cannot drift the way fixtures 9f/11k were written to catch), leave the nine existing keys — in particular the integer `data.projects` count — untouched, then add a short narration step to cogni-projects/skills/projects-dashboard/SKILL.md (after Step 3, lines 92-101) that tells the agent to read those envelope fields and give the partner counts, open roles, the projects carrying a warn/risk flag by name, where the impact weight sits, and any warnings. Step 2's envelope description (lines 74-76) must grow with it, or the agent has no reason to know the fields exist. Per AC-2 the new prose must name flags without restating any trigger condition or threshold, keeping the Notes bullet at lines 105-109 ('The renderer is the source of truth for the flags and aggregates.') true. cogni-projects/tests/test-render-dashboard.sh must gain assertions on the new envelope surface, since this mechanism lives in an executable script; note the issue's cited 'lines 489-501' for the envelope is stale — it is at 580-594 on base.

**Acceptance checklist:**
- [ ] cogni-projects/skills/projects-dashboard/SKILL.md gains a narration step under `## Workflow` (a new `### Step 4:` heading, or an equivalently-headed section) whose body directs the agent to report the projects carrying a warn/risk flag BY NAME and where the strategic-impact weight sits — grep the file for the new step heading plus wording covering both halves.
- [ ] Every envelope field the new SKILL.md step cites exists verbatim as a key in render-dashboard.py's final `print(json.dumps({...}))` block — no invented field name. If the HTML-read route was chosen instead, the step names `output/dashboard.html` explicitly and the skill's `allowed-tools` frontmatter still lists `Read`.
- [ ] All nine pre-existing envelope keys are still emitted: `path`, `portfolio`, `projects`, `consultants`, `avg_allocation`, `fully_allocated`, `open_roles`, `warnings`, `partial` — grep the envelope dict literal in render-dashboard.py.
- [ ] `data["projects"]` remains the INTEGER project count (`len(projects)`) and is not repurposed into a per-project list; the six existing count assertions in cogni-projects/tests/test-render-dashboard.sh (1b:154, 2b:170, 6c:333, 9b:436, 10f:526, 11h:690) are present and unweakened. Any per-project detail lands under a new key.
- [ ] No leaked maintainer breadcrumbs: `python3 scripts/check-breadcrumbs.py` exits 0 — the edited SKILL.md introduces no issue/PR ref (e.g. `#1151`), version tag, or milestone/slice/finding code.
- [ ] The new envelope data is read from `_compute`'s existing return values (`projects_out` / `value_by_impact`, render-dashboard.py:308-393) — the diff adds NO second computation of a health flag, severity, or impact grouping. Check: `_project_health` still has exactly one call site, and none of the literal flag strings (`unstaffed`, `fully staffed`, `staffing unknown`, `no open roles`, `closed`) appear anywhere outside `_project_health` in the script.
- [ ] AC-1 (at-risk half): the narrating agent can name at-risk projects without touching entity records — the new per-project payload carries at minimum the project name plus `health_label` (and/or `health_sev`), verified by grepping the envelope construction for those keys; under the HTML route, the step must name the exact table column that carries the flag.
- [ ] AC-1 (impact half): the strategic-impact grouping reaches the agent as per-impact project counts — `value_by_impact` is present in the envelope, or the step points at the HTML section by its exact heading `Portfolio value — projects by strategic impact`.
- [ ] AC-2: the SKILL.md diff restates no flag trigger condition or threshold — the added lines contain no count/status-to-flag rule (no `filled`, `== 0`, `>= 100`, `100%`, 'when the project is active', 'if it declares no roles' style wording), and the file still contains the string `The renderer is the source of truth for the flags and aggregates.`
- [ ] SKILL.md Step 2's envelope description (base lines 74-76, currently naming only `data.path`) is updated to mention the new fields, so the agent following Step 2 knows they exist — grep Step 2 for the new key names.
- [ ] cogni-projects/tests/test-render-dashboard.sh gains behavioral assertions on the new envelope surface (demanded here because the mechanism lives in an executable script): at least one `assert_json` pinning a named project's health flag through the envelope (e.g. Fixture 1's `unstaffed` / `Compliance Audit`) and at least one pinning the impact grouping counts (Fixture 1: one project at impact 5, one at impact 2).
- [ ] `bash cogni-projects/tests/test-render-dashboard.sh` exits 0 and prints `All render-dashboard tests passed.` — every pre-existing assertion (1a through 11l) still passes, none deleted or loosened.
- [ ] Read-only invariant intact: render-dashboard.py still writes only `<portfolio-dir>/output/dashboard.html` (one `open(..., "w")` site) and never `projects-portfolio.json` or `.metadata/`; assertion 2c ('manifest untouched (read-only)', test lines 172-176) is unchanged and passing.
- [ ] Edge case — JSON key typing: if `value_by_impact` is emitted, `json.dumps` stringifies its int keys, so the new test assertion indexes it with STRING keys (e.g. `d['data']['value_by_impact']['5']`) or the code converts keys via `str(...)` before emission; neither the test nor the SKILL.md text indexes it with an int key.
- [ ] Edge case — closed/unknown projects: a project with `impact` None keeps a null (not a defaulted number) in the new payload, and the added step does not tell the agent to narrate a `closed` project as at-risk or fold closed roles into 'roles still open' (consistent with `_open_role_demand`, render-dashboard.py:289-305). Because this mechanism is prose-carried in SKILL.md, the achievable bar is a grep of the added step's wording plus a documented mutation check: deleting the new step must leave no other place in the skill telling the agent where the flags come from — i.e. the step is load-bearing, not duplicative filler.

**Risk watchlist:**
- Repurposing `data.projects` from an integer count into a per-project list, breaking the six existing assertions — and then rewriting those assertions to match, hiding the envelope-contract break rather than fixing it.
- Re-deriving health flags or the impact grouping in a second place (a new envelope-builder loop, or restated rules in the skill prose), so the rendered HTML table and the JSON envelope can drift apart — precisely the class fixtures 9f and 11k exist to catch.
- Restating a threshold or trigger in SKILL.md ('unstaffed means zero filled roles on an active project'), duplicating `_project_health` (render-dashboard.py:204-222) and violating AC-2 the moment the script changes.
- Emitting `value_by_impact` with int keys and then documenting or asserting int-key access — passes in-process, fails after the JSON round-trip the skill actually performs.
- A skill step that tells the agent to grep or parse `output/dashboard.html` for flags: brittle against any markup change, and no test pins the markup that step would depend on.
- Half-measure: adding the narration step but leaving the envelope unchanged and never saying where the flags come from — the agent silently falls back to re-deriving from entity records, the exact failure AC-1 names.
- Dumping the whole `projects_out` payload (including `health_sev`, a rendering-severity token) into the envelope and then re-explaining colour semantics in the skill, growing both surfaces without improving narration.
- Narrating a `closed` project as at-risk, or counting closed projects' roles as live demand — contradicting `_open_role_demand` and the `projects-staff` set it is deliberately aligned with.
- Skill text that names the new per-project `open_roles` list without disambiguating it from the top-level integer `data.open_roles`, producing a narration that reports the wrong number.
- Judgment-only (no gate can decide it): whether the new step actually yields 'a few lines' a partner can hear, rather than a field-by-field data dump — a human should eyeball the added step's length and tone.

- The plan-time acceptance criteria list 2 criteria (agent can name at-risk projects + impact grouping; no flag trigger/threshold restated), but a base-grounded regeneration derives 15 gating items — including envelope backward-compatibility (`data.projects` must stay an int count against six existing assertions), the single-source-of-truth constraint that the new fields be read from `_compute` rather than recomputed, JSON int-key stringification of `value_by_impact`, the read-only write-path invariant, and the requirement that the test suite gain assertions on the new surface. The acceptance bar was materially under-specified relative to what the change actually touches. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1169

**Summary:** Three of four perspectives passed, but cogni-service:skill-quality-reviewer's needs_revision verdict governs the merge (Rule 3): two minor, un-waived findings remain — a Step 4 worked example that contradicts its own at-risk rule by omitting the staffing-unknown project, and a pre-existing Core-concept line restating _project_health's staffing-unknown trigger condition. That Core-concept line was independently flagged by all three of plugin-validator, skill-reviewer, and skill-quality-reviewer, but only the latter declined to waive it as out-of-scope, arguing AC-2's no-restatement rule applies skill-wide rather than only to the diff. No other findings are major/critical and no INTENT-UNCLEAR signal was raised against the review contract.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:138-143 (Step 4 illustrative example)` — The worked example contradicts the rule directly above it. Step 4 bullet 2 says to name each project whose `health_sev` is `risk` or `warn`, but `staffing unknown` carries `warn` (render-dashboard.py:213-214) — yet the example says "Two projects are flagged" and surfaces Tallow Freight only in a trailing "One warning" sentence. An agent follows the worked example over the prose rule, so the exact case the renderer deliberately gave a warn state — a project that looks healthy but whose coverage cannot be read — is the one most likely to go unnamed, which is what AC-1 asks the skill to prevent. Verified present at the cited location on the PR head. Compounding it: `data.warnings` is populated only by parse failures, `_coerce_impact`, `_text_field`, and unmatched assignment names (render-dashboard.py:155-180, 278) — a missing `open_roles` produces no `data.warnings` entry at all, so the example's "One warning" line points the agent at the wrong envelope field.
  - *Suggested fix:* Fold the staffing-unknown project into the flagged list in the example, e.g. "Three projects are flagged: **Harbour Replatform** is `unstaffed`, **Meridian Rollout** is `2/3 roles open`, **Tallow Freight** is `staffing unknown`." and drop or re-point the trailing warnings sentence so it illustrates a real `data.warnings` entry. (ref: skill-creator §Writing Patterns (examples pattern))

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:33-35` — Core concept prose restates `_project_health`'s trigger condition verbatim: "`staffing unknown` means it never declared `open_roles`" mirrors `if open_roles is None:` (render-dashboard.py:213-214). All three of plugin-validator, skill-reviewer, and skill-quality-reviewer independently flagged this duplication, but disagree on disposition: plugin-validator and skill-reviewer waived it as pre-existing text outside this PR's diff, reasoning that contract item 9 scopes AC-2's no-restatement rule to the added lines only. skill-quality-reviewer did not waive it, arguing AC-2 as stated in the issue scopes the no-restatement rule to the whole skill (not just the diff), that Step 4 and the Notes bullet both commit to that invariant, and that this line is therefore the one place the file contradicts itself and can silently go stale if the script's condition moves.
  - *Suggested fix:* Drop the mechanical clause and keep the semantic contrast the bullet is making: "`staffing unknown` means its coverage cannot be read at all — never narrate it as covered." (ref: skill-creator §Writing Style (explain the why, single source of truth) / §Principle of Lack of Surprise). Note: the two waiving perspectives would treat this as an optional follow-up outside this PR's scope rather than a blocking fix.

- **[minor]** (waived) `cogni-projects/skills/projects-dashboard/SKILL.md:122-127` — The new Step 4 impact bullet restates `_coerce_impact`'s validity rule in prose ("Projects whose `strategic_impact` is missing or outside 1-5 are omitted"). Not a flag trigger or threshold, so it does not breach checklist item 9 or AC-2 as written, but it duplicates a rule that lives in the script — the same drift class AC-2 targets for flags — recording the renderer's coercion semantics a second time.
  - *Justification:* Arguably necessary narration guidance: without it the agent may wrongly assert the impact counts sum to the project total. (skill-reviewer separately noted the explicit 1-5 range is already stated in the unchanged Core concept section.)
  - *Raised by:* plugin-dev:plugin-validator, also flagged by plugin-dev:skill-reviewer

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:118-120` — The at-risk bullet closes with an unconditional instruction to name any project whose flag is `closed` or `no open roles`. The guard is right, but as written it scales linearly with the portfolio: a portfolio with a dozen delivered projects turns the narration into a roll-call — the 'field-by-field data dump' outcome the contract's judgment-only risk item warns about.
  - *Justification:* Judgment call on narration tone, explicitly flagged in the contract as human-eyeball territory; the current wording violates no checklist item and errs toward correctness over brevity.

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-dashboard/SKILL.md:110-112` — Step 4 opens with "do not re-open `output/dashboard.html`" immediately after Step 3 instructs the agent to run `open "<data.path>"`. The intent is clearly 'do not read/parse the HTML for data', but the verb collides with the preceding step's literal `open` command and could read as retracting Step 3.
  - *Justification:* Surrounding context already disambiguates for a careful reader.

- **[minor]** (waived) `cogni-projects/scripts/render-dashboard.py:590` — `"projects_detail": projects` emits the whole `_compute` per-project payload, including `client`, `status`, `roles_total`, `roles_filled`, which Step 4 never narrates. Envelope stdout also grows linearly with portfolio size.
  - *Justification:* Passing the identical object the HTML renderer consumes is what makes drift structurally impossible (checklist item 6); a projection would add a mapping step for no correctness gain.

### Positive observations

- Contract-grading: all 15 committed acceptance-checklist items graded satisfied, with the full test suite (71 assertions) and the breadcrumb gate both executed clean against the checked-out PR head. No checklist item was unadjudicable.
- Checklist item 6 (single source of truth) satisfied structurally, not just textually: the envelope passes the exact objects `_compute` returned and `_render_html` consumes, so HTML and JSON cannot diverge. `_project_health` still has exactly one call site and every flag literal is confined to it.
- Checklist item 4 / the top risk-watchlist item fully avoided: the test diff is +14/-0; all six `data.projects` integer assertions are present and unweakened, and per-project detail landed under a new key rather than repurposing the count.
- Checklist item 14 (JSON int-key stringification) handled on all three surfaces at once — code, test 1m, and SKILL.md prose. This is the failure mode that passes in-process and breaks in the field.
- Checklist item 15 (closed/unknown edge case) handled precisely: `impact` stays null rather than defaulting, closed projects are steered out of the at-risk set, and new assertion 9i pins the closed-project row through the envelope.
- Assertion 5c goes beyond the checklist: it pins that `_esc` stays confined to `_render_html`, so the narrating agent receives raw project names, not HTML entities — a failure mode nothing else in the suite would have caught.
- The `open_roles` name collision is disambiguated at the point of introduction, so the narration cannot report the wrong number — a named watchlist risk, closed.
- The new Step 4 is genuinely load-bearing, not duplicative filler: deleting it would leave Step 2's `projects_detail` mention dangling and no place in the skill telling the agent where the flags come from.
- Step 4 pins the read path to the stdout envelope and explicitly forbids re-deriving from entity records — exactly the AC-1 mechanism, and it forecloses the brittle-HTML-parsing route the contract's risk watchlist flagged.
- Body measures 8244 chars against a 16500 cap (metric_version=2, over_cap=false), with the field contract offloaded to a references/ file verified to exist. Voice is imperative and reason-first, with no CAPS-stacked MUSTs.
- Both deterministic pre-gates ran clean: check-frontmatter.sh reported 0 offenders across 4 files, and check-version-bump.sh reported status ok with 0 violations (1 plugin touched, no stray version bump).

### Recommendation

---

sdh07:
## Human merge decision

Merging at `c574b714`. This PR carried `svc:needs-human-review`, so this is that human call rather than a fourth automated round.

### The two unwaived findings are resolved

1. **Worked example contradicted its own rule** — valid, and fixed. Step 4 says name every project whose `health_sev` is `risk` or `warn`; `staffing unknown` maps to `warn` (`render-dashboard.py:213-214`), so the example's "Two projects are flagged" left out the exact case the renderer gave its own warn state. It now flags three.

   The finding's *compounding* claim was not acted on, because it is factually wrong: it asserts a missing `open_roles` produces no `data.warnings` entry, so the example's warnings line points at the wrong field. `render-dashboard.py:332-333` appends `"%s has no open_roles — staffing status unknown"` for any non-closed project with undeclared roles. The warning is real, so the example now quotes it verbatim — which also demonstrates the relay Step 4 asks for, rather than paraphrasing a rule.

2. **Core-concept restatement** — pre-existing on `main` and outside this delta, waived as out-of-scope by plugin-validator and skill-reviewer, not waived by skill-quality-reviewer. Fixed here anyway: two lines, no behaviour change, and it is the one finding all three perspectives raised independently. Deferring it would have kept it re-firing on every future pass over this file.

### On the `insufficient_coverage` hold

The recorded escalation grades issue #1151's acceptance bar — 2 hand-written criteria against 15 items a base-grounded regeneration derives — not this diff. The contract itself records it as "not holding". Nothing in it identifies a defect in the change, and all 15 derived checklist items were graded satisfied by direct execution across the review cycles.

### Merge basis

Substance was never in question: `render-dashboard.py` is +2/-0, sourcing both new envelope keys from `_compute`'s existing return values so the HTML and the JSON cannot drift; the nine pre-existing envelope keys and the integer `data.projects` are untouched; the suite gained behavioural coverage (5c pinning that `_esc` stays confined to `_render_html`, 9i pinning the closed row through the envelope). Three cycles produced only minor prose findings, each round surfacing a fresh set rather than converging — that is a bar-calibration artifact, not a signal about this change.

Gates green on `c574b714`: 79 assertions pass, `check-breadcrumbs.py` 0 violations, `check-skill-names.sh` clean, no manifest touched.

Issue #1173, opened 2026-08-02
cogni-projects: extract a shared read_frontmatter helper — staffing-score.py is missing the UnicodeDecodeError guard its siblings have


claude-cogni-work:
## Context

Found during a review pass over the assignment referential-integrity change in
`cogni-projects`. Not caused by that change — it is pre-existing divergence the
new code surfaced.

## Problem

The read-frontmatter sequence (open file → read → `parse_frontmatter` → guard)
now exists in three places, and the copies have already diverged on their
exception handling:

- `cogni-projects/scripts/validate-entities.py` — guards `OSError` **and**
  `UnicodeDecodeError`
- `cogni-projects/scripts/render-dashboard.py` (`_read_entities`) — guards both
- `cogni-projects/scripts/staffing-score.py` (`_read_frontmatter`) — guards
  **only `OSError`**

`validate-entities.py` and `render-dashboard.py` both carry an explicit comment
that an entity file saved as latin-1 (a DACH name is the usual way in) is the
case the `UnicodeDecodeError` guard exists for. `staffing-score.py` lacks it, so
one such record raises straight out of the scorer instead of being reported
through the `{success, data, error}` envelope every other path returns.

## Why it matters

The envelope is the contract — callers gate on parsing it, and a traceback on
stderr reads to them as neither success nor a reportable failure. This is the
same class of defect the two sibling scripts already defend against.

## Proposed change

Extract a public `read_frontmatter(path)` next to `parse_frontmatter` in
`validate-entities.py` and have the other two call it. This adds no new
dependency direction: both scripts already load `validate-entities.py` by file
location, so the extraction rides an arrow that exists today. Roughly 5 lines
added, ~20 removed across three files.

## Acceptance criteria

- [ ] A single `read_frontmatter(path)` helper is the only place the
      open/read/parse/guard sequence lives.
- [ ] `staffing-score.py` no longer raises on a non-UTF-8 entity file — it
      reports through the standard envelope like its siblings.
- [ ] `render-dashboard.py` keeps its current warning behaviour for unreadable
      records.
- [ ] All four `cogni-projects/tests/*.sh` suites stay green.

claude-cogni-work:
## Triage Report

- **Severity:** S3-minor
- **Component:** cogni-projects / scripts (`validate-entities.py`, `render-dashboard.py`, `staffing-score.py`)
[...]

Issue #1172, opened 2026-08-02
cogni-projects: run the assignment ref check on the registration write path, not only the validator CLI


claude-cogni-work:
## Context

Deferred from the resolution of the assignment referential-integrity work. That
change added a portfolio-scoped ref check to `cogni-projects/scripts/validate-entities.py`,
but only on the **CLI directory** invocation. A review pass over the diff found
that nothing in the plugin actually invokes the validator with a directory
argument, so the new check only runs when a human types the command out of
`projects-entities/SKILL.md` prose.

## Problem

`register-entity.py` is the layer that owns portfolio-scoped integrity on the
**write** path. It already receives `portfolio_dir` as its first argument, and
already enforces a portfolio-scope invariant elsewhere with the comment
"Validation cannot catch this". But it gates the manifest on
`_ve.validate_file(entity_file)` — a single-file call, which by design runs no
ref check.

Net effect: an assignment carrying a misspelled `consultant` slug is authored,
passes the skill's Step 4 validation, and is registered into
`projects-portfolio.json`. `render-dashboard.py` then filters assignments on a
slug comparison that simply never matches, so the role silently reads as
unfilled with no error — the "surfaces far from the entity that caused it"
outcome the ref check was introduced to prevent.

## Why it was not done in the parent change

The parent issue's acceptance criteria constrain `validate-entities.py`'s
single-file CLI shape ("Step 4 passes a single file and must not start
failing") and say nothing about `register-entity.py`. Adding a check to the
registration path changes registration behaviour and is a scope widening the
plan deliberately did not take.

## Proposed change

At the `validate_file` gate in `register-entity.py`, additionally call
`_ref_errors(_entity_files(portfolio_dir))` and keep the entries whose `file`
is the entity being registered. The module is already loaded by file location,
so this needs no signature change (~4 lines).

## Acceptance criteria

- [ ] Registering an assignment whose `consultant` or `project` does not resolve
      within the portfolio fails with the standard `{entity, file, field, message}`
      error instead of writing the manifest ref.
- [ ] Registering a consultant or project (no refs) is unaffected.
- [ ] Registering an assignment with resolvable refs is unaffected.
- [ ] The existing `test_register_entity.sh` suite stays green.

claude-cogni-work:
## Triage Report

- **Severity:** S3-minor
- **Component:** cogni-projects / `scripts/register-entity.py`
- **Root cause hypothesis:** The registration gate at `register-entity.py:113` calls
[...]

PR #1171, opened 2026-08-02
cogni-projects: referential-integrity check for assignment refs

claude-cogni-work:
Closes #1122.

## Summary
- `validate-entities.py`: a **portfolio-directory** invocation now collects consultant and project slugs and ERRORs on any assignment whose `consultant` or `project` slug does not resolve, in the standard `{entity, file, field, message}` shape (AC1).
- **Single-file** invocation is unchanged (AC2) — `register-entity.py` calls `validate_file()` on a single file, so it is unaffected; the ref pass is gated on a directory argument (`any_dir`).
- Files that already fail shape validation are excluded from the slug sets and the ref check, so a malformed record is reported once, never double-flagged as "dangling".
- Doc anchors updated in lockstep: the module docstring, `projects-entities/SKILL.md` Step 4 scope statement (AC3), and `references/data-model.md` Validation section. The single-file authoring note (SKILL.md Step 2) stays true and is left as-is.
- New `tests/test_validate_entities.sh` covers the positive (directory flags a dangling ref), clean (resolvable refs → success), single-file-exemption, and no-double-report cases.

## Scope decisions
- In: AC1, AC2, AC3, and a dedicated test. All three doc claims moved together to prevent drift.
- Out: nothing deferred. AC4 (ordering before #1114) is moot — #1114 has shipped.
- README component-table label is auto-generated by cogni-docs, so it is not hand-edited here.
- No `plugin.json` / `marketplace.json` version bump — the post-merge workflow owns that.

## Test plan
- [x] `bash cogni-projects/tests/test_validate_entities.sh` passes (new — 14 checks).
- [x] `bash cogni-projects/tests/test_register_entity.sh` still passes (single-file contract unchanged).
- [ ] Manual: run the validator against a real portfolio dir with resolvable refs — `success:true`, no new errors.

---
🤖 Resolved by cogni-service (author instance). Draft, `svc:needs-review` — handed off to the merger instance for independent review.

Issue #1114, opened 2026-07-17
cogni-projects: staffing match engine (availability + profile + strategic impact) — MVP


claude-cogni-work:
## Plugin
cogni-projects

## Epic context
Phase 2 (staffing MVP) of "Roadmap: cogni-projects — partner project-portfolio steering (staffing MVP first)". This is the load-bearing first slice — it consumes the Phase 1 entities to produce staffing recommendations, and the deferred children build on it.

## Use case
A partner staffing a project needs a ranked shortlist of consultants for each open role, balancing who is actually available, who fits the role profile, and how strategically important the project is — the core value of cogni-projects.

## Proposed solution
Add a domain-prefixed `projects-staff` skill and a `scripts/staffing-score.py` (python3 stdlib, `{success,data,error}` JSON). The skill reads the portfolio's consultants, availability, and projects and, for each open project role, ranks candidate consultants on three sub-scores: **availability** (allocation windows overlap the project timeline), **profile fit** (skills/seniority match the role need), and **strategic impact** (the project's `strategic_impact` score weights the match). It writes a staffing-recommendation markdown artifact plus the JSON. AC: (1) output is a ranked candidate list per open role with the three per-factor sub-scores shown; (2) consultants with no availability overlapping the project window are excluded; (3) `staffing-score.py` is deterministic for identical inputs and emits the `{success,data,error}` contract. ~1 skill + 1 scoring script.

## Alternatives considered
A single opaque overall match score — rejected: partners need the per-factor breakdown to defend a staffing call. An LLM-only match with no script — rejected: the deterministic scoring script keeps ranking reproducible and testable.

_Affected components: cogni-projects/skills/projects-staff/SKILL.md, cogni-projects/scripts/staffing-score.py_

_Downstream hint: plugin-dev:skill-development_

_Complexity hint: medium_

## Depends on
- #1112 cogni-projects: scaffold plugin skeleton + marketplace registration + projects-setup
- #1113 cogni-projects: define self-contained data model + entity-authoring skill

## Blocks
- #1115 cogni-projects: backfilling — recommend next assignment when a consultant rolls off

claude-cogni-work:
## Triage Report

- **Severity:** S2-major
- **Component:** cogni-projects / `projects-staff` skill + `scripts/staffing-score.py`
[...]

PR #1174, opened 2026-08-02
cogni-projects: referential-integrity check for assignment refs

### LINKED ISSUE #1122: cogni-projects: referential-integrity check for assignment consultant/project refs
## Context

Deferred from PR #1121 during service-revise. The review verdict raised this as a
minor finding and explicitly noted it is "reasonable to defer to a follow-up for
Phase 1"; the PR took the verdict's suggested cheap fix (documenting the gap
accurately) and defers real enforcement here so the scope is not silently dropped.

**Deferred item:** Referential-integrity enforcement in `validate-entities.py`:
when the argument is a portfolio directory, cross-check every assignment's
`consultant` / `project` slug against the collected `consultants/` and `projects/`
entities and error on a dangling ref.

## Current state (what PR #1121 shipped instead)

`validate-entities.py` checks frontmatter **shape only** — required keys, enums,
slug casing, ISO dates and their ordering, numeric ranges. It does not resolve an
assignment's `consultant` / `project` refs to real entity files, so an assignment
pointing at a nonexistent consultant validates clean.

PR #1121 made that boundary explicit in `projects-entities/SKILL.md` Step 4 rather
than leaving it implied, and keeps the Step 2 "read both referenced entities first"
instruction as the operative guard. That guard is prose an authoring model follows,
not a machine check.

## Why it matters

The staffing match engine (#1114) scores over these refs. A dangling ref that
passes validation surfaces as a failure inside the engine, far from the entity
that caused it.

## Acceptance criteria

- [ ] Given a portfolio directory, `validate-entities.py` collects the consultant
      and project slugs and errors on any assignment whose `consultant` or
      `project` does not resolve, in the standard `{entity, file, field, message}`
      shape.
- [ ] Single-file invocation keeps its current behaviour (no portfolio-wide
      collection available, so no ref check) — the skill's Step 4 passes a single
      file and must not start failing.
- [ ] `projects-entities/SKILL.md` Step 4's scope statement is updated to match
      whatever the validator then guarantees.
- [ ] Lands before the staffing engine (#1114) consumes these refs.

**Parent PR:** #1121
**Linked issue:** #1113

claude-cogni-work:
Closes #1122.

## Summary

- `scripts/validate-entities.py`: when an argv path is a **portfolio directory**, collect that portfolio's consultant and project slugs and emit a `{entity, file, field, message}` error for every assignment whose flat `consultant` or `project` key does not resolve.
- Resolution is **scoped per directory**, so passing two portfolios in one invocation never resolves one portfolio's refs against the other's records.
- Single-file invocation is unchanged — no portfolio-wide collection exists, so no ref check runs. `projects-entities` Step 4 (single file) and `register-entity.py` (direct `validate_file()` call) both bypass the new code path entirely.
- Docs moved with the behaviour: the "frontmatter shape only / refs are not resolved" statements in `skills/projects-entities/SKILL.md` and `references/data-model.md` now describe the actual, invocation-dependent guarantee.
- New `tests/test_validate_entities_refs.sh` (12 assertions); `CLAUDE.md`'s tests listing now names every test file — it was also missing `test_portfolio_init.sh`.

## Design notes

**Placement.** The check lives in `main()`, not inside `validate_file()`. `validate_file()`'s `(errors, warnings)` signature is called directly by `register-entity.py`; threading a pre-collected slug set through it would change that contract for every caller to serve one invocation shape. The cost is one extra read of the batch's frontmatter, bounded by portfolio size — measured at 11–22 ms on portfolios of 300–600 files, flat in body size.

**A missing ref is not double-reported.** An assignment with an absent `consultant` is already a required-key error, so the ref pass skips empty values rather than adding a second, less informative error for the same defect.

**The filename stem is admitted only for a broken record.** A consultant/project whose own `slug` is missing or fails the kebab-case check contributes its filename stem to the known set, so one shape-invalid record does not cascade into dangling-ref errors on every assignment pointing at it. A record with a **valid** slug is deliberately *not* also resolvable by filename — every consumer (the manifest, the dashboard) matches on the slug, so accepting the stem there would pass a ref they then silently drop. Both directions are pinned by tests.

## Scope decisions

- **In:** the directory-scoped ref check in the standard error shape; single-file behaviour preserved; the SKILL.md Step 4 scope statement. `references/data-model.md` is included beyond the literal acceptance criteria because it carries the same boundary sentence — leaving it would ship docs contradicting the validator.
- **Moot:** the criterion about landing before the staffing engine consumes these refs — triage confirmed that engine already shipped.
- **Out:** the assignment's own composite `slug` stays a shape-only check. It is an identity, not a ref field.
- **Out:** no version bump — the post-merge workflow owns it.
- **Incidental:** while rewriting `CLAUDE.md`'s tests list, the pre-existing omission of `test_portfolio_init.sh` was corrected rather than knowingly left wrong.

## Test plan

- [x] `bash cogni-projects/tests/test_validate_entities_refs.sh` — 12/12 assertions pass.
- [x] `bash cogni-projects/tests/test_register_entity.sh` — green (direct `validate_file()` caller unaffected).
- [x] `bash cogni-projects/tests/test-render-dashboard.sh` — green (shares `_entity_files`).
- [x] `bash cogni-projects/tests/test_portfolio_init.sh` — green.
- [x] `scripts/check-breadcrumbs.py` and `scripts/check-version-bump.py` — both clean.

No CI job discovers `tests/*.sh` for any plugin in this repo, so these were run locally — the same as the existing suites.

## Follow-up issues

- #1172 — the ref check added here runs only on the validator's directory CLI invocation, but nothing in the plugin calls it that way. `register-entity.py` (the write path that can introduce the defect) still gates on a single-file `validate_file` call. Closing that needs a change to registration behaviour, outside this issue's acceptance criteria.
- #1173 — pre-existing, not caused by this change: the open/read/parse/guard sequence now exists in three scripts and has diverged; `staffing-score.py` guards only `OSError`, so a non-UTF-8 entity file raises past the envelope its siblings return.

Plan record: https://github.com/cogni-work/insight-wave/issues/1122#issuecomment-5160642790


claude-cogni-work:
## Ship-ready

Author instance is done. This PR is a draft carrying `svc:needs-review` and is
handed off to the merger instance for independent review.

- **Issue:** #1122 — full scope (the one remaining criterion, ordering against
  the staffing engine, is moot: that engine already shipped).
- **Branch:** `feat/issue-1122-assignment-ref-integrity`, forked from
  `origin/main` @ `4d3e15cc`.
- **Tests:** all four `cogni-projects/tests/*.sh` suites green locally, including
  the new 12-assertion ref suite. `check-breadcrumbs.py` and
  `check-version-bump.py` both clean. No version bump in the branch.
- **Not run by the author, by design:** `service-review`, the draft→ready flip,
  and `service-close`. The merger forms its own acceptance contract from the
  issue before reading this diff.

Two things worth the reviewer's attention:

1. The pre-commit review pass found and fixed a real defect in the first draft of
   this change: the filename stem was being admitted to the known-slug set
   unconditionally, which would have made a record resolvable by filename even
   when its declared slug was valid — passing refs that every consumer then
   drops. It is now admitted only when the record's own slug is missing or
   malformed, and both directions are pinned by tests.
2. This lands the check on the validator's directory CLI path only. Nothing in
   the plugin invokes it that way today, so the write path can still introduce a
   dangling ref — tracked in #1172 rather than widened into this PR, because
   closing it changes registration behaviour and is outside this issue's
   acceptance criteria.

Follow-ups filed: #1172, #1173.

sdh07:

_The merger formed this expectation from issue #1122 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds a portfolio-scoped referential-integrity pass to cogni-projects/scripts/validate-entities.py in main() (lines 331-397), after the per-file accumulation loop (369-385) and before the success computation (387), leaving validate_file() (220-328) shape-only. That placement is load-bearing, not stylistic: register-entity.py loads this file by location (lines 45-57) and gates registration by calling _ve.validate_file(entity_file) directly (line 113), and staffing-score.py does the same exec_module dance for parse_frontmatter (lines 100-115) — so the ref logic must not enter validate_file(), must not change its single-positional-arg signature, and must not add top-level execution or I/O that fires on import. The pass should build consultant/project slug sets from the files a *directory* argument expanded to (reusing _entity_files at 202-217 and SUBDIR_TO_TYPE at 109), admitting both the frontmatter `slug` and the filename stem so a shape-invalid consultant record does not cascade into a false dangling-ref error, then emit one error per unresolved flat `consultant`/`project` key on each assignment in the existing {entity, file, field, message} shape (err() at 225-228) with the existing 0/1/2 exit contract intact. Crucially the pass must be gated on the argument having been a directory — not on the slug sets being non-empty — otherwise a portfolio holding assignments and no consultants silently passes, and a single-file invocation (the shape projects-entities/SKILL.md Step 4 emits at lines 94-96) would start reporting every ref as dangling, breaking the skill. The three prose surfaces that today assert unconditional shape-only — the module docstring (lines 9-10, 18-20), projects-entities/SKILL.md:109-113, and references/data-model.md:269-281 — must be rewritten to the argument-shape-dependent guarantee, while SKILL.md Step 2's read-both-referenced-entities guard (lines 74-78) stays, since it remains the only guard on the single-file path. A regression suite at cogni-projects/tests/test_validate_entities_refs.sh in the house idiom of tests/test_register_entity.sh (lines 16-40: set -u, mktemp -d + trap, importlib.util.spec_from_file_location inside a python3 heredoc) and its registration in cogni-projects/CLAUDE.md (tree block 27-29, run block 36-39) complete the change. Note the AC item 'lands before the staffing engine (#1114)' is already moot — cogni-projects/scripts/staffing-score.py exists on base — so it is deliberately not a gate below.

**Acceptance checklist:**
- [ ] cogni-projects/tests/test_validate_entities_refs.sh exists, is bash with `set -u`, uses `mktemp -d` plus a `trap ... EXIT` cleanup, and loads the hyphenated validator via `importlib.util.spec_from_file_location` inside a `python3` heredoc — the same idiom as cogni-projects/tests/test_register_entity.sh lines 16-40 — and exits non-zero on any assertion failure.
- [ ] cogni-projects/CLAUDE.md registers the new test in BOTH the tests/ tree block (currently lines 27-29) and the runnable bash block (currently lines 36-39, `bash cogni-projects/tests/test_validate_entities_refs.sh`); listing it in one but not the other fails this item.
- [ ] No surface still claims the validator is unconditionally shape-only: `grep -rin "shape only\|not resolved to real entity" cogni-projects/` returns no remaining unconditional claim in validate-entities.py's module docstring (currently lines 9-10 and 18-20), projects-entities/SKILL.md (109-113), or references/data-model.md (278-281).
- [ ] projects-entities/SKILL.md Step 4's scope paragraph positively states the new guarantee: refs are resolved only when the validator is given a portfolio directory, and Step 4's own invocation passes a single file so it still checks shape only.
- [ ] projects-entities/SKILL.md Step 2's "read them first, which is also the guard against a dangling reference" instruction (currently lines 74-78) is retained — it remains the only guard on the single-file path and its deletion is a regression, not a cleanup.
- [ ] references/data-model.md's `## Validation` closing paragraph (currently 269-281) states the same argument-shape-dependent guarantee as the SKILL.md and the docstring — the three do not disagree with each other.
- [ ] Running `python3 scripts/check-breadcrumbs.py` exits 0: the SKILL.md edit introduces no new bare issue/PR ref (#NNNN), version tag, milestone, slice, or finding code (the maintainer-breadcrumb guard covers */skills/*/SKILL.md per .github/workflows/lint.yml:12-24).
- [ ] cogni-projects/.claude-plugin/plugin.json is not in the diff — no plugin version was touched, so the version-bump gate (.github/workflows/lint.yml:40-90) passes non-vacuously.
- [ ] Directory invocation over a portfolio whose assignment names a nonexistent consultant exits 1 with `success: false` and a `data.errors[]` entry carrying exactly the four keys {entity, file, field, message}, with entity `"assignment"`, file = the assignment's path, field = `"consultant"`, and a message naming the unresolved slug; the same holds with field `"project"` for a dangling project ref.
- [ ] An assignment whose consultant AND project are both unresolved yields TWO separate errors (one per field), not one merged error — one error per unresolved flat ref key.
- [ ] Single-file invocation of that same dangling assignment file exits 0 with `success: true` and zero errors — the regression the skill's Step 4 depends on — and the new test asserts this case explicitly against the same fixture it used for the directory-invocation failure.
- [ ] The ref pass lives in main() or a new module-level helper, NOT inside validate_file(): validate_file keeps its single-positional-arg signature and shape-only behaviour (register-entity.py:113 calls `_ve.validate_file(entity_file)`), and importing the module via exec_module (register-entity.py:45-57, staffing-score.py:100-115) triggers no new filesystem I/O or execution outside `if __name__ == "__main__"`.
- [ ] `bash cogni-projects/tests/test_register_entity.sh` and `bash cogni-projects/tests/test-render-dashboard.sh` still pass unmodified, and a directory invocation over a portfolio whose refs all resolve still returns `success: true` with exit 0 and no ref errors.
- [ ] The ref pass is gated on the argument having been a directory, not on the collected slug sets being non-empty: a portfolio directory holding an assignment with an empty or absent consultants/ directory still reports the dangling ref rather than silently skipping the check.
- [ ] When more than one portfolio directory is passed in one invocation, slug sets are scoped per directory root (or the code carries an explicit comment justifying a global union), so an assignment in portfolio A cannot resolve against a consultant in portfolio B; and in a mixed invocation only files that came from directory expansion are subjected to the ref pass.
- [ ] No cascade: a consultant record that is itself shape-invalid (e.g. missing `name`) produces only its own shape error — the assignment pointing at it does not additionally report a dangling ref, because the slug set admits the filename stem alongside the frontmatter `slug`.
- [ ] No double-reporting or crashes on degenerate assignments: an assignment missing `consultant` entirely (already "missing required field") or holding an empty/non-string value gets no second dangling-ref error, and a file under assignments/ with no parseable frontmatter is skipped by the ref pass rather than raising through it.
- [ ] Scope holds: the ref pass reads only the flat `consultant` and `project` keys — the assignment's own composite `slug` is not split on `--` and cross-checked, since it is identity rather than a reference field.

**Risk watchlist:**
- The load-bearing regression: collecting slugs from all_files unconditionally rather than from directory-expanded arguments, so a single-file invocation of an assignment finds empty slug sets and reports both refs as dangling — silently breaking projects-entities Step 4 (SKILL.md:94-96) and every authoring run.
- The mirror-image half-measure: skipping the ref pass whenever the collected consultant/project set is empty. That makes the single-file case pass for the wrong reason and blinds the check exactly in the portfolio state (assignments present, no consultant records) where every ref really is dangling.
- Moving the ref logic into validate_file() — or changing its signature to take a slug set — which breaks register-entity.py:113's direct call and therefore the registration gate, silently or with a TypeError swallowed by the catch-all at validate-entities.py:374-383.
- Adding module-level work (walking directories, opening files, calling main) that fires when register-entity.py:45-57 and staffing-score.py:100-115 exec_module this file, turning a cheap parser import into I/O or an exception at import time.
- A false dangling-ref cascade: building the slug set from the frontmatter `slug` alone, so one malformed consultant record turns into N extra assignment errors that point at the wrong file and bury the real shape error.
- Double-reporting a ref that is already flagged as a missing required field, inflating the error count and making the skill's fix-then-rerun loop report a phantom second problem.
- Scope creep into the assignment's composite `slug`: decomposing it and cross-checking both halves would start failing portfolios where the slug legitimately diverges from the flat ref fields, a break the issue never asked for.
- Silent cross-portfolio resolution when two portfolio roots are passed at once — a dangling ref in one portfolio masked by a same-named entity in the other.
- Documentation drift in the opposite direction: overclaiming in SKILL.md/data-model.md that the validator now guarantees referential integrity full stop, which would mislead an authoring model into dropping the Step 2 read-both-entities guard that is still the only protection on the single-file path.
- Deleting Step 2's dangling-ref guard as now-redundant — it is not redundant, because Step 4 invokes the validator on a single file.
- A test that only exercises a new helper function directly and never drives main()/the CLI, so the argument-shape gating (the entire substance of AC 1 and 2) goes untested.
- Introducing a `#1122`-style breadcrumb or a version tag into projects-entities/SKILL.md, tripping the deterministic breadcrumb guard, or bumping cogni-projects/.claude-plugin/plugin.json and tripping the version-bump gate.
- Judgment-only, not gated: whether the new error message wording is legible enough for an authoring model to act on without re-reading the data model. A human should eyeball the message text; no gate can decide it.
- AC item 4 ('lands before the staffing engine #1114 consumes these refs') is already unsatisfiable — cogni-projects/scripts/staffing-score.py exists on base. It is deliberately excluded from the checklist and must not be graded as an author defect.

- The plan-time AC lists 4 criteria (one of which — 'lands before the staffing engine #1114' — is already moot, since cogni-projects/scripts/staffing-score.py exists on base), but a base-grounded regeneration derives 18 gating items, including invariants the AC never named: validate_file()'s signature and shape-only behaviour must hold because register-entity.py:113 calls it directly; the module must stay import-safe because register-entity.py:45-57 and staffing-score.py:100-115 exec_module it; the pass must be gated on the argument being a directory rather than on the slug sets being non-empty; slug collection must not cascade off a shape-invalid consultant record; per-portfolio scoping on multi-directory invocation; the breadcrumb and version-bump CI gates; and regression coverage plus its CLAUDE.md registration. The acceptance bar was materially under-specified relative to what the base code requires. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1174

**Summary:** Three of four perspectives passed and contract-grading-reviewer confirmed all 18 committed review-contract checklist items satisfied with zero findings, but cogni-service:skill-quality-reviewer returned needs_revision on an un-waived minor finding - Step 4's closing clause invites swapping the portfolio directory into the same command that gates registration, which the model could misread as a verdict on the entity just authored. Rule 3 promotes the merged verdict to needs_revision on that single un-waived label regardless of the other three all-pass perspectives. The same underlying defect was independently flagged by plugin-dev:skill-reviewer at the same location and has been deduped. No critical or major findings, and no malformed opt_out waivers, were present.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[minor]** (auto-fixable) `cogni-projects/CLAUDE.md:30` — The new tree-block row breaks the block's description column alignment: test_validate_entities_refs.sh is longer than the column the four sibling rows align on, so its description starts two spaces after the name instead of at the shared column, leaving the tests/ block visually ragged where every other block in the file is aligned.
  - *Suggested fix:* Either widen the column for all four tests/ rows so the descriptions line up again, or shorten the description to keep the row within the existing column width.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:109-118` — The Step 4 scope paragraph does not name the shape of a dangling-ref error, while the preceding paragraph teaches the fix loop off data.errors[] field/message. references/data-model.md:281-282 does state that a dangling ref arrives as an ordinary {entity, file, field, message} error. Adding half a clause here would let the model act on a portfolio-wide run without re-reading the data model.
  - *Suggested fix:* In the last sentence of the Step 4 scope paragraph, note that a dangling ref arrives as an ordinary data.errors[] entry with field 'consultant' or 'project' naming the unresolved slug, so the same fix loop applies.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:116-118` — The closing sentence tells the model to 'run the command above with the portfolio directory ... in place of the single entity file path', but 'the command above' is the registration gate, and the paragraph two above it says 'If success is false ... do not register'. A portfolio-wide run returns success: false for any pre-existing dangling ref anywhere in the portfolio, so a model that follows the invitation literally will read a portfolio-level failure as a verdict on the entity it just authored and refuse to register a perfectly valid record.
  - *Suggested fix:* Replace the final sentence with: 'A portfolio-wide sweep for dangling refs is a separate run, not this gate: pass cogni-projects/<portfolio-slug> instead of the file path. Its success: false covers every entity in the portfolio, so read its data.errors[] for refs naming this entity rather than treating the run as a verdict on the file just authored.' (ref: skill-creator Writing Style - explain the why / Principle of Lack of Surprise)
  - *Raised by:* cogni-service:skill-quality-reviewer, also flagged by plugin-dev:skill-reviewer

- **[minor]** `cogni-projects/tests/test_validate_entities_refs.sh:171` — Three behaviours that the checklist names and that the code gets right are not asserted by the new suite: (a) an assignment with both consultant and project unresolved yields two separate errors, one per field (item 9) - every fixture dangles exactly one field; (b) a portfolio directory with an absent or empty consultants/ directory still reports the dangling ref (item 13) - every fixture seeds a populated consultants/; (c) an assignment file with no parseable frontmatter is skipped by the ref pass rather than raising (item 16). All three were verified manually against the worktree and behave correctly; they are simply untested.
  - *Suggested fix:* Add three short cases in the existing idiom: a fixture whose consultant AND project both dangle asserting two errors with distinct field values; a portfolio with consultants/ absent asserting the ref error is still reported; and a frontmatter-less file under assignments/ asserting the run returns an envelope rather than raising.

### Positive observations

- All 18 committed review-contract checklist items are satisfied by the diff - the ref pass is correctly directory-gated, symmetric across consultant/project, non-cascading, per-portfolio scoped, and validate_file() is untouched; both new and existing regression suites plus the breadcrumb gate pass clean. (contract-grading-reviewer)
- The invocation-shape gate is implemented exactly where the contract requires it - _ref_errors() is a new module-level helper called from main(), and validate_file() is byte-for-byte unchanged in the diff. (plugin-validator)
- The gate is on directory-ness, not slug-set emptiness - confirmed by probe: a portfolio holding an assignment with no consultants/ or projects/ directory still reports both refs as dangling, while single-file invocation of the same assignment returns success. (plugin-validator)
- Checklist item 4 satisfied precisely: the scope paragraph states the new guarantee positively and draws the consequence for this step's own single-file invocation rather than leaving it to inference. (skill-reviewer)
- Checklist item 5 satisfied: Step 2's dangling-reference guard survives untouched, and Step 4 now explicitly reinforces it instead of treating it as redundant. (skill-reviewer)
- Three-surface consistency confirmed: SKILL.md, references/data-model.md and the validator docstring state the same argument-shape-dependent guarantee without disagreement. (skill-quality-reviewer)
- No maintainer breadcrumbs introduced anywhere in the description or body. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1174

**Summary:** All three perspectives independently confirm the four findings from the prior review round are genuinely resolved with no regression; the merged verdict is needs_revision solely because skill-quality-reviewer returned that label on one residual minor finding, which flags that Step 2's dangling-ref guard is stated in filename terms while the validator resolves on frontmatter slug (pinned by test 5b) - text this PR did not touch. Four further minor findings (component-label wording, a file-mode mismatch, and two SKILL.md tightening opportunities) are all auto-fixable.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (5)

- **[minor]** (auto-fixable) `cogni-projects/README.md:86` — The Components table still describes validate-entities.py as 'Entity frontmatter validator'. That label now understates what the script guarantees — the PR updated the module docstring, references/data-model.md and SKILL.md Step 4, but left the two terse component labels (README table row and the CLAUDE.md tree row) describing frontmatter only.
  - *Suggested fix:* Retitle both rows to e.g. 'Entity validator (frontmatter + assignment refs)'.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:118-119` — The sweep is introduced as available but with no trigger condition, so a reader following Step 4 has no criterion for whether to run it during this invocation.
  - *Suggested fix:* Add the occasion to the opening clause, e.g. 'When the user asks for a portfolio-wide audit rather than a single authoring run...'.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:120-126` — The dangling-ref message shape is described twice in adjacent sentences ('whose message names the unresolved slug' and 'its message names the missing consultant or project, not the assignment's own slug'), and the second also re-states the file locator already given. The duplication reads as two passes layered rather than one statement.
  - *Suggested fix:* Collapse into one statement of the shape naming file, field and message together.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:74-77` — Step 4 leans on Step 2 as *the* guard against a dangling reference, but Step 2 states that guard as filename existence ('confirm both referenced entities already exist under consultants/ and projects/'). validate-entities.py::_ref_errors resolves a ref against each record's frontmatter slug and deliberately does not accept the filename stem for a well-formed record (pinned by test 5b). A record whose filename and slug diverge therefore passes Step 2's guard and still dangles in the portfolio sweep. NOTE: pre-existing text — this PR did not touch Step 2.
  - *Suggested fix:* State the guard in slug terms: read the referenced records first and confirm each record's frontmatter slug equals the value going into consultant/project, because the validator resolves on slug, not filename. (ref: skill-creator Writing Style — explain the why, stating the guard in the same terms the validator enforces)

- **[minor]** (auto-fixable) `cogni-projects/tests/test_validate_entities_refs.sh:1` — The new suite is committed mode 100644 despite a bash shebang, while two of its three siblings are 100755. It runs correctly because CLAUDE.md invokes every suite via 'bash <path>', and test-render-dashboard.sh is also 644, so the plugin already mixes both.
  - *Suggested fix:* git update-index --chmod=+x on the new suite.

### Positive observations

- All four prior findings verified by direct measurement rather than taking the commit message at its word: tree columns measured at 40 on every row, and the three new cases mutation-checked for non-vacuity. (plugin-validator)
- The single-file contract is protected structurally, not by convention: portfolio_batches is appended only when not os.path.isfile(path), so no argument shape other than a directory can trigger the ref pass. (plugin-validator)
- Per-directory batching is a non-obvious correctness point a flat concatenation would have broken; case 4 pins it. (plugin-validator)
- The malformed-slug fallback prevents one broken record cascading dangling-ref errors, while refusing to treat the filename stem as an alias for a valid-slug record; case 5b pins the refusal. (plugin-validator)
- No version bump in the branch and plugin.json/marketplace.json agree — correct per the post-merge convention. (plugin-validator)
- The revision commit is scoped to exactly the three files the findings implicated, with no opportunistic production-code edits riding along. (plugin-validator)
- Prior finding closed: the error shape is stated inline, so a reader no longer has to open references/data-model.md to interpret a dangling-ref entry. (skill-reviewer)
- Prior finding closed: the paragraph frames the directory invocation as a separate run and warns its success:false is portfolio-wide rather than a verdict on the entity just authored. (skill-reviewer)
- The scope statement is accurate against the implementation, and the described message wording matches _ref_errors' format string exactly. (skill-reviewer)
- Acceptance criterion 3 is satisfied and criterion 2 is reinforced by 'This step passes a single file, so no ref check runs here'. (skill-reviewer)
- Internal consistency preserved: Step 2's guard is restated as still load-bearing rather than quietly superseded. (skill-reviewer)
- The driving finding is closed cleanly: 'A portfolio-wide sweep for dangling refs is a separate run, not this gate' names the distinction outright. (skill-quality-reviewer)
- The second finding is closed correctly: the file-field locator plus the missing-referent clause matches the emitted shape in _ref_errors exactly. (skill-quality-reviewer)
- Step 4's scope statement now states the validator's guarantee precisely and split by invocation shape, which is what acceptance criterion 3 asked for. (skill-quality-reviewer)
- Progressive disclosure is in good shape: 7653 chars against a 19500 cap, with the canonical field contract offloaded to references/data-model.md behind an explicit precedence rule. (skill-quality-reviewer)
- No maintainer breadcrumbs in the description or body. (skill-quality-reviewer)
- Rules are reasoned rather than commanded throughout. (skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the four findings from the previous verdict in `02ec23fb`.

| Finding | Location | Resolution |
|---|---|---|
| [minor] tree column alignment | `cogni-projects/CLAUDE.md:30` | Every row in this tree aligned its description at column 38, and the new row's name is exactly 30 chars against an 8-char prefix — so it filled the column with no separation. Re-padding only the four `tests/` rows (the suggested first option) would have left the other 13 rows at 38, trading one raggedness for another, so the file-wide column moved 38 → 40. Whitespace only. |
| [minor] error shape not named | `skills/projects-entities/SKILL.md:109-118` | The sweep paragraph now states that a dangling ref arrives as an ordinary `data.errors[]` entry whose `field` is `consultant` or `project`, so the Step 4 fix loop applies without re-reading `references/data-model.md`. |
| [minor] sweep invited re-running the registration gate | `skills/projects-entities/SKILL.md:116-118` | Rewritten with findings 2 and 3 folded into one coherent paragraph: the sweep is now framed as a separate run, and its `success: false` is explicitly scoped to the whole portfolio. Entries are located by each error's `file` field — during the pre-commit pass this was caught as still imprecise, because the `message` names the *missing referent* (`consultant 'x' does not resolve to...`), not the authored entity, so the locator was corrected before commit. |
| [minor] three untested behaviours | `tests/test_validate_entities_refs.sh:171` | Added cases 7–9: both refs dangling reported once per field; a ref still dangling when `consultants/` is absent entirely; a frontmatter-less file skipped by the ref pass rather than raising. Suite goes 12 → 20 assertions, all passing. All three behaviours were verified against `validate-entities.py` rather than assumed. |

`make_portfolio` gained a `project_ref` parameter so either side of a ref can be made to dangle without rebuilding the record set — this replaced a build-then-discard block that duplicated the helper's own assignment frontmatter.

**Verification:** all four `cogni-projects` suites pass; `check-breadcrumbs.py`, `check-frontmatter.sh`, `check-skill-names.sh` and the version-bump gate are clean (1 plugin touched, 0 violations). No production code changed in this revision, and no version bump — the post-merge workflow owns that.

The fresh verdict is **needs_revision** on five residual `minor` findings, all `auto_fixable`. Note the one that drove the label cites `Step 2` text this PR never touched (a pre-existing filename-vs-slug imprecision in the dangling-ref guard).

claude-cogni-work:
## Author instance: holding, not handing off

The bounded auto-fix cycle ran once (the one-shot cap this contract allows) and closed all five residual minor findings in `85993dbb`:

- Component labels in `README.md` and the `CLAUDE.md` tree now name both guarantees rather than frontmatter only.
- Step 2's dangling-ref guard is restated in slug terms, matching what `_ref_errors` actually resolves on.
- Step 4's sweep gained its missing trigger condition and states the error shape once instead of across two overlapping sentences.
- The new suite got the executable bit its shebang implies.

Re-review of that head: `plugin-dev:plugin-validator` **pass** (zero findings, alignment regression checked, README addition confirmed single-line). `cogni-service:skill-quality-reviewer` **needs_revision** on one new residual — Step 4's back-reference still described Step 2's guard as *reading* both referenced entities, while `85993dbb` had just redefined that guard as slug equality. That inconsistency was introduced by the auto-fix itself, so it is fixed in `c40887da` rather than left on the branch.

**State:** `c40887da` is **not** covered by the posted verdict, which reviewed `02ec23fb`. The one-shot cap forbids a second auto-fix cycle in this invocation, so this PR is deliberately **not** flipped to `svc:needs-review` — it keeps `svc:review-needs-revision` and stays re-pickable by the next author drain, which will re-review the current head.

No production code changed in any of the three revision commits; all four `cogni-projects` suites, the breadcrumb, frontmatter, skill-name and version-bump gates are green throughout.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1174

**Summary:** Three of four perspectives passed with only minor, auto-fixable findings. The merged verdict is needs_revision solely because cogni-service:skill-quality-reviewer self-reported that label on one minor, auto-fixable finding — that the portfolio sweep is framed as audit-only, so the one authoring path that writes a ref never reaches the deterministic check. All five findings from the prior verdict round are independently confirmed closed on this head with no regression.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:120-127` — The sweep's trigger condition is scoped to 'when the user asks for a portfolio-wide audit rather than a single authoring run'. Authoring an assignment is the only path that writes a ref, and under that framing it never reaches the deterministic ref check — the guard stays Step 2's read-and-compare, an LLM judgment call, even though a cheap directory-argument run right after Step 5 would confirm it. (ref: skill-creator Writing Style — explain the why)
  - *Suggested fix:* Append to that paragraph: authoring an assignment is the one case worth sweeping unprompted rather than on request — it is the only entity type that carries refs, so a portfolio-directory run right after Step 5 is what turns Step 2's read-and-compare into a checked fact. Keep it after registration, not in place of the Step 4 gate.

- **[minor]** (auto-fixable) `cogni-projects/scripts/validate-entities.py:461` — The ref-pass loop sits outside the defensive per-file try/except that wraps validate_file. An unexpected exception inside the ref pass would fall through to the module-level catch-all, which prints an envelope with an empty errors[] and exit 2 — discarding every per-file error already accumulated. That contradicts the stance the file argues for itself (one unreadable record must not abort the scan). Exposure is low (reads are guarded, parse_frontmatter is regex/string-only), so this is hardening, not a live defect.
  - *Suggested fix:* Wrap the ref-pass loop in the same try/except pattern used for validate_file, appending a single {entity:'unknown', file:<batch root>, field:'<refs>', message:'unexpected failure while resolving refs: ...'} error instead of letting the exception escape.

- **[minor]** (auto-fixable) `cogni-projects/scripts/validate-entities.py:438` — Passing the same portfolio directory twice duplicates the ref errors: 'validate-entities.py p p' returns two byte-identical consultant errors. portfolio_batches appends one batch per argv entry with no de-duplication. This mirrors the pre-existing duplication of shape errors in all_files, so it is not a regression introduced by this PR — but the new pass compounds it.
  - *Suggested fix:* De-duplicate directory arguments by os.path.realpath before appending to portfolio_batches, so a repeated path is validated once.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:100-105` — The fix loop is written as if success:false always implies a populated data.errors[]. validate-entities.py has envelope-level failure paths returning success:false with an EMPTY errors[] and the reason only in the top-level error string ('path not found', and the fail-closed 'no entity files found under ... and it is not a portfolio root'). Both are reachable from the very invocation the new portfolio-sweep paragraph introduces — a mistyped portfolio slug, or passing .../assignments instead of the portfolio root. An agent following the current text would loop over an empty errors[] with nothing to fix.
  - *Suggested fix:* Add one clause to the fix-loop sentence: when data.errors[] is empty, the top-level error holds the reason (a path that does not exist, or a directory that is neither an entity file nor a portfolio root); fix the path rather than the entity.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:79` — 'That slug match is the guard against a dangling reference.' restates the sentence immediately before it, which already gives the mechanism and the consequence. Step 4 then names the same guard a third time. The Step 4 back-reference is load-bearing; the Step 2 label sentence is not.
  - *Suggested fix:* Drop line 79's standalone sentence, or fold the label into the preceding sentence.

- **[minor]** `PR diff` — Diff is 406 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- All five findings from the prior verdict round are independently confirmed closed on this head, verified by direct measurement rather than by reading the commit messages. (plugin-validator, skill-reviewer, skill-quality-reviewer)
- AC1 verified by direct execution: a dangling consultant ref yields exactly one error in the standard four-key shape, keyed on the assignment's file, naming the missing referent. (plugin-dev:plugin-validator)
- AC2 verified: the same dangling assignment passed as a single path returns success:true / exit 0, so SKILL.md Step 4 and register-entity.py are both unaffected; the new test's case 3 is an explicit regression guard. (plugin-dev:plugin-validator)
- AC3 verified: Step 4 now states the invocation-shape dependency exactly as the code behaves, and Step 2 matches the deliberate stem-is-not-an-alias rule and its test case 5b. (plugin-dev:plugin-validator)
- All four cogni-projects suites pass on this head (new suite: 20 assertions, exit 0), so shared-internals consumers are unregressed. (plugin-dev:plugin-validator)
- The ref pass is correctly scoped per directory argument rather than over the flat all_files concatenation, and that scoping is tested — the obvious wrong implementation would have silently cross-resolved. (plugin-dev:plugin-validator)
- Cascade suppression is well-judged and tested: a shape-invalid record is skipped rather than re-reported as N dangling refs, while the underlying error is still emitted exactly once. (plugin-dev:plugin-validator)
- No repo-wide caller regression: no skill or script invokes main() with a portfolio directory as a gate, so no existing workflow starts failing on pre-existing dangling refs. (plugin-dev:plugin-validator)
- Doc coherence is complete rather than partial — module docstring, data-model.md, CLAUDE.md, README.md and SKILL.md all describe the same guarantee; no stale 'frontmatter shape only' claim survives. (plugin-dev:plugin-validator)
- Repo conventions hold: no branch version bump, stdlib-only, breadcrumb and skill-name guards clean, py_compile and bash -n clean, new test file mode 755 matching siblings. (plugin-dev:plugin-validator)
- Multilingual encoding convention respected end to end: a ref value 'müller-groß' renders intact via ensure_ascii=False rather than as escapes. (plugin-dev:plugin-validator)
- Prior finding closed: the portfolio sweep now opens with an explicit trigger condition and an explicit contrast to the authoring gate, so it cannot be misread as a second step of every authoring run. (plugin-dev:skill-reviewer)
- Prior finding closed: the dangling-ref error shape is stated exactly once and is accurate against _ref_errors. (plugin-dev:skill-reviewer)
- Prior finding closed and now mutually consistent: Step 2's guard is stated in frontmatter-slug terms, matching _ref_errors' slug-set construction, and Step 4's back-reference uses the same slug framing. (plugin-dev:skill-reviewer)
- Step 5's parenthetical is exactly calibrated: register-entity.py calls validate_file only, never _ref_errors — no over-claim of ref integrity at registration time. (plugin-dev:skill-reviewer)
- Scope prose is consistent across the surrounding docs at this head, so the skill does not contradict the canonical field contract it points readers at. (plugin-dev:skill-reviewer)
- Body is 7870 chars against a 19500 cap for complexity 9, with the canonical field contract offloaded to references/data-model.md and an explicit precedence rule on the quick-reference duplication. (skill-reviewer, skill-quality-reviewer)
- Voice is imperative and reason-first throughout — no CAPS-stacked MUSTs; each gate states why it exists. (cogni-service:skill-quality-reviewer)
- No maintainer-meta breadcrumbs despite the change originating from a numbered issue. (cogni-service:skill-quality-reviewer)

### Recommendation

⚠️ **Revise before merge.** 6 finding(s) need attention.

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1174
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator, cogni-service:skill-quality-reviewer

**Summary:** The bounded auto-fix cycle closed all five findings from the previous verdict, confirmed by execution rather than by reading the commit message. Two perspectives re-reviewed the new head: plugin-validator returned pass, and skill-quality-reviewer returned needs_revision on one fresh minor. Every finding on this head is minor and auto-fixable; none blocks on correctness, and all four issue acceptance criteria remain independently verified met.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md — Step 5 closing paragraph` — The new unprompted-sweep instruction is correct but is stated three paragraphs upstream of where it executes. A model working the file top-down finishes Step 4's gate, moves to Step 5 (register), then Step 6 — and Step 5 never mentions the sweep. The forward reference across a step boundary is the one place the assignment path can still silently skip the deterministic ref check the change exists to reach.
  - *Suggested fix:* Append one sentence to Step 5's closing paragraph, after the data.action sentence: 'When the entity was an assignment, follow the successful registration with the portfolio-directory sweep described in Step 4 — that run is what turns the Step 2 slug comparison into a checked fact.' Leave the Step 4 paragraph as written; this is a back-pointer, not a relocation.

- **[minor]** (auto-fixable) `cogni-projects/tests/test_validate_entities_refs.sh:236` — Neither new behaviour introduced by the auto-fix commit is codified as a regression test. The suite's 9 groups still pass unchanged, but nothing asserts that a repeated directory argument reports a dangling ref once, nor that an exception inside _ref_errors surfaces as an ordinary error rather than an empty errors[]. Both are silent-reintroduction risks under a future refactor of the batching loop — precisely the loop this PR rewrites.
  - *Suggested fix:* Append two cases to the existing harness, which already loads the module as `ve` and drives main() directly. (a) Reuse the group-2 dangling-consultant portfolio and call main([p, p]), asserting exactly one error with field == 'consultant'. (b) Monkeypatch ve._ref_errors to raise, call main([portfolio]), and assert the envelope carries the pre-existing per-file errors plus one entry with field == '<refs>' and return code 1 (not 2 with an empty errors[]). Restore the original afterwards.

- **[minor]** `PR diff` — Diff is 438 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- The ref-pass try/except covers the exact failure mode claimed, confirmed by fault injection on BOTH heads: monkeypatching _ref_errors to raise on c40887da let the exception escape to the module-level catch-all (empty errors[], exit 2, every accumulated per-field error discarded); the same injection on bb3dabe1 returns exit 1 with the pre-existing shape errors intact plus one {entity:'unknown', field:'<refs>'} entry. (plugin-dev:plugin-validator)
- The try is placed per batch rather than around the loop, so one failing portfolio does not suppress the other's ref pass — a two-portfolio injection produced two independent '<refs>' errors. (plugin-dev:plugin-validator)
- The batch[0]-else-'<portfolio>' locator is safe and its fallback is genuinely reachable rather than dead: a portfolio root holding projects-portfolio.json with no entity files batches an empty list, and injecting a raise there emitted file='<portfolio>' with no IndexError. (plugin-dev:plugin-validator)
- The realpath de-duplication is correctly scoped inside the not-isfile branch, so no single-file path reaches it. Byte-compared old vs new head: single-file invocation identical (AC2 intact), repeated single-file argument identical, and the distinct-two-portfolios case identical at 8 errors with a sibling-only slug still dangling — portfolio isolation preserved. Only the repeated-directory case moves, 2 ref errors to 1. (plugin-dev:plugin-validator)
- De-duplication is by realpath rather than by string, so a trailing-slash alias and a symlink to the same portfolio also collapse to one ref pass — both verified. all_files is deliberately left unchanged, so `checked` and shape-error behaviour are bit-identical to the prior head. (plugin-dev:plugin-validator)
- The argument-error paths the new Step 4 prose describes were confirmed to behave as documented: a nonexistent path, a non-portfolio directory, and an entity subdirectory passed where the root belongs each return success:false with an empty errors[] and the reason in the top-level error at exit 2, while an empty-but-valid portfolio root remains a vacuous success. (plugin-dev:plugin-validator)
- Prior skill-quality finding closed on the merits: the new paragraph names assignment-authoring as the one case worth sweeping unprompted, states the reason, pins the ordering after registration, and explicitly ring-fences the Step 4 gate as unchanged — no scope creep into the single-file path. (cogni-service:skill-quality-reviewer)
- The Step 4 empty-errors edit reads as reasoned diagnosis rather than a rule: it tells the model what an empty data.errors[] with success:false means (argument problem, not entity problem) and where the reason lives. (cogni-service:skill-quality-reviewer)
- The Step 2 fold left no seam — the guard rationale now reads as one continuous sentence pair, and nothing elsewhere in the body references the removed standalone label. (cogni-service:skill-quality-reviewer)
- Body is 8471 chars against a 19500 cap (complexity 9); the commit did not inflate it and progressive disclosure to references/data-model.md is intact. (cogni-service:skill-quality-reviewer)
- No regression in the four suites: test-render-dashboard, test_portfolio_init, test_register_entity and test_validate_entities_refs are all green on bb3dabe1, and group 4 ('refs never resolve across two portfolios') is the direct guard on the de-duplication change. (plugin-dev:plugin-validator)
- No version bump: neither cogni-projects/.claude-plugin/plugin.json nor .claude-plugin/marketplace.json is touched, both read 0.0.14 and agree, and the version-bump gate reports 0 violations vs origin/main. (plugin-dev:plugin-validator)
- Breadcrumb, frontmatter, external-dispatch and skill-name gates all exit 0; validate-entities.py compiles clean and the delta is confined to the two files claimed. (plugin-dev:plugin-validator)

### Recommendation

---

claude-cogni-work:
## Revision posted — and held, not handed off

This author drain picked the PR up at `c40887da`, the head the previous invocation left uncovered by any verdict.

**Step 1 — re-review of the uncovered head.** All five findings from the verdict on `02ec23fb` were independently confirmed closed against the tree, by reading it rather than trusting the commit messages: both component labels name the full guarantee, the sweep has its trigger condition, the message shape is stated once, Step 4's back-reference matches Step 2's slug framing, and the new suite is mode 755. Three perspectives ran; two passed, and the merged verdict was `needs_revision` on one auto-fixable minor.

**Step 2 — bounded auto-fix cycle (`bb3dabe1`).** All five findings from that verdict were closed:

| Finding | Resolution |
|---|---|
| `validate-entities.py` ref pass outside the defensive try/except | Wrapped per batch; an exception now becomes an ordinary `{entity:"unknown", field:"<refs>"}` error instead of reaching the module catch-all |
| Repeated directory argument duplicated every ref error | Directory arguments de-duplicated by `realpath` before batching |
| Step 4 fix loop assumed a populated `data.errors[]` | Now covers the envelope-level failures where the reason is in the top-level `error` |
| Sweep framed as audit-only | Names assignment-authoring as the one unprompted-sweep case, after registration, Step 4 gate ring-fenced |
| Step 2 guard label restated redundantly | Folded into the sentence that already gives the mechanism |

Verified by execution, not assertion: a repeated directory argument now yields one error rather than two, and fault injection confirms the try/except preserves accumulated per-file errors that previously would have been discarded. Single-file invocation still returns `success: true` with no ref check, so AC2 holds. All four `cogni-projects` suites and the breadcrumb, frontmatter, skill-name and version-bump gates are green.

**Why this is held rather than handed to the merger.** The re-review of `bb3dabe1` came back `needs_revision` again — `plugin-validator` passed, `skill-quality-reviewer` raised one *new* auto-fixable minor (a Step 5 back-pointer to the sweep). The one-shot auto-fix cap for this invocation is spent, so the PR keeps `svc:review-needs-revision` and is not flipped to `svc:needs-review`.

**Worth a human's attention before another drain runs.** `skill-quality-reviewer` has now returned `needs_revision` on three consecutive heads — each time on a single, different, auto-fixable prose nit, with `structural_issues[]` empty every time. Each observation is individually reasonable, but the pattern means the PASS gate the author handoff requires may not be reachable by further auto-fix rounds: every drain spends a full cycle and ends in the same hold. The two open findings are the Step 5 back-pointer and two regression tests for the behaviours `bb3dabe1` introduced; neither blocks on correctness, and all four acceptance criteria on #1122 are independently verified met.

**Also flagged:** #1171 is a second open PR resolving the same issue (#1122) on a different branch, `feat/issue-1122-assignment-ref-check`. This PR is the more advanced of the two. Deciding which to keep is a human call — no drain action was taken on either.

sdh07:
## Human decision: loop broken, handing off to the merger

`92b2c87f` closes the two open findings from the verdict on `bb3dabe1`. **This PR
is handed off regardless of any fresh minor the next review surfaces** — that is a
human ruling, recorded here so the next drain does not restart the cycle.

**Why the ruling.** `skill-quality-reviewer` has returned `needs_revision` on three
consecutive heads (`02ec23fb`, `c40887da`, `bb3dabe1`), each time on a single,
different, auto-fixable prose nit, with `structural_issues[]` empty every time.
Each observation was individually reasonable and each was closed on the merits. But
a gate that produces a fresh minor on every head it sees is not converging, and the
PASS the author handoff requires is not reachable by spending another cycle. All
four acceptance criteria on #1122 are independently verified met, and the last
verdict's own summary says every finding on this head is minor and auto-fixable
with none blocking on correctness.

**What `92b2c87f` changes**

| Finding | Resolution |
|---|---|
| Step 5 never mentions the sweep — the forward reference crosses a step boundary, so a model working top-down can silently skip the deterministic ref check | Back-pointer appended to Step 5 after the `data.action` sentence. Step 4's paragraph is unchanged, as the finding asked. Placed after registration because the sweep is only worth running once the assignment is in the manifest. |
| Neither behaviour `bb3dabe1` introduced is codified — nothing pins repeated-directory de-duplication or the ref-pass try/except | Two groups added to the existing harness: group 10 (repeated directory argument → exactly one ref error) and group 11 (a raising `_ref_errors` → one ordinary `{field: "<refs>"}` error, accumulated per-file errors intact, exit 1). |

**Both new guards were mutation-verified, not assumed.** Each was run against a
copy of the tree with the guarded property reverted:

- Remove the `realpath` de-duplication → group 10 goes RED with exactly the
  duplicate-error symptom (two byte-identical `consultant` errors).
- Remove the per-batch `try`/`except` → group 11 goes RED on 5 assertions.

One thing worth noting about group 11: because the harness drives `main()` directly
rather than as a subprocess, an escaped exception would abort the run instead of
failing an assertion. The escape is therefore caught and asserted on explicitly, so
a regression reads as a named failure rather than a traceback. The suite's header
enumeration was extended to cover both new edges — leaving it listing only the
original ones would have understated what the file guards.

**Gates, all green on `92b2c87f`:** all four `cogni-projects/tests/*.sh` suites
(28 assertions in the ref suite, up from 20); `check-breadcrumbs.py`,
`check-external-dispatch.py` and `check-version-bump.py` all clean
(`14 plugin(s) scanned, 1 touched, 0 violation(s)`). No version bump —
`git diff origin/main -- '*/.claude-plugin/plugin.json'` is empty; insight-wave
bumps post-merge. Skill body 8821 chars against the 19500 cap.

**Also handled:** #1171, the duplicate PR on #1122, is closed and its branch
deleted. This PR is the survivor — it fixes all three majors #1171 carried
(global slug pooling across directory arguments, the argument-shape-scoped
single-file exemption, and the shape-error cascade), and graded 18/18 contract
items against #1171's 15/18.

Applying `svc:needs-review`, removing `svc:review-needs-revision`.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1174
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer

**Summary:** All five perspectives passed independently, including contract-grading-reviewer at 18/18 with zero findings. The only un-waived finding is the routine oversized-PR-diff advisory (Rule 5b: an advisory minor, non-gating). Three further minors are explicitly waived with written justification - two of them (skill-reviewer, skill-quality-reviewer) are the same underlying duplicate sweep-trigger prose defect and are merged into one finding here; the third (plugin-validator) is a pre-existing dedup asymmetry disclosed in code.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** `PR diff` — Diff is 509 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

- **[minor]** (waived) `cogni-projects/scripts/validate-entities.py:407-448` — The realpath de-duplication added for repeated directory arguments covers only the ref pass; all_files is deliberately left unchanged, so passing the same portfolio twice still emits byte-identical duplicate SHAPE errors for every malformed record. The asymmetry is disclosed in an inline comment and regression-covered on the ref side (test group 10).
  - *Justification:* Pre-existing behaviour on base (duplicate paths already duplicated shape errors before this PR), no contract item covers it, the boundary is documented in code, and the new dedup strictly improves the situation rather than introducing the asymmetry. Out of scope for this change.

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:132-136 and :154-158` — The 'when to run the portfolio-wide sweep' instruction is dual-sourced: Step 4's final paragraph and Step 5's new closing paragraph each state the same three claims - that an assignment is the case worth sweeping unprompted, that the sweep belongs after registration, and that it never displaces the Step 4 single-file gate - down to a near-verbatim repeat of the 'turns Step 2's read-and-compare / slug comparison into a checked fact' clause. A future edit to either paragraph silently diverges from the other. NOTE: the Step 5 paragraph is the MERGER's own edit, so this waiver is self-assessed; tracked as follow-up issue #1178.
  - *Justification:* Functionality-neutral prose duplication. No contract item requires the sweep prose at all - items 4 and 5 are both fully satisfied by the Step 4 scope paragraph and the Step 2 guard, untouched by this fix - and both copies are individually accurate against the implementation. Waived by the merger with the conflict of interest stated openly, on the same grounds plugin-dev:skill-reviewer independently reached, and tracked as a follow-up rather than gating the merge.

### Positive observations

- 18/18 checklist items satisfied with cited evidence; the suite was run (28/28) rather than trusted. (contract-grading-reviewer)
- item 12: validate_file(filepath) keeps its single-positional-arg signature, so register-entity.py:113 is unaffected. (contract-grading-reviewer)
- item 15: portfolio_batches holds one file-list per directory argument, so slug sets never merge across directories. (contract-grading-reviewer)
- item 16: the filename stem is admitted only when the record's own slug is missing or malformed, so one shape-broken record cannot cascade dangling-ref errors. (contract-grading-reviewer)
- Item 12 verified empirically: exec_module with builtins.open and os.listdir instrumented opens zero files and lists zero directories. (plugin-validator)
- Item 14 verified on the exact shape the risk watchlist named: a portfolio with no consultants/ directory still reports the dangling ref and exits 1. (plugin-validator)
- The two groups added by the head commit (group 10 realpath dedup, group 11 ref-pass containment) each go RED under their own targeted mutation - the prior verdict's two named findings are genuinely closed. (plugin-validator)
- CLAUDE.md registration is complete on both surfaces and opportunistically fixes a pre-existing gap: test_portfolio_init.sh was registered in neither block on base. (plugin-validator)
- Item 4 is met positively, not by omission: :114-121 states that a passing single-file run is not evidence the refs exist. (skill-reviewer)
- Item 5 is not merely retained but sharpened toward the code: the guard is upgraded to frontmatter-slug equality, matching _ref_errors' actual resolution key. (skill-reviewer)
- The three prose surfaces genuinely converge rather than each merely being edited; no surface overclaims full referential integrity. (skill-reviewer)
- structural_issues[] empty - no structural drift, introduced or pre-existing, on this head. (skill-quality-reviewer)
- Body is 8,820 chars against a 19,500-char complexity-scaled budget - progressive disclosure intact and not eroded by this change. (skill-quality-reviewer)

### Recommendation

---

Issue #1182, opened 2026-08-03
cogni-projects follow-up (#1173): extract a shared tests/ assertion harness — three suites now inline it byte-identically


claude-cogni-work:
## Context

Surfaced by the pre-commit quality pass on #1173, which added
`cogni-projects/tests/test_staffing_score.sh` — the plugin's fifth suite.

## Problem

The bash assertion harness is now inlined byte-identically across suites:

- `pass()` / `fail()` / the `failures` counter and `assert_json()` are defined in
  **three** of the five suites (`test_portfolio_init.sh`, `test-render-dashboard.sh`,
  `test_staffing_score.sh`).
- `assert_no_traceback()`, `seed_portfolio()` and `write_entity()` are duplicated
  between `test-render-dashboard.sh` and `test_staffing_score.sh`, along with the
  `set -u  # NOT -e` preamble and the `TMPROOT` / `trap` idiom.
- The latin-1 fixture writer — the `python3 -c "... .encode('latin-1')"` block and
  its "bash printf parses a leading `---` as options" comment — is duplicated
  verbatim, and is exactly what a third suite would need next.

This is the same defect class #1173 fixed on the script side: a shared sequence
copied per consumer, which then drifts.

## The drift has already started

`test_staffing_score.sh`'s `run_score()` redirects stdout to a file rather than
piping into `tail`, because after a `$(... | tail -n 1)` command substitution `$?`
is *tail's* exit status — always 0 — so any exit-code assertion passes vacuously.
`test-render-dashboard.sh`'s `run()` still has the piped shape, which is why that
suite asserts no exit codes at all and has no `assert_exit` helper.

So the two copies already disagree about a correctness-relevant detail, and the
newer one is the better primitive.

## Proposed change

Extract `cogni-projects/tests/fixtures/test_helpers.sh` (or `tests/lib.sh`) and
source it from all five suites. There is an in-repo precedent with the same
rationale in its header: `cogni-knowledge/tests/fixtures/test_helpers.sh`, sourced
by 10+ suites.

Carry `run_score`'s file-redirect shape and `assert_exit` forward as the canonical
runner — not `run()`'s — and let `test-render-dashboard.sh` gain exit-code coverage
it currently cannot express.

Note that inlining is the dominant convention repo-wide today (cogni-consult,
cogni-workspace, cogni-portfolio and cogni-trends all define `pass`/`fail`
per-suite), so this is a deliberate departure for a plugin that has crossed the
[...]

Issue #1181, opened 2026-08-03
cogni-projects follow-up (#1173): collapse the three duplicated importlib validator-loader blocks


claude-cogni-work:
## Context

Deferred from #1173. Noted while extracting the shared `read_frontmatter` helper.

## Problem

The "load `validate-entities.py` by file location" `importlib` block is duplicated
three times, and the copies have deliberately diverged:

| Site | Scope | Extracts | On load failure |
|---|---|---|---|
| `render-dashboard.py:37-50` | module scope | whole module as `_ve` | prints envelope, `sys.exit(2)` at **import** time |
| `register-entity.py:45-57` | module scope | whole module as `_ve` | prints envelope, `sys.exit(2)` at **import** time |
| `staffing-score.py:100-116` | function scope | a single attribute | returns `None`, caller `_fail`s with exit **1** |

`validate-entities.py` is not an importable module name (hyphens), which is why
every consumer loads it by path in the first place.

## Why this is not a mechanical extraction

The three differ in ways that look like drift but are partly deliberate:

- **Exit code.** `staffing-score.py` reserves exit 2 for usage errors and treats a
  broken install as exit 1 (domain failure); the other two exit 2 at import time.
- **Failure timing.** Import-time `sys.exit(2)` fires before argparse/`main()` runs;
  `staffing-score.py`'s deferred check runs *after* its argument and portfolio-dir
  preflight, so a bad-args invocation still reports the usage error rather than the
  install error.
- **What is bound.** Two need the whole module (`validate_file`, `SCHEMA`,
  `_entity_files`); one needs a single callable.

So unifying them requires deciding which of those behaviours is canonical — a
behaviour-design task, not a refactor. Filing it separately rather than smuggling
the decision into a bug fix.

## Proposed change

Agree a single loader contract and apply it to all three sites. Suggested shape:
one helper returning the loaded module, with the caller choosing what to pull off
it and how to report failure — preserving each script's current exit code and
failure timing unless there is a reason to change them.

## Acceptance criteria

- [ ] One loader implementation, used by all three scripts.
- [ ] Each script's existing exit code on a missing/broken validator is preserved,
      or the change is deliberate and stated.
- [ ] All `cogni-projects/tests/*.sh` suites stay green.

claude-cogni-work:
## Triage Report

- **Severity:** S4-enhancement
- **Component:** cogni-projects / scripts
[...]

PR #1183, opened 2026-08-03
cogni-projects: shared read_frontmatter helper (UTF-8 guard)

### LINKED ISSUE #1173: cogni-projects: extract a shared read_frontmatter helper — staffing-score.py is missing the UnicodeDecodeError guard its siblings have

### LINKED ISSUE #1180: cogni-projects follow-up (#1173): route register-entity.py's fourth read-frontmatter site through the shared helper
## Context

Deferred from #1173, which extracted a public `read_frontmatter(path) -> (fm, exc)`
helper in `cogni-projects/scripts/validate-entities.py` and routed the three read
sites the issue named through it.

## Problem

`cogni-projects/scripts/register-entity.py:121-125` is a **fourth** copy of the
open/read/parse/guard sequence, and it guards only `OSError`:

```python
    try:
        with open(entity_file, "r", encoding="utf-8") as f:
            fm = _ve.parse_frontmatter(f.read())
    except OSError as exc:
        return _fail("cannot read entity file: %s" % exc)
```

## Why this was deferred rather than fixed in #1173

**It is not a live defect.** `register-entity.py:113` calls `_ve.validate_file()`
first, which *does* guard `UnicodeDecodeError` and returns an error, and line 114
short-circuits to `_fail`. So an undecodable file can never reach line 121 — the
missing guard is unreachable, unlike the `staffing-score.py` site #1173 fixed,
where the exception escaped to a raw traceback with no envelope at all.

What remains is textual duplication: #1173's AC 1 asks that the helper be "the
only place the open/read/parse/guard sequence lives", and on a strict reading this
site still counts. That is why #1173's PR links `Refs`, not `Closes`.

## Proposed change

Replace the block with `fm, exc = _ve.read_frontmatter(entity_file)` +
`if exc is not None: return _fail("cannot read entity file: %s" % exc)`. The
message stays unchanged and `register-entity.py:45-57` already binds the whole
validator module as `_ve`, so no loader change is needed.

## Acceptance criteria

- [ ] `register-entity.py` no longer contains its own open/read/parse/guard sequence.
- [ ] The `cannot read entity file: %s` message is unchanged.
- [ ] `bash cogni-projects/tests/test_register_entity.sh` stays green.
- [ ] With this landed, #1173's AC 1 holds literally and #1173 can be closed.

claude-cogni-work:
Refs #1173. Closes #1180.

## Summary

- Adds a public `read_frontmatter(path) -> (fm, exc)` beside `parse_frontmatter` in `scripts/validate-entities.py`. It owns the open/read/parse sequence and returns the exception instead of raising it, so no caller can be surprised by `OSError` or `UnicodeDecodeError`.
- Routes the three read sites named in the issue through it: `validate_file` and `_ref_errors` (both `validate-entities.py`) and `_read_entities` (`render-dashboard.py`).
- Fixes the actual defect: `staffing-score.py` drops its private `_read_frontmatter` (which guarded `OSError` only) for the shared helper. A latin-1 entity record is now skipped like any other unreadable one instead of raising a bare traceback with **no JSON on stdout at all** — `staffing-score.py` has no top-level `except Exception` envelope guard, unlike its two siblings, so the escape was total.
- Adds `tests/test_staffing_score.sh`, the plugin's first suite for this script.

### Behaviour preserved

- `validate_file` keeps two distinct messages — `cannot read file:` vs `cannot decode file as UTF-8 — re-save it as UTF-8:` — now selected by `isinstance(exc, UnicodeDecodeError)` rather than by two `except` clauses. Text is byte-identical, em dash (U+2014) included. The decode branch must come **first**: the branch below it is the catch-all, so ordering is what keeps the two messages distinct now that they are ordinary `if`s.
- `render-dashboard.py` keeps its `cannot read %s: %s` warning, its `no frontmatter in %s — skipped` warning, and its skip-and-continue degrade. `test-render-dashboard.sh` Fixture 6 is the proof.
- `_ref_errors` keeps silently continuing on both failure modes.
- `parse_frontmatter` stays public and unchanged, which is what keeps `register-entity.py` unaffected.

## On the acceptance criteria

AC 4 asks that all four existing suites stay green. They do — but that bar is **necessary and not sufficient**, because none of the four invokes `staffing-score.py` at all (`grep -l staffing.score cogni-projects/tests/*` returns nothing on `main`). A green four-suite run could not have demonstrated AC 2. That is why a fifth suite is in this diff rather than left for a reviewer to notice.

The new suite was verified to actually catch the bug, not merely to pass: run against `origin/main`'s pre-fix scripts it fails on exactly the reported symptom —

```
FAIL 2a latin-1 record does not fail the run: want exit 0, got 1
FAIL 2b envelope still returned: expr false: d['success'] is True | json=
FAIL 2e no traceback on stderr: traceback present: Traceback (most recent call last):
```

— and passes after. Its fixture C pins that usage and preflight errors still report through the envelope, so AC 2 cannot be satisfied by swallowing every failure.

## Scope decisions

**Included:** the three read sites the issue enumerates, plus `_ref_errors` in `validate-entities.py` — a fourth copy of the same sequence, but it lives in a file this diff already opens, so folding it in costs no extra blast radius and makes AC 1 hold across all three named files. `render-dashboard.py`'s module docstring is corrected in the same file the edit already touches: line 148 was its only `parse_frontmatter` call, so the docstring would otherwise name a function the file never calls. `CLAUDE.md` gains the new suite in its test list; the suites have no CI wiring and are run manually from that list, so an undocumented fifth suite would never actually run.

**Deferred — why this is `Refs`, not `Closes`:** `register-entity.py:121-125` is a fourth *script* carrying the same `OSError`-only guard. It is **not a live defect**: line 113 runs `_ve.validate_file()`, which does guard `UnicodeDecodeError` and returns an error, and line 114 short-circuits to `_fail`, so an undecodable file can never reach line 121. It is pure textual duplication. Because a strict reading of AC 1 ("the only place the sequence lives") could still count that site, this PR does not close the issue.

**No version bump** — the post-merge workflow owns it.

## Test plan

All five suites run green on this branch:

- [x] `bash cogni-projects/tests/test_staffing_score.sh` — 20 assertions, incl. the latin-1 regression and the deterministic candidate ordering
- [x] `bash cogni-projects/tests/test-render-dashboard.sh` — Fixture 6 (6a / 6c / 6f / 6g) still green, which is AC 3
- [x] `bash cogni-projects/tests/test_validate_entities_refs.sh`
- [x] `bash cogni-projects/tests/test_register_entity.sh`
- [x] `bash cogni-projects/tests/test_portfolio_init.sh`
- [x] Manual: a latin-1 entity file in a scratch portfolio → `staffing-score.py` emits one JSON line, exit 0, **0 bytes on stderr**
- [x] Manual: `validate-entities.py` on the same dir still reports `cannot decode file as UTF-8 — re-save it as UTF-8`, distinct from the `cannot read file` wording
- [x] Manual: `render-dashboard.py` on the same dir still emits `cannot read consultants/cesar.md: ...` and `partial: true`
- [x] `scripts/check-breadcrumbs.py` clean
- [x] `git diff --stat` shows no change to `.claude-plugin/plugin.json` or `.claude-plugin/marketplace.json`

## Review notes

A four-angle quality pass ran before the commit. Applied: dropped a redundant `isinstance(fm, dict)` conjunct in `staffing-score.py` (defensive residue from the deleted private wrapper — `parse_frontmatter` returns dict-or-None, so a truthy `fm` is always a dict) and the equivalent redundant disjunct in `_ref_errors`; trimmed comments that restated the helper's docstring or narrated the diff rather than the code; earned the second consultant fixture with a deterministic-ordering assertion; and converted three discarded stderr captures in fixture C into `assert_no_traceback` coverage.

Two findings were **deliberately not applied**, both because they widen scope past a read-guard fix:

1. `_load_read_frontmatter` narrows the loaded module to a single callable and threads it as a parameter through three signatures, where both sibling scripts bind the whole module as `_ve` and call through it. Removing that plumbing belongs with #1181, which already owns the loader blocks.
2. The new suite re-implements ~50 lines of harness (`pass`/`fail`/`assert_json`/`assert_no_traceback`/`seed_portfolio`) byte-identically from `test-render-dashboard.sh`; those helpers are now defined in three of the five suites. Inlining is the current house convention here, and `cogni-knowledge/tests/fixtures/test_helpers.sh` is the in-repo precedent for outgrowing it. Filed as #1182 rather than touching three green suites in this PR.

One thing worth a reviewer's eye: `run_score` in the new suite redirects stdout to a file rather than piping into `tail`, because after a `$(... | tail -n 1)` substitution `$?` is *tail's* status and every exit-code assertion passes vacuously. `test-render-dashboard.sh`'s `run()` still has that shape, which is why that suite asserts no exit codes. Not fixed here — it is the same three-suites-touched problem as #1182.

**On `Refs` vs `Closes` for #1173 — a merger call.** This PR kept `Refs #1173` because AC 1 ("a single `read_frontmatter(path)` helper is the only place the open/read/parse/guard sequence lives") was not strictly satisfied while `register-entity.py` held a fourth copy. Revision `4701fc9f` converts that copy, so on my reading all four of #1173's acceptance criteria are now met and the linking line could be upgraded to `Closes #1173`. I have deliberately **not** upgraded it: #1181 and #1182 are deferred follow-ups that came out of this work, and whether the parent stays open to umbrella them is a scope judgment I would rather the merger make than assume. If the answer is "close it", the edit is one word in the first line of this body.

## Follow-up issues

- ~~#1180~~ — route `register-entity.py`'s fourth read-frontmatter site through the shared helper — **now shipped in this PR** (revision `4701fc9f`, answering the merger verdict's item-8 finding), so it closes on merge
- #1186 — give `staffing-score.py` the sibling module-level catch-all envelope (deferred from the verdict's second, non-gating finding)
- #1181 — collapse the three duplicated importlib validator-loader blocks
- #1182 — extract a shared `tests/` assertion harness

### DIFF
--- a/cogni-projects/CLAUDE.md
+++ b/cogni-projects/CLAUDE.md
@@ -28,6 +28,7 @@ cogni-projects/
 │   ├── test_portfolio_init.sh          Portfolio scaffolder idempotency + no-debris suite
 │   ├── test_register_entity.sh         Atomic-write + idempotency regression suite
 │   ├── test_validate_entities_refs.sh  Assignment ref-integrity + single-file regression suite
+│   ├── test_staffing_score.sh          Staffing scorer envelope + non-UTF-8 record regression suite
 │   └── test-render-dashboard.sh        Dashboard render regression test (stdlib-only)
 └── references/
     └── data-model.md                   Consultant / project / assignment entity schemas
@@ -39,6 +40,7 @@ Run the tests directly — bash + python3, no pytest or pip:
 bash cogni-projects/tests/test_portfolio_init.sh
 bash cogni-projects/tests/test_register_entity.sh
 bash cogni-projects/tests/test_validate_entities_refs.sh
+bash cogni-projects/tests/test_staffing_score.sh
 bash cogni-projects/tests/test-render-dashboard.sh
 ```
 
--- a/cogni-projects/scripts/register-entity.py
+++ b/cogni-projects/scripts/register-entity.py
@@ -118,10 +118,12 @@ def main(argv):
             code=1,
         )
 
-    try:
-        with open(entity_file, "r", encoding="utf-8") as f:
-            fm = _ve.parse_frontmatter(f.read())
-    except OSError as exc:
+    # Route through the shared reader, not a local open(): it guards
+    # UnicodeDecodeError too, so a latin-1 record reports "cannot read entity
+    # file" rather than the __main__ catch-all's generic message — and the
+    # validate_file gate above rejecting it first stops being load-bearing.
+    fm, exc = _ve.read_frontmatter(entity_file)
+    if exc is not None:
         return _fail("cannot read entity file: %s" % exc)
 
     # `type` is required for every type and the validator errors when it
--- a/cogni-projects/scripts/render-dashboard.py
+++ b/cogni-projects/scripts/render-dashboard.py
@@ -14,8 +14,8 @@ malformed entity field yields a partial snapshot with a surfaced warning, never
 a hard failure, because a partial snapshot is more useful than no snapshot to a
 partner reviewing a portfolio that is still being authored.
 
-Stdlib-only (no PyYAML). Reuses validate-entities.py's parse_frontmatter and
-_entity_files by file-path load rather than re-implementing a parser or a
+Stdlib-only (no PyYAML). Reuses validate-entities.py's read_frontmatter and
+_entity_files by file-path load rather than re-implementing a reader or a
 directory walk — the same idiom register-entity.py uses.
 
 Usage:
@@ -143,15 +143,10 @@ def _read_entities(portfolio_dir):
     warnings = []
     for path in _ve._entity_files(portfolio_dir):
         rel = os.path.relpath(path, portfolio_dir)
-        try:
-            with open(path, "r", encoding="utf-8") as f:
-                fm = _ve.parse_frontmatter(f.read())
-        # UnicodeDecodeError is a ValueError, not an OSError, so it escaped the
-        # original guard: one non-UTF-8 entity file must degrade to a single
-        # warning, not abort the whole render. Named explicitly rather than
-        # widening to ValueError, so a future parser fault still surfaces
-        # instead of being silently reported as an unreadable file.
-        except (OSError, UnicodeDecodeError) as exc:
+        # One unreadable entity file must degrade to a single warning, not abort
+        # the whole render.
+        fm, exc = _ve.read_frontmatter(path)
+        if exc is not None:
             warnings.append("cannot read %s: %s" % (rel, exc))
             continue
         if not fm:
--- a/cogni-projects/scripts/staffing-score.py
+++ b/cogni-projects/scripts/staffing-score.py
@@ -30,9 +30,9 @@ ranking. Output is deterministic for identical inputs — scores are rounded to 
 fixed precision, projects order by strategic impact then slug, candidate ties
 break on the consultant slug, and no wall-clock value enters the result.
 
-Reads entity frontmatter with the same stdlib parser the validator uses
-(`validate-entities.py:parse_frontmatter`, loaded by file location as
-`register-entity.py` does) — no duplicated parser, no PyYAML.
+Reads entity frontmatter with the same stdlib reader the validator uses
+(`validate-entities.py:read_frontmatter`, loaded by file location as
+`register-entity.py` does) — no duplicated reader, no PyYAML.
 
 Usage:
   python3 staffing-score.py <portfolio-dir>
@@ -97,12 +97,12 @@ def _fail(message, code):
     return code
 
 
-def _load_parse_frontmatter():
-    """Load parse_frontmatter from the sibling validate-entities.py by file path.
+def _load_read_frontmatter():
+    """Load read_frontmatter from the sibling validate-entities.py by file path.
 
     validate-entities.py is not an importable module name (hyphens), so — exactly
     as register-entity.py does — load it by location to reuse the one canonical
-    frontmatter parser rather than duplicating the schema-shaped parsing rules.
+    frontmatter reader rather than duplicating the open/read/parse/guard sequence.
     Returns the callable, or None if the validator module cannot be loaded.
     """
     v_path = os.path.join(
@@ -113,18 +113,7 @@ def _load_parse_frontmatter():
         return None
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
-    return getattr(module, "parse_frontmatter", None)
-
-
-def _read_frontmatter(parse_frontmatter, path):
-    """Return the parsed frontmatter dict for an entity file, or {} on any miss."""
-    try:
-        with open(path, "r", encoding="utf-8") as f:
-            text = f.read()
-    except OSError:
-        return {}
-    fm = parse_frontmatter(text)
-    return fm if isinstance(fm, dict) else {}
+    return getattr(module, "read_frontmatter", None)
 
 
 def _overlap_fraction(cf, cu, ps, pe):
@@ -276,7 +265,7 @@ def _round(x):
     return round(x, SCORE_PRECISION)
 
 
-def _load_entities(parse_frontmatter, portfolio_dir, manifest, kind, subdir):
+def _load_entities(read_frontmatter, portfolio_dir, manifest, kind, subdir):
     """Read every entity of one kind into a list of frontmatter dicts.
 
     Prefers the manifest's summary refs (the index later skills scan) for the
@@ -300,13 +289,15 @@ def _load_entities(parse_frontmatter, portfolio_dir, manifest, kind, subdir):
                     files.append(path)
     entities = []
     for path in files:
-        fm = _read_frontmatter(parse_frontmatter, path)
+        # An unreadable or frontmatter-less record is dropped silently — this
+        # script has no warnings channel to report it through.
+        fm, _exc = read_frontmatter(path)
         if fm:
             entities.append(fm)
     return entities
 
 
-def score_portfolio(portfolio_dir, parse_frontmatter):
+def score_portfolio(portfolio_dir, read_frontmatter):
     """Build the ranked-shortlist data structure for a portfolio directory."""
     manifest_path = os.path.join(portfolio_dir, "projects-portfolio.json")
     try:
@@ -317,10 +308,10 @@ def score_portfolio(portfolio_dir, parse_frontmatter):
                            % (manifest_path, exc))
 
     consultants = _load_entities(
-        parse_frontmatter, portfolio_dir, manifest, "consultants", "consultants"
+        read_frontmatter, portfolio_dir, manifest, "consultants", "consultants"
     )
     projects = _load_entities(
-        parse_frontmatter, portfolio_dir, manifest, "projects", "projects"
+        read_frontmatter, portfolio_dir, manifest, "projects", "projects"
     )
 
     # A closed project is delivered or lost — its roles are not open work to
@@ -420,17 +411,17 @@ def main(argv):
             "(run /cogni-projects:projects-setup first)" % portfolio_dir, 1
         )
 
-    parse_frontmatter = _load_parse_frontmatter()
-    if parse_frontmatter is None:
+    read_frontmatter = _load_read_frontmatter()
+    if read_frontmatter is None:
         # A broken/partial install (validator missing), not a caller usage
         # error — so exit 1 (domain failure), reserving exit 2 for bad args.
         return _fail(
-            "cannot load validate-entities.py:parse_frontmatter (expected sibling "
+            "cannot load validate-entities.py:read_frontmatter (expected sibling "
             "of this script) — the cogni-projects install looks broken", 1
         )
 
     try:
-        data = score_portfolio(portfolio_dir, parse_frontmatter)
+        data = score_portfolio(portfolio_dir, read_frontmatter)
     except RuntimeError as exc:
         return _fail(str(exc), 1)
 
--- a/cogni-projects/scripts/validate-entities.py
+++ b/cogni-projects/scripts/validate-entities.py
@@ -202,6 +202,34 @@ def _split_inline_list(inner):
     return [p.strip() for p in parts if p.strip() != ""]
 
 
+def read_frontmatter(path):
+    """Read an entity file and parse its frontmatter. Returns ``(fm, exc)``.
+
+    The one place the open/read/parse sequence lives, so the guard cannot drift
+    between the scripts that need it. It never raises for a read or decode
+    failure: ``exc`` carries the exception instead, so a caller reports it
+    however its own output contract requires — and one that needs the
+    read-vs-decode distinction tests ``isinstance(exc, UnicodeDecodeError)``.
+
+    ``exc`` is None whenever the file was read; ``fm`` is then whatever
+    ``parse_frontmatter`` returned — a dict, or None if the file carries no
+    ``---`` fence. On failure the pair is ``(None, exc)``.
+
+    Entity records are UTF-8 by contract, but a record saved as latin-1 (a DACH
+    name is the usual way in) would otherwise raise past the envelope every path
+    returns. UnicodeDecodeError is a ValueError, not an OSError, so it escapes a
+    bare OSError guard — it is named explicitly rather than widening to
+    ValueError, so a future parser fault still surfaces instead of being
+    silently reported as an unreadable file.
+    """
+    try:
+        with open(path, "r", encoding="utf-8") as f:
+            text = f.read()
+    except (OSError, UnicodeDecodeError) as exc:
+        return None, exc
+    return parse_frontmatter(text), None
+
+
 def _entity_files(path):
     """Expand a path to the entity .md files it covers.
 
@@ -241,11 +269,7 @@ def _ref_errors(files):
         etype = SUBDIR_TO_TYPE.get(os.path.basename(os.path.dirname(path)))
         if etype is None:
             continue
-        try:
-            with open(path, "r", encoding="utf-8") as f:
-                fm = parse_frontmatter(f.read())
-        except (OSError, UnicodeDecodeError):
-            continue
+        fm, _exc = read_frontmatter(path)
         if fm is None:
             # Unreadable or frontmatter-less records are already reported by
             # validate_file; naming the same defect a second way here would
@@ -295,20 +319,16 @@ def validate_file(filepath):
             {"entity": entity, "file": rel, "field": field, "message": message}
         )
 
-    try:
-        with open(filepath, "r", encoding="utf-8") as f:
-            text = f.read()
-    except OSError as exc:
-        err("<file>", "cannot read file: %s" % exc)
-        return errors, warnings
-    except UnicodeDecodeError as exc:
-        # Entity records are UTF-8 by contract, but a record saved as latin-1
-        # (a DACH name is the usual way in) would otherwise raise past the
-        # envelope every other path returns.
+    fm, exc = read_frontmatter(filepath)
+    # Decode first: the branch below it is the catch-all, so order is what keeps
+    # the two messages distinct.
+    if isinstance(exc, UnicodeDecodeError):
         err("<file>", "cannot decode file as UTF-8 — re-save it as UTF-8: %s" % exc)
         return errors, warnings
+    if exc is not None:
+        err("<file>", "cannot read file: %s" % exc)
+        return errors, warnings
 
-    fm = parse_frontmatter(text)
     if fm is None:
         err("<frontmatter>", "no YAML frontmatter block found (expected leading --- ... --- fence)")
         return errors, warnings
--- a/cogni-projects/tests/test_staffing_score.sh
+++ b/cogni-projects/tests/test_staffing_score.sh
@@ -1,0 +1,224 @@
+#!/usr/bin/env bash
+# Test staffing-score.py — the ranked-shortlist scorer.
+#
+# Covers:
+#   A. a clean portfolio scores, pinning the counts B compares against,
+#   B. an entity record that is not valid UTF-8 is skipped like any other
+#      unreadable one — the run still returns its {success, data, error}
+#      envelope rather than a bare traceback,
+#   C. usage and preflight errors still report through that envelope, so B
+#      cannot be satisfied by swallowing every failure.
+#
+# stdlib-only (bash + python3, no pytest/pip), matching the house convention.
+#
+# Usage: bash cogni-projects/tests/test_staffing_score.sh
+# Exits non-zero on any assertion failure.
+
+set -u  # NOT -e: a failing assertion must not abort the per-fixture counter.
+
+TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
+PLUGIN_DIR="$(cd "$TESTS_DIR/.." && pwd)"
+SCRIPT="$PLUGIN_DIR/scripts/staffing-score.py"
+
+if [ ! -f "$SCRIPT" ]; then
+  echo "FAIL: staffing-score.py not found at $SCRIPT" >&2
+  exit 1
+fi
+
+TMPROOT="$(mktemp -d)"
+trap 'rm -rf "$TMPROOT"' EXIT
+
+failures=0
+pass() { printf 'OK   %s\n' "$1"; }
+fail() { printf 'FAIL %s: %s\n' "$1" "$2" >&2; failures=$((failures + 1)); }
+
+# assert_json <label> <python-bool-expr over `d`> — pipes the last stdout line
+# (the envelope) into python3 and checks the expression is truthy.
+assert_json() {
+  local label="$1" expr="$2"
+  if printf '%s' "$LAST_JSON" | python3 -c "import json,sys
+d = json.loads(sys.stdin.read())
+sys.exit(0 if ($expr) else 1)" 2>/dev/null; then
+    pass "$label"
+  else
+    fail "$label" "expr false: $expr | json=$LAST_JSON"
+  fi
+}
+
+# The crash signature belongs in one place — widening it later should be one
+# edit. This is the assertion the whole suite exists for.
+assert_no_traceback() {
+  local label="$1" file="$2"
+  if grep -qF 'Traceback' "$file"; then
+    fail "$label" "traceback present: $(head -3 "$file")"
+  else
+    pass "$label"
+  fi
+}
+
+assert_exit() {
+  local label="$1" want="$2" got="$3"
+  if [ "$want" = "$got" ]; then
+    pass "$label"
+  else
+    fail "$label" "want exit $want, got $got"
+  fi
+}
+
+seed_portfolio() {
+  # $1 = portfolio dir. Seeds a manifest + entity subdirs. The manifest's ref
+  # lists stay empty on purpose: _load_entities then falls back to scanning the
+  # subdirectories, which is the path a hand-authored record takes.
+  local pf="$1"
+  mkdir -p "$pf"/{consultants,projects,assignments,.metadata}
+  cat > "$pf/projects-portfolio.json" <<'EOF'
+{"slug":"test","name":"Test Portfolio","language":"en","consultants":[],"projects":[],"assignments":[],"created":"2026-01-01","updated":"2026-01-01"}
+EOF
+}
+
+write_entity() { # $1 = path; body heredoc'd by the caller on stdin
+  cat > "$1"
+}
+
+# run_score <stderr-file> [args...] — invoke the scorer, setting LAST_JSON to the
+# last stdout line (the envelope) and RUN_CODE to the script's own exit status.
+# Redirecting to a file rather than piping into `tail` is load-bearing: after a
+# `$(... | tail -n 1)` substitution, `$?` is tail's status, which is always 0, so
+# every exit-code assertion would pass vacuously.
+run_score() {
+  local errfile="$1"; shift
+  local outfile="$TMPROOT/stdout.txt"
+  python3 "$SCRIPT" "$@" >"$outfile" 2>"$errfile"
+  RUN_CODE=$?
+  LAST_JSON="$(tail -n 1 "$outfile")"
+}
+
+# Seed two valid consultants and one active project with an open role. Shared by
+# both fixtures so the only difference between them is the latin-1 file.
+seed_entities() {
+  local pf="$1"
+  write_entity "$pf/consultants/anna.md" <<'EOF'
+---
+type: consultant
+slug: anna
+name: Anna Schmidt
+seniority: senior
+skills: [cloud, architecture]
+available_from: 2026-01-01
+available_until: 2026-12-31
+---
+# anna
+EOF
+  write_entity "$pf/consultants/bruno.md" <<'EOF'
+---
+type: consultant
+slug: bruno
+name: Bruno Weber
+seniority: consultant
+skills: [cloud, data]
+available_from: 2026-01-01
+available_until: 2026-12-31
+---
+# bruno
+EOF
+  write_entity "$pf/projects/migration.md" <<'EOF'
+---
+type: project
+slug: migration
+name: Cloud Migration
+client: GoodCo
+strategic_impact: 4
+status: active
+start_date: 2026-02-01
+end_date: 2026-06-30
+open_roles: [cloud]
+---
+# migration
+EOF
+}
+
+# ---------------------------------------------------------------------------
+# Fixture A — control. A clean, fully decodable portfolio scores normally. This
+# is what pins the counts Fixture B compares against: without it, an assertion
+# that the latin-1 run still finds two consultants could pass for the wrong
+# reason (e.g. the scorer silently finding none at all).
+# ---------------------------------------------------------------------------
+PFA="$TMPROOT/clean"
+seed_portfolio "$PFA"
+seed_entities "$PFA"
+
+STDERRA="$TMPROOT/stderr-a.txt"
+run_score "$STDERRA" "$PFA"
+
+assert_exit  "1a clean portfolio exits 0"        0 "$RUN_CODE"
+assert_json  "1b clean portfolio succeeds"       "d['success'] is True"
+assert_json  "1c both consultants loaded"        "d['data']['consultant_count'] == 2"
+assert_json  "1d project loaded"                 "d['data']['project_count'] == 1"
+assert_json  "1e both consultants ranked"        "d['data']['ranked_candidate_count'] == 2"
+# Anna and Bruno share the 'cloud' skill and differ only in seniority, so the
+# ranking is what tells them apart — without this the second consultant is a
+# fixture nothing distinguishes, and the documented deterministic ordering goes
+# unpinned.
+assert_json  "1f candidates ranked deterministically" \
+  "[c['consultant'] for c in d['data']['projects'][0]['open_roles'][0]['candidates']] == ['anna', 'bruno']"
+assert_no_traceback "1g no traceback on stderr"  "$STDERRA"
+
+# ---------------------------------------------------------------------------
+# Fixture B — the regression. Same portfolio plus one consultant record saved as
+# latin-1, which is invalid UTF-8. Before the shared read_frontmatter helper,
+# this raised UnicodeDecodeError out of the scorer: no JSON on stdout at all,
+# just a traceback on stderr, which reads to a caller as neither success nor a
+# reportable failure. The record must now be skipped like any other unreadable
+# one while the rest of the portfolio still scores.
+# ---------------------------------------------------------------------------
+PFB="$TMPROOT/latin1"
+seed_portfolio "$PFB"
+seed_entities "$PFB"
+
+# Written via python3 in wb mode because bash printf parses a leading '---' as
+# options; \xe9 is 'é' in latin-1 and an invalid UTF-8 start byte here.
+python3 -c "
+import sys
+open(sys.argv[1], 'wb').write(
+    '---\ntype: consultant\nslug: cesar\nname: C\xe9sar Rossi\nseniority: senior\nskills: [cloud]\navailable_from: 2026-01-01\navailable_until: 2026-12-31\n---\n'.encode('latin-1')
+)" "$PFB/consultants/cesar.md"
+
+STDERRB="$TMPROOT/stderr-b.txt"
+run_score "$STDERRB" "$PFB"
+
+assert_exit  "2a latin-1 record does not fail the run"  0 "$RUN_CODE"
+assert_json  "2b envelope still returned"               "d['success'] is True"
+# The undecodable consultant is dropped; the two decodable ones still load.
+assert_json  "2c undecodable consultant skipped"        "d['data']['consultant_count'] == 2"
+assert_json  "2d project still scored"                  "d['data']['project_count'] == 1"
+# The assertion this suite exists for: the guard gap surfaced as a raw traceback.
+assert_no_traceback "2e no traceback on stderr"         "$STDERRB"
+
+# ---------------------------------------------------------------------------
+# Fixture C — usage and preflight errors still report through the envelope, so
+# the fix above cannot have been achieved by swallowing every failure.
+# ---------------------------------------------------------------------------
+STDERRC="$TMPROOT/stderr-c.txt"
+
+run_score "$STDERRC"
+assert_exit  "3a no-args exits 2"                2 "$RUN_CODE"
+assert_json  "3b no-args reports usage"          "d['success'] is False and 'usage' in d['error']"
+assert_no_traceback "3c no-args stderr is clean" "$STDERRC"
+
+run_score "$STDERRC" "$TMPROOT/does-not-exist"
+assert_exit  "3d missing dir exits 1"            1 "$RUN_CODE"
+assert_json  "3e missing dir reports envelope"   "d['success'] is False and d['error'] != ''"
+assert_no_traceback "3f missing dir stderr is clean" "$STDERRC"
+
+# A directory that exists but is not a portfolio.
+mkdir -p "$TMPROOT/not-a-portfolio"
+run_score "$STDERRC" "$TMPROOT/not-a-portfolio"
+assert_exit  "3g non-portfolio dir exits 1"      1 "$RUN_CODE"
+assert_json  "3h non-portfolio reports envelope" "d['success'] is False and 'projects-portfolio.json' in d['error']"
+assert_no_traceback "3i non-portfolio stderr is clean" "$STDERRC"
+
+if [ "$failures" -gt 0 ]; then
+  printf '\n%d assertion(s) failed\n' "$failures" >&2
+  exit 1
+fi
+printf '\nAll staffing-score.py assertions passed\n'


claude-cogni-work:

#1173 end-to-end to a draft PR. Handing off to the merger identity for
never reviews its own work."

`svc:needs-review` applied to this PR; #1173 marked `svc:resolved`.

sdh07:

_The merger formed this expectation from issue #1173 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds a public module-level read_frontmatter(path) to cogni-projects/scripts/validate-entities.py next to parse_frontmatter (base lines 130-182) that does open(encoding='utf-8') -> read -> parse_frontmatter guarded by (OSError, UnicodeDecodeError), and returns a shape that lets callers tell three outcomes apart: read/decode failure with the exception still reachable, read-OK-but-no-fence (parse_frontmatter returns None), and a parsed dict. It then converts the four in-scope copies: validate_file (base validate-entities.py:298-311, which must keep its two distinct messages 'cannot read file: %s' at :302 and 'cannot decode file as UTF-8 - re-save it as UTF-8: %s' at :305-308, and its separate <frontmatter> error at :313), _ref_errors (base :244-253), render-dashboard.py:_read_entities (base :146-158, whose 'cannot read %s: %s' warning with the portfolio-relative path and separate 'no frontmatter in %s - skipped' warning are pinned by tests/test-render-dashboard.sh:343 and :349-350), and staffing-score.py:_read_frontmatter (base :119-127, the OSError-only copy). Because staffing-score reaches the validator through _load_parse_frontmatter (base :100-116) and threads the callable into score_portfolio/_load_entities (base :279-306, :309, :423-433), the loader and those signatures must be widened to surface read_frontmatter while preserving the exit-1 broken-install envelope whose 'validate-entities.py' wording skills/projects-staff/SKILL.md:85 instructs on. Finally a new bash suite cogni-projects/tests/test_staffing_score.sh (the house shape of tests/test-render-dashboard.sh:1-15, latin-1 fixture idiom at :331-335) proves the scorer skips an undecodable record rather than raising, and cogni-projects/CLAUDE.md registers it in both the tests/ tree listing (base :27-31) and the run-commands block (base :36-43).

**Acceptance checklist:**
- [ ] read_frontmatter is defined as a public (no leading underscore) module-level function in cogni-projects/scripts/validate-entities.py adjacent to parse_frontmatter (base lines 130-182); check by grepping that file for 'def read_frontmatter'.
- [ ] cogni-projects/tests/test_staffing_score.sh exists and follows the house suite shape - '#!/usr/bin/env bash' shebang, 'set -u' (not -e), a mktemp -d TMPROOT with a trap-EXIT cleanup, and a non-zero exit on any assertion failure - matching tests/test-render-dashboard.sh:1-31 and :711-717.
- [ ] cogni-projects/CLAUDE.md registers the new suite in BOTH surfaces: the tests/ tree listing (base lines 27-31, four entries today) and the runnable bash block (base lines 36-43, four commands today); check that the string 'test_staffing_score.sh' appears at least twice in CLAUDE.md.
- [ ] No plugin version line is touched: neither cogni-projects/.claude-plugin/plugin.json nor the repo-root .claude-plugin/marketplace.json appears in the diff, so scripts/check-version-bump.py stays clean (CONTRIBUTING.md:82-95 forbids authoring-time bumps).
- [ ] staffing-score.py holds no entity open/read/parse sequence of its own: the base _read_frontmatter body (lines 119-127, 'except OSError' only) is deleted or reduced to a delegation, and the only remaining open( ) call in that file is the manifest read (base line 313).
- [ ] render-dashboard.py:_read_entities routes through the already-loaded validator module (_ve.read_frontmatter) instead of its own open( )/try/except (base lines 146-155); check that no bare parse_frontmatter call remains in that file and that the surviving open( ) calls are only the theme file (base :110), the manifest (base :558) and the dashboard write (base :575).
- [ ] Both in-file copies inside validate-entities.py are converted: validate_file (base :298-311) and _ref_errors (base :244-253) each call read_frontmatter rather than pairing their own open( ) with parse_frontmatter; check that exactly one open(...)+parse_frontmatter pair remains in the file, inside the helper.
- [ ] register-entity.py's fourth copy (base lines 121-125: open + _ve.parse_frontmatter guarded by OSError only) is either converted to the shared helper or its retention is justified in an in-code comment naming the reason it is unreachable for the decode case (the _ve.validate_file gate at base line 113 fails first); check by grepping register-entity.py lines ~110-130 for 'read_frontmatter' or such a comment. Leaving it silently untouched contradicts the acceptance criterion that the helper is the only place the sequence lives.
- [ ] validate-entities.py still distinguishes a decode failure from a read failure: grep the file for both 'cannot read file:' (base :302) and 'cannot decode file as UTF-8' together with the 're-save it as UTF-8' remediation (base :308), both still reported under the '<file>' field. A helper that collapses the two into one message is a user-facing regression, not a refactor.
- [ ] The helper reports 'read OK but no frontmatter fence' as a state distinct from both failure and an empty dict, so validate_file still emits its '<frontmatter>' error 'no YAML frontmatter block found' (base :313) exactly once for a fence-less file; pinned by tests/test_validate_entities_refs.sh:233-239, which asserts a single <frontmatter>-field error for the orphan.
- [ ] render-dashboard.py keeps its unreadable-record warning verbatim in shape - 'cannot read %s: %s' with the portfolio-relative path (base :155) - plus the separate 'no frontmatter in %s - skipped' warning (base :158), and still continues past the record. Directly pinned by tests/test-render-dashboard.sh:349-350 (matches on 'cannot read' AND 'latin1') and :343 (projects == 4); substituting a helper-supplied 'cannot decode' string here breaks 6f.
- [ ] The latin-1/DACH-name rationale survives the move rather than being dropped: read_frontmatter carries a docstring or comment naming a non-UTF-8 (latin-1) entity record as the case the UnicodeDecodeError guard exists for (present today at validate-entities.py:305-307 and render-dashboard.py:149-153), and the module docstrings that name what they borrow (staffing-score.py:33-35, render-dashboard.py:17-19, both saying parse_frontmatter) are updated to name read_frontmatter where that is now the call.
- [ ] The new suite proves the reported defect by acting, not by assertion of success alone: it writes a latin-1-encoded entity file (the .encode('latin-1') idiom at tests/test-render-dashboard.sh:331-335), invokes staffing-score.py with stderr captured to a file, and asserts all four of - the last stdout line parses as {success,data,error} with success True; exit status 0; stderr contains no 'Traceback'; and a count assertion showing the remaining valid records are still scored (e.g. consultant_count / project_count / ranked_candidate_count), so the undecodable record is skipped rather than the run aborted.
- [ ] The converted staffing-score call site never appends a non-dict entity: _load_entities (base :301-306) still guards with a truthiness/dict check before appending, so a failed or fence-less read cannot put None into consultants/projects and raise AttributeError later in _availability_score/_profile_fit_score.
- [ ] staffing-score.py's broken-install path is preserved end to end: its loader surfaces read_frontmatter (not just parse_frontmatter), and when the validator module cannot be loaded main still returns exit code 1 (not 2) with a failure envelope whose 'error' string names 'validate-entities.py' - skills/projects-staff/SKILL.md:85 tells the agent to read exactly that as a broken install rather than a portfolio-data problem.
- [ ] Scoring semantics are untouched by the refactor: the weight constants (W_AVAILABILITY 0.55, W_PROFILE_FIT 0.45, W_OVERLAP 0.60, W_HEADROOM 0.40, W_SKILL_MATCH 0.75, W_SENIORITY 0.25, TOKEN_OVERLAP_DISCOUNT 0.5, SCORE_PRECISION 3) and the score_portfolio envelope keys (weights, project_order, consultant_count, project_count, ranked_candidate_count, projects) are unchanged in the diff.
- [ ] All five cogni-projects/tests/*.sh suites run green with recorded output - test_portfolio_init.sh, test_register_entity.sh, test_validate_entities_refs.sh, test-render-dashboard.sh, test_staffing_score.sh. test-render-dashboard.sh and test_validate_entities_refs.sh are the two that must be shown green explicitly: they pin the exact warning text and the <frontmatter> error the extraction most threatens.

**Risk watchlist:**
- The helper returns only (fm) or (fm, message) and loses the OSError-vs-UnicodeDecodeError distinction, so validate-entities.py silently drops the actionable 're-save it as UTF-8' remediation and reports every miss as 'cannot read'.
- The helper formats its own message and render-dashboard adopts it, changing the warning from 'cannot read <rel>: ...' to 'cannot decode ...' - which fails tests/test-render-dashboard.sh:349 (6f) and quietly breaks the AC that render-dashboard keeps its current warning behaviour.
- The helper collapses 'no frontmatter fence' into the failure branch or returns {} for it, so validate_file stops emitting the '<frontmatter>' error and test_validate_entities_refs.sh:237 (which asserts exactly one such error) goes red - or worse, goes green while _ref_errors starts reporting the same defect twice.
- The helper is made private (_read_frontmatter) and reached across module boundaries anyway, or is placed in staffing-score.py instead of validate-entities.py, inverting the existing dependency arrow (both siblings load validate-entities.py by file location; nothing loads staffing-score.py).
- A half-measure: the UnicodeDecodeError is simply added to staffing-score.py's own except clause with no extraction, leaving four copies of the sequence and failing the AC that a single helper is the only place it lives.
- The staffing-score loader is widened to return the whole module or a tuple but the exit-1 broken-install envelope naming 'validate-entities.py' is reworded or downgraded to exit 2, silently breaking the signal skills/projects-staff/SKILL.md:85 instructs the narrating agent to key on.
- A new test that only asserts success is True on a latin-1 fixture would pass against a scorer that swallowed the whole portfolio (zero consultants, zero projects scored) - the count assertion is what makes the test discriminating.
- register-entity.py keeps a fourth OSError-only copy (base :121-125) that no item in the issue's own enumeration mentions; it is currently unreachable for the decode class only because _ve.validate_file gates first (base :113), so any future reordering re-opens the same defect there.
- Judgment-only, not gated: staffing-score.py is the one script of the three with no module-level try/except catch-all (contrast validate-entities.py:499-511 and render-dashboard.py:600-608). This fix closes the specific non-UTF-8 path but leaves the class open. A human should decide whether adding the sibling catch-all belongs in this PR or a follow-up; it is outside the issue's acceptance bar either way.
- Scope creep: an extraction PR that also normalizes text fields, unifies the three case-sensitive 'closed' comparisons render-dashboard.py:225-240 explicitly defers, or retunes scoring weights - all of which the base code documents as deliberately deferred decisions.

- The plan-time acceptance criteria list 4 bullets, but a base-grounded regeneration derives ~17 gating items. The AC names none of: registering the new suite in cogni-projects/CLAUDE.md (both the tree listing and the run-commands block, the plugin's only test-discovery surface); the repo-wide no-version-bump gate (CONTRIBUTING.md:82-95, scripts/check-version-bump.py); preserving validate-entities.py's decode-specific 'cannot decode file as UTF-8 - re-save it as UTF-8' remediation distinct from 'cannot read file:'; preserving the exact render-dashboard warning string that tests/test-render-dashboard.sh:349-350 pins (the AC says 'keeps its current warning behaviour' without naming the pinned check); the loader/signature ripple in staffing-score.py (_load_parse_frontmatter -> score_portfolio -> _load_entities) and the exit-1 broken-install envelope naming validate-entities.py that skills/projects-staff/SKILL.md:85 depends on; or that register-entity.py:121-125 holds an unmentioned fourth copy of the sequence that the AC's 'only place it lives' wording would sweep in while the issue body enumerates only three sites. AC bullet 4 also says 'all four tests/*.sh suites' while this change adds a fifth. The acceptance bar was materially under-specified. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1183

**Summary:** Both perspectives independently confirm 16 of 17 review-contract checklist items are satisfied with all five test suites passing green, but both also converge on the same clearly-adjudicated violation of item 8: register-entity.py:121-125 still carries its own untouched, OSError-only open+parse_frontmatter copy with no call to the new read_frontmatter helper and no justifying comment.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[major]** (auto-fixable) `cogni-projects/scripts/register-entity.py:121-125` — Checklist item 8 violated: the fourth copy of the open/read/parse sequence is left silently untouched. The file is not in the diff at all — it still does `with open(entity_file, "r", encoding="utf-8") as f: fm = _ve.parse_frontmatter(f.read())` guarded by `except OSError`, with no `read_frontmatter` call and no comment explaining why the decode case cannot reach it. The decode path is in fact unreachable today only because the `_ve.validate_file(entity_file)` gate at line 113 fails first on a latin-1 record, so any future reordering of that gate re-opens exactly the defect this PR closes. The checklist required either conversion or an in-code justification comment; neither is present. (Independently confirmed by contract-grading-reviewer against the review contract, item 8: register-entity.py does not appear in changed_files at all, and the contract's own text treats this outcome as a violation.)
  - *Suggested fix:* Convert the block to the shared helper: `fm, exc = _ve.read_frontmatter(entity_file)` followed by `if exc is not None: return _fail("cannot read entity file: %s" % exc)`. (The existing behaviour is preserved — the message stays the same and validate_file has already rejected undecodable files upstream.) If retention is preferred instead, add a comment at line 121 naming the validate_file gate at line 113 as the reason the decode case is unreachable here.

- **[minor]** `cogni-projects/scripts/staffing-score.py:429-430` — staffing-score.py remains the only one of the three scripts with no module-level try/except catch-all around main (contrast validate-entities.py's `except Exception` block that converts any escape into a {success:false} envelope with exit 2, and render-dashboard.py's equivalent). This PR closes the specific non-UTF-8 path, and the new suite's assert_no_traceback checks pin that, but the class of 'traceback instead of envelope' stays open here — e.g. an exec_module failure inside _load_read_frontmatter still escapes raw. The contract lists this as judgment-only and explicitly outside the acceptance bar, so it is not gating.
  - *Suggested fix:* Human decision: either add the sibling catch-all envelope to staffing-score.py's `__main__` block in this PR, or file a follow-up. Do not treat as a merge blocker for this extraction.

### Positive observations

- read_frontmatter is public, module-level, and lives in validate-entities.py (the module both siblings already load by file location), so the existing dependency arrow is preserved rather than inverted (item 1).
- The (fm, exc) return shape keeps all three outcomes distinguishable: (None, exc) for read/decode failure with the exception still reachable, (None, None) for read-OK-but-no-fence, (dict, None) for parsed (items 9, 10).
- render-dashboard.py:_read_entities keeps 'cannot read %s: %s' and the separate 'no frontmatter in %s — skipped' warning, continuing past the record; test-render-dashboard.sh passes green (item 11).
- Exactly one open()+parse_frontmatter pair survives in validate-entities.py; staffing-score.py's only remaining open() is the manifest read; render-dashboard.py's surviving open() calls are only the theme file, manifest, and dashboard write (items 5, 6, 7).
- The staffing-score loader ripple is done properly (_load_parse_frontmatter to _load_read_frontmatter), and the broken-install path still returns exit 1 naming 'validate-entities.py', matching skills/projects-staff/SKILL.md:85 (item 15).
- _load_entities still guards `if fm:` before appending, so a failed or fence-less read cannot put None into consultants/projects (item 14).
- The new test_staffing_score.sh is discriminating rather than assertion-of-success-only: fixture A pins clean-portfolio counts, fixture B writes a genuine latin-1 record and asserts exit 0 + success True + counts + no Traceback, fixture C proves failures still report correctly (item 13). House shape matched.
- The latin-1/DACH rationale survived into read_frontmatter's docstring; both borrowing module docstrings were updated to name read_frontmatter, with no stale 'parse_frontmatter' reference remaining (item 12).
- Scope discipline held: only the 5 declared files touched, no version line in plugin.json or marketplace.json, and no scoring weights or normalization logic touched (items 4, 16).
- All five suites were executed directly against the PR-head worktree and each exited 0 with no genuine FAIL lines, including the pinned-warning (test-render-dashboard.sh 6c/6f) and frontmatter-error (test_validate_entities_refs.sh test 9) assertions the contract flagged as most threatened (item 17). Scripts compile clean; plugin.json manifest valid.
- item 2 satisfied: test_staffing_score.sh uses `#!/usr/bin/env bash`, `set -u  # NOT -e`, `mktemp -d TMPROOT` with a trap-EXIT cleanup, and exits 1 on any assertion failure — matches the house shape.
- item 3 satisfied: 'test_staffing_score.sh' appears in cogni-projects/CLAUDE.md's tree listing and run-commands block — two registrations.

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1183

**Summary:** The sole queued perspective (plugin-dev:plugin-validator) returned pass on this re-review, confirming the revision commit resolves the prior major finding (AC 1 fully met, all four read sites unified through read_frontmatter) with all five test suites green. One un-waived minor finding remains (staffing-score.py silently drops an undecodable record with no warnings-channel signal) — this is an advisory observation, not a merge blocker, and is explicitly called out by the reviewer as non-gating residual scope adjacent to already-deferred issue #1186, not a regression introduced by this PR.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** `cogni-projects/scripts/staffing-score.py:292-296` — An undecodable entity record is now dropped with no signal at all — `fm, _exc = read_frontmatter(path)` discards the exception and the run returns success with a silently short consultant_count. AC 2's first half (no longer raises) is met and pinned by the new suite, but its second half ('reports through the standard envelope like its siblings') is only partly met: the validator reports the same record as an error and render-dashboard as a warning, while the scorer's shortlist just quietly omits the consultant. NON-GATING: residual scope adjacent to already-deferred issue #1186, not a regression introduced by this PR — the pre-PR behaviour was strictly worse (a raw traceback).
  - *Suggested fix:* Optional follow-up, not required for this merge: add a `warnings[]` key to the score_portfolio data dict and append 'cannot read <rel>: <exc>' when `_exc is not None`, mirroring render-dashboard.py:150. Fold into the #1186 work rather than re-opening this PR.

### Positive observations

- AC 1 verified exhaustively, not just at the reported site: a repo-wide grep confirms parse_frontmatter is now called from exactly one place (validate-entities.py:230, inside read_frontmatter), and all four read sites — validate_file, _ref_errors, render-dashboard._read_entities, register-entity.main — go through the shared helper.
- AC 2 verified behaviourally: a latin-1 fixture through staffing-score.py exits 0 with a clean envelope and no traceback.
- AC 3 verified behaviourally: render-dashboard.py on a latin-1 record still returns success=True with its 'cannot read ...' warning, byte-identical in shape to the pre-PR message.
- AC 4 verified: all five suites exit 0 on this head.
- The (fm, exc) return shape is the right design call — it lets validate_file keep its two distinct messages while render-dashboard collapses both to one warning, so unifying the read did not flatten any caller's output contract.
- read_frontmatter guards only the open/read, leaving parse_frontmatter outside the try — so a future parser fault still surfaces instead of being mislabelled 'unreadable file'.
- The new test_staffing_score.sh is well-constructed: Fixture A pins the counts Fixture B compares against, and Fixture C proves the fix was not achieved by swallowing every failure.
- No plugin.json or marketplace.json version was touched; check-version-bump reports 0 violations vs origin/main.
- Scripts remain stdlib-only across all four files — no pip dependency introduced.
- CLAUDE.md's architecture tree and test-run block were both updated for the new suite, so the developer guide did not drift.
- check-breadcrumbs and check-skill-names both pass — no maintainer breadcrumbs leaked in.

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressing the merger verdict at
https://github.com/cogni-work/insight-wave/pull/1183#issuecomment-5164023580
(graded head `9cbc71c9`, which equalled the head at the time — both findings were
against the live tree).

Revision commit: `4701fc9f` — one file, +6/−4.

### Findings addressed

**1. [major] `cogni-projects/scripts/register-entity.py:121-125` — checklist item 8: the fourth copy of the open/read/parse sequence left untouched.** ✅ Resolved in `4701fc9f`.

Took the conversion option (the one the finding lists first) rather than the
justification-comment option, because it removes the latent defect instead of
documenting it:

```python
fm, exc = _ve.read_frontmatter(entity_file)
if exc is not None:
    return _fail("cannot read entity file: %s" % exc)
```

Behaviour is unchanged — same message, same exit code, same `fm` value from
`parse_frontmatter`, and the `validate_file` gate at line 113 is still what
guarantees `fm` is a dict at line 130. No loader change was needed (`_ve` is
already bound at lines 45-57). All four read sites now share one guard, so the
gate-reordering hazard the finding named is closed rather than annotated.

**2. [minor] `cogni-projects/scripts/staffing-score.py:429-430` — no module-level catch-all envelope.** ⏭️ Deferred, not waived.

The verdict marks this judgment-only, explicitly outside the acceptance bar, and
explicitly not a merge blocker, and its own suggested fix offers "file a
follow-up" as an accepted resolution. Wrapping `main` in an envelope is a
different change class from extracting a read helper. Filed as **#1186** with
the sibling pattern and acceptance criteria written out, so it is tracked rather
than lost.

### One correction worth flagging

The first draft of this revision shipped a comment claiming the old bare
`OSError` guard "let UnicodeDecodeError escape past the JSON envelope". That is
false for this file — `register-entity.py:284-288` has a module-level
`except Exception` catch-all, so the exception was always enveloped; what the
change actually buys is the specific `cannot read entity file: ...` message
instead of the generic `unexpected failure: UnicodeDecodeError: ...`. The
comment also said "the other three" call sites when there are four. Both were
caught by the pre-commit quality pass and corrected before the commit — the
committed comment is accurate.

### Also in this revision

- PR body linking line is now `Refs #1173. Closes #1180.` — converting the fourth
  site *is* the entirety of #1180's scope, so it closes on merge.
- A review note was added to the body on whether #1173 itself should now be
  upgraded from `Refs` to `Closes`: with the fourth site converted, all four of
  its acceptance criteria read as met. I deliberately did not make that call
  unilaterally — it is a scope judgment for the merger, and it is one word.

### Verification

All five `cogni-projects` suites green on `4701fc9f`:
`test_portfolio_init`, `test_register_entity`, `test_validate_entities_refs`,
`test_staffing_score`, `test-render-dashboard` — each exit 0.
Both deterministic pre-gates clean (`check-frontmatter` schema v1, 4 files;
`check-version-bump` status `ok`, 0 violations vs `origin/main`).

### Follow-up issues

- #1186 — give `staffing-score.py` the sibling module-level catch-all envelope

sdh07:

_The merger formed this expectation from issue #1173 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The base clone shows the open/read/parse/guard sequence in FIVE places, not the three the issue names: validate-entities.py:298-314 (validate_file, guards OSError and UnicodeDecodeError with two distinct messages), validate-entities.py:244-248 (_ref_errors, guards both, skips), render-dashboard.py:146-159 (_read_entities, guards both and emits two distinct warnings 'cannot read %s: %s' vs 'no frontmatter in %s — skipped'), staffing-score.py:119-127 (_read_frontmatter, guards only OSError — the defect), and register-entity.py:121-125 (guards only OSError; reachable only because validate_file at :113 already rejects an undecodable file first). A correct fix adds one public read_frontmatter(path) beside parse_frontmatter in validate-entities.py that opens with an explicit encoding='utf-8' (no lenient errors=), catches OSError and UnicodeDecodeError, and returns enough information for callers to keep their current diagnostics — i.e. the exception (or its class) alongside the parsed dict, since a helper collapsing to dict-or-None would make render-dashboard report a latin-1 file as 'no frontmatter' (breaking test-render-dashboard.sh assertion 6f at :349-350) and would cost validate_file its 're-save it as UTF-8' guidance (:308). All five sites then call it; staffing-score.py must also switch its file-location loader (_load_parse_frontmatter, :100-116) and the threaded callable through main/score_portfolio/_load_entities to the new symbol, and update the broken-install message at :427-430 that names 'validate-entities.py:parse_frontmatter'. Because the mechanism lives in executable .py files, a behavioral regression test is demandable: cogni-projects/tests/test_staffing_score.sh (new, in changed_files) should seed a latin-1 entity the way test-render-dashboard.sh:331-335 does and assert the scorer prints a success envelope with exit 0 and no Traceback; cogni-projects/CLAUDE.md:27-43 is the only registry of the suites (there is no .github test runner) so it must gain the fifth entry in both the tree block and the run-commands block.

**Acceptance checklist:**
- [ ] validate-entities.py defines a module-level PUBLIC read_frontmatter(path) (no leading underscore) beside parse_frontmatter, with a docstring stating its return contract — grep the diff for 'def read_frontmatter(' in cogni-projects/scripts/validate-entities.py.
- [ ] The helper opens with an explicit encoding="utf-8" and passes NO errors= argument — grep its body for 'encoding="utf-8"' and confirm the absence of errors="replace"/"ignore"/"surrogateescape"; lenient decoding would silently corrupt the DACH name instead of reporting it.
- [ ] The helper's guard names BOTH exception types explicitly and is not widened to bare Exception or ValueError — grep for 'except (OSError, UnicodeDecodeError)' (or two separate named handlers) inside read_frontmatter.
- [ ] Every duplicate of the sequence is gone from all FIVE base sites, not just the three the issue enumerates: after the diff, an open() of an entity .md followed by parse_frontmatter() appears exactly once in cogni-projects/scripts/ — check validate-entities.py validate_file (base :298-311) and _ref_errors (base :244-246), render-dashboard.py _read_entities (base :146-148), staffing-score.py _read_frontmatter (base :121-126), and register-entity.py (base :121-123) each now delegate. Manifest/JSON/theme opens are out of scope.
- [ ] staffing-score.py no longer carries its own _read_frontmatter guard: the function is removed or reduced to a pass-through, and the callable threaded through main() -> score_portfolio() -> _load_entities() is the loaded read_frontmatter, not a stale parse_frontmatter parameter — confirm every parameter name and call site along that chain agrees.
- [ ] staffing-score.py's broken-install failure message no longer names a symbol it does not load: the string at base :427-430 ('cannot load validate-entities.py:parse_frontmatter') names whatever getattr target the loader now requires.
- [ ] A new suite cogni-projects/tests/test_staffing_score.sh exists, is stdlib-only bash + python3 (grep shows no pytest, no pip, no 'import yaml'), and follows the house idiom: a per-assertion pass/fail counter and a non-zero exit on any failure, as in test-render-dashboard.sh:29-31.
- [ ] cogni-projects/CLAUDE.md registers the new suite in BOTH places it lists tests — the tests/ tree block (base :27-31) and the runnable command block (base :38-43), which now shows five 'bash cogni-projects/tests/*.sh' lines; CLAUDE.md is the only registry (no .github workflow enumerates them).
- [ ] The new suite writes a genuinely non-UTF-8 entity file (a latin-1 byte, e.g. via python3 ... .encode('latin-1') as test-render-dashboard.sh:331-335 does), runs staffing-score.py against that portfolio, and asserts: the last stdout line parses as JSON with success is True, the process exit status is 0, and stderr contains no 'Traceback'.
- [ ] That same fixture asserts the readable records STILL score — e.g. consultant_count / project_count / ranked_candidate_count are non-zero and match the decodable entities — so the fix is a per-record skip, not an abort that returns an empty portfolio.
- [ ] render-dashboard.py keeps both distinct warnings verbatim — grep for the format literals 'cannot read %s: %s' and 'no frontmatter in %s — skipped' — so an undecodable file is still reported as UNREADABLE, never as frontmatter-less (test-render-dashboard.sh:349-350 requires one warning containing both 'cannot read' and 'latin1').
- [ ] validate-entities.py keeps both distinct diagnostics — grep for the literals 'cannot read file: ' and 'cannot decode file as UTF-8 — re-save it as UTF-8: ' (base :302, :308) — proving the helper's return contract carries the read-vs-decode distinction to callers instead of collapsing it; validate_file's separate '<frontmatter>' error 'no YAML frontmatter block found' (base :313) still fires only when the read succeeded and the fence is absent.
- [ ] register-entity.py still validates BEFORE reading: the _ve.validate_file gate and its exit-1 'entity failed validation' envelope (base :113-119) stay ahead of the frontmatter read, and its read-failure path still returns the 'cannot read entity file' envelope at the base exit code 2 — the refactor must not reorder these or change either code.
- [ ] No new dependency direction and no third-party import: staffing-score.py, render-dashboard.py and register-entity.py each still load the validator via importlib.util.spec_from_file_location on a sibling path, validate-entities.py imports none of its consumers, and read_frontmatter adds no module-scope side effect (it is exec_module'd by three consumers on every invocation).
- [ ] All five suites pass from the repo root with exit 0 and no 'FAIL' lines: test_portfolio_init.sh, test_register_entity.sh, test_validate_entities_refs.sh, test-render-dashboard.sh, and the new test_staffing_score.sh.

**Risk watchlist:**
- The obvious half-measure: adding UnicodeDecodeError to staffing-score.py's own _read_frontmatter guard and extracting nothing — that satisfies the issue's AC-2 and AC-4 while failing AC-1 and leaving the next divergence free to happen.
- A helper returning only dict-or-None (or {} on any miss) loses the read-vs-decode distinction: render-dashboard would then warn 'no frontmatter in ... — skipped' for a latin-1 file (test-render-dashboard.sh 6f fails) and validate-entities would drop its actionable 're-save it as UTF-8' guidance.
- 'Fixing' the decode leniently — errors='replace', reading bytes with a best-effort decode, or a latin-1 fallback — which silently writes a corrupted DACH name into the manifest and the partner dashboard instead of reporting it.
- Widening the guard to bare `except Exception` or `except ValueError`; render-dashboard.py:149-153 argues explicitly against that widening because it would bury a future parser fault as an unreadable-file report.
- Trusting the issue's 'three places' enumeration: base has five sites (validate_file, _ref_errors, _read_entities, staffing _read_frontmatter, register-entity.main). A fix scoped to the three named ones leaves two live copies and does not actually satisfy 'the only place the sequence lives'.
- Partial rename churn in staffing-score.py, where the parser is a parameter threaded main -> score_portfolio -> _load_entities -> _read_frontmatter: a half-done rename leaves score_portfolio(portfolio_dir, parse_frontmatter) receiving one callable and invoking another, which only surfaces at runtime.
- Judgment-only, not gated: staffing-score.py's __main__ (base :443-444) has NO envelope catch-all, unlike validate-entities.py:503-510 and render-dashboard.py:604-607. Closing only the decode path leaves the broader 'traceback instead of envelope' class the issue's 'Why it matters' section describes still open for any other unexpected failure — a human should decide whether this PR is the place to close it.
- Judgment-only, not gated: the scorer's envelope has no warnings key, so a skipped record vanishes from every shortlist silently — a partner cannot distinguish an unreadable consultant from an unavailable one. Adding one would change the documented data shape, so it is a design call rather than a gate.
- A new test that asserts only 'no traceback' would pass against a fix that returns a vacuously empty portfolio — which is why the positive-count assertion is a separate checklist item.
- CLAUDE.md drifting from tests/: the doc is the sole registry (no .github runner), so a new suite that is not listed there is a suite nobody runs.
- Scope creep beyond the extraction — e.g. retiring render-dashboard.py's _text_field coercion class (documented as a separate change at :270-273) or unifying the three case-sensitivity checks _is_closed warns about at :233-238 — which would change rendering behaviour under cover of a refactor.

- The issue's acceptance criteria are materially thinner than the base code requires. (1) They enumerate three copies of the read/parse sequence, but the base has five — register-entity.py:121-125 and validate-entities.py:244-248 (_ref_errors) are unnamed, yet AC-1's 'only place' wording silently binds them. (2) They require the four existing suites to stay green but never require a regression test for the very defect being fixed, so a fix could satisfy every stated criterion with zero coverage of the latin-1 staffing path (the PR does add cogni-projects/tests/test_staffing_score.sh, which the AC did not ask for). (3) They pin render-dashboard.py's warning behaviour but say nothing about validate-entities.py's own two distinct diagnostics ('cannot read file' vs 'cannot decode file as UTF-8 — re-save it as UTF-8'), which the same refactor can just as easily collapse. (4) They are silent on decode discipline (explicit encoding='utf-8', no lenient errors=), which is the difference between reporting a DACH name and corrupting it. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1183

**Summary:** Both perspectives passed against all 15 review-contract checklist items; the only un-waived finding is an advisory minor and a second minor is explicitly waived with a written justification tying it to tracked follow-up #1186 — non-gating, safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) `cogni-projects/scripts/register-entity.py:121-124` — The new comment claims a consequence the control flow does not produce: that routing through the shared reader makes a latin-1 record report 'cannot read entity file' rather than the __main__ catch-all's generic message. The _ve.validate_file gate at :113-119 still runs first and, for a latin-1 record, returns a non-empty errors list, so main() exits via the 'entity failed validation' envelope at code=1 and never reaches the read. The read-failure path is unreachable for a latin-1 file. Only the comment's second clause is accurate — this is defence in depth, not an observable message change. Behaviour and exit codes are correct and unchanged (contract item 13 satisfied); only the rationale text overstates.
  - *Suggested fix:* Reword to the accurate claim, e.g. 'Route through the shared reader, not a local open(): it guards UnicodeDecodeError too, so this path no longer depends on the validate_file gate above having rejected an undecodable record first.'

- **[minor]** (waived) `cogni-projects/scripts/staffing-score.py:442-443` — Judgment-only, explicitly not gated by the contract (risk_watchlist items 8 and 9): staffing-score.py's __main__ still has no envelope catch-all, unlike its two siblings, and _load_entities drops an unreadable record silently because the scorer's envelope has no warnings key. The PR closes the decode path (which is what was asked) and its comment is honest about the silent drop, but a partner still cannot distinguish an unreadable consultant from an unavailable one. Flagged for human awareness, not as a defect in this PR — adding a warnings key would change the documented data shape and belongs in its own change (tracked as #1186).
  - *Justification:* Contract risk_watchlist marks both items 'Judgment-only, not gated' and says a human should decide whether this PR is the place to close them; deferring is a legitimate scope decision, and #1186 already tracks it.

### Positive observations

- Contract items 1-3 (decode discipline): read_frontmatter is module-level and public, docstring states the (fm, exc) return contract, opens with encoding=utf-8, and grep for errors= across cogni-projects/scripts/ returns nothing — no lenient decoding and no latin-1 fallback, so a DACH name is reported rather than corrupted. The guard is 'except (OSError, UnicodeDecodeError)', not widened.
- Contract item 4 (all FIVE sites, not the issue's three): after the diff the only open() of an entity .md followed by parse_frontmatter() in cogni-projects/scripts/ is inside read_frontmatter itself. The two sites the issue's AC never enumerated are both converted — validate-entities._ref_errors and register-entity.py.
- Contract items 11-12 (no collapsed diagnostics — the top risk on the watchlist): the (fm, exc) pair rather than dict-or-None is what preserves every caller's diagnostics. render-dashboard.py keeps both warning literals verbatim so a latin-1 file is still UNREADABLE and test-render-dashboard.sh assertion 6f still passes. validate-entities.validate_file keeps both diagnostics, ordered decode-first-then-catch-all with a comment explaining the order is load-bearing.
- Contract items 5-6 (no partial rename churn): the callable name agrees at every point of the main -> score_portfolio -> _load_entities chain, and the broken-install message now names the symbol the loader actually requires.
- Contract items 7, 9, 10 (the regression test the issue's AC never asked for): verified as a real regression test, not a tautology — against the base scripts 5 assertions fail (exit 1, empty stdout, UnicodeDecodeError traceback); against the PR head it exits 0. Fixture B writes a genuine latin-1 byte and asserts exit 0, success is True, no Traceback, AND positive counts — so a vacuously-empty-portfolio 'fix' would not pass. Stdlib-only, house pass/fail counter idiom, mode 0755 like its siblings.
- The suite's run_score helper documents and avoids a real trap: it redirects to a file instead of a $(... | tail -n 1) substitution, under which $? is tail's status and every exit-code assertion would pass vacuously.
- Contract items 8, 13, 14: CLAUDE.md gains the fifth suite in BOTH registries (tree block and runnable command block), which matters because CLAUDE.md is the only registry. register-entity.py's validate gate and its code=1 envelope stay ahead of the read, and the read-failure path still returns code=2. No new dependency direction and no third-party import; read_frontmatter has no module-scope side effect.
- Contract item 15 verified directly: all five suites exit 0 from the repo root with zero FAIL lines. All four .py files compile clean under py_compile, the new suite passes bash -n, and plugin.json is untouched and valid JSON.
- No scope creep of the kind the watchlist warned about: render-dashboard.py's _text_field coercion class and the _is_closed case-sensitivity checks are untouched, so no rendering behaviour changes under cover of a refactor.
- item 1 satisfied: validate-entities.py:200 defines a module-level public read_frontmatter(path) with a docstring stating the (fm, exc) return contract.
- item 2 satisfied: the helper's open(path, 'r', encoding='utf-8') at :221 has no errors= argument anywhere in its body.
- item 3 satisfied: :223 guards with except (OSError, UnicodeDecodeError) as exc — both types named explicitly, not widened.
- item 4 satisfied: all five base sites delegate — validate_file, _ref_errors, render-dashboard._read_entities, staffing-score (function deleted), and register-entity.py:121 (the fourth-site gap this PR's own revise commit 4701fc9f closed). A repo-wide grep confirms no remaining parse_frontmatter(f.read()) call outside the new helper.
- item 5 satisfied: the callable is renamed read_frontmatter consistently through the whole threaded chain — every parameter name and call site agrees.
- item 6 satisfied: staffing-score.py's broken-install message now names read_frontmatter, matching the getattr target.
- item 7 satisfied: test_staffing_score.sh is new, stdlib-only, and uses the house failures counter idiom with exit 1 on any failure.
- item 8 satisfied: CLAUDE.md registers the new suite in both the tests/ tree block and the runnable bash-command block, each now listing five suites.
- item 9 satisfied: Fixture B writes cesar.md via .encode('latin-1'), runs the scorer, and asserts exit 0, success is True, and no Traceback on stderr.
- item 10 satisfied: the same fixture asserts consultant_count == 2 and project_count == 1, matching the two decodable consultants and one project — proving a per-record skip, not an abort to an empty portfolio.
- item 11 satisfied: render-dashboard.py keeps both verbatim format literals.
- item 12 satisfied: validate-entities.py keeps both distinct diagnostics verbatim, checked decode-first, and the separate no-YAML-frontmatter error still only fires after a successful read.
- item 13 satisfied: register-entity.py's validate_file gate and its exit-1 envelope remain ahead of the read_frontmatter call, and the read-failure path still returns _fail's default code=2, unchanged from base.
- item 14 satisfied: all three consumers still load validate-entities.py via importlib.util by sibling file path; only stdlib imports appear; validate-entities.py imports none of its consumers; read_frontmatter has no module-scope side effect.
- item 15 satisfied: all five suites were run from the worktree root and each exited 0 with no FAIL lines.

### Recommendation

---

PR #1188, opened 2026-08-03
cogni-projects: portfolio-scoped ref gate on entity registration

### LINKED ISSUE #1172: cogni-projects: run the assignment ref check on the registration write path, not only the validator CLI

claude-cogni-work:
Closes #1172.

`register-entity.py` owns portfolio-scoped integrity on the **write** path, but gated only on `_ve.validate_file(entity_file)` — a single-file entry point that by design runs no cross-entity ref resolution. An assignment carrying a misspelled `consultant` slug passed that gate, was registered into `projects-portfolio.json`, and then read as an unfilled role in `render-dashboard.py` with no error anywhere.

## Summary

- **`register-entity.py`** — at the registration gate, resolve the entity's refs against the whole portfolio via `_ve._ref_errors(_ve._entity_files(portfolio_dir))`, keeping only the entries belonging to the entity being registered. `_fail` gains an optional `data=` payload so the raw `{entity, file, field, message}` dicts ride in `data.errors[]`; every existing call site stays byte-identical.
- **`tests/test_register_entity.sh`** — the suite had **zero** assignment coverage (no `make_project` / `make_assignment`, never created an `assignments/` dir). Adds both fixtures and a block covering rejection, the path-spelling guard, and the two unaffected paths.
- **`tests/test_validate_entities_refs.sh`** — comment-only. Its header and section-3 comment framed `register-entity.py` as ref-free, which this change falsifies. No assertion touched.
- **`skills/projects-entities/SKILL.md`** — corrects the Step 4 and Step 5 prose that documented the write path as unable to check refs.

## The three decisions worth reviewing

**1. Scoped to assignments.** `_ref_errors` structurally can only emit `entity == "assignment"` errors, and the filter keeps only the registered file — so running the batch for a consultant or project reads every record in the portfolio to build a result that cannot be non-empty. The gate sits after `entity_type` is derived and guards on `entity_type == "assignment"`. Measured on a synthetic 60-record portfolio, the ref pass is ~3.0 ms of the 5.1 ms `main()` does; scoping removes that entirely for two of the three entity types. Verified live: 0 `_ref_errors` calls when registering a consultant, 1 when registering an assignment.

**2. The batch must be the whole portfolio.** `_ref_errors([entity_file])` would resolve nothing — the known-slug sets are built from the consultant and project records *in the batch*, so a single-file expansion leaves them empty and every ref spuriously dangles. `_entity_files(portfolio_dir)` is required, not incidental.

**3. Path identity is `samefile`, not `==` and not `realpath`.** `_ref_errors` builds its `file` key from the raw `portfolio_dir` argument, while `entity_file` is an independent raw argument — so the two spellings of one file routinely differ. A string compare is the obvious bug. Less obviously, **bare `realpath` equality is also wrong**: `realpath` does not case-fold, so on a case-insensitive filesystem a case-variant path yields inequality, the filter drops the entity's own ref error, and the gate **fails open** — registering exactly the dangling assignment it exists to refuse. `os.path.samefile` compares `(st_dev, st_ino)`, which is also hardlink-correct, with a `realpath` fallback on `OSError` (only reachable under a concurrent unlink) to keep the helper raise-free for the imported-`main()` callers the `__main__` catch-all does not cover.

A test registers through a `<root>/.` portfolio argument specifically to pin this. It has teeth — verified directly: with that argument `_entity_files` emits `<root>/./assignments/x.md`, so a naive `==` matches **0 of 1** errors and the gate would pass the dangling assignment through.

## Ordering

The gate sits after the shape gate (a missing `consultant` key is a required-key error, not a dangling ref, and must report as the former) and before the portfolio-escape check. That ordering is benign in the escape direction: an entity outside the portfolio never appears in `_entity_files`' walk, so the filter is empty and the escape check still raises its own correct error.

## Test plan

- [x] `test_register_entity.sh` — 27 checks pass (20 pre-existing + 7 new)
- [x] `test_validate_entities_refs.sh`, `test_portfolio_init.sh`, `test_staffing_score.sh`, `test-render-dashboard.sh` — all green
- [x] AC 1 — dangling `consultant` returns `success: false`, exit 1, `data.errors[0]` carrying all four keys with `entity == "assignment"` / `field == "consultant"`; manifest `assignments[]` untouched
- [x] AC 2 — a project registers cleanly *while a dangling assignment sits in the same portfolio*
- [x] AC 3 — an assignment whose refs both resolve registers `created`, with the dangling one still on disk (so this also pins that the filter reports only the registered entity)
- [x] AC 4 — existing suite green, unmodified
- [x] Version-bump gate: 14 plugins scanned, 0 touched. Breadcrumb guard: 0 violations.

## Note for the reviewer (not addressed here)

This makes `register-entity.py` the second out-of-module consumer of `validate-entities.py`'s underscore-prefixed helpers (`render-dashboard.py` already uses `_entity_files`). The honest fix is a rename to drop the underscore on both symbols — smaller than a wrapper, and it corrects the actual defect, which is naming rather than layering. Deliberately left out of this PR as unrelated to the issue's scope; flagging it rather than filing it, since it is a judgment about the module's public surface.

An alternative plan adding a public `portfolio_ref_errors(portfolio_dir, entity_file=None)` seam was considered and rejected: it would serve exactly one caller, and `validate-entities.py`'s own `main()` would not use it (it already holds a realpath-deduped batch reused across both passes, so routing through the seam would re-walk every directory).

### DIFF
--- a/cogni-projects/scripts/register-entity.py
+++ b/cogni-projects/scripts/register-entity.py
@@ -183,22 +183,25 @@ def main(argv):
     # Placed after the type is known but before the portfolio-escape check below,
     # which is harmless: an entity outside the portfolio never appears in the
     # walk, so the filter is empty and that check still raises its own error.
-    # _ref_errors itself is raise-free — it reads through read_frontmatter,
-    # which returns (None, exc) rather than raising. _entity_files is not: its
-    # os.listdir walk is unguarded, so an unreadable portfolio subdirectory
-    # escapes as an OSError. The __main__ catch-all below would convert that to
-    # an envelope for CLI callers, but not for the in-process callers (the test
-    # harness among them) that call main() directly, so the guard belongs here.
+    # Neither helper is trusted to be raise-free, matching validate-entities.py's
+    # own main(), which wraps this identical call because the envelope is the
+    # contract. _entity_files' os.listdir walk is unguarded, so an unreadable
+    # subdirectory raises OSError; and read_frontmatter calls parse_frontmatter
+    # outside its try, so a parser fault (int() on an over-long digit scalar)
+    # surfaces as ValueError. The __main__ catch-all below would convert either
+    # to an envelope for CLI callers, but not for the in-process callers (the
+    # test harness among them) that call main() directly — hence the guard here.
     if entity_type == "assignment":
         try:
             ref_errors = [
                 e for e in _ve._ref_errors(_ve._entity_files(portfolio_dir))
                 if _same_file(e["file"], entity_file)
             ]
-        except OSError as exc:
+        except Exception as exc:  # noqa: BLE001 — the envelope is the contract
             # Default code 2 (environment failure), never code 1 — the entity
             # did not fail validation; the portfolio could not be read.
-            return _fail("cannot scan portfolio for refs: %s" % exc)
+            return _fail("cannot scan portfolio for refs: %s: %s"
+                         % (type(exc).__name__, exc))
         if ref_errors:
             return _fail(
                 "entity failed ref validation (%d error(s)) — a consultant or "
--- a/cogni-projects/scripts/register-entity.py
+++ b/cogni-projects/scripts/register-entity.py
@@ -24,6 +24,11 @@ swap, since mkstemp would otherwise force every target to 0600.
 
 Validates before registering, via validate-entities.py, so the manifest cannot
 take an entity the validator rejects even when this script is invoked directly.
+An assignment's consultant / project refs are additionally resolved against the
+whole portfolio directory — the pass the validator runs only for a directory
+argument, never for the single file this script hands it — so a dangling ref is
+refused here rather than surfacing later as a role that silently reads unfilled.
+Each unresolved ref is reported in data.errors[].
 
 Stdlib-only (no PyYAML).
 
@@ -68,11 +73,37 @@ REF_FIELDS = {
 }
 
 
-def _fail(message, code=2):
+def _fail(message, code=2, data=None):
     print(json.dumps(
-        {"success": False, "data": {}, "error": message}, ensure_ascii=False
+        {"success": False, "data": data or {}, "error": message},
+        ensure_ascii=False,
     ))
     return code
+
+
+def _same_file(a, b):
+    """True when `a` and `b` name the same file on disk.
+
+    Never a string compare: the ref errors carry the path _entity_files built
+    from the caller's raw portfolio-dir argument, while the entity file is an
+    independent raw argument, so the two spellings of one file routinely differ
+    (relative vs absolute, a trailing "/.", a symlinked portfolio root).
+
+    samefile rather than the realpath equality used for the portfolio-escape
+    check below, because realpath does not case-fold: on a case-insensitive
+    filesystem a case-variant path to the same file compares unequal, which
+    would drop the registered entity's own ref error and let the gate fail
+    OPEN — registering exactly the dangling assignment it exists to refuse.
+    Comparing (st_dev, st_ino) is also correct through a hard link.
+
+    Both paths exist whenever this is called, so the OSError branch is only a
+    concurrent-unlink race; falling back keeps the helper raise-free for the
+    imported-main callers the __main__ catch-all does not cover.
+    """
+    try:
+        return os.path.samefile(a, b)
+    except OSError:
+        return os.path.realpath(a) == os.path.realpath(b)
 
 
 def _intended_mode(path):
@@ -136,6 +167,37 @@ def main(argv):
             % (entity_type, ", ".join(sorted(REF_FIELDS))),
             code=1,
         )
+
+    # Resolve this assignment's refs against the portfolio. validate_file is
+    # single-file by design and so runs no ref pass; without this, an assignment
+    # naming a misspelled consultant passes the gate above, lands in the
+    # manifest, and then reads as an unfilled role with no error anywhere.
+    #
+    # The batch must be the whole portfolio, never [entity_file]: the known-slug
+    # sets are built from the consultant and project records in the batch, so a
+    # single-file expansion resolves nothing and every ref would dangle. That is
+    # also why this is scoped to assignments — they are the only type carrying
+    # refs, and the walk is O(portfolio), so running it for a consultant or a
+    # project would read every record to build a result that cannot be non-empty.
+    #
+    # Placed after the type is known but before the portfolio-escape check below,
+    # which is harmless: an entity outside the portfolio never appears in the
+    # walk, so the filter is empty and that check still raises its own error.
+    # _ref_errors reads through read_frontmatter, which returns (None, exc)
+    # rather than raising, so no guard is needed here.
+    if entity_type == "assignment":
+        ref_errors = [
+            e for e in _ve._ref_errors(_ve._entity_files(portfolio_dir))
+            if _same_file(e["file"], entity_file)
+        ]
+        if ref_errors:
+            return _fail(
+                "entity failed ref validation (%d error(s)) — a consultant or "
+                "project ref does not resolve within this portfolio; fix each "
+                "data.errors[] entry before registering" % len(ref_errors),
+                code=1,
+                data={"errors": ref_errors},
+            )
 
     array_name = _ve.SCHEMA[entity_type]["subdir"]
     slug = str(fm["slug"])
--- a/cogni-projects/scripts/register-entity.py
+++ b/cogni-projects/scripts/register-entity.py
@@ -183,13 +183,22 @@ def main(argv):
     # Placed after the type is known but before the portfolio-escape check below,
     # which is harmless: an entity outside the portfolio never appears in the
     # walk, so the filter is empty and that check still raises its own error.
-    # _ref_errors reads through read_frontmatter, which returns (None, exc)
-    # rather than raising, so no guard is needed here.
+    # _ref_errors itself is raise-free — it reads through read_frontmatter,
+    # which returns (None, exc) rather than raising. _entity_files is not: its
+    # os.listdir walk is unguarded, so an unreadable portfolio subdirectory
+    # escapes as an OSError. The __main__ catch-all below would convert that to
+    # an envelope for CLI callers, but not for the in-process callers (the test
+    # harness among them) that call main() directly, so the guard belongs here.
     if entity_type == "assignment":
-        ref_errors = [
-            e for e in _ve._ref_errors(_ve._entity_files(portfolio_dir))
-            if _same_file(e["file"], entity_file)
-        ]
+        try:
+            ref_errors = [
+                e for e in _ve._ref_errors(_ve._entity_files(portfolio_dir))
+                if _same_file(e["file"], entity_file)
+            ]
+        except OSError as exc:
+            # Default code 2 (environment failure), never code 1 — the entity
+            # did not fail validation; the portfolio could not be read.
+            return _fail("cannot scan portfolio for refs: %s" % exc)
         if ref_errors:
             return _fail(
                 "entity failed ref validation (%d error(s)) — a consultant or "
--- a/cogni-projects/skills/projects-entities/SKILL.md
+++ b/cogni-projects/skills/projects-entities/SKILL.md
@@ -39,7 +39,10 @@ are separate tool calls, not a transaction — an interrupted run can leave an
 entity file on disk that the manifest does not yet reference. That state is
 repaired by **re-running the skill for the same slug**: registration is
 idempotent (`register-entity.py` upserts keyed on `slug`), so a repeated run
-never double-registers.
+never double-registers. A registration *refused* for an unresolved consultant /
+project ref is a different case — a precondition failure, not an interrupted
+run — and re-running fails identically until the ref itself resolves; Step 5
+covers what to fix.
 
 ## Workflow
 
@@ -76,7 +79,8 @@ assignment, read the referenced consultant and project records first and confirm
 each record's frontmatter `slug` equals the value going into `consultant` /
 `project`. The validator resolves a ref against the record's `slug`, not its
 filename, so a file that merely exists under a similar name still dangles — that
-slug match, not the file's existence, is the guard.
+slug match, not the file's existence, is what makes the ref resolve; Step 5's
+registration is what refuses one that does not.
 
 ### Step 3: Read siblings, then write the entity file
 
@@ -159,8 +163,10 @@ It returns the same `{"success", "data", "error"}` envelope; `data.action` is
 `created` or `updated`, which is how a re-run reports that it replaced an
 existing ref rather than adding a second one.
 
-A registration that fails on a dangling ref wrote nothing — fix the misspelled
-slug (or author the missing record) and re-run this same command.
+A registration that fails on a dangling ref leaves the manifest and the
+execution log untouched; the entity file written in Step 3 stays on disk. So only
+the ref needs fixing — correct the misspelled slug (or author the missing
+record) and re-run this same command.
 
 When the entity was an assignment, follow the successful registration with the
 portfolio-directory sweep described in Step 4 — it is only worth running once
--- a/cogni-projects/skills/projects-entities/SKILL.md
+++ b/cogni-projects/skills/projects-entities/SKILL.md
@@ -116,9 +116,10 @@ enums, slug casing, ISO dates and their ordering, numeric ranges. It resolves an
 assignment's `consultant` / `project` slugs against the `consultants/` and
 `projects/` records only when given a whole portfolio directory — that is what
 collects the records to resolve against. This step passes a single file, so no
-ref check runs here and a passing run is not evidence the refs exist; confirming
-in Step 2 that each record's frontmatter `slug` equals the ref remains the guard
-against a dangling reference.
+ref check runs here and a passing run is not evidence the refs exist. Step 5's
+registration is the gate that enforces ref resolution; confirming in Step 2 that
+each record's frontmatter `slug` equals the ref catches the problem earlier,
+before a file is written.
 
 When the user asks for a portfolio-wide audit rather than a single authoring
 run, sweep for dangling refs with a separate invocation — not this gate: pass
@@ -130,18 +131,25 @@ unchanged. Its `success: false` covers every entity in the portfolio, not only
 the one just authored — locate the relevant entries by each error's `file`.
 
 Authoring an assignment is the one case worth sweeping unprompted rather than on
-request: it is the only entity type that carries refs, so a portfolio-directory
-run right after Step 5 is what turns Step 2's read-and-compare into a checked
-fact. Keep it after registration and never in place of the Step 4 gate — the
-single-file gate's behaviour is unchanged.
+request: it is the only entity type that carries refs. Step 5's registration
+already checks the refs of the assignment it registers, so the sweep covers the
+*rest* of the portfolio, where a record renamed or removed since it was authored
+can leave an older assignment dangling. Run it after registration, never in
+place of the Step 4 gate — that gate runs no ref check.
 
 ### Step 5: Register in the manifest and log the transition
 
 Once validation passes, register the entity. Do not hand-edit
 `projects-portfolio.json` — run the script, which upserts the summary ref keyed
 on `slug`, bumps the manifest `updated` date, and appends the
-`.metadata/execution-log.json` transition in a single invocation (it re-runs the
-validator itself, so it refuses an entity the Step 4 gate would have caught):
+`.metadata/execution-log.json` transition in a single invocation. It re-runs the
+validator, so it refuses an entity the Step 4 gate would have caught — but it
+reports those shape failures only as an error count, which is why Step 4 is
+still the run that surfaces the per-field detail. For an assignment it
+additionally resolves the `consultant` / `project` refs against the whole
+portfolio, catching a dangling ref the single-file gate passes; each unresolved
+ref comes back as a full `data.errors[]` entry naming the offending field and
+the missing referent.
 
 ```bash
 python3 "${CLAUDE_PLUGIN_ROOT:-$(ls -td "$HOME"/.claude/plugins/cache/*/cogni-projects/*/ 2>/dev/null | head -1)}/scripts/register-entity.py" "cogni-projects/<portfolio-slug>" "cogni-projects/<portfolio-slug>/<subdir>/<entity>.md"
@@ -151,11 +159,12 @@ It returns the same `{"success", "data", "error"}` envelope; `data.action` is
 `created` or `updated`, which is how a re-run reports that it replaced an
 existing ref rather than adding a second one.
 
+A registration that fails on a dangling ref wrote nothing — fix the misspelled
+slug (or author the missing record) and re-run this same command.
+
 When the entity was an assignment, follow the successful registration with the
-portfolio-directory sweep described in Step 4 — that run is what turns Step 2's
-slug comparison into a checked fact. The Step 4 gate above stays as written; this
-is the sweep, and it belongs here because it is only worth running once the
-assignment is in the manifest.
+portfolio-directory sweep described in Step 4 — it is only worth running once
+the assignment is in the manifest.
 
 ### Step 6: Summarize
 
--- a/cogni-projects/tests/test_register_entity.sh
+++ b/cogni-projects/tests/test_register_entity.sh
@@ -365,6 +365,26 @@ else:
     finally:
         os.chmod(os.path.join(root5, "consultants"), 0o755)
 
+# The scan is not trusted to raise only OSError: read_frontmatter calls
+# parse_frontmatter outside its own try, so a parser fault surfaces as
+# ValueError. The envelope is the contract for every exception type, not just
+# the one that happened to be found first.
+_real_ref_errors = reg._ve._ref_errors
+try:
+    def _boom(_files):
+        raise ValueError("simulated parser fault")
+    reg._ve._ref_errors = _boom
+    code, env = register(root5, resolving)
+    check("refs: a non-OSError from the ref scan still returns the envelope",
+          env.get("success") is False and code == 2, "code=%s env=%s" % (code, env))
+    check("refs: the envelope names the failing exception type",
+          "ValueError" in (env.get("error") or ""), str(env))
+except Exception as exc:
+    check("refs: a non-OSError from the ref scan still returns the envelope",
+          False, "main() raised %s: %s" % (type(exc).__name__, exc))
+finally:
+    reg._ve._ref_errors = _real_ref_errors
+
 print()
 if failures:
     print("%d check(s) failed." % failures)
--- a/cogni-projects/tests/test_register_entity.sh
+++ b/cogni-projects/tests/test_register_entity.sh
@@ -82,6 +82,45 @@ def make_consultant(root, slug, name):
     return path
 
 
+def make_project(root, slug, name):
+    # make_portfolio creates only consultants/ and .metadata/, so the subdir has
+    # to be made here — and exist_ok because a portfolio takes several records.
+    os.makedirs(os.path.join(root, "projects"), exist_ok=True)
+    path = os.path.join(root, "projects", slug + ".md")
+    with open(path, "w", encoding="utf-8") as f:
+        f.write(
+            "---\n"
+            "type: project\n"
+            "slug: %s\n"
+            "name: %s\n"
+            "client: Northwind\n"
+            "strategic_impact: 4\n"
+            "---\n\n# %s\n" % (slug, name, name)
+        )
+    return path
+
+
+def make_assignment(root, consultant_ref, project_ref):
+    """Write an assignment. Its refs are whatever the caller passes, so a
+    dangling one is written without disturbing the records it points at."""
+    os.makedirs(os.path.join(root, "assignments"), exist_ok=True)
+    slug = "%s--%s" % (consultant_ref, project_ref)
+    path = os.path.join(root, "assignments", slug + ".md")
+    with open(path, "w", encoding="utf-8") as f:
+        f.write(
+            "---\n"
+            "type: assignment\n"
+            "slug: %s\n"
+            "consultant: %s\n"
+            "project: %s\n"
+            "role: Lead Engineer\n"
+            "start_date: 2026-02-01\n"
+            "end_date: 2026-08-01\n"
+            "---\n\n# %s\n" % (slug, consultant_ref, project_ref, slug)
+        )
+    return path
+
+
 def register(root, entity_file):
     """Run register-entity.main and return (exit_code, parsed_envelope)."""
     buf = io.StringIO()
@@ -199,6 +238,59 @@ check("modes: upsert succeeded", env.get("success") is True, str(env))
 man_mode2 = os.stat(os.path.join(root4, "projects-portfolio.json")).st_mode & 0o777
 check("modes: upsert preserves the existing mode",
       man_mode2 == 0o640, "%o != 0o640" % man_mode2)
+
+# --- Portfolio ref resolution on the write path ------------------------------
+# validate_file is single-file and runs no ref pass, so before the gate an
+# assignment naming a misspelled consultant registered clean and then read as an
+# unfilled role. Every record below is on disk before the first register call:
+# the project case only proves the gate leaves non-assignments alone if a
+# dangling assignment is already sitting in the same portfolio.
+root5 = make_portfolio("refs")
+make_consultant(root5, "ana-silva", "Ana Silva")
+proj = make_project(root5, "nordic-erp", "Nordic ERP")
+# Misspelled but still kebab-valid: an underscore would fail the composite-slug
+# rule and the assertions below would be measuring a shape error, not a ref one.
+dangling = make_assignment(root5, "ana-silvaa", "nordic-erp")
+
+
+def registered_assignments():
+    with open(os.path.join(root5, "projects-portfolio.json"), encoding="utf-8") as f:
+        return json.load(f)["assignments"]
+
+
+code, env = register(root5, proj)
+check("refs: a project still registers while a dangling assignment exists",
+      env.get("success") is True, str(env))
+
+code, env = register(root5, dangling)
+errs = env.get("data", {}).get("errors", [])
+assigns = registered_assignments()
+check("refs: dangling consultant ref fails registration",
+      env.get("success") is False and code == 1, "code=%s env=%s" % (code, env))
+check("refs: the offending ref comes back in data.errors[]", len(errs) == 1, str(env))
+check("refs: the error keeps the standard {entity, file, field, message} shape",
+      bool(errs) and set(errs[0]) == {"entity", "file", "field", "message"}
+      and errs[0]["entity"] == "assignment" and errs[0]["field"] == "consultant",
+      str(errs))
+check("refs: nothing was written to the manifest", assigns == [], str(assigns))
+
+# The filter compares files, not path strings. Passing the portfolio as
+# "<root>/." makes _entity_files emit "<root>/./assignments/..." — a string
+# compare matches none of the errors, the gate fails open, the registration
+# succeeds, and this goes red.
+code, env = register(os.path.join(root5, "."), dangling)
+assigns = registered_assignments()
+check("refs: still rejected when the portfolio is passed as <root>/.",
+      env.get("success") is False and code == 1 and assigns == [],
+      "code=%s assignments=%s env=%s" % (code, assigns, env))
+
+# The dangling assignment stays on disk, so this also proves the filter reports
+# only the entity being registered rather than the whole portfolio's errors.
+resolving = make_assignment(root5, "ana-silva", "nordic-erp")
+code, env = register(root5, resolving)
+check("refs: an assignment whose refs resolve registers unaffected",
+      env.get("success") is True
+      and env.get("data", {}).get("action") == "created", str(env))
 
 print()
 if failures:
--- a/cogni-projects/tests/test_register_entity.sh
+++ b/cogni-projects/tests/test_register_entity.sh
@@ -258,10 +258,37 @@ def registered_assignments():
         return json.load(f)["assignments"]
 
 
+# A rejected registration must leave both write targets untouched. The parsed
+# assignments[] check alone is too weak: it would still pass if the manifest were
+# rewritten with equivalent JSON, and it never looks at the execution log at all.
+def manifest_bytes():
+    with open(os.path.join(root5, "projects-portfolio.json"), "rb") as f:
+        return f.read()
+
+
+def transition_count():
+    # No missing-file fallback on purpose: every caller runs after a successful
+    # registration, so an absent log is a real regression and should raise here
+    # rather than read as a legitimate count of zero.
+    path = os.path.join(root5, ".metadata", "execution-log.json")
+    with open(path, encoding="utf-8") as f:
+        return len(json.load(f).get("transitions", []))
+
+
+def check_untouched(tag, man_before, trans_before):
+    """Assert a rejected registration wrote to neither the manifest nor the log."""
+    trans_after = transition_count()
+    check("refs: %s leaves the manifest byte-identical" % tag,
+          manifest_bytes() == man_before, "manifest changed")
+    check("refs: %s leaves execution-log transitions[] unchanged" % tag,
+          trans_after == trans_before, "%d != %d" % (trans_after, trans_before))
+
+
 code, env = register(root5, proj)
 check("refs: a project still registers while a dangling assignment exists",
       env.get("success") is True, str(env))
 
+man_before, trans_before = manifest_bytes(), transition_count()
 code, env = register(root5, dangling)
 errs = env.get("data", {}).get("errors", [])
 assigns = registered_assignments()
@@ -273,24 +300,70 @@ check("refs: the error keeps the standard {entity, file, field, message} shape",
       and errs[0]["entity"] == "assignment" and errs[0]["field"] == "consultant",
       str(errs))
 check("refs: nothing was written to the manifest", assigns == [], str(assigns))
+check_untouched("a dangling consultant ref", man_before, trans_before)
 
 # The filter compares files, not path strings. Passing the portfolio as
 # "<root>/." makes _entity_files emit "<root>/./assignments/..." — a string
 # compare matches none of the errors, the gate fails open, the registration
 # succeeds, and this goes red.
+man_before, trans_before = manifest_bytes(), transition_count()
 code, env = register(os.path.join(root5, "."), dangling)
 assigns = registered_assignments()
 check("refs: still rejected when the portfolio is passed as <root>/.",
       env.get("success") is False and code == 1 and assigns == [],
       "code=%s assignments=%s env=%s" % (code, assigns, env))
+check_untouched("a non-canonical portfolio path", man_before, trans_before)
 
-# The dangling assignment stays on disk, so this also proves the filter reports
+# _ref_errors loops over both ref fields, so the consultant case above exercises
+# only half of it. A resolving consultant with a misspelled project covers the
+# other half — and again kebab-valid, so this measures a ref error, not a shape one.
+dangling_project = make_assignment(root5, "ana-silva", "nordic-erpp")
+man_before, trans_before = manifest_bytes(), transition_count()
+code, env = register(root5, dangling_project)
+errs = env.get("data", {}).get("errors", [])
+check("refs: dangling project ref fails registration",
+      env.get("success") is False and code == 1, "code=%s env=%s" % (code, env))
+check("refs: the dangling project is the single reported error",
+      len(errs) == 1 and errs[0]["entity"] == "assignment"
+      and errs[0]["field"] == "project", str(errs))
+assigns = registered_assignments()
+check("refs: nothing was written to the manifest for a dangling project",
+      assigns == [], str(assigns))
+check_untouched("a dangling project ref", man_before, trans_before)
+
+# The dangling assignments stay on disk, so this also proves the filter reports
 # only the entity being registered rather than the whole portfolio's errors.
 resolving = make_assignment(root5, "ana-silva", "nordic-erp")
 code, env = register(root5, resolving)
 check("refs: an assignment whose refs resolve registers unaffected",
       env.get("success") is True
       and env.get("data", {}).get("action") == "created", str(env))
+# Length, not just success: a duplicate append would satisfy the check above.
+assigns = registered_assignments()
+check("refs: the resolving assignment is registered exactly once",
+      len(assigns) == 1, str(assigns))
+
+# The ref scan walks the portfolio, so an unreadable subdirectory raises inside
+# _entity_files. main() must still answer with the envelope: the __main__
+# catch-all only covers CLI callers, and register() here calls main() directly —
+# exactly the in-process caller that would otherwise get a traceback.
+if os.geteuid() == 0:
+    print("SKIP: refs: unreadable subdir (running as root — mode bits do not deny)")
+else:
+    os.chmod(os.path.join(root5, "consultants"), 0o000)
+    try:
+        code, env = register(root5, resolving)
+        # code 2 (environment failure), never 1 — the entity did not fail
+        # validation; the portfolio could not be read.
+        check("refs: an unreadable portfolio subdir returns the envelope, not a traceback",
+              env.get("success") is False and code == 2, "code=%s env=%s" % (code, env))
+        check("refs: the unreadable-subdir error names the ref scan",
+              "cannot scan portfolio for refs" in (env.get("error") or ""), str(env))
+    except OSError as exc:
+        check("refs: an unreadable portfolio subdir returns the envelope, not a traceback",
+              False, "main() raised %s: %s" % (type(exc).__name__, exc))
+    finally:
+        os.chmod(os.path.join(root5, "consultants"), 0o755)
 
 print()
 if failures:
--- a/cogni-projects/tests/test_validate_entities_refs.sh
+++ b/cogni-projects/tests/test_validate_entities_refs.sh
@@ -6,7 +6,10 @@
 # from that same directory, while a single file argument — which carries no
 # portfolio-wide collection — keeps its shape-only behaviour. The single-file
 # case is the regression guard for projects-entities Step 4 and for
-# register-entity.py, which both hand the validator one file at a time.
+# register-entity.py, which both hand the validator one file at a time. That
+# still describes what they hand validate_file; register-entity.py additionally
+# runs its own portfolio-scoped ref pass on top of this gate, so a clean result
+# here is not evidence that registration accepts the entity.
 #
 # Also covers the pass's edges: an assignment dangling on both fields is
 # reported once per field; a portfolio missing the consultants/ subdirectory
@@ -141,9 +144,11 @@ if errs:
           "grace-hopper" in e["message"], e["message"])
 
 # --- 3. Single-file invocation on the SAME dangling assignment -----------
-# The guard for projects-entities Step 4 and register-entity.py: with no
-# portfolio-wide collection there is nothing to resolve against, so the file
-# must still validate clean.
+# The guard for projects-entities Step 4 and for the validate_file call inside
+# register-entity.py: with no portfolio-wide collection there is nothing to
+# resolve against, so the file must still validate clean. Registration refuses
+# this assignment anyway — it runs its own portfolio-scoped ref pass after this
+# gate — which is precisely why that behaviour is asserted there, not here.
 code, env = run(assignment)
 check("single-file invocation runs no ref check",
       env["success"] is True and code == 0, json.dumps(env["data"]["errors"]))


claude-cogni-work:
## Ship-ready

Author instance finished authoring. This draft PR is handed off to the merger instance for independent review — it carries `svc:needs-review`, and no automated review verdict has been formed on this side by design (`--role author`: the merger forms its own expectation from the issue before it reads the diff).

**Landed:** issue #1172 → branch `fix/issue-1172-register-entity-ref-gate` → PR #1188 (draft). Issue is `svc:resolved`; the `Closes #1172` link closes it on merge.

**Pre-commit gates run on this side:**

| Gate | Result |
|---|---|
| `cogni-projects` test suites (5) | all green |
| New assignment coverage | 7 new assertions, suite 20 → 27 checks |
| `/simplify` (code) | 4 agents; 1 converged finding applied — gate scoped to assignments |
| Prose simplify (window-scoped) | applied to the 3 changed SKILL.md regions only |
| `skill-quality-reviewer` | round 1 `needs_revision` (3 auto-fixable) → all applied → round 2 **pass** |
| Version-bump gate | 14 plugins scanned, 0 touched |
| Maintainer-breadcrumb guard | 0 violations |
| Token-scope preflight | exit 0 |

**Where a reviewer's attention is best spent** — the two things that are easy to get wrong and that the tests deliberately pin:

1. **The path-identity filter must not be `==` or bare `realpath`.** Bare `realpath` fails *open* on a case-insensitive filesystem (it does not case-fold), dropping the entity's own ref error and registering the dangling assignment. The `<root>/.` test case exists to catch a regression to a naive compare — verified it has teeth: a naive `==` matches 0 of 1 errors there.
2. **The batch must be the whole portfolio.** `_ref_errors([entity_file])` resolves nothing, because the known-slug sets come from the consultant/project records in the batch — every ref would spuriously dangle.

**Deliberately not done, flagged for judgment rather than filed:** this makes `register-entity.py` the second out-of-module consumer of `validate-entities.py`'s underscore-prefixed helpers. The PR body's closing note argues the honest fix is a rename dropping the underscore, not a wrapper — a call about the module's public surface that belongs to a maintainer, not to this issue's scope.

sdh07:

_The merger formed this expectation from issue #1172 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds ~4 lines to `main()` in `cogni-projects/scripts/register-entity.py` at the existing `_ve.validate_file(entity_file)` gate (base line 113), calling the already-loaded validator module's `_ve._entity_files(portfolio_dir)` and `_ve._ref_errors(...)` (defined at `validate-entities.py` base lines 233-248 and 251-309) and keeping only the returned `{entity, file, field, message}` entries that name the entity being registered, so a dangling `consultant`/`project` ref routes into the existing `_fail(..., code=1)` branch (base lines 115-119) before the manifest read (base line 161) and before either `os.replace` (base line 242). The filter must compare normalized paths, because `_entity_files` constructs `os.path.join(portfolio_dir, subdir, name)` (base lines 244-247) while `entity_file` is whatever the caller typed — the same reason the file already uses `os.path.realpath` for the portfolio-scope invariant at base lines 148-157 ('Validation cannot catch this'). No signature change is needed or wanted: `validate_file` keeps its `(errors, warnings)` contract, which `_ref_errors`' own docstring (base lines 258-263) calls out as load-bearing for this caller, and `validate-entities.py`'s single-file shape-only behaviour must stay exactly as `test_validate_entities_refs.sh` section 3 (base lines 143-149) asserts. The prose that now becomes false must move with the code: `register-entity.py`'s docstring (base lines 25-26), `test_validate_entities_refs.sh`'s header claim that the single-file case guards `register-entity.py` (base lines 7-9), and `projects-entities/SKILL.md`'s Step 4/Step 5 passages that name the Step 2 slug comparison as the remaining dangling-ref guard (base lines 118-121, 132-136, 144, 154-158). `test_register_entity.sh` (which today has no validation-failure case at all) must gain error-path coverage proving the manifest and `.metadata/execution-log.json` are untouched on rejection.

**Acceptance checklist:**
- [ ] The new ref check is added inside `main()` of `cogni-projects/scripts/register-entity.py` at or adjacent to the existing `_ve.validate_file(entity_file)` gate (base line 113) — i.e. before the manifest is read (base line 161) and before either `os.replace` (base line 242); the diff shows no ref logic after the temp-file staging block.
- [ ] The check reuses the validator's helpers through the already-loaded module handle (`_ve._entity_files(...)` and `_ve._ref_errors(...)`, or their renamed equivalents), and `register-entity.py` gains NO re-implementation of ref resolution: grep the diff for a new slug-collection loop, a `SUBDIR_TO_TYPE` walk, or a `KEBAB_RE` use in that file and find none.
- [ ] No signature change to the validator's API: `validate_file(filepath)` still returns `(errors, warnings)`, and `_ref_errors(files)` / `_entity_files(path)` keep their single-parameter shapes. If either private helper is renamed to a public name, every existing reference is updated in the same diff — specifically `test_validate_entities_refs.sh`'s monkeypatch block (base lines 267, 274, 285, 302, all spelled `ve._ref_errors`).
- [ ] `validate-entities.py`'s single-file contract is preserved: `test_validate_entities_refs.sh` section 3 ('single-file invocation runs no ref check', base lines 143-149) is still present and still asserts `env["success"] is True and code == 0`; the parent issue's 'Step 4 passes a single file and must not start failing' constraint is not relaxed.
- [ ] `register-entity.py`'s module docstring (base lines 25-26) is updated to state that registration additionally resolves an assignment's `consultant`/`project` refs within the target portfolio.
- [ ] `cogni-projects/tests/test_validate_entities_refs.sh`'s header comment (base lines 7-9) no longer claims the single-file case 'is the regression guard for projects-entities Step 4 and for register-entity.py, which both hand the validator one file at a time' — that sentence is now false for `register-entity.py` and must be corrected.
- [ ] `cogni-projects/skills/projects-entities/SKILL.md` Step 5's parenthetical (base line 144) is updated to say registration also refuses an assignment whose `consultant`/`project` does not resolve within the portfolio; the file's YAML frontmatter block (base lines 1-12: `name`, `description`, `allowed-tools`) remains present and valid.
- [ ] `projects-entities/SKILL.md` no longer asserts anywhere that the Step 2 slug comparison is what guards a dangling ref at registration time — base lines 118-121, 132-136, and 154-158 are corrected or removed rather than left contradicting the new write-path gate.
- [ ] The ref batch is portfolio-scoped from argv: the check is fed `_entity_files(portfolio_dir)` using `main()`'s first argument (base line 99), NOT `os.path.dirname(entity_file)` and not a set spanning two portfolios — confirm the argument expression in the diff.
- [ ] Only entries naming the entity being registered are blocking: the diff shows an explicit filter over `_ref_errors`' returned list keyed on the entity file, so a pre-existing dangling assignment elsewhere in the same portfolio cannot block an unrelated registration.
- [ ] That filter compares NORMALIZED paths on both sides (`os.path.realpath`, `os.path.samefile`, or `os.path.normpath`) rather than raw `==` between two independently-constructed strings — `_entity_files` builds `os.path.join(portfolio_dir, subdir, name)` (base lines 244-247) while `entity_file` is caller-supplied, and the file already uses `realpath` for exactly this class of comparison at base line 148.
- [ ] A dangling ref routes into the existing entity-failed-validation branch: the run prints one JSON line `{"success": false, "data": {...}, "error": ...}` and returns exit code 1 (the `_fail(..., code=1)` path at base lines 115-119), not 2 and not 0.
- [ ] The ref errors reach the caller in `_ref_errors`' standard `{entity, file, field, message}` shape (produced at base lines 302-308) — the offending `field` (`consultant` or `project`) and the unresolved value are surfaced; `register-entity.py` constructs no bespoke ref-error dict of its own.
- [ ] The failure text stays actionable: it does NOT direct the user to run `validate-entities.py` on the single entity file for a ref failure (that invocation is shape-only by design — base lines 12-13 and test section 3 — so it would report zero errors and dead-end the fix loop). It either names the unresolved `consultant`/`project` value inline or points at the portfolio-directory invocation.
- [ ] Nothing is written on the ref failure: after a rejected registration, `projects-portfolio.json` is byte-identical to its pre-run bytes AND `.metadata/execution-log.json` gains no new entry in `transitions[]` (the log is written first, base line 224, so both must be asserted).
- [ ] AC-2 and AC-3 hold unchanged: registering a consultant or project still returns `success: true` / `data.action == "created"`, and registering an assignment whose `consultant` and `project` both resolve still succeeds with the manifest ref carrying `REF_FIELDS` plus `file` (base lines 64-68, 142-158).
- [ ] `cogni-projects/tests/test_register_entity.sh` gains a dangling-`consultant` case AND a dangling-`project` case, each asserting `env["success"] is False`, exit code 1, the manifest byte-identical, and `transitions[]` unchanged.
- [ ] `test_register_entity.sh` gains a resolvable-assignment case (asserting `action == "created"` and `assignments[]` length 1) and a scoping case: a portfolio that already holds a dangling assignment still registers an unrelated consultant successfully.
- [ ] `test_register_entity.sh` gains a non-canonical-path case — the entity file passed with a `./` prefix, as an absolute path while `portfolio_dir` is relative, or through a symlinked portfolio directory — asserting the dangling ref STILL blocks. This is the regression guard for the normalization item; without it a raw-`==` filter passes every other test.
- [ ] Both suites are green with pre-existing coverage intact: `bash cogni-projects/tests/test_register_entity.sh` prints 'All checks passed.' and exits 0 with all six existing sections still present (happy, idempotency, encoding, replace-fail, notadir, modes), and `bash cogni-projects/tests/test_validate_entities_refs.sh` prints 'all assertions passed' and exits 0. Both remain stdlib-only bash + python3 loading the script by file location — no pytest, no pip — per `cogni-projects/CLAUDE.md` lines 37-45.

**Risk watchlist:**
- The decisive failure mode: filtering `_ref_errors`' output with raw `e["file"] == entity_file`. `_entity_files` builds `os.path.join(portfolio_dir, subdir, name)` while `entity_file` is caller-supplied, so any non-identical-but-equivalent spelling drops every entry and the gate silently never fires. Worse, a test written with the same canonical spelling the implementation assumes would still go green — a passing suite is not by itself evidence the gate works.
- Re-implementing ref resolution inside `register-entity.py` and diverging from `_ref_errors`' deliberate anti-cascade rules: the filename-stem admission for a record whose own slug is missing/malformed (base lines 284-292) and leaving an ABSENT ref to the required-key error rather than reporting it as dangling (base lines 298-300).
- Widening `validate_file`'s `(errors, warnings)` return to hand frontmatter back — `_ref_errors`' docstring (base lines 258-263) explicitly rejects this as a contract change for every caller.
- Making the gate portfolio-wide instead of entity-scoped: blocking registration because some OTHER assignment in the portfolio dangles would let one unrelated defect halt all authoring.
- Running the full portfolio scan for consultant/project registrations that carry no refs is wasteful but correct; a short-circuit on `entity_type == "assignment"` is a legitimate variant, not a defect. Do not fail the diff for either choice.
- `_ref_errors` raising inside `register-entity.main()` is NOT caught there — unlike `validate-entities.py`'s per-batch guard (base lines 493-503) — and the test harness drives `main()` directly (base lines 85-90), bypassing the `__main__` catch-all (base lines 284-287). An exception would escape as a traceback rather than the `{success, data, error}` envelope every caller parses.
- Changing `_fail`'s hardcoded `"data": {}` (base lines 71-75) to carry error lists would alter the failure envelope every consumer parses. AC-1's wording invites exactly this over-reach; a human should eyeball whether the envelope shape moved.
- Ordering the new check before the existing `os.path.isfile(manifest_path)` / `os.path.isfile(entity_file)` guards (base lines 102-108) would change which failure a bad invocation reports and could make `_entity_files` walk a nonexistent directory.
- Deleting the post-registration portfolio sweep from SKILL.md Step 5 entirely: the per-entity guarantee now lives in the script, but the sweep still has value as a portfolio-wide audit. Editorial judgment no gate can decide.
- An assignment file placed at the portfolio root rather than under `assignments/` is invisible to `_entity_files`, so the new gate passes vacuously for it. Pre-existing and out of scope — but the fix must not deepen the hole.

- The plan-time acceptance criteria list 4 items, all happy/sad-path behavioral. A base-grounded regeneration derives 20 gating items, including several the AC never asked for and that a wrong fix would plausibly miss: the path-normalization requirement in the entity-scoping filter (the single most likely silent no-op), the dead-end remediation message, the assertion that `.metadata/execution-log.json` is also left untouched on failure, and three stale-prose corrections the change strands. The acceptance bar was materially under-specified. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1188

**Summary:** The production ref gate itself is correct — plugin-validator independently probed nine edge cases against the live script and every one behaved as the contract requires — but three test-coverage gaps (contract items 15, 17, 18) were independently flagged by both contract-grading-reviewer and plugin-validator, plugin-validator additionally found an unguarded os.listdir letting PermissionError escape main() for in-process callers, and both skill reviewers found stale SKILL.md prose left in tension with the new write-path gate. Revision needed for the test regression-guards and the SKILL.md residue; no change to the gate logic itself is required.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (11)

- **[major]** (auto-fixable) `review-contract:item-17 (cogni-projects/tests/test_register_entity.sh:243-290)` — VIOLATED. Item 17 requires BOTH a dangling-consultant AND a dangling-project case. Only one dangling assignment is constructed — make_assignment(root5, "ana-silvaa", "nordic-erp"), a misspelled consultant with a resolving project. No assignment with a valid consultant and a dangling project appears anywhere in the diff, so the project half of _ref_errors' two-field loop is untested on the write path. Behaviour verified correct out-of-band, so this is a regression-guard gap, not a live defect — but the gate exists so a future edit cannot silently move the check below the log write. Flagged independently by plugin-dev:plugin-validator.
  - *Suggested fix:* Add a second assignment with a resolvable consultant and a misspelled project (e.g. "nordic-erpp"), asserting success is False, code 1, errs[0]["field"] == "project", manifest byte-identical and transitions[] unchanged.

- **[major]** (auto-fixable) `review-contract:item-15 (cogni-projects/tests/test_register_entity.sh:266,282)` — VIOLATED. Item 15 requires asserting BOTH that projects-portfolio.json is byte-identical AND that .metadata/execution-log.json's transitions[] is unchanged after a rejected ref-failure registration. The refs block only checks registered_assignments() == [] (a parsed-array check, not byte-identical); no assertion reads execution-log.json or transitions[] at all. The behaviour holds — both files were confirmed byte-identical after a rejection — the assertions are simply not there. Flagged independently by plugin-dev:plugin-validator.
  - *Suggested fix:* Snapshot both projects-portfolio.json's raw bytes and .metadata/execution-log.json's transitions[] length before each failing register() call and assert both unchanged after, matching the before/after byte comparison the replace-fail section already uses.

- **[major]** (auto-fixable) `review-contract:item-18 (cogni-projects/tests/test_register_entity.sh:288-292)` — VIOLATED in part. The resolvable-assignment case asserts success and action == "created" but never calls the registered_assignments() helper defined two lines earlier to confirm assignments[] has length 1, so a duplicate append would pass. Flagged independently by plugin-dev:plugin-validator.
  - *Suggested fix:* Add len(registered_assignments()) == 1 to the resolvable-assignment check.

- **[minor]** (auto-fixable) `cogni-projects/scripts/register-entity.py:186-192` — The raise-free posture is incomplete and its justifying comment is narrower than it reads. The comment covers _ref_errors (true — parse_frontmatter is pure regex and cannot raise) but not _entity_files, whose os.listdir is unguarded. Reproduced: with consultants/ at mode 000, main() raises PermissionError straight out to the caller. Via the CLI the __main__ catch-all converts it to the envelope, but the test harness and any other in-process caller — the exact audience _same_file's own docstring cites — get a traceback instead of {success, data, error}. Before this diff main() had no unguarded OSError source.
  - *Suggested fix:* Wrap the two calls: try: ref_errors = [...] except OSError as exc: return _fail("cannot scan portfolio for refs: %s" % exc). One line, and it makes the docstring's raise-free claim true for the whole block.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:37-42` — The recovery claim 'That state is repaired by re-running the skill for the same slug: registration is idempotent' was written for the interrupted-run case. The change introduces a second, now more common way to reach file-on-disk-but-unregistered: an assignment rejected by the new ref gate. For that case a bare re-run fails identically, so the stated repair is incomplete and will send a model into a loop.
  - *Suggested fix:* Append: '— unless registration was refused for an unresolved consultant / project ref, in which case resolve the ref first (author the missing record or correct the slug), since a re-run otherwise fails the same way.'

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:76-79` — Step 2 still ends with 'that slug match, not the file's existence, is the guard' — the exact claim the change moved onto the write path, while line 119-120 now says 'Step 5's registration is the gate that enforces ref resolution'. A model reading Step 2 and Step 4 in one pass gets two different answers to 'what stops a dangling ref?'. skill-reviewer judged it does not breach contract item 8 on a strict reading, since the clause's real contrast is slug-vs-filename resolution, which remains accurate.
  - *Suggested fix:* Reword to describe resolution rather than enforcement, naming Step 5's registration as what refuses it — e.g. '— that slug match, not the file's existence, is what makes the ref resolve.'

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:162-163` — 'A registration that fails on a dangling ref wrote nothing' is the right guarantee but ambiguous about scope — the entity file was already written in Step 3, so 'wrote nothing' can be read as 'the file is gone' and invite a needless rewrite.
  - *Suggested fix:* State the exact effect: manifest and execution-log untouched, entity file stays on disk as written in Step 3.

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:106-108` — Unchanged Step 4 line asserts unqualified that 'Step 5's script reports only an error count and points back here', while new lines 148-152 correctly state ref failures come back as full data.errors[] entries. The implementation confirms the split.
  - *Justification:* Lines 145-148 already scope the claim to 'those shape failures' downstream.

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:118-119,121-122,137-138` — 'Step 4 runs no ref check' is stated three times in about twenty lines, and the third instance attaches it as a justification it does not support — it gives a reason the sweep is not redundant, not a reason Step 4 is still required.
  - *Justification:* Redundancy and a weak connective; no incorrect instruction.

- **[minor]** (waived) (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:165-167` — The post-registration sweep keeps its base-era rationale 'it is only worth running once the assignment is in the manifest', which no longer follows from the sweep's new stated purpose. The sweep reads entity files on disk, not the manifest, so manifest membership is irrelevant to what it sees.
  - *Justification:* Vestigial justification only; the instruction it attaches to is right.

- **[minor]** (waived) (auto-fixable) `review-contract:item-14 (cogni-projects/scripts/register-entity.py:195-197)` — Partially satisfied. The hard prohibition is met — the message does NOT send the user to validate-entities.py on the single file. But its disjunction (name the unresolved value inline or point at the portfolio-directory invocation) is met by neither branch literally; the text points at data.errors[]. In practice the unresolved value is on the same stdout line, so the fix loop is intact for any caller parsing the envelope — but a human reading only `error` in a terminal gets a count, not the slug.
  - *Justification:* Satisfied in intent; every documented caller parses the envelope, where the offending slug is present.

### Positive observations

- The production ref gate is correct. plugin-validator probed the live script on nine edge cases — dangling project, both-fields-dangling, abs-vs-relative, symlinked portfolio root, case-variant path, hard-linked out-of-portfolio file, root-level assignment, unreadable subdir, and both happy paths — and every one behaved as the contract requires: exit 1, {success:false}, and BOTH projects-portfolio.json and .metadata/execution-log.json byte-identical after rejection.
- Contract item 11 — the single most likely silent no-op — is genuinely defeated, not decoratively. _same_file uses os.path.samefile (st_dev/st_ino), and the docstring's case-folding argument is not theoretical: passing the entity as <root>/ASSIGNMENTS/<file>.md on the case-insensitive default macOS filesystem still fired the gate, where a realpath == filter would have dropped the error and failed OPEN.
- Item 19's non-canonical-path test is a real guard: register(os.path.join(root5, "."), dangling) makes a raw-== filter match nothing, so the gate would fail open and the check goes red.
- The ref check sits after the validate_file gate and before both the manifest read and every os.replace (item 1), is scoped by entity_type == "assignment" (item 16), and is fed _entity_files(portfolio_dir) from main()'s first argument (item 9).
- _fail's additive data=None -> data or {} is backward-compatible for all nine existing call sites; the shape-failure and missing-file envelopes still emit "data": {} verbatim. The over-reach the contract's escalation warned about did not happen.
- No ref resolution leaked into register-entity.py — the anti-cascade rules stay solely in _ref_errors and errors reach the caller verbatim in {entity, file, field, message}.
- validate-entities.py is not touched at all; the single-file shape-only contract and its monkeypatch test block are preserved exactly.
- Every stranded prose claim moved with the code (items 5-8): the module docstring, the test header's now-false 'regression guard for register-entity.py' sentence, and all three SKILL.md passages were rewritten rather than left contradicting the write-path gate.
- The portfolio-wide sweep is preserved rather than deleted and re-justified under the new division of labour — the right editorial call on an item no automated gate could decide.
- Failure guidance routes the fix loop back to the registration command and pointedly does not send the user to the shape-only single-file validator, which would report zero errors and dead-end.
- Both test suites pass with all six pre-existing test_register_entity.sh sections green; structural validation clean (plugin.json parses, semver version, 4 skills with valid frontmatter, no credentials, both suites stdlib-only).
- No INTENT-UNCLEAR findings were required — every contract item was adjudicable.

### Recommendation

⚠️ **Revise before merge.** 7 finding(s) need attention.

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1188

**Summary:** All three perspectives passed; the only two findings (both minor, from plugin-validator) are explicitly waived with substantive justifications, so the PR is clean to ship. This revision closed all 7 actionable findings from the prior NEEDS_REVISION verdict — three test regression-guards (contract items 15, 17, 18), a raise-free portfolio ref scan, and three SKILL.md prose fixes — plus one further guard-widening raised during this review pass.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (waived) (auto-fixable) `cogni-projects/references/data-model.md` — The Manifest Registration section does not mention that registration refuses an assignment whose consultant/project ref does not resolve within the portfolio.
  - *Justification:* Waived: pre-existing documentation omission (the section was already silent on the existing validate_file gate), separable doc-sync work.

- **[minor]** (waived) `cogni-projects/scripts/register-entity.py` — Registration fails open for an assignment file that _entity_files does not enumerate (off-convention placement): such a file never appears in the portfolio walk, so the _same_file filter is empty and the ref gate passes it through.
  - *Justification:* Waived: the gap belongs in _entity_files, not in this PR's register-time gate, and is out of scope for the portfolio-scoped ref gate. Directory-mode validation shares the same blind spot.

### Positive observations

- The widened guard is an alignment with the sibling call site, not a blanket loosening: validate-entities.py wraps the same _ref_errors call in `except Exception ... # noqa: BLE001`, and register-entity.py now carries the identical comment token. (plugin-dev:plugin-validator)
- The new test is a genuine regression guard, not a tautology: reverting only the guard to `except OSError` produces exactly one failure. (plugin-dev:plugin-validator)
- Refusal is genuinely side-effect-free: manifest and execution log verified byte-identical after both a dangling-consultant and a dangling-project refusal. (plugin-dev:plugin-validator)
- Both ref fields are covered independently, exercising both arms of _ref_errors' per-field loop rather than half of it. (plugin-dev:plugin-validator)
- plugin.json correctly left at v0.0.16 for the post-merge bump workflow, with no marketplace.json desync. (plugin-dev:plugin-validator)
- Core concept draws the interrupted-run vs precondition-failure distinction precisely, and the re-run advice is now scoped to the case it actually repairs. (plugin-dev:skill-reviewer)
- The division of labour between Step 4 and Step 5 is stated consistently in both directions, so a reader arriving at either step learns what the other does and does not cover. (plugin-dev:skill-reviewer)
- Progressive disclosure is sound: per-type field lists marked a quick reference with references/data-model.md declared canonical and given precedence on disagreement. (plugin-dev:skill-reviewer)
- All three prior prose findings closed cleanly with no new drift; file remains well under its char budget (9799 vs 19500 cap), breadcrumb-free. (cogni-service:skill-quality-reviewer)
- Acceptance-criteria fit is complete against issue #1172: AC1 covered by Step 5's data.errors[] description plus the untouched-manifest sentence, AC2/AC3 by the scoping qualifier. (cogni-service:skill-quality-reviewer)
- Voice holds per skill-creator: imperative throughout, no CAPS-heavy MUST stacks, each constraint carrying its rationale. (cogni-service:skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addresses the 7 actionable findings from the NEEDS_REVISION verdict (which graded head `2d2069d`). The 4 findings marked *(waived)* in that verdict were left byte-identical, as waived.

**Commits:** `7573cf9c` (the 7 findings) · `3f2da88c` (one further finding raised during the re-review)

### Findings resolved

| Finding | Resolution | Commit |
|---|---|---|
| **[major]** item-17 — only a dangling-*consultant* case; the project half of `_ref_errors`' two-field loop untested on the write path | Added a resolving-consultant / misspelled-project fixture (`ana-silva` / `nordic-erpp`, kebab-valid so it fails as a ref error, not a composite-slug shape error), asserting `success False`, code 1, a single error with `field == "project"` | `7573cf9c` |
| **[major]** item-15 — rejected registrations only checked `registered_assignments() == []`; no byte-identical manifest assertion, no read of `execution-log.json` at all | Added `manifest_bytes()` / `transition_count()` snapshots around **all three** failing `register()` calls via a `check_untouched()` helper | `7573cf9c` |
| **[major]** item-18 — the resolvable case never confirmed `assignments[]` length, so a duplicate append would pass | Added `len(registered_assignments()) == 1` | `7573cf9c` |
| **[minor]** `register-entity.py:186-192` — `_entity_files`' unguarded `os.listdir` let `PermissionError` escape `main()` for in-process callers | Wrapped the ref scan, returning the standard envelope at **code 2** (environment failure, never code 1 — the entity did not fail validation). Reproduced against the old head, then pinned by a new test | `7573cf9c` |
| **[minor]** `SKILL.md:37-42` — the idempotent-re-run recovery claim was incomplete for a ref-gate rejection | A refusal is now named a *precondition failure* a re-run does not repair, pointing at Step 5 rather than restating the remedy | `7573cf9c` |
| **[minor]** `SKILL.md:76-79` — Step 2's "is the guard" contradicted Step 4 on what enforces | Step 2 now describes **resolution**; Step 5 is named as what **refuses** | `7573cf9c` |
| **[minor]** `SKILL.md:162-163` — "wrote nothing" ambiguous, since Step 3 already wrote the file | Now states the exact effect: manifest and execution log untouched, the Step 3 entity file stays on disk | `7573cf9c` |
| **[minor]** *(raised during re-review)* the new guard caught only `OSError`, so a `ValueError` from `parse_frontmatter` still escaped `main()` | Widened to `except Exception` with the same posture and `# noqa: BLE001 — the envelope is the contract` token as `validate-entities.py`'s own `main()` at the identical call site | `3f2da88c` |

### Waived findings — unchanged, as waived

`SKILL.md:106-108` (error-count claim scoped downstream) · `SKILL.md:118-122,137-138` (thrice-stated "Step 4 runs no ref check") · `SKILL.md:165-167` (vestigial sweep rationale) · `review-contract:item-14` (failure message satisfied in intent)

### Verification

Both new guards were confirmed **load-bearing**, not decorative — each goes red against the pre-fix script:

- unreadable subdir → old head raises `PermissionError` out of `main()`; new head returns `{success: false}` at code 2
- monkeypatched `ValueError` → red against the narrower `except OSError`

All five `cogni-projects` suites green at `3f2da88c`. `plugin.json` deliberately **not** bumped — the post-merge workflow owns that.

Two waived minors remain on the fresh verdict: a pre-existing `data-model.md` doc omission, and a fail-open for off-convention assignment placement that belongs in `_entity_files` rather than this gate.

sdh07:

_The merger formed this expectation from issue #1172 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds a portfolio-scoped ref gate to `main()` in cogni-projects/scripts/register-entity.py at the existing `_ve.validate_file(entity_file)` gate (base lines 110-119), reusing the already-loaded validator module (loaded by file location at base lines 45-57): call `_ve._ref_errors(_ve._entity_files(portfolio_dir))` (defined in cogni-projects/scripts/validate-entities.py at base lines 233-248 and 251-309) and keep only the entries whose `file` is the entity being registered. The batch must be built from `portfolio_dir`, never from `entity_file` — `_entity_files` on a file returns `[path]` (base validate-entities.py line 239-240), which collects no consultant/project slugs and would make every assignment ref spuriously dangle. The filter's path comparison cannot be `==` on the raw argv strings, because `_ref_errors` builds `file` by joining `portfolio_dir` (validate-entities.py line 247) while `entity_file` is an independent argv string that may be relative, absolute, or symlinked; an inode-level `os.path.samefile` check (guarded against the `OSError` it raises) is the shape that does not fail open. The gate must fire before the manifest read at base register-entity.py line 161 and before the atomic temp-dump/`os.replace` block (base lines 224-243), fail with code 1 (the docstring's "the entity failed validation" code, base line 35), and surface the raw `{entity, file, field, message}` dicts machine-readably — which means `_fail` (base lines 71-75, currently hardcoding `"data": {}`) gains an optional data argument with a default so every existing call site is unchanged. Behavioural coverage belongs in cogni-projects/tests/test_register_entity.sh (whose `make_portfolio` helper, base lines 54-67, creates only `consultants/` and has no assignment case at all), and the stale prose in cogni-projects/skills/projects-entities/SKILL.md (base lines 114-136 and 144) plus the header of cogni-projects/tests/test_validate_entities_refs.sh (base lines 7-9) must be reconciled, since both currently assert that the registration path performs no ref check.

**Acceptance checklist:**
- [ ] register-entity.py's `main()` gains a call to the validator's ref pass whose batch argument is the portfolio expansion — grep shows `_ref_errors(` receiving `_entity_files(portfolio_dir)` (or the equivalent on `portfolio_dir`), and NOT `[entity_file]` / `_entity_files(entity_file)`; a single-file expansion collects no consultant or project slugs, so that variant would dangle every ref.
- [ ] The new ref gate is placed before any write: in the diff the `_ref_errors` call and its `return _fail(...)` appear above the manifest `open(manifest_path, "r")` (base line 161) and above the temp-dump / `os.replace` block (base lines 224-243), and it is not nested inside a branch that a given entity type can skip.
- [ ] The returned errors are filtered down to the entity being registered — the diff shows a per-error comparison of each dict's `file` against `entity_file`, so a dangling ref on some *other* assignment in the same portfolio is dropped rather than blocking this registration.
- [ ] That comparison is inode/realpath-normalizing, not string equality: the diff uses `os.path.samefile` (or an equivalent normalizing check) rather than `==` on the raw argv strings, and the `OSError` `samefile` raises for a missing path is caught rather than allowed to escape.
- [ ] `_fail` keeps every existing call site byte-identical: its new parameter carries a default (e.g. `data=None`) and emits `data or {}`, and grep of the ~12 pre-existing `_fail(` call sites in register-entity.py shows none of them gained an argument — each still prints `"data": {}`.
- [ ] The dangling-ref failure envelope has `success: false`, returns exit code 1 (the docstring's "entity failed validation" code, not 2), and carries the raw ref-error dicts machine-readably in the envelope (e.g. `data.errors[]`), each with exactly the keys `entity`, `file`, `field`, `message` — not merely interpolated into the `error` string.
- [ ] test_register_entity.sh gains a case registering an assignment whose `consultant` does not resolve in the portfolio: it asserts `success is False`, exit code 1, an error dict with `field == "consultant"` and `file ==` the assignment path, AND that `projects-portfolio.json` is byte-identical to before (assignments[] gains no ref) with no new `.metadata/execution-log.json` transition and no `.tmp` debris.
- [ ] test_register_entity.sh also covers a dangling `project` ref (AC 1 names both fields) with a case asserting an error dict whose `field == "project"`.
- [ ] test_register_entity.sh gains the positive assignment case that no existing test provides: an assignment whose `consultant` and `project` both resolve registers with `success is True`, `data.action == "created"`, and the manifest `assignments[]` gaining the ref — this is the assertion that catches a batch built from the single entity file.
- [ ] test_register_entity.sh proves the filter's scoping behaviourally: registering a consultant or project into a portfolio that already holds a *dangling* assignment still succeeds (the new gate must not fail closed on refs the caller is not registering).
- [ ] test_register_entity.sh proves the path comparison does not fail open: the same dangling assignment is still refused when `portfolio_dir` and `entity_file` are passed in differing-but-equivalent forms (absolute dir with relative file, a trailing-slash or `./`-prefixed dir, or a symlinked portfolio root).
- [ ] Both suites are green on the branch: `bash cogni-projects/tests/test_register_entity.sh` and `bash cogni-projects/tests/test_validate_entities_refs.sh` exit 0, and no pre-existing assertion in either file was weakened or deleted — in particular test_validate_entities_refs.sh case 3 ("single-file invocation runs no ref check", base lines 143-149) is retained unchanged, because `validate_file`'s single-file contract is deliberately untouched.
- [ ] test_validate_entities_refs.sh's header comment (base lines 7-9), which currently names register-entity.py as a consumer that "hands the validator one file at a time" and treats the single-file case as its regression guard, is corrected so it no longer implies the registration path performs no ref check.
- [ ] register-entity.py's module docstring (base lines 25-26 and the Exit line 35) is updated to state that registration additionally resolves the assignment's `consultant`/`project` refs within the target portfolio and that exit 1 covers that refusal.
- [ ] projects-entities/SKILL.md Step 5 documents the new refusal: registering an assignment with an unresolvable `consultant`/`project` fails, naming the `data.errors[]` four-key entries and the fix loop (author or correct the referenced record, then re-run).
- [ ] No surviving SKILL.md sentence contradicts the new behaviour: base lines 118-121 ("a passing run is not evidence the refs exist … confirming in Step 2 … remains the guard against a dangling reference"), base lines 132-136 (the unprompted post-Step-5 sweep "is what turns Step 2's read-and-compare into a checked fact") and base line 144 ("it re-runs the validator itself, so it refuses an entity the Step 4 gate would have caught") are reconciled, while the still-true statement that the *single-file* Step 4 call runs no ref check is preserved; the frontmatter fence, `name: projects-entities`, `description`, and `allowed-tools` keys are unchanged in shape.

**Risk watchlist:**
- The batch is built from the entity file instead of the portfolio directory, so every assignment registration fails with both refs dangling — a fix with no positive assignment test would ship this undetected, since the existing suite registers only consultants.
- The filter is dropped entirely, so any pre-existing dangling assignment anywhere in the portfolio blocks registration of an unrelated consultant — a fail-closed regression the current tests (portfolios with no assignments) cannot catch.
- The filter compares `e["file"] == entity_file` on raw argv strings, or uses bare `realpath` equality: a relative-vs-absolute, trailing-slash, symlinked, or case-variant path (APFS does not case-fold under realpath) drops the entity's own error and the gate silently fails open — the most dangerous wrong fix, because it passes every happy-path test.
- The check is placed after the manifest read or after the execution-log write, so a refused assignment still mutates portfolio state or leaves `.tmp` debris.
- `_fail` is changed in a way that shifts existing call sites (positional insertion, required parameter), altering the envelope of unrelated failures the atomic-write suite asserts on.
- The failure is reported only as a prose `error` string with no machine-readable `{entity, file, field, message}` dicts, so the skill's documented fix loop cannot name the offending field — AC 1's shape requirement met in letter, not in effect.
- `validate_file`'s single-file behaviour is widened to run a ref pass, breaking Step 4 and the explicit single-file regression guard in test_validate_entities_refs.sh — the exact scope constraint the parent issue imposed.
- Exit code drift: the ref refusal returns 2 (environment failure) instead of 1, so a caller distinguishing "bad entity" from "bad environment" mis-routes it.
- An exception raised inside `_ref_errors` on the registration path escapes to register-entity.py's module-level catch-all and reports as a generic "unexpected failure" at exit 2 rather than a contained, actionable error — judgment call whether the registration path deserves the same per-batch containment validate-entities.py's `main()` has (base lines 493-503).
- Every registration now reads the frontmatter of every entity in the portfolio — an O(n) cost added to a previously O(1) write path; acceptable at portfolio scale but worth a human eyeball if a portfolio is expected to hold thousands of records.
- An assignment file not named `*.md` is not collected by `_entity_files`, so the filter finds no error and the ref check silently does not apply to it — a narrow inherited hole, not introduced here, but a wrong fix could widen it.
- Documentation half-measure: the code gate lands but SKILL.md still tells the author that the Step 2 slug comparison is the only guard, leaving the skill's prose actively misleading about where the integrity boundary now sits.

- The plan-time acceptance criteria list 4 outcomes (dangling ref refused, consultants/projects unaffected, resolvable assignment unaffected, existing suite green), but a base-grounded regeneration derives 16 gating items — including several substantive behaviours the AC never asked for: the filter must not block registration of an unrelated entity when some *other* assignment in the portfolio dangles (a fail-closed regression the stated AC would pass); the path comparison must not fail open on relative/absolute/symlinked/case-variant path forms; the failure must carry the four-key dicts machine-readably at exit 1 rather than only in the prose `error` string; and the now-false prose in projects-entities/SKILL.md (base lines 118-121, 132-136, 144), register-entity.py's docstring, and test_validate_entities_refs.sh's header must be reconciled. The acceptance bar was materially under-specified. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1188

**Summary:** Two of four perspectives passed cleanly, but contract-grading-reviewer independently flags two committed checklist items (item 7's dangling-consultant test case and item 14's module-docstring Exit line) as major, concrete gaps rather than intent ambiguity, and skill-reviewer/skill-quality-reviewer both flag the same SKILL.md Step 5 sentence for omitting the data.errors[] four-key shape — Rule 3 (explicit needs_revision verdicts) and Rule 4 (major findings) both fire.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (8)

- **[major]** (auto-fixable) `review-contract:item-14` — register-entity.py's module docstring prose was updated to describe the new ref resolution (and does say 'Each unresolved ref is reported in data.errors[].'), but the 'Exit:' line (line 40: "Exit: 0 ok / 1 the entity failed validation / 2 usage or environment failure.") is byte-identical to base and does not state that exit code 1 also covers a dangling-ref refusal, as the contract item requires.
  - *Suggested fix:* Update the Exit line to name the refusal, e.g. 'Exit: 0 ok / 1 the entity failed validation, including an unresolved assignment ref / 2 usage or environment failure.'

- **[major]** (auto-fixable) `review-contract:item-7` — The new dangling-consultant test case in cogni-projects/tests/test_register_entity.sh (the 'refs: dangling consultant ref fails registration' block, lines 295-303) asserts success is False, code == 1, set(errs[0]) == {entity,file,field,message}, errs[0]["entity"] and errs[0]["field"], but never asserts errs[0]["file"] equals the dangling assignment's path, and never calls the file's own tmp_debris() helper (used at lines 148/189/208 for other cases) against root5 — both explicitly required by the contract item.
  - *Suggested fix:* Add `errs[0]["file"]` compared via os.path.samefile against `dangling` to the shape check at line 299, and add a `check("refs: dangling consultant ref leaves no tmp debris", tmp_debris(root5) == [], ...)` assertion alongside the existing check_untouched() call for that case.

- **[minor]** (auto-fixable) `cogni-projects/references/data-model.md:269-288` — Progressive-disclosure gap (outside the contract's explicit scope, but the skill calls this file canonical). The reference's Validation section describes referential integrity purely as a property of validate-entities.py's directory argument, and the Manifest Registration section describes register-entity.py as an upsert-plus-date-bump with no mention of the new ref gate. Nothing there is now false, but a reader routed to the canonical document learns nothing about where the integrity boundary actually sits.
  - *Suggested fix:* Add one sentence to the Manifest Registration section noting that registering an assignment additionally resolves its consultant/project refs against the target portfolio and refuses a dangling ref with the same {entity, file, field, message} errors, exit 1, leaving the manifest untouched.

- **[minor]** (auto-fixable) `cogni-projects/scripts/register-entity.py:40` — Checklist item 14 asks for the module docstring's Exit line to be updated so exit 1 explicitly covers the new ref refusal. The prose block at lines 27-31 was added and does describe the portfolio-scoped ref resolution, but line 40 is byte-identical to base: 'Exit: 0 ok / 1 the entity failed validation / 2 usage or environment failure.' A reader mapping the new 'cannot scan portfolio for refs' (code 2) against the ref refusal (code 1) has to infer the split from the code, since the docstring's exit legend never names it.
  - *Suggested fix:* Extend line 40 to name the refusal, e.g. 'Exit: 0 ok / 1 the entity failed shape or ref validation / 2 usage or environment failure (including an unreadable portfolio during the ref scan).'

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md body §Step 4 (lines 118-142)` — The single fact 'the single-file gate runs no ref check; Step 5's registration is the ref gate' is now restated in four separate places. Each restatement is individually accurate — this is not a contradiction — but the accumulation makes Step 4 the longest step in the skill while its actual operational instruction is two paragraphs from the top, and it costs context on every trigger for no added decision.
  - *Suggested fix:* Compress the second sweep paragraph to one sentence that preserves the instruction without re-deriving the rationale, and drop the trailing 'never in place of the Step 4 gate — that gate runs no ref check', which line 124 already establishes.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:110-112` — Step 4's unchanged sentence 'Run this step even though Step 5's script validates again: that script reports only an error count and points back here' is now overbroad. After this PR, register-entity.py reports shape failures as a count but ref failures as full per-field dicts in data.errors[]. Step 5 correctly qualifies this as 'those shape failures'; Step 4 does not, so read alone it contradicts the new behaviour.
  - *Suggested fix:* Qualify the Step 4 sentence the same way Step 5 does: '…that script reports shape failures only as an error count and points back here…'.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-entities/SKILL.md:152-156` — Two gaps in one sentence. (a) The contract's SKILL.md item asks Step 5 to name the data.errors[] four-key entries; the text says only 'a full data.errors[] entry naming the offending field and the missing referent' and never enumerates {entity, file, field, message} — the four keys appear nowhere in the file. (b) 'resolves the consultant / project refs against the whole portfolio' describes the batch scope, not the failure scope, and reads as though any dangling assignment anywhere in the portfolio blocks this registration; line 139 states the opposite, so the two sentences pull against each other.
  - *Suggested fix:* Replace the sentence with one that separates batch scope from failure scope and enumerates the four keys: '…resolves the refs against the whole portfolio — the portfolio supplies the records to resolve against, but only the refs of the assignment being registered can fail the run. Each unresolved ref comes back as a data.errors[] entry with the four keys entity, file, field, message.'
  - *Raised by:* cogni-service:skill-quality-reviewer, also flagged by plugin-dev:skill-reviewer

- **[minor]** (auto-fixable) `cogni-projects/tests/test_register_entity.sh:296-302` — Checklist item 7 enumerates four assertions for the dangling-consultant case; two are missing. (a) The error dict's file is never compared against the assignment path — the test asserts only set(errs[0]) == {entity,file,field,message} plus entity/field, so a filter that returned some other file's error with the right field would still pass. (b) No .tmp debris assertion, even though the suite already defines tmp_debris(root) at line 132 and uses it in the happy, replace-fail, and notadir cases. Both gaps are latent rather than live: the gate returns before any mkstemp call, so debris is impossible by construction, and the _same_file filter is what selects the error in the first place. Test completeness only.
  - *Suggested fix:* Add errs[0]["file"] compared via os.path.samefile(errs[0]["file"], dangling) to the shape check at line 298, and add check("refs: no tmp debris on a refused registration", tmp_debris(root5) == [], str(tmp_debris(root5))) inside check_untouched so all three rejection cases cover it.

### Positive observations

- Item 1 satisfied exactly as specified: register-entity.py:197 calls _ve._ref_errors(_ve._entity_files(portfolio_dir)). The mutant that swaps portfolio_dir for entity_file fails 6 checks, including the positive assignment case — the primary risk-watchlist failure mode is caught. (plugin-dev:plugin-validator)
- Item 4 satisfied with the strongest available shape: _same_file (register-entity.py:84-106) uses os.path.samefile, catches the OSError, and falls back to realpath equality. The mutant replacing it with e["file"] == entity_file fails 5 checks, led by 'refs: still rejected when the portfolio is passed as <root>/.' — the fail-open case the watchlist calls the most dangerous wrong fix is behaviourally guarded. (plugin-dev:plugin-validator)
- Item 3 and item 10 satisfied together: dropping the per-entity filter fails 3 checks, including 'refs: a project still registers while a dangling assignment exists'. The fail-closed regression that the plan-time AC would have passed is covered. (plugin-dev:plugin-validator)
- Item 5 satisfied: _fail gained data=None with data or {} (lines 76-81), and all 13 pre-existing call sites are unchanged. (plugin-dev:plugin-validator)
- Item 12 satisfied: both suites exit 0 (42 and 29 checks); test_validate_entities_refs.sh case 3 'single-file invocation runs no ref check' is retained byte-identical, so validate_file's single-file contract is provably untouched. (plugin-dev:plugin-validator)
- The escaping-exception risk on the watchlist is handled deliberately: the try/except at lines 195-204 returns code 2 with 'cannot scan portfolio for refs', covered by two tests (unreadable consultants/ with a root-user skip guard, and an injected non-OSError via monkeypatching, restored in a finally). (plugin-dev:plugin-validator)
- Item 16 satisfied thoroughly: all three contradicting SKILL.md passages are reconciled while the still-true single-file statement survives verbatim; frontmatter untouched. (plugin-dev:plugin-validator)
- Plugin structure is clean and the branch correctly does NOT bump version, so the version-touched gate stays green. (plugin-dev:plugin-validator)
- Contract item 16 fully satisfied: all three named stale passages are reconciled and none of the replacements over-corrects. (plugin-dev:skill-reviewer)
- The still-true single-file statement is preserved verbatim in both places it matters, so the prose does not accidentally imply validate_file was widened — the exact scope constraint issue #1172 imposed. (plugin-dev:skill-reviewer)
- Contract item 15's fix loop is present and correctly grounded in the implementation: the refusal leaves the manifest and execution log untouched and the Step 3 file on disk. (plugin-dev:skill-reviewer)
- The new Core-concept paragraph pre-empts the most likely misread of the new failure — that a refused registration is the same repairable interrupted-run state. (plugin-dev:skill-reviewer)
- Frontmatter is byte-identical to base: fence, name, the multi-line description with its trigger-phrase list, and allowed-tools untouched. (plugin-dev:skill-reviewer)
- Step 2's edit is a genuine improvement, not just a contradiction patch: it demotes the slug comparison from 'the guard' to an earlier catch and names Step 5 as the enforcer. (plugin-dev:skill-reviewer)
- Body length (9,799 chars) stays well under the 19,500-char cap despite the additions, and the linked reference resolves correctly. (plugin-dev:skill-reviewer)
- The contract's hardest SKILL.md item is met: no surviving sentence asserts the registration path performs no ref check. (cogni-service:skill-quality-reviewer)
- Core concept distinguishes a refused registration from an interrupted run as a precondition failure, which keeps the idempotent-rerun repair advice from being misapplied to a dangling ref. (cogni-service:skill-quality-reviewer)
- The fix loop is present and actionable: manifest and execution log untouched, entity file stays on disk, correct the slug or author the missing record and re-run. (cogni-service:skill-quality-reviewer)
- Description is third-person, states both what and when, and carries seven concrete trigger phrases plus an explicit under-trigger hedge. Frontmatter shape unchanged. (cogni-service:skill-quality-reviewer)
- No maintainer-meta breadcrumbs found in the body or description. (cogni-service:skill-quality-reviewer)
- Voice is imperative and reasoned throughout — no CAPS-stacked MUSTs; emphasis is sparing and always paired with the reason. (cogni-service:skill-quality-reviewer)
- item 2 satisfied: the ref gate (lines 194-212) sits above the manifest open at line 235 and above the temp-dump/os.replace block at lines 298-337, in main()'s single linear flow. (cogni-service:contract-grading-reviewer)
- item 3 satisfied: line 198 filters with if _same_file(e["file"], entity_file); the 'a project still registers while a dangling assignment exists' and 'single reported error' checks confirm it behaviourally. (cogni-service:contract-grading-reviewer)
- item 6 satisfied: the dangling-ref path returns _fail(..., code=1, data={"errors": ref_errors}) and _ref_errors builds each dict with exactly the four keys. (cogni-service:contract-grading-reviewer)
- item 8 satisfied: the dangling-project case asserts errs[0]["field"] == "project". (cogni-service:contract-grading-reviewer)
- item 9 satisfied: the resolving-assignment case checks success True, data.action == 'created', and assignments[] moving 0 -> 1. (cogni-service:contract-grading-reviewer)
- item 11 satisfied: 'still rejected when the portfolio is passed as <root>/.' proves the comparison does not fail open on a non-canonical path form. (cogni-service:contract-grading-reviewer)
- item 12 satisfied: both suites run against the PR-head worktree exit 0; no existing assertion removed, and test_validate_entities_refs.sh case 3 is byte-identical. (cogni-service:contract-grading-reviewer)
- item 13 satisfied: the test_validate_entities_refs.sh header now states register-entity.py runs its own portfolio-scoped ref pass on top of this gate. (cogni-service:contract-grading-reviewer)

### Recommendation

---

sdh07:
## Merged by human override of the NEEDS_REVISION verdict

@sdh07 authorized merging this PR as-is, over the verdict posted at
https://github.com/cogni-work/insight-wave/pull/1188#issuecomment-5170207934. Recording the
reasoning so the override is auditable.

### Why the verdict is not being acted on

The two `major` findings that produced it — `review-contract:item-7` and `review-contract:item-14` —
are **requirements that did not exist in the contract this PR was built against.** This PR carries
two review contracts with zero textual overlap, and those two items are exactly where the second
one diverged:

| | Contract 1 (16:22Z, the author's bar) | Contract 2 (18:15Z, the graded bar) |
|---|---|---|
| docstring | module docstring "(base lines 25-26)" updated to state the ref resolution | "(base lines 25-26 **and the Exit line 35**) … **and that exit 1 covers that refusal**" |
| dangling test | asserts `success is False`, exit 1, manifest byte-identical, `transitions[]` unchanged | …plus **`file ==` the assignment path** and **no `.tmp` debris** |

The head commit satisfies contract 1 on both points. The added clauses are the entire content of
both `major` findings.

This is contract drift, not a defect in the change. `service-gatekeeper`'s sticky-contract
adoption (substep 2.5) exists to prevent exactly this, but its precondition query
(`gh-issues.sh issue-edited-since`) returns `query_failed` unconditionally — `gh api graphql --jq`
renders a JSON null as an empty string, never the literal `null`, so the `never_edited` branch at
`gh-issues.sh:1383` is unreachable and every pass re-derives. Tracked as
cogni-work/managed-service#709.

### Why the change itself is trusted

Independent of either contract, `plugin-validator` mutation-tested the three failure modes on the
risk watchlist against the live tree:

- swapping `portfolio_dir` for `entity_file` in the batch → **6 checks fail**, including the
  positive assignment case
- replacing `_same_file` with `e["file"] == entity_file` → **5 checks fail**, led by
  `refs: still rejected when the portfolio is passed as <root>/.` — the fail-open case
- dropping the per-entity filter → **3 checks fail**, including
  `refs: a project still registers while a dangling assignment exists`

Both suites are green (`test_register_entity.sh` 42 checks, `test_validate_entities_refs.sh` 29),
and `test_validate_entities_refs.sh` case 3 (`single-file invocation runs no ref check`) is retained
byte-identical, so `validate_file`'s single-file contract is provably untouched — the scope
constraint #1172 imposed.

### Known gaps accepted at merge

Two real coverage gaps ship with this, both cheap and both worth closing later:

1. `cogni-projects/scripts/register-entity.py:40` — the `Exit:` legend does not name the new
   dangling-ref refusal, so exit 1 now covers a case the docstring never mentions.
2. `cogni-projects/tests/test_register_entity.sh` — the dangling-consultant case asserts the error
   dict's key *set* but not that `errs[0]["file"]` is the assignment being registered, so a filter
   returning another file's error with the right `field` would still pass. It also omits the
   `tmp_debris(root5)` assertion the suite applies to its three other rejection cases.

Neither is a live bug — the gate returns before any `mkstemp`, so debris is impossible by
construction, and `_same_file` is what selects the error in the first place. They are test- and
doc-completeness only.

### Do not route this branch through the gatekeeper again

Until managed-service#709 is fixed, a further merger pass would derive a third independent bar with
no memory of the first two, and this PR is already at bounce 2 of the 3-bounce hold threshold.

PR #1189, opened 2026-08-03
cogni-workspace: pin every sanitize-theme denylist character

### LINKED ISSUE #1184: cogni-workspace: pin every denylist character in sanitize-theme's three profiles — 19 of 21 pairs are still vacuously covered
## Context

Found during the resolution of the `@`-clause anti-vacuity gap in the font-aware
denylist. That work added one isolating check for one character in one profile.
The same reasoning applies to the rest of the table, and the rest is not pinned.

## The gap

`_PROFILES` in `cogni-workspace/scripts/sanitize-theme.py` declares three denylist
character sets — 21 (profile, character) pairs in total. Only **two** of those pairs
are independently pinned by the suite: `@`/`font-aware` (the new check) and
`\`/`import-aware` (existing check `2w`). For the other 19, deleting the character
from its profile leaves `bash cogni-workspace/tests/test-sanitize-theme.sh` fully
green — every existing case that would have caught it is also rejected by some
*other* character in the same payload, so no case isolates the clause.

`strict` is the worst of the three, and it is the default profile guarding `colors`
and `status`. **All nine** of its characters are independently deletable with the
suite green.

### Reproduced

Dropping `;` from `strict` (`set("<>{}();@\\")` to `set("<>{}()@\\")`):

- full suite exits 0, zero `FAIL` lines
- `is_safe_value("red;position:fixed", "strict")` becomes `True` — declaration
  chaining straight into the `:root` block, from a `colors` value

Dropping `}` from `strict` admits `"red}"`, which closes the `:root` block early.
Same result: suite green.

## Proposed change

One data-driven check that pins every pair at once, subsuming the single-character
case. Shape:

```bash
assert_py "1mh every denylist character is independently rejected" \
  'all(g.is_safe_value("a" + c, p) is False for p, cs in {"strict": "<>{}();@\\", "font-aware": "<>{};@\\", "import-aware": "<>{};@\\"}.items() for c in cs)'
```

Verified green on unmutated code, and RED under each of the 21 single-character
mutations.

### The trap this must avoid

The expected denylists have to be **hardcoded in the test**, as above. Deriving them
from the module under test — `for p, (cs, *_) in g._PROFILES.items()` — is vacuous by
construction: the loop reads the same table the mutation edits, so a removed character
simply stops being tested and the assertion stays green against exactly the defect it
exists to catch. That is the original gap one level up.

## Acceptance criteria

- [ ] A single check pins every (profile, denylist character) pair in `_PROFILES`.
- [ ] The expected character sets are hardcoded in the test, not read from `_PROFILES`.
- [ ] Removing any one character from any one profile turns that check RED.
- [ ] Suite still green end to end.

## Related

- The parent follow-up that closed the `@`/`font-aware` pair and surfaced this class.

claude-cogni-work:
Closes #1184.

## Summary

- Adds one data-driven check `1mh` to `cogni-workspace/tests/test-sanitize-theme.sh`, inserted after `1mg` inside the font-aware/import-aware block, asserting `g.is_safe_value("a" + c, profile) is False` for every character `c` of every profile's denylist.
- The expected character sets are **hardcoded literals in the test**, never read from `g._PROFILES` — deriving them from the module under test would be vacuous by construction, since the loop would walk the same table a mutation edits.
- Leaves `1mg` in place as the named regression for the `@`/font-aware clause.
- No guard change: the denylists in `scripts/sanitize-theme.py` are correct. This closes a test-coverage gap, not a guard defect.

**Counting note.** `_PROFILES` holds 9 + 7 + 7 = **23** pairs, not the 21 the issue title states. Mutation-testing the suite during exploration also found a third pair already pinned (`;`/font-aware, caught incidentally by 11 checks) alongside `@`/font-aware (`1mg`) and `\`/import-aware (`2w`). So the real gap was **20 of 23**, not 19 of 21. The new check pins all 23 either way, and no file in the repo states a pair count, so nothing else needed correcting.

## Verification

Acceptance criteria, each checked:

| Criterion | Result |
|---|---|
| A single check pins every (profile, denylist character) pair | 23/23 |
| Expected sets hardcoded, not read from `_PROFILES` | the `1mh` expression contains no `_PROFILES` reference |
| Removing any one character from any one profile turns it RED | 23/23 (in-process, evaluating the check's verbatim expression against a mutated table) |
| Suite still green end to end | `All sanitize-theme tests passed.` |

Four of the 23 mutations were additionally driven end to end at the source level — dropping `;` from `strict`, the backslash from `strict`, `<` from `font-aware`, and `@` from `import-aware` — each making the real suite exit 1 with `FAIL 1mh`. The backslash case matters because it is the one the fragile quoting protects.

**Quoting mechanic, deliberate — please do not "fix" it.** The expression stays in bash **single** quotes with **two** backslashes per literal. `assert_py` splices `$expr` into an unquoted heredoc via parameter expansion, whose result is not re-scanned for escapes, so single quotes mean only Python's own escaping applies and two source backslashes yield the one literal backslash `_PROFILES` carries. Nearby check `2w` is double-quoted and therefore needs four; copying that spelling here would be wrong.

## Scope decisions

- **Out:** the independent hand-rolled copy of the strict denylist at `cogni-projects/scripts/render-dashboard.py:88` — a different plugin that does not import this guard. A real pre-existing drift vector, but not an acceptance criterion here.
- **Out:** the legacy-stub fixture's `_FORBIDDEN` copy later in this same test file. It fabricates a *stale* guard for the degradation tests, so it must deliberately not track the live table.
- **Out:** the prose restatements in the module docstring and `references/design-variables-pattern.md`.
- **Not bumped:** no `plugin.json` / `marketplace.json` version change — the post-merge workflow owns the patch bump.

## Known coverage boundary

`1mh` pins **removals**, not additions: a character or a whole profile *added* to `_PROFILES` later would go unpinned with the suite still green. The check's comment says so explicitly rather than overselling itself. The root-cause plan considered in the plan record proposed a companion check pinning the hardcoded table against `_PROFILES`' key set to close that direction; it was not adopted here because both candidate plans touched the identical file set, so the pick rubric's tie-breaks resolved to the minimal-surgical plan. It remains a coherent follow-up if the profile table grows.

## Open questions

- *(avoidable — recorded, not blocking)* `cogni-projects/scripts/render-dashboard.py:88` keeps an independent hand-rolled copy of the strict denylist (`_THEME_VALUE_FORBIDDEN = set("<>{}();@\\")`) and does not import `sanitize-theme.py`, so it can silently drift from `_PROFILES`. Deliberately left out of this PR (different plugin, not an acceptance criterion); surfaced so a human can decide whether it warrants its own issue.

Plan record: https://github.com/cogni-work/insight-wave/issues/1184#issuecomment-5169251574

### DIFF
--- a/cogni-workspace/tests/test-sanitize-theme.sh
+++ b/cogni-workspace/tests/test-sanitize-theme.sh
@@ -120,6 +120,15 @@ assert_py "1mf radius section rejects the injection payload" \
 # exactly one character with the denylist, so it fails iff '@' is dropped.
 assert_py "1mg font-aware rejects a bare @ carrying no other denylisted character" \
   'g.is_safe_value("@import url(https://evil.example/x.css)", "font-aware") is False'
+# Generalises 1mg to the whole table: "a" + c shares exactly one character with
+# the profile's denylist, so it is rejected iff that character is still listed.
+# Hardcoded on purpose — reading the sets from g._PROFILES would walk the table
+# a mutation edits, so a removed character would silently stop being tested.
+# Two-char payloads clear every max_len, and import-aware's shape gate (anchored
+# on '@import url(' / 'https://') cannot match a value starting with 'a'.
+# Pins removals only: a character or profile ADDED to _PROFILES is not covered.
+assert_py "1mh every denylist character is independently rejected" \
+  'all(g.is_safe_value("a" + c, p) is False for p, cs in {"strict": "<>{}();@\\", "font-aware": "<>{};@\\", "import-aware": "<>{};@\\"}.items() for c in cs)'
 assert_py "1n style breakout still rejected" \
   'g.is_safe_value("</style><script>alert(1)</script>", "font-aware") is False'
 assert_py "1o declaration injection rejected" \


claude-cogni-work:

#1184 end-to-end to a draft PR. Handing off to the merger identity for
never reviews its own work."

`svc:needs-review` applied to this PR; #1184 marked `svc:resolved`.

sdh07:

_The merger formed this expectation from issue #1184 before checking out the PR head — the review grades the diff against it._

**Expected approach:** This is a test-coverage gap, not a guard defect: `_PROFILES` in cogni-workspace/scripts/sanitize-theme.py:159-163 is correct as shipped (strict `<>{}();@\` = 9 chars, font-aware `<>{};@\` = 7, import-aware `<>{};@\` = 7 — 23 (profile, character) pairs, not the 21 the issue body states), and only two of those pairs are isolated today by checks `1mg` (tests/test-sanitize-theme.sh:121-122) and `2w` (line 230-231). A correct fix adds one data-driven `assert_py` check to cogni-workspace/tests/test-sanitize-theme.sh that iterates a set of expected denylists **hardcoded in the test** and asserts `g.is_safe_value(<single-denylisted-char payload>, profile) is False` for all 23 pairs, leaving sanitize-theme.py untouched. The payload must share exactly one character with the profile's denylist (e.g. `"a" + c`) and must not look like an `@import`/`https://` value, or the import-aware rows would be adjudicated by `_FONT_IMPORT_RE` (sanitize-theme.py:121-124, consulted at line 216) rather than by the denylist. The load-bearing mechanical hazard is the backslash: `assert_py` interpolates the expression into an **unquoted** heredoc (`<<PY`, tests/test-sanitize-theme.sh:31-39), so bash halves backslash pairs before Python parses the expression — the issue's own proposed snippet, pasted verbatim, yields an unterminated Python string literal. The check must also be non-vacuous: `all()` over an empty or truncated iteration is True, which is the same trap the file already guards against at lines 258-260 and 300-304.

**Acceptance checklist:**
- [ ] Scope: the diff touches only cogni-workspace/tests/test-sanitize-theme.sh. cogni-workspace/scripts/sanitize-theme.py is unmodified — check the changed-file list contains exactly that one path and the diff has no hunk under cogni-workspace/scripts/ (the shipped _PROFILES at sanitize-theme.py:159-163 is correct; only its coverage is thin).
- [ ] The new check goes through the existing `assert_py` helper (tests/test-sanitize-theme.sh:29-41), not an open-coded `python3 -c ... && pass || fail`, and carries a label unique in the file — grep the chosen label (the issue proposes `1mh`) in the base file returns 0 hits and in the head file exactly 1.
- [ ] Additive, not substitutive: `grep -c '^assert_' cogni-workspace/tests/test-sanitize-theme.sh` is 111 at base and >= 111 at head; no existing labelled assertion is deleted or weakened. The only defensible removal is the now-subsumed `1mg` (lines 121-122), and only if the new check provably covers the `@`/font-aware pair.
- [ ] AC#2 — hardcoded, not derived: the expected character sets appear as literal strings in the added lines, and the added expression does not read `g._PROFILES` (or any other attribute of the module under test) as the source of the iterated characters. Check by grepping the added lines for `_PROFILES` — it must not appear as the iteration source.
- [ ] AC#1 — completeness: the hardcoded sets are exactly strict `<>{}();@\` (9 chars, both parens present), font-aware `<>{};@\` (7 chars, no parens), import-aware `<>{};@\` (7 chars) — 23 (profile, character) pairs total, matching sanitize-theme.py:160-162. The reviewer counts characters per profile in the diff. The issue body's '21 pairs' / 'other 19' undercounts by two; a check pinning fewer than 23 pairs fails this item.
- [ ] Non-vacuity guard: `all(...)` over an empty or truncated iteration returns True, so the check must be un-passable vacuously — either an explicit pair count in the assertion (e.g. `and sum(1 for ...) == 23`) or the three literals visibly carrying 9/7/7 characters per the previous item. This is the file's own arity-first convention (line 258-260: 'Arity first, so the all()-loops below cannot pass vacuously on an empty dict'; line 300-304 pairs a negative with a `checked` count for the same reason).
- [ ] Isolating payload: each pair is probed with a value sharing exactly one character with that profile's denylist (e.g. `"a" + c`), so no other denylist clause can account for the rejection — read the payload expression in the diff. The payload must additionally not begin with `@import` or `https://`, or the import-aware rows would be routed through the shape gate `_FONT_IMPORT_RE` (sanitize-theme.py:121-124, reached at line 216) instead of the denylist, which tolerates `@` and `;` inside a URL.
- [ ] The backslash pair is exercised as a real backslash (U+005C), not lost to shell escaping. `assert_py`'s heredoc delimiter is unquoted (`<<PY`, line 31), so bash collapses a backslash pair to one backslash before Python sees the expression; producing a Python string literal that contains one backslash therefore needs FOUR literal backslashes inside a single-quoted bash expression — compare existing check `2w` (lines 230-231), which spends four inside a double-quoted string to put exactly one backslash in its payload. Pasting the issue's snippet verbatim (two backslashes inside single quotes) yields Python source ending `@\"`, an unterminated string literal, and the check reports FAIL. Confirm via the suite run and the backslash mutations below.
- [ ] AC#4 — suite green end to end, verified by acting: `bash cogni-workspace/tests/test-sanitize-theme.sh` exits 0, prints zero lines starting `FAIL`, prints `All sanitize-theme tests passed.`, and prints an `OK` line for the new label. Total OK lines >= 114 at head (base = 111 `assert_*` invocations + 3 open-coded pass/fail at lines 277, 282, 286).
- [ ] AC#3 — mutation RED, verified by acting: for each of at least `;` from strict, `}` from strict, `(` from strict (a strict-only character), `@` from font-aware, and the backslash from import-aware, delete that one character from that one profile's set in sanitize-theme.py:160-162, rerun the suite, and confirm the new check is the reported `FAIL` line and the suite exits 1; restore the file after each. A full 23-of-23 sweep is the ideal bar. Any mutation that leaves the suite green fails this item.
- [ ] Rationale comment recorded: the diff adds a comment above the new check stating why the sets are hardcoded — deriving them from `_PROFILES` would be vacuous because the loop reads the same table the mutation edits — matching this file's convention of documenting every anti-vacuity trap in place (lines 74-77, 117-120, 190-191, 220-225, 300-302). Check that the added comment names `_PROFILES` and the derivation trap.

**Risk watchlist:**
- Verbatim copy of the issue's proposed snippet: the backslash collapses in the unquoted heredoc, so the check either dies as a Python SyntaxError (visible FAIL) or — worse — gets 'fixed' by dropping the backslash from the sets, silently leaving 3 of 23 pairs unpinned while the suite reads green.
- Deriving the expected sets from `g._PROFILES` (the trap the issue itself names): vacuous by construction, since a removed character simply stops being tested.
- Hardcoding only 21 pairs to match the issue body's miscount, leaving two pairs (the strict-only parens, or an omitted character) unpinned.
- Using a richer, realistic payload (e.g. a full `@import url(...)` string) that carries more than one denylisted character, so some other clause accounts for the rejection and the check stops isolating each pair — the exact defect one level up that this issue exists to close.
- A payload beginning with `https://` or `@import`, which routes the import-aware rows through the shape gate rather than the denylist; the gate deliberately tolerates `@` and `;` inside a URL, so such rows could pass regardless of the denylist's contents.
- An `all()` comprehension over a mangled or empty literal passing vacuously because no arity/count guard was added.
- Deleting `1mg` and/or `2w` as 'subsumed' and net-losing either coverage or the rationale comments that explain why those payloads were chosen.
- Editing sanitize-theme.py to 'fix' the guard — the guard is correct; the defect is purely in the suite's ability to detect a mutation of it.
- Judgment-only (no gate can decide it): the new check keys on hardcoded profile *names*, so renaming or deleting a whole profile key in `_PROFILES` still leaves `_profile()` (sanitize-theme.py:169-171) failing closed onto `strict`, under which all 7 relaxed-profile characters are still rejected — a profile-level mutation can therefore stay green. That is one level up again and outside this issue's scope; a human should note it rather than the gate blocking on it.
- Judgment-only: placement of the new check (inside the '=== font-aware profile ===' section, where the `1mh` label sorts, versus a section of its own given it spans all three profiles) is cosmetic and must not gate.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1189

**Summary:** Both perspectives passed with a clean 9-line, test-file-only diff; the two minor findings raised by plugin-validator are pre-existing, out-of-scope conditions explicitly waived with non-empty justifications, so Rule 5 (all-pass, all-findings-waived) applies and the merge is safe.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (waived) `.github/workflows/lint.yml` — No CI workflow discovers or executes cogni-workspace/tests/*.sh, so the new 1mh check — like every check already in the file — is only exercised when a human or agent runs the suite manually.
  - *Justification:* Pre-existing repo-level CI gap affecting all 115 checks equally; the acceptance criterion asks only that the suite be green end to end, which it is.

- **[minor]** (waived) `cogni-workspace/skills/{install-mcp,manage-themes,manage-workspace,pick-theme,workspace-dashboard,workspace-status}/SKILL.md` — Six SKILL.md files carry a non-standard top-level `version` frontmatter key. Pre-existing context only — this PR touches no SKILL.md file and did not introduce the key. Claude Code ignores unknown frontmatter keys, so nothing is broken; the practical risk is drift against the authoritative plugin.json version.
  - *Justification:* Pre-existing; not introduced, touched, or worsened by this PR's 9-line test-only diff.

### Positive observations

- item 1 satisfied: exactly one changed path, cogni-workspace/tests/test-sanitize-theme.sh; sanitize-theme.py byte-identical to base. (contract-grading-reviewer)
- item 2 satisfied: label 1mh is 0 hits at base, 1 at head; goes through the existing assert_py helper (test-sanitize-theme.sh:130-131), not open-coded. (contract-grading-reviewer)
- item 3 satisfied: 9 added lines, zero deletions; assert count additive by exactly one — the contract's base-count prediction was inaccurate but the deletion-set check independently confirms nothing was weakened. (contract-grading-reviewer)
- item 4 satisfied: _PROFILES appears only in the rationale comment, never as the iteration source in the assert expression. (contract-grading-reviewer)
- item 5 satisfied: the three literals parse to exactly 9, 7, 7 logical characters (23 pairs), matching sanitize-theme.py byte-for-byte. (contract-grading-reviewer)
- item 6 satisfied: sets spelled as complete literals rather than derived or truncated; a 23-way mutation sweep confirms no character can go missing without the check catching it. (contract-grading-reviewer)
- item 7 satisfied: payload construction ensures exactly one denylist character overlaps per probe and import-aware rows are adjudicated by the denylist rather than the font-import regex. (contract-grading-reviewer)
- item 8 satisfied — verified empirically that the heredoc/single-quote splice preserves exactly one real backslash per set in Python, refuting the contract's own four-backslash prediction as a contract error rather than a diff defect. (contract-grading-reviewer)
- item 9 satisfied — verified by acting: the suite exits 0, zero FAIL lines, includes an OK line for 1mh. (contract-grading-reviewer)
- item 10 satisfied — verified by acting with a full 23-of-23 mutation sweep (beyond the required 5), each mutation restored and worktree left clean. (contract-grading-reviewer)
- item 11 satisfied: the added comment block names _PROFILES explicitly and states the derivation trap verbatim. (contract-grading-reviewer)
- plugin.json valid JSON, semver version 0.6.44 in sync with marketplace.json, correctly left unbumped per the post-merge bump convention. (plugin-validator)
- Scope exactly as declared: 1 file changed, 9 insertions, 0 deletions; sanitize-theme.py byte-identical to main, as are all skills, hooks, and manifests. (plugin-validator)
- All 9 skills have well-formed frontmatter with directory names matching their name field, so auto-discovery resolves cleanly. (plugin-validator)
- hooks/hooks.json valid, both referenced scripts present, executable, bash -n clean, addressed via ${CLAUDE_PLUGIN_ROOT}. (plugin-validator)
- Both test suites pass end to end; running them left the worktree clean. (plugin-validator)
- The new check 1mh coexists with the retained narrower 1mg rather than replacing it. (plugin-validator)
- Mutation sweep: 23 mutants, 23 kills, 0 survivors — the check is genuinely RED for every single-character removal. (plugin-validator)
- No security concerns introduced: inert two-character test payloads only. (plugin-validator)

### Recommendation

---

PR #1192, opened 2026-08-03
cogni-projects: shared tests/ harness with an exit-code-safe runner

### LINKED ISSUE #1182: cogni-projects follow-up (#1173): extract a shared tests/ assertion harness — three suites now inline it byte-identically

claude-cogni-work:
Closes #1182.

## Summary

Three of the five `cogni-projects/tests/` suites inlined the same bash assertion harness byte-identically, and the copies had **already drifted on a correctness-relevant detail**: `test_staffing_score.sh`'s runner redirects stdout to a file, while `test-render-dashboard.sh`'s piped into `tail`. After a `$(... | tail -n 1)` substitution `$?` is tail's status — always 0 — so `test-render-dashboard.sh` could not express an exit-code assertion at all, and any it had grown would have passed vacuously.

- Adds `cogni-projects/tests/fixtures/test_helpers.sh`, following the in-repo precedent `cogni-knowledge/tests/fixtures/test_helpers.sh` — **pattern, not content**: a `fixtures/` helper sourced via the POSIX dot idiom, bash 3.2 compatible, no shebang. It lives under `fixtures/` so a non-recursive `tests/test_*.sh` sweep can never collect it as a suite.
- Exposes `failures`/`pass()`/`fail()`, `assert_json()`, `assert_no_traceback()`, `assert_exit()`, `seed_portfolio()`, `write_entity()`, `write_latin1()`, `init_tmproot()` and the canonical runner `run_script`.
- `run_script` captures the invoked script's **own** exit status into `RUN_CODE` by redirecting stdout to a file instead of piping into `tail`, and takes its whole command line positionally — so it serves the two `python3` suites and the `bash portfolio-init.sh` suite alike.
- `test-render-dashboard.sh` gains its first exit-code assertions: `1o` (happy render exits 0) and `7d` (no-args usage exits 2). `render-dashboard.py` pins `Exit: 0 ok / 2 usage or environment failure`, and the no-args path reaches 2 through `_fail(...)`, not argparse's own exit.

Net: three copies of the harness collapse to one (−154/+200 lines, of which 139 are the new shared file), and the suite that structurally could not assert an exit code now does.

## Why the call sites are untouched

The two pre-existing runner signatures collide: in `run()` arg 1 is the first CLI argument to the target script, while in `run_score()` arg 1 is consumed as the stderr sink. Merging onto one positional signature would have silently collapsed every render-dashboard fixture into the no-args usage path. So the sink and the optional cwd move into **one-shot globals** (`RUN_ERRFILE`, `RUN_CWD` — consumed and cleared by `run_script`, so a fixture cannot leak its setting into the next run), and each suite keeps a thin wrapper. All 9 `run` and 5 `run_score` call sites are byte-identical to before.

## Scope decision — AC 1 is met as "every suite that uses the bash assertion harness"

The issue's AC 1 says "all five suites". Only **three** are bash-assertion suites. `test_register_entity.sh` and `test_validate_entities_refs.sh` hand off entirely to a single inline `python3 - <<'PY'` heredoc whose assertions are a Python `check()`; neither defines any bash `pass`/`fail`/`assert_json`. A sourced bash helper cannot be reached from inside that subprocess, so a `source` line there would be inert. They are deliberately left untouched, and `cogni-projects/CLAUDE.md` now records which suites source the harness and why the other two do not. This was flagged at triage before any code was written — see the triage comment on #1182.

**Reviewer note:** this is the one place the PR diverges from the issue text as literally written. Every other acceptance criterion is delivered in full, so the PR links `Closes`.

## Verification

- All five suites green: **136 assertions** (134 baseline + the 2 new exit-code ones), every suite exit 0.
- **Negative control** — flipping `7d` to expect exit 0 makes the suite fail (`want exit 0, got 2`) and exit non-zero; reverting restores 136/exit 0. This is the proof the new assertion is not vacuous, which the old piped `run()` could not have provided.
- `check-breadcrumbs.py` 0 violations · `check-version-bump.py` 0 touched, 0 violations · `check-skill-names.sh` clean.
- No `plugin.json` / `marketplace.json` change — the post-merge workflow owns the bump.

A pre-commit quality pass raised, and this PR fixes, one defect the extraction itself introduced: `write_latin1` originally swallowed a failed encode, leaving a 0-byte file that the loaders skip as frontmatter-less — so the very regression fixture that exists to prove non-UTF-8 records are tolerated would have gone green while testing nothing. It now encodes before opening the file and reports the failure, verified against a Polish `ł` payload.

## Out of scope, noted for future work

Not deferred slices of this issue — every acceptance criterion above is delivered here:

- Deduplicating the identical Python `check()` across the two heredoc suites (needs a shared *Python* module, not a bash helper).
- Giving `test_portfolio_init.sh` real stderr capture so it could assert `assert_no_traceback` — it currently discards stderr. A latent coverage hole, not a regression.
- Aligning this harness's grep-assertion argument order with `cogni-knowledge`'s `assert_grep(pattern, file, description)`, and folding `test-render-dashboard.sh`'s local `assert_html`/`assert_html_lacks` into that shape.

### DIFF
--- a/cogni-projects/CLAUDE.md
+++ b/cogni-projects/CLAUDE.md
@@ -26,6 +26,7 @@ cogni-projects/
 │   ├── staffing-score.py               Deterministic staffing scorer (availability/fit/impact)
 │   └── render-dashboard.py             Portfolio health + value HTML render (read-only, stdlib-only)
 ├── tests/
+│   ├── fixtures/test_helpers.sh        Shared bash assertion harness + exit-code-safe runner
 │   ├── test_portfolio_init.sh          Portfolio scaffolder idempotency + no-debris suite
 │   ├── test_register_entity.sh         Atomic-write + idempotency regression suite
 │   ├── test_validate_entities_refs.sh  Assignment ref-integrity + single-file regression suite
@@ -44,6 +45,13 @@ bash cogni-projects/tests/test_validate_entities_refs.sh
 bash cogni-projects/tests/test_staffing_score.sh
 bash cogni-projects/tests/test-render-dashboard.sh
 ```
+
+`test_portfolio_init.sh`, `test_staffing_score.sh` and `test-render-dashboard.sh`
+source `tests/fixtures/test_helpers.sh` for their assertions and their script
+runner; the runner captures the invoked script's own exit status rather than a
+pipeline's, which is what makes an exit-code assertion meaningful. The other two
+suites assert inside a single inline `python3 - <<'PY'` heredoc and define no
+bash-level assertions, so they do not source it.
 
 Planned (not yet scaffolded — see the roadmap epic):
 
--- a/cogni-projects/tests/fixtures/test_helpers.sh
+++ b/cogni-projects/tests/fixtures/test_helpers.sh
@@ -1,0 +1,139 @@
+# test_helpers.sh — shared bash assertion harness sourced by cogni-projects/tests/*.
+#
+# Source via:
+#   . "$TESTS_DIR/fixtures/test_helpers.sh"
+#
+# The POSIX dot, not `source`, and no shebang: this file is sourced, never
+# executed. It lives under fixtures/ rather than beside the suites so a
+# `tests/test_*.sh` sweep never collects a non-suite file as a suite.
+#
+# `set -u` is deliberately NOT set here — it is a per-suite preamble line that
+# must precede this source. Nor is there a shared summary/exit tail: the suites
+# print different completion lines.
+#
+# Bash 3.2 compatible (the macOS default these suites run under).
+
+failures=0
+pass() { printf 'OK   %s\n' "$1"; }
+fail() { printf 'FAIL %s: %s\n' "$1" "$2" >&2; failures=$((failures + 1)); }
+
+# init_tmproot — set TMPROOT to a fresh temp dir and arm its cleanup trap.
+# TMPROOT is a global on purpose: run_script defaults its stdout/stderr sinks
+# underneath it, and the suites reference it directly when building fixtures.
+# The chmod is what lets a fixture leave a read-only (0555) directory behind
+# without defeating cleanup; it is a harmless no-op for suites that never do.
+init_tmproot() {
+  TMPROOT="$(mktemp -d)"
+  trap 'chmod -R u+rwx "$TMPROOT" 2>/dev/null; rm -rf "$TMPROOT"' EXIT
+}
+
+# assert_json <label> <python-bool-expr over `d`> — pipes the last stdout line
+# (the envelope) into python3 and checks the expression is truthy.
+assert_json() {
+  local label="$1" expr="$2"
+  if printf '%s' "$LAST_JSON" | python3 -c "import json,sys
+d = json.loads(sys.stdin.read())
+sys.exit(0 if ($expr) else 1)" 2>/dev/null; then
+    pass "$label"
+  else
+    fail "$label" "expr false: $expr | json=$LAST_JSON"
+  fi
+}
+
+# The crash signature belongs in one place — widening it later should be one
+# edit. Every "the script degraded instead of crashing" fixture needs it.
+assert_no_traceback() {
+  local label="$1" file="$2"
+  if grep -qF 'Traceback' "$file"; then
+    fail "$label" "traceback present: $(head -3 "$file")"
+  else
+    pass "$label"
+  fi
+}
+
+assert_exit() {
+  local label="$1" want="$2" got="$3"
+  if [ "$want" = "$got" ]; then
+    pass "$label"
+  else
+    fail "$label" "want exit $want, got $got"
+  fi
+}
+
+seed_portfolio() {
+  # $1 = portfolio dir. Seeds a manifest + entity subdirs. The manifest's ref
+  # lists stay empty on purpose: _load_entities then falls back to scanning the
+  # subdirectories, which is the path a hand-authored record takes.
+  #
+  # Never call this from a suite that exercises portfolio-init.sh — creating
+  # this layout is that script's job, and seeding it would pre-empt the
+  # behaviour under test.
+  local pf="$1"
+  mkdir -p "$pf"/{consultants,projects,assignments,.metadata}
+  cat > "$pf/projects-portfolio.json" <<'EOF'
+{"slug":"test","name":"Test Portfolio","language":"en","consultants":[],"projects":[],"assignments":[],"created":"2026-01-01","updated":"2026-01-01"}
+EOF
+}
+
+write_entity() { # $1 = path; body heredoc'd by the caller on stdin
+  cat > "$1"
+}
+
+# write_latin1 <path> — body heredoc'd by the caller on stdin as UTF-8, written
+# out latin-1-encoded. python3 does the write in wb mode because bash printf
+# parses a leading '---' as options; the encoded bytes are invalid UTF-8 start
+# bytes by design, which is the whole point of the fixture. stdin is read as
+# binary and decoded explicitly so a C/POSIX locale cannot change the result.
+#
+# The encode happens before the file is opened, and a failure is reported rather
+# than swallowed. Both matter: a character outside latin-1 (say 'ł') would
+# otherwise leave a 0-byte file, which the loaders skip as frontmatter-less — so
+# the very regression fixture that exists to prove non-UTF-8 records are
+# tolerated would pass while testing nothing.
+write_latin1() {
+  if ! python3 -c "
+import sys
+body = sys.stdin.buffer.read().decode('utf-8')
+blob = body.encode('latin-1')
+open(sys.argv[1], 'wb').write(blob)" "$1"; then
+    fail "write_latin1 $1" "payload is not latin-1 encodable"
+    return 1
+  fi
+}
+
+# run_script <command> [args...] — invoke a script and capture its envelope and
+# its own exit status. The whole command line is positional, so the runner is
+# interpreter-agnostic: `run_script python3 "$SCRIPT" ...` and
+# `run_script bash "$SCRIPT" ...` both work.
+#
+# Sets LAST_JSON (the last stdout line) and RUN_CODE (the command's status).
+# Redirecting to a file rather than piping into `tail` is load-bearing: after a
+# `$(... | tail -n 1)` substitution, `$?` is tail's status, which is always 0,
+# so every exit-code assertion would pass vacuously.
+#
+# Two optional globals tune a single call, and both are ONE-SHOT — consumed and
+# cleared here, so a fixture cannot leak its setting into the next run:
+#   RUN_ERRFILE  stderr sink for this call (default "$TMPROOT/stderr.txt")
+#   RUN_CWD      run the command in this directory, leaving the caller's cwd
+#                untouched; the redirections and the status capture stay in the
+#                caller's shell, so RUN_CODE is still the command's own.
+# Clearing them here is what keeps every caller on one discipline: set it on the
+# line above the call, and never think about resetting it. The sink FILE of
+# course survives — only the variable is cleared — so an assertion reading it
+# after the call still works.
+#
+# init_tmproot must have run first — the default sink dereferences TMPROOT.
+run_script() {
+  local outfile="$TMPROOT/stdout.txt"
+  local errfile="${RUN_ERRFILE:-$TMPROOT/stderr.txt}"
+  local cwd="${RUN_CWD:-}"
+  RUN_ERRFILE=""
+  RUN_CWD=""
+  if [ -n "$cwd" ]; then
+    ( cd "$cwd" && "$@" ) >"$outfile" 2>"$errfile"
+  else
+    "$@" >"$outfile" 2>"$errfile"
+  fi
+  RUN_CODE=$?
+  LAST_JSON="$(tail -n 1 "$outfile")"
+}
--- a/cogni-projects/tests/test-render-dashboard.sh
+++ b/cogni-projects/tests/test-render-dashboard.sh
@@ -23,25 +23,9 @@ if [ ! -f "$SCRIPT" ]; then
   exit 1
 fi
 
-TMPROOT="$(mktemp -d)"
-trap 'rm -rf "$TMPROOT"' EXIT
+. "$TESTS_DIR/fixtures/test_helpers.sh"
 
-failures=0
-pass() { printf 'OK   %s\n' "$1"; }
-fail() { printf 'FAIL %s: %s\n' "$1" "$2" >&2; failures=$((failures + 1)); }
-
-# assert_json <label> <python-bool-expr over `d`> — pipes the last stdout line
-# (the envelope) into python3 and checks the expression is truthy.
-assert_json() {
-  local label="$1" expr="$2"
-  if printf '%s' "$LAST_JSON" | python3 -c "import json,sys
-d = json.loads(sys.stdin.read())
-sys.exit(0 if ($expr) else 1)" 2>/dev/null; then
-    pass "$label"
-  else
-    fail "$label" "expr false: $expr | json=$LAST_JSON"
-  fi
-}
+init_tmproot
 
 assert_html() {
   local label="$1" needle="$2" file="$3"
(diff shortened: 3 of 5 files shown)


claude-cogni-work:
## Ship-ready

Authored by the cogni-service author instance and handed off for independent review. The PR is draft and carries `svc:needs-review`; the merger instance owns the next move (independent review, ready flip, merge).

**What landed:** a shared bash assertion harness at `cogni-projects/tests/fixtures/test_helpers.sh` with an exit-code-safe, interpreter-agnostic runner, sourced by the three bash-assertion suites. `test-render-dashboard.sh` gains its first two exit-code assertions.

**Verification:** all five suites green at 136 assertions (134 baseline + 2 new); negative control confirms the new exit-code assertion is not vacuous; breadcrumb, version-bump and skill-name gates clean; no manifest version touched.

**One point needing a reviewer call:** the issue's AC 1 says "all five suites", and this PR sources the harness from three. The other two assert entirely inside an inline `python3 - <<'PY'` heredoc and define no bash assertions, so a `source` line there would be inert. This was identified at triage before any code was written, is recorded in `cogni-projects/CLAUDE.md`, and is set out in the PR body under "Scope decision". Every other acceptance criterion is delivered in full, so the PR links `Closes #1182`.

sdh07:

_The merger formed this expectation from issue #1182 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix extracts the byte-identical bash assertion harness into a plugin-local `cogni-projects/tests/fixtures/test_helpers.sh`, modelled on the in-repo precedent `cogni-knowledge/tests/fixtures/test_helpers.sh` (no shebang, no `set`, bash-3.2 compatible, sourced as `. "$(dirname "$0")/fixtures/test_helpers.sh"` — the form used at e.g. `cogni-knowledge/tests/test_root_index.sh:54`). It must absorb the definitions currently triplicated at `cogni-projects/tests/test_staffing_score.sh:31-93`, `test-render-dashboard.sh:29-94` and `test_portfolio_init.sh:35-50` (`failures`/`pass`/`fail`, `assert_json`, `assert_no_traceback`, `assert_exit`, `seed_portfolio`, `write_entity`, the latin-1 fixture writer duplicated at `test_staffing_score.sh:180-184` and `test-render-dashboard.sh:331-335`), and the *inline copies must be deleted*, not shadowed. The canonical runner must be `run_score`'s shape (`test_staffing_score.sh:88-94`): redirect stdout to a file, take `RUN_CODE=$?` from the script itself, then `tail -n 1` the file — never `run()`'s piped `$(... | tail -n 1)` (`test-render-dashboard.sh:92-94`), after which `$?` is tail's and always 0. `test-render-dashboard.sh` must then gain a real exit-code assertion; since `render-dashboard.py`'s `_fail` returns 2 on every failure path (`cogni-projects/scripts/render-dashboard.py:513`, `:537`, `:541`, `:548`), the non-vacuous assertion is a non-zero expectation on the no-args (Fixture 7) or missing-manifest (Fixture 4) path. The refactor must lose no existing assertion, must preserve `test-render-dashboard.sh`'s three direct-invoke stderr sites (lines 338, 526, 683, feeding `assert_no_traceback` 6g/10g/11i) and `test_portfolio_init.sh`'s `chmod -R u+rwx` EXIT trap (line 33) and root-user SKIP, and must update the tests/ tree in `cogni-projects/CLAUDE.md:28-46`. No CI workflow runs these suites (`.github/workflows/` holds only cla/lint/version-bump), so greenness must be proven by running them.

**Acceptance checklist:**
- [ ] `cogni-projects/tests/fixtures/test_helpers.sh` exists at exactly that path (plugin-local, under `fixtures/` so a non-recursive `tests/*.sh` sweep never collects it), matching the precedent `cogni-knowledge/tests/fixtures/test_helpers.sh`.
- [ ] The harness carries no shebang and no `set -e`/`set -o errexit`: grep the file for `^#!`, `set -e`, `errexit` returns zero matches. A sourced errexit would abort each suite on its first failing assertion, contradicting the `set -u  # NOT -e` preamble every suite documents.
- [ ] The harness is bash-3.2 compatible (macOS default, and the stated bar in the precedent header): grep returns zero matches for `declare -A`, `local -n`, `mapfile`, `readarray`, `${*,,}`-style case modification.
- [ ] Each of `test_portfolio_init.sh`, `test-render-dashboard.sh`, `test_staffing_score.sh` contains exactly one source line of the precedent form `. "$(dirname "$0")/fixtures/test_helpers.sh"` — 3 files, 1 match each.
- [ ] The duplication is REMOVED, not shadowed: grepping the five files in `cogni-projects/tests/*.sh` (excluding `fixtures/`) for the definitions `pass()`, `fail()`, `assert_json()`, `assert_no_traceback()`, `assert_exit()`, `seed_portfolio()`, `write_entity()`, and the `failures=0` initializer returns ZERO matches; each is defined exactly once, inside `fixtures/test_helpers.sh`.
- [ ] The duplicated latin-1 fixture writer is absorbed too: the `python3 -c "... .encode('latin-1')"` block appears zero times inline in `test-render-dashboard.sh` and `test_staffing_score.sh` and exactly once in the harness (as a helper such as `write_latin1`), and its "bash printf parses a leading `---` as options" rationale comment is carried across rather than dropped.
- [ ] AC 2 — the shared runner captures the script's own exit status: in the harness runner body, stdout is redirected to a file (`>"$outfile"` or equivalent), `RUN_CODE=$?` is the statement immediately after the invocation, the invocation line contains no `| tail` (grep it), and `LAST_JSON` is assigned from `tail -n 1 "$outfile"` as a separate later statement.
- [ ] The harness header states the anti-drift rationale semantically — that after a `$(... | tail -n 1)` substitution `$?` is tail's status, always 0, so every exit-code assertion would pass vacuously — so the knowledge that prevented re-drift survives the move (grep the harness comments for `tail`).
- [ ] AC 3 — `test-render-dashboard.sh` contains at least one `assert_exit` call site (grep count >= 1 in that file).
- [ ] At least one new `test-render-dashboard.sh` exit assertion expects a NON-ZERO status (render-dashboard's `_fail` returns 2 on the no-args and missing-manifest paths — `scripts/render-dashboard.py:513`, `:537`). An all-`0` set is vacuous: it would also pass under the old piped `run()`, so it cannot prove the shared runner works where the old one could not.
- [ ] Documented mutation check on the new runner: temporarily restore the piped shape (`LAST_JSON="$(python3 "$SCRIPT" "$@" | tail -n 1)"; RUN_CODE=$?`) in the harness and confirm at least one `test-render-dashboard.sh` exit assertion then FAILS; restore afterwards. If every assertion still passes, AC 3 is met only in letter.
- [ ] No assertion coverage was lost: per-suite assertion call-site counts are >= the base counts (`test-render-dashboard.sh` 74, `test_staffing_score.sh` 21, `test_portfolio_init.sh` 4 `assert_*` calls plus its hand-rolled `if …; then pass … else fail` checks labelled 1a/1c/2a/2e/2f, and render-dashboard's hand-rolled 2c). Spot-check that named labels survive verbatim: `11l clean sibling project still rendered`, `3i non-portfolio stderr is clean`, `2f no temp-file debris left behind`.
- [ ] Stderr capture is preserved, not downgraded: `test-render-dashboard.sh`'s three direct-invoke sites (base lines 338, 526, 683 — Fixtures 6, 10, 11) still route the run's stderr into the file that `assert_no_traceback` 6g/10g/11i inspects — not `/dev/null`, and not a file left stale from an earlier run whose contents no longer correspond to the invocation being asserted.
- [ ] `test_portfolio_init.sh` keeps its fixture-specific idioms: the EXIT trap still runs `chmod -R u+rwx "$TMPROOT"` before `rm -rf` (base line 33 — a generic shared `rm -rf` trap would leave the 0555 `.metadata` fixture undeletable), the root-user SKIP guard `[ "$(id -u)" -eq 0 ]` is intact, and the script is still invoked with its `cd "$WORK…"` working-directory scoping.
- [ ] `cogni-projects/CLAUDE.md` documents the harness: its tests/ architecture tree (base lines 28-33) gains a `fixtures/test_helpers.sh` entry with a one-line description, and the "Run the tests directly" block (base lines 41-45) still lists all five suite commands.
- [ ] AC 4 — all five suites are actually run and each exits 0 with no `FAIL` line: `bash cogni-projects/tests/test_portfolio_init.sh`, `test_register_entity.sh`, `test_validate_entities_refs.sh`, `test_staffing_score.sh`, `test-render-dashboard.sh`. No CI workflow runs them (`.github/workflows/` has only cla/lint/version-bump), so greenness must be executed, not inferred from the diff.

**Risk watchlist:**
- The half-measure the AC literally permits: add the harness, source it from three suites, but leave the inline definitions in place. Every function then resolves to the local (shadowing) copy, all four AC criteria read as satisfied, and nothing is actually deduplicated — the drift the issue reports is untouched.
- Carrying `run()`'s piped shape forward as the canonical runner (or keeping `| tail` anywhere in the runner's invocation line), which reinstates the exact vacuous-exit-code defect the issue names.
- A new `assert_exit` in test-render-dashboard.sh that only ever expects 0, so it would pass under the old broken runner too — AC 3 satisfied in letter, not in substance.
- Silent assertion loss: fixtures or `assert_*` call sites dropped while 'simplifying' the suites, so all five stay green because they now assert less. Counts and named labels are the guard.
- Stderr regressions: the shared runner discarding stderr (as `run()` does with `2>/dev/null`) would make `assert_no_traceback` 6g/10g/11i assert against an empty or stale file and pass unconditionally — a traceback-detection suite that can no longer detect tracebacks.
- A single shared stdout/stderr temp file reused across runs, so an assertion reads output from the previous invocation; especially dangerous for the repeated-`run_score`-into-one-stderr-file pattern at `test_staffing_score.sh:201-218`.
- `TMPROOT` unset (or set after sourcing) when the harness runner references it: under `set -u` this aborts the whole suite rather than failing one assertion.
- Bash 3.2: `"${args[@]}"` on an empty array under `set -u` errors 'unbound variable', which would break `test-render-dashboard.sh`'s zero-argument Fixture 7 `run` on macOS while passing on bash 5 in CI-less local runs.
- A generic `init_tmproot`/trap in the harness overwriting `test_portfolio_init.sh`'s `chmod -R u+rwx` trap, leaving an undeletable 0555 temp tree behind on every run.
- Cargo-cult sourcing of the harness into `test_register_entity.sh` / `test_validate_entities_refs.sh` to satisfy AC 1's literal 'all five' — their assertions live inside a `python3 - <<'PY'` heredoc and cannot reach a bash helper, so a source line there is dead weight that misrepresents the dedup.
- Scope creep into the repo-wide inlining convention (cogni-consult, cogni-workspace, cogni-portfolio, cogni-trends) that the issue explicitly rules out; or a plugin version bump, which `scripts/check-version-bump.py` fails on feature branches.
- Judgment-only, not gated: whether the harness's documented rationale is good enough to stop a sixth suite from re-inlining, and whether the helper API names (`run_script`, `RUN_ERRFILE`, `RUN_CWD`) read clearly enough to be adopted rather than worked around. A human should eyeball the harness header.

- The plan-time AC has 4 criteria; a base-grounded regeneration derives ~15 gating items. The decisive gap: no AC criterion requires the inlined definitions to be REMOVED — a change that adds the harness, sources it, and leaves the local copies shadowing it satisfies all four criteria verbatim while fixing none of the reported duplication. The AC also never requires the new exit-code assertion to be non-zero (an all-`0` assertion is vacuous under the old piped runner), never protects the existing ~99 assertions from silent loss, and never mentions the latin-1 fixture writer the issue body names as duplicated verbatim. The acceptance bar was materially under-specified. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1192

**Summary:** Both perspectives (plugin-validator, contract-grading-reviewer) independently confirmed all 16 committed review-contract checklist items are satisfied, including live mutation-testing and execution of all five suites. The only un-waived findings are two minor, auto-fixable sourcing/documentation nits from plugin-validator; per Rule 5b they render as advisory and do not gate merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) `cogni-projects/tests/fixtures/test_helpers.sh:127` — The stdout sink is hard-coded to $TMPROOT/stdout.txt with no RUN_OUTFILE knob mirroring RUN_ERRFILE, and the header documents only RUN_ERRFILE and RUN_CWD — it never names the stdout path. Not a defect today (LAST_JSON is captured immediately after each run and no assertion reads full stdout), but a future fixture needing more than the last line has to reach into an undocumented internal path.
  - *Suggested fix:* Add a one-line mention of $TMPROOT/stdout.txt to the run_script header block, or add a symmetric one-shot RUN_OUTFILE alongside RUN_ERRFILE.

- **[minor]** (auto-fixable) `cogni-projects/tests/test_portfolio_init.sh:32` — The source line is `. "$TESTS_DIR/fixtures/test_helpers.sh"` (same at test-render-dashboard.sh:26 and test_staffing_score.sh:28), not the in-repo precedent form `. "$(dirname "$0")/fixtures/test_helpers.sh"` used at all ~20 cogni-knowledge call sites. Functionally equivalent (TESTS_DIR is already absolute and cwd-independent) but the repo now has two documented sourcing idioms for the same pattern.
  - *Suggested fix:* Either switch the three suites to the precedent `. "$(dirname "$0")/fixtures/test_helpers.sh"`, or keep the TESTS_DIR form and add one clause to the harness header noting the deliberate deviation.

### Positive observations

- Item 5 (the checklist's decisive gap over the plan-time AC) is genuinely met: grep for the seven definitions plus failures=0 across cogni-projects/tests/*.sh returns zero matches — the inline copies are deleted, not shadowed. (plugin-dev:plugin-validator)
- Item 11's mutation check passes for real, not by inspection: restoring the piped shape in a scratchpad copy produced a failing assertion; the worktree was never modified. (plugin-dev:plugin-validator)
- The runner correctly propagates the invoked command's own exit status even under RUN_CWD scoping, preserving the exit-code fidelity the whole change is about. (plugin-dev:plugin-validator)
- Stderr handling is a net upgrade, not parity: every call now writes stderr to a real, per-call-truncated file instead of discarding it, so assert_no_traceback sites cannot read stale output. (plugin-dev:plugin-validator)
- The one-shot semantics of RUN_ERRFILE/RUN_CWD removes a whole class of fixture-leakage bugs, documented at the point of use. (plugin-dev:plugin-validator)
- write_latin1 improves on both blocks it absorbs: carries the rationale comment forward and adds a guard against silently leaving a 0-byte file on non-encodable payloads. (plugin-dev:plugin-validator)
- init_tmproot folds the chmod -R u+rwx into the shared EXIT trap without losing the root-user SKIP guard or cd-scoped invocation. (plugin-dev:plugin-validator)
- cogni-projects/CLAUDE.md documents why the two heredoc suites do not source the harness, resolving the checklist's 'all five' item on the record. (plugin-dev:plugin-validator)
- Harness has no shebang and no set -e, confirmed bash-3.2 clean via grep for declare -A/local -n/mapfile/readarray/case-modification expansions. (cogni-service:contract-grading-reviewer)
- run_script's RUN_CODE=$? is the statement immediately following the redirected invocation, with no pipe on that line — LAST_JSON is a separate, later read. (cogni-service:contract-grading-reviewer)
- The documented mutation check was independently performed: reverting run_script to the piped shape made an exit-code assertion fail as expected, then the harness was restored byte-identical. (cogni-service:contract-grading-reviewer)
- Assertion call-site counts are at or above baseline across all three suites and the spot-check labels survive verbatim. (cogni-service:contract-grading-reviewer)
- The three direct-invoke sites each set a fresh per-fixture RUN_ERRFILE that assert_no_traceback inspects — never /dev/null, never stale. (cogni-service:contract-grading-reviewer)
- All five suites were executed from the worktree during review: each exits 0 with zero FAIL lines. (cogni-service:contract-grading-reviewer)

### Recommendation

---

PR #1194, opened 2026-08-03
cogni-projects: catch-all envelope for staffing-score.py

### LINKED ISSUE #1186: cogni-projects: give staffing-score.py the sibling module-level catch-all envelope
## Context

Deferred from PR #1183 (which resolved #1173 by extracting the shared
`read_frontmatter(path) -> (fm, exc)` helper). Raised as a **minor,
explicitly non-gating** finding by `plugin-dev:plugin-validator` in the
merger verdict on that PR.

## Problem

`cogni-projects/scripts/staffing-score.py` is the only one of the three
sibling scripts whose `__main__` block has no catch-all envelope:

```python
if __name__ == "__main__":
    sys.exit(main(sys.argv[1:]))
```

Both siblings convert any escaping exception into the house
`{"success": false, ...}` envelope with exit 2 — see
`validate-entities.py`'s `except Exception as _exc:  # noqa: BLE001 —
deliberate catch-all` block and `render-dashboard.py`'s equivalent.

#1173 closed the *specific* non-UTF-8 read path in this script, and
`tests/test_staffing_score.sh`'s `assert_no_traceback` checks pin that. What
stays open is the *class*: any other escape still surfaces as a raw traceback
instead of a readable envelope — e.g. an `exec_module` failure inside
`_load_read_frontmatter()`.

## Why this was deferred rather than fixed in #1183

The verdict marks it judgment-only and explicitly outside that PR's
acceptance bar ("Do not treat as a merge blocker for this extraction"), and
"wrap main in an envelope" is a different change class from "extract a
shared read_frontmatter helper". Widening a five-line revise delta to carry
it would have blurred the PR's scope.

## Proposed change

Add the sibling catch-all to `staffing-score.py`'s `__main__` block, matching
`validate-entities.py`'s shape (message format `unexpected failure: %s: %s`
with the exception type name, exit code 2). Keep the existing `_fail`
contract for the errors `main()` already handles.

## Acceptance criteria

- [ ] `staffing-score.py`'s `__main__` converts an escaping exception into a
      `{"success": false, "data": {...}, "error": "unexpected failure: ..."}`
      envelope with exit 2, matching the two siblings
- [ ] A regression case in `tests/test_staffing_score.sh` proves a raised
      exception reports as an envelope rather than a traceback
- [ ] `bash cogni-projects/tests/test_staffing_score.sh` passes green

claude-cogni-work:
Closes #1186.

## Summary

`staffing-score.py` was the last of the sibling scripts whose `__main__` called
`main()` bare, so any exception `main()` has no guard for reached the caller as
a traceback instead of the house `{success, data, error}` envelope.

- **The envelope.** `__main__` now wraps `main()` in the same deliberate
  catch-all the siblings carry, emitting
  `{"success": false, "data": {}, "error": "unexpected failure: <Type>: <msg>"}`
  at exit 2. The structural template is `render-dashboard.py`, **not**
  `validate-entities.py` as the issue suggested: the latter's failure `data` is
  `{"errors": [], "warnings": []}`, schema-specific to its own success shape,
  where this script's `_fail` emits bare `{}`. Only the message format and the
  exit code come from the issue's named sibling. This script's `_fail` has no
  default `code`, so `2` is passed explicitly.
- **The regression fixture** pins the *class*, not one path. A manifest holding
  valid JSON of the wrong type (`[]`) parses cleanly, so it survives
  `score_portfolio`'s `except (OSError, ValueError)` manifest-read guard and
  only fails later at `_load_entities`' `manifest.get(...)` with an
  `AttributeError` — which `main()`'s `except RuntimeError` does not catch.
  Verified both ways: the fixture fails red on the unpatched script (exit 1,
  empty stdout, traceback) and green with the change.
- **Two exit-code statements this diff falsifies** are corrected in the same PR:
  the module docstring and `skills/projects-staff/SKILL.md` Step 2 both scoped
  exit 2 to "usage (wrong argument count)".

## Scope decisions

**Included beyond the acceptance criteria:** the docstring and SKILL.md
exit-code corrections. This PR is what makes them wrong — shipping the
behaviour change and deferring the one-clause documentation fix would leave the
plugin's own contract documentation contradicting its code. The SKILL.md edit is
a single in-place clause substitution plus the reflow it forces.

**Considered and deliberately excluded — a shared `unexpected_failure(exc)`
formatter in `_projects_lib.py`.** Four scripts now restate the same message
literal, and `_projects_lib.py` is the declared home for shared primitives, so
this is a real consolidation. It is not done here because it would touch
`validate-entities.py`, `render-dashboard.py`, `register-entity.py` and
`_projects_lib.py` — four files outside this change's scope — and the
`try/except` *structure* genuinely resists sharing, since the four sites do not
agree on a `_fail` signature or a failure `data` shape. Worth its own change.

**Also excluded:** a dedicated fixture for the `exec_module` failure the issue
names as an example. The catch-all now covers it as a class, but forcing it in a
test means corrupting the real sibling script on disk, which makes it a poor
regression vector.

**No version bump** — the post-merge workflow owns that, per repo convention.

## A note for the reviewer

Fixture D asserts the *generic* catch-all message for a failure that arguably
deserves a typed guard of its own: an `isinstance(manifest, dict)` check inside
`score_portfolio`'s existing `try` would report "manifest is not a JSON object"
rather than `AttributeError: 'list' object has no attribute 'get'`. That would
be a genuine improvement, and it would break `4a`/`4b`. The fixture's comment
says so explicitly, so a future editor re-points it at an escape that still has
no typed home rather than reading the failure as a regression. Flagging it here
because it is the one place this PR pins behaviour that a later change should be
free to improve.

## Test plan

- [x] `bash cogni-projects/tests/test_staffing_score.sh` — green, including the
      new `4a`/`4b`/`4c`.
- [x] Red-check: with only the `__main__` change reverted, `4a`/`4b`/`4c` all
      fail (exit 1, empty stdout, traceback on stderr). The fixture does not
      pass vacuously.
- [x] The four sibling suites still pass — `test-render-dashboard.sh`,
      `test_portfolio_init.sh`, `test_register_entity.sh`,
      `test_validate_entities_refs.sh`. `tests/fixtures/test_helpers.sh` is
      untouched.
- [x] Manual: a wrong-shaped manifest prints one envelope line on stdout, zero
      bytes on stderr, exit 2.
- [x] Handled paths unchanged: no-args still exits 2 with the `usage:` envelope,
      a missing directory still exits 1 — the catch-all did not swallow the
      existing `_fail` contract.
- [x] Gates: `check-breadcrumbs.py` clean (0 violations), `check-version-bump.py`
      `ok` (0 plugins touched), `check-frontmatter.sh` clean (0 offenders),
      `measure-skill-length.sh` 9,335 / 18,000 chars (`over_cap: false`).

### DIFF
--- a/cogni-projects/scripts/staffing-score.py
+++ b/cogni-projects/scripts/staffing-score.py
@@ -43,7 +43,8 @@ projects-portfolio.json manifest.
 Output: a single JSON line following the repo contract
   {"success": bool, "data": {...}, "error": str}
 Exit: 0 ok / 1 domain failure (bad portfolio / unreadable manifest / broken
-install) / 2 usage (wrong argument count).
+install) / 2 usage (wrong argument count) or an unexpected failure the
+module-level catch-all converted into the envelope.
 """
 
 import datetime
@@ -412,7 +413,7 @@ def main(argv):
     read_frontmatter = _load_read_frontmatter()
     if read_frontmatter is None:
         # A broken/partial install (validator missing), not a caller usage
-        # error — so exit 1 (domain failure), reserving exit 2 for bad args.
+        # error — so exit 1 (domain failure), never exit 2.
         return _fail(
             "cannot load validate-entities.py:read_frontmatter (expected sibling "
             "of this script) — the cogni-projects install looks broken", 1
@@ -430,4 +431,15 @@ def main(argv):
 
 
 if __name__ == "__main__":
-    sys.exit(main(sys.argv[1:]))
+    # Same envelope contract as the sibling scripts: the errors main() knows
+    # about already route through _fail, but anything else — a hand-edited
+    # manifest of the wrong shape, say — would otherwise escape as a
+    # traceback, which reads to callers as neither success nor a reportable
+    # failure.
+    try:
+        _code = main(sys.argv[1:])
+    except Exception as _exc:  # noqa: BLE001 — deliberate catch-all
+        _code = _fail(
+            "unexpected failure: %s: %s" % (type(_exc).__name__, _exc), 2
+        )
+    sys.exit(_code)
--- a/cogni-projects/skills/projects-staff/SKILL.md
+++ b/cogni-projects/skills/projects-staff/SKILL.md
@@ -80,15 +80,16 @@ python3 "${CLAUDE_PLUGIN_ROOT:-$(ls -td "$HOME"/.claude/plugins/cache/*/cogni-pr
 ```
 
 It returns `{"success", "data", "error"}` (exit 0 ok / 1 domain failure / 2
-usage). On `success: false`, surface `error` to the user and stop. A domain
-failure (exit 1) usually means the portfolio directory or its manifest is
-unreadable — but an `error` naming `validate-entities.py` instead means a
-broken or partial plugin install, not a portfolio-data problem; say so rather
-than pointing the user at their entities. On success, `data.projects[]` holds
-only the non-closed projects (closed projects are skipped) and, per project, an
-`open_roles[]` array where each role carries a `candidates[]` list already ranked
-by combined score (best first) and a per-role `excluded_count` of consultants
-dropped for no availability overlap with that project's window.
+usage or unexpected failure). On `success: false`, surface `error` to the user
+and stop. A domain failure (exit 1) usually means the portfolio directory or
+its manifest is unreadable — but an `error` naming `validate-entities.py`
+instead means a broken or partial plugin install, not a portfolio-data
+problem; say so rather than pointing the user at their entities. On success,
+`data.projects[]` holds only the non-closed projects (closed projects are
+skipped) and, per project, an `open_roles[]` array where each role carries a
+`candidates[]` list already ranked by combined score (best first) and a
+per-role `excluded_count` of consultants dropped for no availability overlap
+with that project's window.
 
 ### Step 3: Write the staffing-recommendation artifact
 
--- a/cogni-projects/tests/test_staffing_score.sh
+++ b/cogni-projects/tests/test_staffing_score.sh
@@ -7,7 +7,9 @@
 #      unreadable one — the run still returns its {success, data, error}
 #      envelope rather than a bare traceback,
 #   C. usage and preflight errors still report through that envelope, so B
-#      cannot be satisfied by swallowing every failure.
+#      cannot be satisfied by swallowing every failure,
+#   D. an exception main() has no guard for still reports as an envelope
+#      rather than a bare traceback — the class B pins one instance of.
 #
 # stdlib-only (bash + python3, no pytest/pip), matching the house convention.
 #
@@ -165,6 +167,29 @@ run_score "$STDERRC" "$TMPROOT/not-a-portfolio"
 assert_exit  "3g non-portfolio dir exits 1"      1 "$RUN_CODE"
 assert_json  "3h non-portfolio reports envelope" "d['success'] is False and 'projects-portfolio.json' in d['error']"
 assert_no_traceback "3i non-portfolio stderr is clean" "$STDERRC"
+
+# ---------------------------------------------------------------------------
+# Fixture D — a manifest holding valid JSON of the wrong type parses cleanly,
+# so it survives the manifest-read guard and only fails later, deeper in, with
+# an exception main() has no guard for. Only the module-level catch-all keeps
+# that inside the envelope. No entities are seeded: the failure precedes any
+# entity I/O, so seeding them would imply a dependency that is not there.
+#
+# If a dedicated guard is ever added for a wrong-typed manifest — an
+# isinstance check in score_portfolio would be the natural home, and would
+# report better than the generic catch-all does — this fixture must be
+# re-pointed at an escape that still has no typed home, or it will fail on an
+# improvement rather than a regression.
+# ---------------------------------------------------------------------------
+PFD="$TMPROOT/bad-manifest"
+mkdir -p "$PFD"
+printf '[]' > "$PFD/projects-portfolio.json"
+STDERRD="$TMPROOT/stderr-d.txt"
+
+run_score "$STDERRD" "$PFD"
+assert_exit  "4a wrong-shaped manifest exits 2" 2 "$RUN_CODE"
+assert_json  "4b wrong shape reports envelope"  "d['success'] is False and d['error'].startswith('unexpected failure:')"
+assert_no_traceback "4c wrong shape stderr is clean" "$STDERRD"
 
 if [ "$failures" -gt 0 ]; then
   printf '\n%d assertion(s) failed\n' "$failures" >&2


claude-cogni-work:

#1186 end-to-end to a draft PR. Handing off to the merger identity for
never reviews its own work."

`svc:needs-review` applied to this PR; #1186 marked `svc:resolved`.

sdh07:

_The merger formed this expectation from issue #1186 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix replaces the bare `__main__` at cogni-projects/scripts/staffing-score.py:432-433 (`sys.exit(main(sys.argv[1:]))`) with the house catch-all envelope its siblings already carry — render-dashboard.py:594-602 and register-entity.py:356-364 (`try: _code = main(...)` / `except Exception as _exc:  # noqa: BLE001 — deliberate catch-all` / `_code = _fail("unexpected failure: %s: %s" % (type(_exc).__name__, _exc), 2)` / `sys.exit(_code)`), with validate-entities.py:518-530 supplying only the message format and exit code 2. The load-bearing local detail is that this script's `_fail` (staffing-score.py:94-99) is `def _fail(message, code)` with NO default, unlike render-dashboard.py:513 (`code=2`) and register-entity.py:78 (`code=2`), so a copy-pasted one-argument `_fail(...)` call would raise TypeError inside the handler and re-emit the very traceback the fix removes; its failure `data` is bare `{}`, so validate-entities.py:527's schema-specific `{"errors": [], "warnings": []}` must NOT be copied. `main()` (400-429) and its `except RuntimeError` guard (421-424) stay untouched so manifest failures remain exit 1, and the two exit-code statements the change falsifies — the docstring `Exit:` clause at staffing-score.py:45-46 and the `reserving exit 2 for bad args` comment at 414-415 — plus skills/projects-staff/SKILL.md:82 (`exit 0 ok / 1 domain failure / 2 usage`) must be widened. The regression fixture in cogni-projects/tests/test_staffing_score.sh must force a non-RuntimeError escape that survives main()'s isdir/isfile preflights and score_portfolio's `except (OSError, ValueError)` json.load guard (301-306) — e.g. a valid-JSON-but-non-dict manifest, which reaches `manifest.get(kind, [])` at _load_entities:275 and raises AttributeError — and assert exit 2 + envelope + clean stderr using only the existing primitives in tests/fixtures/test_helpers.sh.

**Acceptance checklist:**
- [ ] Scope containment: the diff changes exactly the three files cogni-projects/scripts/staffing-score.py, cogni-projects/skills/projects-staff/SKILL.md, cogni-projects/tests/test_staffing_score.sh, creates no new files, and does NOT modify cogni-projects/tests/fixtures/test_helpers.sh (five suites source it: test_portfolio_init.sh, test_register_entity.sh, test_validate_entities_refs.sh, test-render-dashboard.sh, test_staffing_score.sh).
- [ ] staffing-score.py's `if __name__ == "__main__":` block now assigns `_code = main(sys.argv[1:])` inside a `try:`, and `sys.exit(_code)` is the final statement OUTSIDE that try — grep the block for both `_code = main(` and a trailing `sys.exit(_code)`; a diff that keeps `sys.exit(main(...))` inside the try, or drops the `sys.exit` entirely, fails.
- [ ] The handler is `except Exception as _exc:` — not a bare `except:` and not `except BaseException:` — so KeyboardInterrupt and SystemExit still propagate; confirm the literal token `Exception` and the absence of `BaseException` in the new block.
- [ ] The new handler carries the house rationale marker `# noqa: BLE001 — deliberate catch-all` on the except line, matching render-dashboard.py:600 and register-entity.py:362, plus a semantic comment above the `try:` explaining the envelope contract (both siblings carry one at render-dashboard.py:595-597 and register-entity.py:357-359).
- [ ] The failure path yields exit code 2 EXPLICITLY: either the `_fail(...)` call passes `2` as a second positional/keyword argument, or `_fail`'s signature at staffing-score.py:94 gains a `code=2` default. Grep the new call site — a bare `_fail("unexpected failure: ...")` copied from render-dashboard.py:601 is a TypeError against this script's two-required-argument `_fail` and must fail this item.
- [ ] If `_fail`'s signature at line 94 was changed to add a default, all four pre-existing call sites (402, 405, 407, 416) still pass their explicit codes (2, 1, 1, 1) and are unmodified — the exit-code semantics of usage vs domain failure must not shift.
- [ ] The emitted failure `data` is the bare `{}` this script's `_fail` already produces (line 97) — the diff must NOT inline validate-entities.py:527's `{"errors": [], "warnings": []}`, which is schema-specific to that script's success shape. Confirm no `"errors"`/`"warnings"` literals appear in staffing-score.py.
- [ ] The error string uses the sibling format `"unexpected failure: %s: %s" % (type(_exc).__name__, _exc)` — grep for the literal `unexpected failure: %s: %s` and for `type(_exc).__name__`, so the exception class name is present in the message as the issue's Proposed change specifies.
- [ ] `main()` (staffing-score.py:400-429) is otherwise unchanged — in particular the `try/except RuntimeError` at 421-424 still returns `_fail(str(exc), 1)`, so an unreadable/invalid manifest remains a domain failure at exit 1 and is NOT reclassified as an `unexpected failure` exit 2. The catch-all must be additive, never a swallow-everything that subsumes handled paths.
- [ ] Documentation truthfulness — all three exit-code statements the change falsifies are widened: (a) the docstring `Exit:` clause at staffing-score.py:45-46 now states exit 2 covers a usage error OR a catch-all-converted unexpected failure; (b) the in-file comment at staffing-score.py:414-415 (`reserving exit 2 for bad args`) no longer claims exit 2 is exclusive to bad args; (c) skills/projects-staff/SKILL.md:82 (`exit 0 ok / 1 domain failure / 2 usage`) is widened the same way. Confirm each of the three sites in the diff. (Verified cogni-projects/README.md:89 and cogni-projects/CLAUDE.md:26 carry only a one-line description with no exit-code claim, so no further doc site is in scope.)
- [ ] The SKILL.md edit is a minimal in-place clause substitution inside Step 2 — no new section, no new workflow step, and no change to the frontmatter block (SKILL.md:1-14) or to the Step 3 artifact template; the surrounding Step 2 guidance about `success: false` and the `validate-entities.py`-naming broken-install hint (SKILL.md:83-87) is preserved.
- [ ] skills/projects-staff/SKILL.md still has valid frontmatter with all required keys (name, description, allowed-tools) — adjudicate by running the frontmatter gate against the PR head and requiring `data.clean`.
- [ ] The diff introduces no breadcrumb tokens — no issue number (#1186, #1183, #1173), branch name, PR reference, or version marker leaks into SKILL.md or into the source/test comments; adjudicate by running the breadcrumb gate against the PR head.
- [ ] skills/projects-staff/SKILL.md remains within its skill length / body char budget after the edit — adjudicate by running the skill length gate against the PR head.
- [ ] test_staffing_score.sh gains a new fixture whose assertions are inserted BEFORE the summary block at lines 169-173 (`if [ "$failures" -gt 0 ]; then ... exit 1; fi`). Assertions placed after that block are counted after the exit decision and would pass vacuously — check the line ordering in the diff.
- [ ] The new fixture forces a genuine escape: it drives staffing-score.py past main()'s `len(argv)`, `os.path.isdir`, `os.path.isfile(projects-portfolio.json)` and `_load_read_frontmatter()` preflights (400-419) AND past score_portfolio's `except (OSError, ValueError)` json.load guard (301-306), so the raised exception is NOT a RuntimeError and is not already handled at 421-424. A valid-JSON-but-non-dict manifest (e.g. `[]`) reaching `manifest.get(kind, [])` at _load_entities:275 is the base-derived vector; any vector that demonstrably tracebacks without the new catch-all satisfies this. A fixture that merely re-tests an already-handled path (bad args, missing dir, unreadable manifest) does NOT satisfy it.
- [ ] The new fixture asserts all four properties, using the existing harness helpers: `assert_exit` for exit code 2, `assert_json` for `d['success'] is False`, `assert_json` for `'unexpected failure' in d['error']`, and `assert_no_traceback` against that fixture's own stderr file — the last is what proves the envelope replaced the traceback rather than accompanying it.
- [ ] The new fixture calls `run_score` with a stderr sink as its FIRST argument (e.g. `run_score "$STDERRD" "$PFD"`), matching run_score's `RUN_ERRFILE="$1"; shift` contract at test_staffing_score.sh:34-37 — a call that omits the sink silently consumes the portfolio dir as the stderr file and the fixture tests nothing.
- [ ] The suite header `Covers:` block (test_staffing_score.sh:4-10) is extended with the new fixture's bullet, keeping the file's stated coverage honest — the header currently enumerates A, B, C only.
- [ ] Every pre-existing assertion is preserved unweakened: labels 1a-1g, 2a-2e and 3a-3i still exist with the same expected values (notably `1a clean portfolio exits 0`, `3a no-args exits 2`, `3d missing dir exits 1`, `3g non-portfolio dir exits 1`) — the catch-all must not have been made to pass by relaxing an existing pin.
- [ ] `bash cogni-projects/tests/test_staffing_score.sh` exits 0 and prints `All staffing-score.py assertions passed` with no `FAIL` lines, and the four sibling suites that share test_helpers.sh (test_portfolio_init.sh, test_register_entity.sh, test_validate_entities_refs.sh, test-render-dashboard.sh) still pass — proving the change is contained.

**Risk watchlist:**
- The signature trap: `_fail(message, code)` at staffing-score.py:94 has NO default, unlike render-dashboard.py:513 and register-entity.py:78. A verbatim copy of render-dashboard.py:601's one-argument `_fail(...)` raises TypeError inside the except handler — producing a traceback and a Python-default exit status instead of the envelope, i.e. the exact defect the issue is about, now hidden behind a fix that looks right on a skim.
- Copying validate-entities.py's shape too literally: its handler inlines `data: {"errors": [], "warnings": []}`, which is schema-specific to the validator's success envelope and wrong for this script, whose failure data is bare `{}`. The issue names validate-entities.py as the template, but only its message format and exit code transfer.
- Over-broad catching: a bare `except:` or `except BaseException:` would swallow KeyboardInterrupt and SystemExit, converting a Ctrl-C into a false `unexpected failure` envelope. Unlike render-dashboard.py, this script has no argparse, so there is no legitimate SystemExit to convert here.
- Catch-all as swallow-all: widening the guard at main():421-424 from `except RuntimeError` to `except Exception`, or moving the manifest-failure path into the new handler, would silently reclassify handled exit-1 domain failures as exit-2 unexpected failures — regressing assertions 3e/3h in spirit even if they still pass on their string checks.
- A vacuous regression fixture: one that raises through an ALREADY-handled path (bad args, missing dir, unreadable/invalid-JSON manifest — all caught by main() or by score_portfolio's `except (OSError, ValueError)`) would pass green today with or without the catch-all, proving nothing. Equally vacuous: assertions appended after the `failures` summary block at lines 169-173, which runs `exit 1` before they are counted.
- Losing `sys.exit(_code)`: if the diff sets `_code` but never exits with it, the script would return 0 on every path and fixtures 3a/3d/3g would fail — but a diff that instead keeps `sys.exit(main(...))` inside the try and only adds the except would still exit correctly on the happy path while never assigning `_code` on the failure path.
- Doc drift left behind: the issue's acceptance criteria say nothing about documentation, so a diff that satisfies all three criteria can still leave staffing-score.py:45-46, :414-415 and SKILL.md:82 asserting that exit 2 means 'usage' only — three now-false statements shipped alongside the fix.
- Scope creep into tests/fixtures/test_helpers.sh: adding a helper there (a new assert, a manifest-corrupting seeder) touches five suites for one fixture's convenience. The existing primitives (seed_portfolio, write_entity, run_script/run_score, assert_exit, assert_json, assert_no_traceback) are sufficient.
- Judgment-only, not gated: the issue names an `exec_module` failure inside `_load_read_frontmatter()` (via _projects_lib.py:36, which calls `spec.loader.exec_module(module)` unguarded) as its motivating example, but the issue phrases it as 'e.g.'. Whether shipping without a fixture for that specific vector is acceptable coverage — versus adding one — is a human call, not a gate.
- Judgment-only, not gated: whether the catch-all's exit code should be 2 (sibling parity, as the issue directs) or 1 (this script's own convention, where 2 is reserved for usage and every non-usage failure is 1) is a taste question the issue already decided; a reviewer may still want to eyeball that the resulting exit-code taxonomy reads coherently in the widened docstring.

- The plan-time acceptance criteria list 3 items (catch-all envelope, a regression case, suite green), but a base-grounded regeneration derives 21 gating items — including the `_fail(message, code)` no-default arity trap that would make a literal sibling copy raise TypeError inside the handler, the three exit-code statements the change falsifies (staffing-score.py:45-46, :414-415, SKILL.md:82), and the fixture-placement-before-the-summary-block requirement. Notably, skills/projects-staff/SKILL.md appears in changed_files but no acceptance criterion mentions documentation at all, so the acceptance bar was materially under-specified relative to the change's actual footprint. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1194

**Summary:** All four perspectives passed and all 21 committed review-contract checklist items were graded satisfied; the two remaining findings are un-waived minors on SKILL.md documentation clarity, surfaced as advisory notes (Rule 5b) rather than gating the merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-staff/SKILL.md:82-83` — Reading nit: `exit 0 ok / 1 domain failure / 2 usage or unexpected failure` scans ambiguously across the slash-delimited list — "or unexpected failure" can momentarily attach to the list as a fourth item rather than to "2 usage". The script docstring disambiguates with "2 usage (wrong argument count) or an unexpected failure..."; SKILL.md's compressed form loses that anchor.
  - *Suggested fix:* Write `2 usage error or unexpected failure` so the noun binds the disjunction to code 2 unambiguously.

- **[minor]** (auto-fixable) `cogni-projects/skills/projects-staff/SKILL.md:82-87` — Exit 2 now has two very different meanings, but the surrounding triage guidance still only explains exit 1. A reader who hits the newly reachable exit 2 with an `unexpected failure: <ExcName>: ...` error has no instruction distinguishing it from a caller usage error, and will most likely conclude they invoked the script wrongly rather than that they hit a scorer bug worth reporting. The truthfulness requirement (item 10c) is met, but the actionability of the widened statement lags it.
  - *Suggested fix:* Extend the existing Step 2 sentence in place (no new section, preserving item 11's minimality) with one clause noting that an `error` beginning `unexpected failure:` means the scorer hit an unhandled condition — report the exception name rather than re-running with different arguments.

### Positive observations

- Item 1 (scope containment): diff against merge-base 5e49b9e9 shows exactly the three declared files, all modifications, zero new files; tests/fixtures/test_helpers.sh untouched. (plugin-validator)
- Item 5 (the arity trap avoided): the new call passes an explicit second positional `2` — the contract's headline risk, since this script's `_fail(message, code)` has no default unlike its siblings. (plugin-validator)
- Item 16 (genuine escape, independently verified): running the BASE script at 5e49b9e9 against a `[]` manifest emits a raw traceback (AttributeError at _load_entities:275, exit 1), so the fixture is a real regression pin, not a re-test of an already-handled path. (plugin-validator)
- Item 9 (additive, not swallow-all): main()'s `except RuntimeError` guard is untouched, so manifest failures remain exit-1 domain failures. (plugin-validator)
- Item 21 (suite green + containment): the suite exits 0 with zero FAIL lines, and all four sibling suites sharing test_helpers.sh also pass. (plugin-validator)
- Plugin hygiene: plugin.json is valid, version 0.0.20 correctly NOT bumped on the feature branch; all five scripts byte-compile and the test script passes `bash -n`. (plugin-validator)
- Item 10c satisfied: SKILL.md:82 widened from the base's `2 usage` so the doc no longer asserts a now-false exclusivity, consistent with the two sibling doc sites in staffing-score.py. (skill-reviewer)
- Item 11 satisfied: single in-place clause substitution with only reflow; frontmatter, Step 3 artifact template, and the validate-entities.py broken-install hint all survive verbatim. (skill-reviewer)
- Items 12/13/14 satisfied: valid frontmatter with all three required keys, no breadcrumb tokens, chars_body=8516 against an 18000 cap. (skill-reviewer)
- The widened claim is true of the shipped code: the block assigns `_code` inside the try with `sys.exit(_code)` outside, catches `Exception` not `BaseException`, and passes the explicit `2`. (skill-reviewer)
- Contract items 10c, 11, 12, 13, 14 all satisfied. (skill-quality-reviewer)
- No orphaned exit-code claim survives the change: `exit` appears only at line 82, so the widening leaves no second site asserting the old `2 = usage` taxonomy. (skill-quality-reviewer)
- Step 2's existing `On success: false, surface error to the user and stop` already covers a catch-all envelope, and the exit-1 diagnostic hint is explicitly scoped to `A domain failure (exit 1)`. (skill-quality-reviewer)
- Description follows skill-creator conventions: third-person, nine concrete trigger phrases, pushy tail countering under-triggering. (skill-quality-reviewer)
- Body explains the why rather than stacking MUSTs; detail is offloaded to references/data-model.md (progressive disclosure applied correctly). (skill-quality-reviewer)
- All 21 contract checklist items graded satisfied with cited diff evidence; zero violated, zero unclear (no INTENT-UNCLEAR findings). (contract-grading-reviewer)
- item 16 independently confirmed: the `[]`-manifest vector clears main()'s preflights and score_portfolio's json.load guard, then raises AttributeError (not RuntimeError) — the suite yields exit 2, proving the vector was genuinely unhandled before the catch-all. (contract-grading-reviewer)
- item 21 confirmed by re-running the suite against the PR head worktree: exit 0, 25/25 OK including new 4a/4b/4c, plus all four sibling suites green. (contract-grading-reviewer)
- item 13 confirmed by re-running check-breadcrumbs.py directly against the PR head: 0 violations. (contract-grading-reviewer)

### Recommendation

---

PR #1204, opened 2026-08-04
cogni-help: report a failed repo-label query as its own error

### LINKED ISSUE #1198: cogni-issues: a failed gh label list is reported as "label(s) missing from repo"
## Problem

`cogni-issues`' create path validates requested labels against the target repo before invoking `gh issue create`. The query that fetches the repo's label set silences its own failure:

`cogni-help/skills/cogni-issues/scripts/gh-issues-helper.sh:140`

```bash
EXISTING_LABELS=$(gh label list --repo "$REPO" --json name --jq '.[].name' 2>/dev/null || true)
```

`2>/dev/null || true` collapses **every** failure mode — expired auth, network loss, a repo that does not exist, a token without `repo` scope — into an empty `EXISTING_LABELS`. The membership check at `:146` then finds no match for any requested label, and `:153` reports:

```json
{"error": "create: label(s) missing from repo", "repo": "...", "missing_labels": "bug,S3-minor", "hint": "Create the label on GitHub or omit it from --labels"}
```

An empty result and a failed query are indistinguishable to the code, so a transient auth or network fault is reported as a *data* problem on the remote repo, and the hint tells the caller to go create labels that already exist. The suggested remedy is the opposite of what the situation needs.

## Why this matters

The failure is worst exactly when it is hardest to diagnose. The call sits on the create path, so the issue is never filed — `:153` exits 1 before `gh issue create` runs. A caller reading the error has no signal that the cause was their credential rather than the repo's label set, and the natural response (creating a duplicate label, or dropping `--labels` to get past the gate) either fails again or files an untyped issue.

It also compounds with the label-count bug fixed in #1197: before that fix a >30-label repo produced this same wrong error for a *third* unrelated reason. Two distinct causes collapsing onto one misleading message is why the create path reads as flaky rather than broken.

## Proposed change

Distinguish a failed query from a genuinely empty label set, and give the failure its own error.

Capture the exit status rather than swallowing it — for example keep stderr, check the status explicitly, and emit a distinct `create: could not read repo labels` error carrying gh's own stderr and a hint pointing at auth/network/repo-existence, while leaving the existing `label(s) missing from repo` error for the case where the query genuinely succeeded and returned a set that lacks the requested label.

A repo with zero labels is a legitimate empty set and should still produce the existing "missing" error, not the new one — the discriminator is the query's exit status, not the emptiness of its output.

## Scope

- `cogni-help/skills/cogni-issues/scripts/gh-issues-helper.sh:140` (and the `:150-154` error branch)

Deliberately out of scope for #1197, which fixed the truncation half of the same line and named this defect in its "Not fixed here" section.

## Notes

Worth checking whether any sibling path in the same script silences a `gh` call the same way; `:208` and `:241` return gh's status directly, so this looks isolated to the create-path validator.

claude-cogni-work:
Closes #1198.

## Summary

The `create` path validated `--labels` against the repo's label set with a query whose failure was swallowed:

```bash
EXISTING_LABELS=$(gh label list --repo "$REPO" --limit 1000 --json name --jq '.[].name' 2>/dev/null || true)
```

`2>/dev/null` discarded gh's stderr and `|| true` discarded its exit status, so every failure mode — network loss, a repo the token cannot see, a scopeless or SSO-unauthorized token, an API 5xx — produced an empty `EXISTING_LABELS`. Every requested label then failed the whole-line `grep -qx` check, and the user was told `create: label(s) missing from repo` with a hint to go create labels that already exist. The remedy suggested was the opposite of the one needed, and because the branch exits before `gh issue create`, the issue was never filed.

The query now discriminates on **exit status, not emptiness**, using the same `if ! VAR=$(...); then` form the file already uses for `gh issue create`:

- A **failed** query emits a new, distinct `create: could not read repo labels`, carrying gh's own stderr in `detail` and a hint naming network access, repo visibility, and token scope.
- A **successful** query that returns nothing is unchanged — a repo whose label set is genuinely empty is a legitimate empty set and still reaches the existing `create: label(s) missing from repo` error.
- `--limit 1000` and its three-line comment are preserved verbatim. That flag is load-bearing: this repo carries 50 labels against gh's default page size of 30.

`SKILL.md`'s "Common errors" list gains a matching bullet, placed first so the list stays in the order the errors can actually fire.

## Verified behaviour

Both paths were exercised against the real `gh` CLI, not reasoned about:

| Case | Result |
|---|---|
| Failed query (nonexistent repo) | `{"error": "create: could not read repo labels", "repo": "…", "detail": "GraphQL: Could not resolve to a Repository with the name '…'. (repository)", "hint": "Check network access, …"}` · exit 1 |
| Genuinely absent label (real repo) | `{"error": "create: label(s) missing from repo", "repo": "cogni-work/insight-wave", "missing_labels": "…", …}` · exit 1 — unchanged |

Before this change the first row produced the second row's error.

## Scope decisions

**Kept the merged-stream `2>&1`; deliberately added no temp-file machinery.** This deserves recording, because the idiom is borrowed from a site where it is safe into a site where it might not be. At the `gh issue create` precedent the capture is only regex-parsed for a URL, but here the capture *becomes* the label set that `grep -qx` matches line by line — so a stderr line leaking in on a **successful** call could in principle read as an existing label. Adjudicated as negligible on evidence:

- In this exact invocation shape — command substitution, so stdout is a pipe — `gh label list --json name --jq` writes zero bytes to stderr on success. gh's update notifier prints only when stdout and stderr are *both* terminals, which command substitution structurally prevents.
- Even given a leak, `grep -qx` demands a whole-line exact match against a requested label name, and gh's notice lines are multi-word sentences and URLs.
- A leaked blank line cannot match either — empty requested labels are skipped by the guard above the check.

The failure mode is bounded to a false *accept*, never a false reject, and only against a label literally named after a gh notice line. An `mktemp` + cleanup dance would add its own failure mode, and its own JSON-error path for `mktemp` itself failing, to defend a compound improbability. The reasoning is recorded in a comment at the call site so the next reader does not have to re-derive it.

**Hint wording narrowed.** `check_gh` already runs `command -v gh` and `gh auth status` and exits before the label query is reached, so plain "not authenticated" is not a reachable cause here. The hint names network, repo visibility, and token scope instead of leading with `gh auth status`.

**Audited, not widened.** Every `gh` call site in the file was re-checked. The `gh --version` / `gh api user` probes swallow failure *by design* — they sit inside the `check` command, documented as the one command that probes without exiting on failure. The `gh auth status`, `gh issue create`, `gh issue list`, `gh issue view`, and search calls all handle status correctly. The label query was the sole masking site.

**No test added.** `cogni-help` has no `tests/` directory, no suite anywhere in this repo mocks or stubs the `gh` CLI, and CI runs no test job at all. Building the first `gh`-mocking harness is new infrastructure rather than a rider on a one-line bug fix — so this ships with the manual verification above instead. Flagging the gap explicitly rather than leaving it implied.

**Out of repo.** The same masking pattern reported in a sibling repository is a different repository, as the issue itself states.

## Checks run

| Check | Result |
|---|---|
| `bash -n` on the script | pass |
| Failed-query path | new error, exit 1 — verified live |
| Genuine-miss path | unchanged error, exit 1 — verified live |
| `--limit 1000` preserved | confirmed in diff |
| `check-breadcrumbs.py` | 0 violations |
| `check-version-bump.py` | `ok` — 14 plugins scanned, 0 touched, 0 violations |
| `check-frontmatter.sh cogni-help` | clean |
| `measure-skill-length.sh` | `chars_body` 17382 / cap 45000, `over_cap: false` |
| `plugin.json` / `marketplace.json` | absent from the diff — the post-merge workflow owns the bump |

### DIFF
--- a/cogni-help/skills/cogni-issues/SKILL.md
+++ b/cogni-help/skills/cogni-issues/SKILL.md
@@ -313,6 +313,9 @@ The helper returns:
 Capture `number` and `url` for the next step. If the helper exits non-zero, surface the
 JSON error to the user and stop — never retry blindly. Common errors:
 
+- `could not read repo labels` → the label query itself failed, so nothing is known about
+  the repo's label set; surface the `detail` field and have the user check network access,
+  that the repo exists, and that their token has repo scope.
 - `label(s) missing from repo` → tell the user which labels are missing and ask whether
   to file without them or to create the labels first.
 - `gh issue create failed` → surface the `detail` field; usually a network or auth issue.
--- a/cogni-help/skills/cogni-issues/scripts/gh-issues-helper.sh
+++ b/cogni-help/skills/cogni-issues/scripts/gh-issues-helper.sh
@@ -140,7 +140,17 @@ print(json.dumps({
       # --limit is required: gh label list defaults to 30, so on a repo with more
       # labels than that the ones past the first page read as missing and the
       # check below rejects a label that exists.
-      EXISTING_LABELS=$(gh label list --repo "$REPO" --limit 1000 --json name --jq '.[].name' 2>/dev/null || true)
+      #
+      # A failed query is told apart by exit status, never by an empty result: a
+      # repo whose label set is genuinely empty is a legitimate empty set and must
+      # still reach the "missing from repo" error below, not this one. Merging
+      # stderr into the capture is safe here even though it becomes the label set
+      # the whole-line match below reads, because gh's notices only print when
+      # stdout and stderr are both terminals, which command substitution prevents.
+      if ! EXISTING_LABELS=$(gh label list --repo "$REPO" --limit 1000 --json name --jq '.[].name' 2>&1); then
+        emit_error "create: could not read repo labels" repo "$REPO" detail "$EXISTING_LABELS" hint "Check network access, that the repo exists and is visible to you, and that the token has repo scope (gh auth status)"
+        exit 1
+      fi
       MISSING=""
       IFS=',' read -r -a LABEL_ARR <<< "$LABELS_CSV"
       for LBL in "${LABEL_ARR[@]}"; do


claude-cogni-work:
## Ship-ready

Authored by the cogni-service **author** instance and handed off for independent review. The author instance deliberately ran no review gate and did not flip this PR out of draft — forming the acceptance expectation is the merger's job, and grading a diff against the author's own PR body would defeat the two-identity split.

**Issue:** #1198 · **Branch:** `fix/issue-1198-label-query-failure-error` · **Scope:** full (`Closes #1198`)

### What the merger should know

- **The issue's line numbers are stale.** The label-count fix landed on this same line and shifted everything: the masking query is at `:143`, not `:140`, and the error branch at `:153-157`, not `:150-154`. All anchors were re-verified against `origin/main` before planning.
- **Both behavioural paths were exercised against the real `gh` CLI**, not reasoned about — a failed query against a nonexistent repo now returns the new error with gh's GraphQL message in `detail`, and a genuinely absent label against a real repo still returns the unchanged missing-label error. Both exit 1.
- **The one judgement call worth grading** is keeping the merged-stream `2>&1` rather than redirecting stderr to a temp file. The capture *becomes* the label set that the whole-line match reads, so the borrowed idiom is not automatically safe here. The reasoning for judging the risk negligible is in the PR body and recorded in a comment at the call site. A competing plan proposed the temp-file route; it was considered and rejected on evidence, not overlooked.
- **No test ships with this fix.** `cogni-help` has no test directory, nothing in the repo mocks the `gh` CLI, and CI runs no test job — so the gap is stated rather than papered over. Verification is the manual matrix in the PR body.

### Gates already run by the author

`bash -n` · `check-breadcrumbs` (0 violations) · `check-version-bump` (ok, 0 touched) · `check-frontmatter cogni-help` (clean) · `measure-skill-length` (17382 / 45000, under cap) · `skill-quality-reviewer` (pass, 0 auto-fixable, 0 structural)

These are the author's own checks and are **not** a substitute for the merger's independent contract.

sdh07:

_The merger formed this expectation from issue #1198 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The single masking line is at `cogni-help/skills/cogni-issues/scripts/gh-issues-helper.sh:143` in the base (the issue's `:140` is stale — the sibling `--limit 1000` fix, whose comment sits at `:140-142`, shifted it), and a correct fix replaces only that line's redirection/`|| true` with an explicit exit-status check in the exact shape the same file already uses at `:168-171` for `gh issue create` (`if ! VAR=$(cmd 2>&1); then emit_error ...; exit 1; fi`) — that form is also what keeps `set -euo pipefail` (`:23`) from aborting the script before the status can be inspected, so no `set +e` toggling should appear. On failure it must call `emit_error` (`:37-48`, which writes JSON to stderr and does NOT exit, so an explicit `exit 1` must follow) with a distinct message such as `create: could not read repo labels`, a field carrying gh's own stderr, and a hint naming network / repo existence-visibility / token scope rather than "create the label". The `MISSING` loop at `:144-152` and the `create: label(s) missing from repo` branch at `:153-157` must be left byte-identical, so a query that succeeds and returns an empty or lacking label set still produces the old error — the discriminator is exit status only, never emptiness, and no `[ -z "$EXISTING_LABELS" ]` test may be introduced. Companion doc work belongs in `cogni-help/skills/cogni-issues/SKILL.md` "Common errors" (`:313-318`), which currently lists only `label(s) missing from repo` and `gh issue create failed` and would otherwise leave the skill with no guidance for a new error whose correct remedy is the opposite of the existing one.

**Acceptance checklist:**
- [ ] Exactly two files change — `cogni-help/skills/cogni-issues/scripts/gh-issues-helper.sh` and `cogni-help/skills/cogni-issues/SKILL.md`; no `.claude-plugin/plugin.json` or root `.claude-plugin/marketplace.json` version edit appears in the diff (per repo policy the post-merge workflow owns the bump).
- [ ] SKILL.md frontmatter remains valid with its required keys (`name`, `description`) intact — confirm via the frontmatter gate on the PR head, not by eyeballing.
- [ ] Plugin structure / manifest health is unchanged — confirm via the structure/manifest gate on the PR head.
- [ ] No new maintainer breadcrumb is introduced in SKILL.md or the script — no `#1198`/`#NNNN` issue or PR ref, version tag, or slice/finding code; confirm via the breadcrumb gate, and the rationale for the change is stated semantically instead.
- [ ] SKILL.md stays within its skill length / body cap after the added prose — confirm via the skill-length (body cap / char budget) gate on the PR head; if it exceeds, the base measurement is cited to show whether the cap was already exceeded before the change.
- [ ] The `--limit 1000` flag survives verbatim on the label query, and the three-line comment at base `:140-142` explaining why `--limit` is required is preserved unchanged — grep the head for `--limit 1000` in the `gh label list` call; dropping it silently re-opens the >30-label bug fixed by the sibling issue.
- [ ] The literal `2>/dev/null || true` no longer appears on the `gh label list` call — `grep -n '2>/dev/null || true' gh-issues-helper.sh` returns no hit on the create-path label query.
- [ ] `set -euo pipefail` at base `:23` is unchanged and the diff introduces no `set +e` / `set -e` toggling anywhere in the script — the status check must work with errexit on, not by disabling it.
- [ ] The capture is wrapped in a construct that makes the assignment's exit status inspectable under errexit — an `if ! VAR=$(...)` (or equivalent `if VAR=$(...); then ... else`) form matching the same-file precedent at base `:168`; a bare assignment followed by a separate `$?` test is a defect, because errexit aborts before the test runs.
- [ ] On query failure the script calls the existing `emit_error` helper (not a hand-rolled `echo '{...}'`), so gh stderr containing quotes, newlines, or backslashes is JSON-escaped by `emit_error`'s `json.dumps` at base `:39-47`.
- [ ] The new `emit_error` invocation passes an EVEN number of key/value arguments after the message string — `emit_error` iterates `pairs[i], pairs[i+1]` in steps of 2 (base `:44-45`), so an odd count raises IndexError and replaces the error JSON with a Python traceback.
- [ ] The new error's `error` string is distinct from `create: label(s) missing from repo` (e.g. `create: could not read repo labels`), so a caller branching on the JSON `error` field can tell the two apart.
- [ ] The new error payload carries gh's own captured stderr in a detail-style field (grep the diff for the captured variable being passed to `emit_error`), and a `hint` field naming query-failure causes — at least two of {network access, repo existence/visibility, token scope/auth}. The hint must NOT tell the caller to create a label.
- [ ] An explicit `exit 1` follows the new `emit_error` call — `emit_error` writes to stderr and returns, it does not exit (base `:37-48`); without it execution falls through to `gh issue create` and the issue is filed unvalidated.
- [ ] The failure branch fires before the missing-label loop, so a failed query can never reach `gh issue create` — confirm by reading the head's control flow order in the `create` case.
- [ ] The `MISSING` accumulation loop and the `create: label(s) missing from repo` branch (base `:144-157`) are semantically unchanged — same message string, same `repo` / `missing_labels` / `hint` fields, same `exit 1`.
- [ ] The discriminator is exit status ONLY: the diff introduces no `[ -z "$EXISTING_LABELS" ]`, `-n` emptiness test, or line-count check that would route a genuinely zero-label repo (gh exits 0, prints nothing) into the new error. Verify by grepping the head for any emptiness test on the captured variable between the query and the loop.
- [ ] The whole label-validation block stays guarded by `[ -n "$LABELS_CSV" ]` (base `:139`/`:158`), so a `create` invoked without `--labels` still performs no label query and no new error path.
- [ ] If the fix merges stderr into the same capture with `2>&1`, the merged text is consumed only by the whole-line `grep -qx` match at base `:149` (empty entries already skipped at `:148`) and the PR states why a stderr line cannot fabricate a label match; if instead a temp file / separate fd is used, it is created with `mktemp` and removed on BOTH the success and the failure path.
- [ ] The change is bash 3.2 compatible per the file's stated target at base `:16` — no `mapfile`/`readarray`, no `${VAR@Q}`, no associative arrays, no `local -n` introduced.
- [ ] SKILL.md's "Common errors" list (base `:316-318`, currently two bullets) gains a bullet naming the new error and prescribing a remedy distinct from the existing one — retry / check network, repo visibility, or token scope — rather than "create the label"; grep SKILL.md for the new error string.
- [ ] Verification evidence is present and specific: because `cogni-help` has no `tests/` directory and no gh mock harness, the achievable bar is an executed reproduction recorded in the PR (e.g. running the create path against a nonexistent/unreadable repo to show the new error, plus a successful-query case to show the old `label(s) missing from repo` error still fires) — a claim of "verified" with no transcript, command, or observed JSON does not satisfy this item.
- [ ] Sibling gh call sites in the same script are addressed as an audit, not widened: the deliberate failure-swallowing inside the `check` command (base `:95`, `:98`) is left intact, and the PR says so — `check`'s whole job is to report readiness without exiting, so 'fixing' those is a regression.

**Risk watchlist:**
- A fix that drops `|| true` without wrapping the assignment in `if !` — under `set -euo pipefail` the script then dies silently with no JSON error at all, which is a worse failure than the misleading one being fixed.
- A fix that uses emptiness (`[ -z "$EXISTING_LABELS" ]`) as the discriminator: it produces the right error for the reported symptom while breaking the legitimate zero-label repo case the issue explicitly calls out.
- Silently dropping `--limit 1000` or its explanatory comment while rewriting the line, re-opening the >30-label truncation bug the sibling issue just closed.
- Hand-rolling the new error JSON with `echo`/`printf` instead of `emit_error`: gh stderr routinely contains quotes and newlines, so the output stops being parseable JSON exactly when a caller most needs to parse it.
- Forgetting the trailing `exit 1` after `emit_error` — the create then proceeds to `gh issue create` with an unvalidated (empty) label set, converting a loud wrong error into a silent wrong success path.
- Scope creep into the sibling repository (`cogni-service/scripts/gh-issues.sh`, `label-manager.sh`) named in the issue comment — those live in a different repo and cannot be touched from this PR; a diff claiming to fix them would be fabricated.
- Over-reach into `:95`/`:98` inside the `check` command, where swallowing failure is the documented contract (base `:84-87`).
- Rewriting `emit_error`, `check_gh`, or the `usage()` command list as incidental cleanup, enlarging the blast radius of a one-line reliability fix.
- SKILL.md edits that restate the whole error-handling story rather than adding the one missing bullet, pushing the skill body over its cap for no user-visible gain.
- Judgment-only, not gated: whether merging stderr into the label capture (`2>&1`) is the right trade versus a temp file is a design call — the reasoning should be legible in the PR, but no gate can decide it.
- Judgment-only, not gated: the exact wording of the new hint (which of network / visibility / scope to name, and in what order) is a UX call a human should eyeball.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1204
**Perspectives consulted:** cogni-service:frontmatter-schema-gate, plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer

**Summary:** All five perspectives passed; the committed 23-item review contract was independently satisfied in full (zero unclear items) by both the contract-grading-reviewer and plugin-validator via live execution. The only un-waived finding is an advisory minor on a pre-existing, out-of-scope doc inconsistency; a second minor (pre-existing, systemic, empirically-disproved gate premise) is explicitly waived with justification. Neither gates the merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| cogni-service:frontmatter-schema-gate | ✅ pass |

### Findings (2)

- **[minor]** (waived) `cogni-help/skills/cogni-issues/SKILL.md:3` — check-frontmatter.sh unknown-key: non-standard top-level frontmatter key 'version' (schema_version 1 allows only allowed-tools, description, name for a skill). This PR did not touch the frontmatter; the key is pre-existing and systemic (22 of 120 SKILL.md files across 5 plugins carry it). The gate's premise -- that a non-standard key means the skill 'silently fails to register' -- is disproved here: every cogni-help skill carrying this key is loaded and available in the reviewing session. The authoritative plugin-dev:skill-development guide both carries 'version: 0.1.0' in its own frontmatter and prescribes it in its canonical frontmatter template, so the gate's schema contradicts the guide it enforces.
  - *Justification:* Pre-existing, PR-untouched, systemic across 5 plugins, and the gate's stated harm is empirically disproved (the flagged skills are registered and available). Waived to advisory so it annotates for a human without bouncing a correctly-scoped change; the gate defect is the thing that needs fixing, not this PR.
  - *Raised by:* cogni-service:frontmatter-schema-gate

- **[minor]** (auto-fixable) `cogni-help/skills/cogni-issues/SKILL.md:72` — Pre-existing, not introduced by this change (verified present on the PR head): the row the file labels the canonical command list documents `search <repo> "keywords" [--state open|all]`, but the helper accepts `[--state open|closed|all] [--limit N]` (usage line 11, arg loop lines 247-253), and this same SKILL.md contradicts its own table at line 189 by invoking `search ... --limit 5`.
  - *Suggested fix:* Change the row's command cell to: `bash gh-issues-helper.sh search <repo> "keywords" [--state open|closed|all] [--limit N]`. Out of scope for this PR -- the committed contract names widening the SKILL.md footprint as a watched risk; worth a follow-up. (ref: skill-creator Principle of Lack of Surprise / Writing Patterns)

### Positive observations

- Exit-status-only discrimination proven empirically: a gh stub returning exit 0 with zero output still yields `create: label(s) missing from repo`, and grep confirms no -z/-n/line-count test on EXISTING_LABELS between the query and the loop. (plugin-dev:plugin-validator)
- The regression is demonstrable: the base script against a nonexistent repo returns the inverted remedy, while the head returns the new error with real GraphQL stderr. The PR's verification table matches an independent reproduction verbatim, so its evidence is genuine rather than asserted. (plugin-dev:plugin-validator)
- emit_error is reused rather than hand-rolled, with an even argument count (6), so the IndexError-traceback failure mode cannot fire; confirmed live with adversarial stderr. (plugin-dev:plugin-validator)
- The explicit `exit 1` follows emit_error and the failure branch sits above the MISSING loop, so a failed query structurally cannot reach `gh issue create`. (plugin-dev:plugin-validator)
- Bash 3.2 compatibility is real, not assumed: /bin/bash -n (3.2.57) parses clean and every reproduction ran under that interpreter. (plugin-dev:plugin-validator)
- Scope held tight: the `check` command's deliberate failure-swallowing is untouched and the PR states why; emit_error/check_gh/usage() unmodified; the sibling repository is correctly left alone. (plugin-dev:plugin-validator)
- item 6: `--limit 1000` survives verbatim at :150 and its three-line comment is unchanged diff context, not a +/- line. (cogni-service:contract-grading-reviewer)
- item 7: grep for `2>/dev/null || true` on the head returns no hit. (cogni-service:contract-grading-reviewer)
- item 15: control flow at :150-153 runs before MISSING= at :154, so a failed query can never reach the missing-label loop or gh issue create. (cogni-service:contract-grading-reviewer)
- item 16: the MISSING loop and missing-labels branch diff byte-identical against base. (cogni-service:contract-grading-reviewer)
- item 17: grep for -z/-n tests on EXISTING_LABELS returns no hit -- the discriminator is exit status only. (cogni-service:contract-grading-reviewer)
- item 20: no mapfile/readarray/@Q/declare -A/local -n -- bash 3.2 compatible. (cogni-service:contract-grading-reviewer)
- item 22: the PR body records observed JSON for both paths plus a Checks-run table, not a bare 'verified' claim -- the achievable bar given no tests/ dir or gh mock harness exists. (cogni-service:contract-grading-reviewer)
- item 23: the `check` command block diffs byte-identical; the deliberate probe failure-swallowing is untouched and the PR says so. (cogni-service:contract-grading-reviewer)
- The new bullet names the exact error emitted by the sibling change, with the same `create: ` prefix elision the two existing bullets use, so a user pasting the message can find the entry. (plugin-dev:skill-reviewer)
- The remedy is genuinely distinct from the adjacent bullet and mirrors the script's own hint field, so agent-surfaced text and script-emitted text cannot contradict each other. (plugin-dev:skill-reviewer)
- Bullet ordering is correct rather than merely appended: the label-query failure is placed above `label(s) missing from repo`, matching the script's actual execution order. (plugin-dev:skill-reviewer)
- Placement is first in the list, matching the order the three errors can actually fire (label read -> label validation -> create), so the list reads as a sequence rather than an arbitrary set. (cogni-service:skill-quality-reviewer)
- Body is 17382 chars against a 45000 cap with detail already offloaded to references/issue-templates.md; the change adds ~230 chars. (cogni-service:skill-quality-reviewer)
- Frontmatter is valid and unchanged, with a description carrying both what and when plus English and German trigger phrases. (cogni-service:skill-quality-reviewer)

### Recommendation

---

Issue #1217, opened 2026-08-04
CI: run plugin tests/*.sh, and record the plugin-qualified hook-matcher rule as a cross-plugin convention


claude-cogni-work:
**Plugin:** repo-wide (`.github/workflows`, root `CLAUDE.md`)
**Marketplace:** insight-wave

## Current behavior

No CI job discovers or runs any plugin's `tests/*.sh`. `.github/workflows/lint.yml` runs exactly three jobs, all `python3 scripts/*.py` (breadcrumb guard, external-dispatch guard, version-bump gate). `cogni-consult` alone ships 8 standalone bash test scripts that only ever run when a human remembers to invoke them; other plugins ship more.

This surfaced while fixing a hook whose failure mode was silent by construction — the hook script exits 0 on every path, and a matcher that never matches raises nothing anywhere. A regression test was added for it, but a test nothing runs is documentation, not a guard, and the recurrence would be exactly as undetectable as the original.

## Desired behavior

1. A CI job that discovers and runs plugin `tests/*.sh`, so the existing suites become actual guards. Needs a decision on the contract those scripts must satisfy (they already share a convention: `set -u`, `pass`/`fail` counters, non-zero exit on any failure) and on how long the full sweep may take.
2. A one-line entry under root `CLAUDE.md` § Cross-Plugin Conventions recording the lesson that motivated the fix: **a hook matcher that matches on an agent name must accept the plugin-qualified `plugin:agent` form**, because that is how a plugin-supplied agent is dispatched. Today that fact lives only mid-paragraph in one plugin's dev guide, so the next plugin to add an agent-name matcher hits the same silent no-op.

## Motivation

Both items are the "carry the lesson / enforce it" half of the hook fix, and both are genuinely separable from it: item 1 is repo-wide CI policy touching every plugin's test suite, item 2 edits the shared root guide. Neither belongs inside a single-plugin hook fix, and bundling them would have widened that PR's blast radius from one plugin to the whole repo.

## Acceptance criteria

- [ ] A CI job runs plugin `tests/*.sh` (or a documented, deliberate decision not to, recorded where the convention lives)
- [ ] Root `CLAUDE.md` § Cross-Plugin Conventions states the plugin-qualified agent-name matcher rule
- [ ] `cogni-consult/CLAUDE.md`'s "not CI-enforced, run it by hand" caveat is updated if item 1 lands

claude-cogni-work:
## Triage Report

- **Severity:** S4-enhancement
[...]

PR #1218, opened 2026-08-04
cogni-consult: fix the dead SubagentStart hook matcher

### LINKED ISSUE #1215: cogni-consult: verify the SubagentStart hook event actually fires
## Problem

`cogni-consult/hooks/hooks.json` declares its hook on the event `SubagentStart`:

```json
{
  "hooks": {
    "SubagentStart": [
      {
        "matcher": "^(consult-persona-challenger|consult-empathy-mapper|consult-framework-adherence-reviewer|consult-dashboard-refresher)$",
        ...
      }
    ]
  }
}
```

This is the **only** occurrence of that event name in the repo — every other `hooks.json`
(cogni-portfolio, cogni-visual, cogni-workspace) uses `PreToolUse` or `SessionStart`.
`plugin-dev:plugin-validator` flagged it twice during the review of #1203 and could not
confirm `SubagentStart` is a supported Claude Code hook event. The documented event set
includes `SubagentStop`, which is the near-miss worth ruling in or out.

If the event never fires, the hook is a silent no-op — and the consequence is not cosmetic.
`cogni-consult/CLAUDE.md` states that language reaches subagents *only* through this hook:
a subagent's system prompt is its own body plus a notes block and the environment info, with
no settings `language` key, no `CLAUDE.md`, and no active output style. So a dead event means
the four `consult-*` agents silently lose both the workspace interaction language and
`references/subagent-output-contract.md`.

## Why now

Pre-existing — the hook predates #1203 and that PR did not touch `hooks.json`. It is filed
now because #1203 added a `hooks` keyword to `cogni-consult/.claude-plugin/plugin.json` and
mirrored it into `marketplace.json`, so the plugin now *advertises* this capability at the
marketplace level. Advertising a hook that may not fire is worse than the prior silence.

## Proposed change

1. Confirm whether `SubagentStart` is a supported event on the targeted Claude Code version —
   empirically, not from documentation alone: dispatch one of the four matched agents and check
   whether `hooks/on-subagent-start.sh` actually runs.
2. If it fires: no code change. Record the confirmation in `cogni-consult/CLAUDE.md` next to
   the existing language-delivery note so the next reviewer does not re-raise it.
3. If it does not fire: migrate to a supported seam and re-verify that the interaction language
   plus `references/subagent-output-contract.md` still reach all four agents. Keep the doctrine
   in the reference file, not in the hook script or the agent bodies.

## Acceptance criteria

- [ ] A recorded empirical result — the hook script observably ran (or did not) on a real
      dispatch of one matched `consult-*` agent
- [ ] If unsupported: the four agents receive the interaction language and the output contract
      through a seam that is verified to fire, with the same reference-file indirection
- [ ] If supported: a note in `cogni-consult/CLAUDE.md` recording the verification, so the
      finding stops recurring in plugin-validator runs
- [ ] `cogni-consult` keeps the `hooks` keyword in both `plugin.json` and the `marketplace.json`
      mirror only if the hook demonstrably fires

claude-cogni-work:
Closes #1215.

## Summary

- `hooks/hooks.json`: widen the `SubagentStart` matcher to `^(cogni-consult:)?(consult-persona-challenger|consult-empathy-mapper|consult-framework-adherence-reviewer|consult-dashboard-refresher)$`. A plugin-supplied agent is dispatched under its plugin-qualified name; the shipped bare-name-anchored regex never matched, so the hook was a silent no-op and the four `consult-*` agents received neither the interaction language nor `references/subagent-output-contract.md`.
- `tests/test_subagent_start_hook.sh` (new): standalone bash regression test in the existing `cogni-consult/tests/` convention. Reads the regex back out of `hooks.json` (never a hardcoded copy), derives the agent list from `agents/` (so a fifth agent cannot drift past the loops), pins both the bare and the qualified form, and asserts on the emitted envelope rather than the exit status.
- `CLAUDE.md`: record the empirical verification next to the existing language-delivery doctrine, so the finding stops recurring in `plugin-dev:plugin-validator` runs.
- No version bump; no manifest edit. The `hooks` keyword stays in `plugin.json` and its `marketplace.json` mirror.

## The issue's premise was false

The issue asked whether `SubagentStart` is a supported event. It is — the defect is a matcher that never matches.

| Check | Method | Result |
|---|---|---|
| Is `SubagentStart` supported? | Claude Code changelog: added 2.0.43, refined 2.1.142 / 2.1.199; installed build 2.1.220 | **Yes** |
| Is `on-subagent-start.sh` correct? | Run standalone with `CLAUDE_PLUGIN_ROOT` set | **Yes** — valid envelope, exit 0, carrying the language block, `## Audience`, `## Register`. No silent-empty path even with no workspace resolvable |
| Does the hook reach a matched agent? | Live dispatch of `cogni-consult:consult-dashboard-refresher`, asked to report its injected context | **No** — it received none of it |
| Does the matcher match the dispatched name? | Regex test | Matches bare `consult-dashboard-refresher`; **does not match** `cogni-consult:consult-dashboard-refresher` |

`SubagentStop`, the near-miss the issue suggested, is a different lifecycle point (subagent *finishing*) and is unsuited to dispatch-time context injection — ruled out.

The recurring "could not confirm" signal came from `plugin-dev`'s `scripts/validate-hook-schema.sh`, whose `VALID_EVENTS` whitelist predates the event; it prints `Unknown event type: SubagentStart` and then **exits 0**. That is a warning about the validator, not evidence about the event. `plugin-dev` is an external plugin and cannot be fixed from this repo — flagged upstream, out of scope here.

## What is proven, and what is not

Everything in the table above was established by execution before this PR, and the regression test was confirmed to actually bite: reverting `hooks.json` to the old regex fails exactly the four qualified-name assertions and nothing else.

**Not proven, and not provable in the authoring session:** that the widened matcher fires. Hooks load at session start, and the session loads the plugin from the installed cache rather than from this branch — so post-fix confirmation requires the merge, a plugin-cache refresh, and a *new* session. That check is the last item of the test plan below, not a claim made here. The qualified-name explanation is well-evidenced (script always emits + nothing arrived + regex demonstrably rejects the qualified string), but it remains an inference. If the post-merge check still shows nothing, reopen rather than treating this as shipped.

## Scope decisions

- **`consult-dashboard-refresher` stays in the matcher.** It is the one matched agent with no `interaction_language` input and no `## Output Contract` section, and it emits no free user-facing prose — so dropping it is arguable. Not done here: nothing is defective about its inclusion, the injection is small and non-blocking, and it is the cheapest live probe for the post-merge verification.
- **Widened rather than replaced the seam.** A `"*"` matcher was rejected — it would inject the cogni-consult register contract into every other plugin's subagents. Migrating to a skill-level seam (passing the contract at each of the ~7 dispatch call sites) was rejected too: the hook and script are correct and the call sites already pass `interaction_language` correctly, so a one-line matcher fix has a far smaller blast radius than editing five SKILL.md files.
- **No follow-up issue filed.** The remaining work is one post-merge observation, listed in the test plan — not a separable deliverable.

## Test plan

- [x] `bash cogni-consult/tests/test_subagent_start_hook.sh` — 15 assertions, all `OK`, exit 0.
- [x] Negative control: restoring the old regex fails exactly the 4 qualified-form assertions, exit 1.
- [x] All 7 sibling tests in `cogni-consult/tests/` still pass.
- [x] `hooks.json` parses as JSON.
- [x] `git diff origin/main -- cogni-consult/.claude-plugin/plugin.json .claude-plugin/marketplace.json` is empty — no version touched, `hooks` keyword retained.
- [x] Repo guards clean: `check-breadcrumbs.py`, `check-version-bump.py`, `check-external-dispatch.py` all exit 0 with zero violations.
- [ ] **Post-merge, new session (the one open item):** once the plugin cache refreshes to the merged version, dispatch a matched agent (e.g. `cogni-consult:consult-dashboard-refresher`) and ask it to report the context it was given. Expect `# Interaction language and output register`, `## Language (workspace default)`, `## Audience`, `## Register`. If absent, reopen — the qualified-name hypothesis is wrong and the next step is to capture the literal string the harness matches against.

## Follow-up issues

- #1217 — no CI job discovers plugin `tests/*.sh`, so this regression test (and the 7 existing cogni-consult suites) only run by hand; and the plugin-qualified matcher rule belongs in root `CLAUDE.md` § Cross-Plugin Conventions where the next plugin author will find it. Both are repo-wide and deliberately kept out of this single-plugin fix.


claude-cogni-work:

Authored under the two-instance posture (`--role author`), so this PR stays draft and carries `svc:needs-review`. The merger instance forms its own acceptance checklist from the issue before reading this diff.

**Issue:** #1215 · **Follow-up filed:** #1217

### What a reviewer should check hardest

The fix rests on an **inference**, and the PR says so: the hook was proven dead (script emits correctly standalone, a live matched-agent dispatch received nothing) and the shipped regex was proven not to match the plugin-qualified name — but that the *widened* regex fires cannot be shown from the authoring session, because hooks load at session start from the installed plugin cache, not from this branch. The post-merge check is the one unchecked box in the test plan. If a reviewer wants that closed before merge rather than after, that is a legitimate call to make here.

### Verification already run

- New test: 15 assertions, exit 0.
- Negative control: reverting `hooks.json` to the old regex fails exactly the 4 qualified-name assertions, exit 1 — the test provably bites.
- All 7 sibling `cogni-consult` tests still pass.
- `check-breadcrumbs.py`, `check-version-bump.py`, `check-external-dispatch.py` — all exit 0, zero violations. No version touched.

### Deliberate scope boundaries

- `consult-dashboard-refresher` kept in the matcher (argued in the PR body).
- Bare-name matching kept as an explicitly-labelled hedge, not an observed requirement.
- CI test discovery and the root `CLAUDE.md` convention line went to #1217 rather than widening this single-plugin fix repo-wide.

sdh07:

_The merger formed this expectation from issue #1215 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix must establish by observation (not documentation) whether the SubagentStart hook actually delivers its additionalContext to the four consult-* agents, leave that delivery path working, and record the observation durably next to the existing doctrine bullet in cogni-consult/CLAUDE.md:143. The issue frames this as a binary (fires -> doc note only; does not fire -> migrate seam), but the base code points at a third and more likely diagnosis the issue does not name: cogni-consult/hooks/hooks.json:5 anchors its matcher fully (^(...)$) on BARE agent names, while every cross-plugin reference in this plugin uses the plugin-qualified cogni-consult:<name> form (cogni-consult/skills/consult-design-thinking/SKILL.md:183, skills/consult-resume/SKILL.md:205, README.md:57-64) — so a qualified dispatch name would never match and the hook is a silent no-op even though the event itself is fine. Under that diagnosis the expected change is a one-line widening of the matcher to accept both name forms while staying anchored and staying narrow (never '*' or '.*', which would inject cogni-consult's register contract into every other plugin's subagents), plus a standalone regression test under cogni-consult/tests/ following the test_register_generator.sh convention (set -u not set -e, mktemp -d + trap, pass/fail counters, stdlib-only per cogni-consult/CLAUDE.md:175) that reads the regex OUT of hooks.json rather than hardcoding a copy. Invariants the fix must preserve: the register doctrine stays in references/subagent-output-contract.md (read at hooks/on-subagent-start.sh:46) and is never inlined into the hook script or the four agents/consult-*.md bodies; the event key in hooks.json stays equal to the hookEventName emitted at on-subagent-start.sh:116; and no version is touched in cogni-consult/.claude-plugin/plugin.json or the root .claude-plugin/marketplace.json mirror (root CLAUDE.md:107-112, CI 'Version-bump gate').

**Acceptance checklist:**
- [ ] cogni-consult/hooks/hooks.json still parses as valid JSON: running `python3 -m json.tool cogni-consult/hooks/hooks.json` from the repo root exits 0.
- [ ] The hooks.json event key still equals the hookEventName string emitted by the script: grep the same literal (base: `SubagentStart`) in both cogni-consult/hooks/hooks.json and cogni-consult/hooks/on-subagent-start.sh (base line 116). If either changed, BOTH changed on this commit — a mismatch is a dead delivery path and reproduces the bug the issue reports.
- [ ] The hook entry still invokes the unchanged script path `${CLAUDE_PLUGIN_ROOT}/hooks/on-subagent-start.sh` with a `timeout` key present (base hooks.json lines 9-10) — the fix must not silently relocate, rename, or drop the timeout on the delivery script.
- [ ] All four agent names remain covered: `consult-persona-challenger`, `consult-empathy-mapper`, `consult-framework-adherence-reviewer`, `consult-dashboard-refresher` each appear in the matcher (or in the replacement seam). Count = 4; no agent silently dropped from delivery.
- [ ] The file cogni-consult/tests/test_subagent_start_hook.sh exists and follows the tests convention established by cogni-consult/tests/test_register_generator.sh: a header comment containing a `Usage: bash cogni-consult/tests/test_subagent_start_hook.sh` line, `set -u` and NOT `set -e` (test_register_generator.sh:23-24, so one failing assertion does not abort the run), pass/fail helper functions with a failure counter, and a non-zero exit when the counter is non-zero.
- [ ] No plugin version was touched on this branch: cogni-consult/.claude-plugin/plugin.json `version` and the cogni-consult entry in the root .claude-plugin/marketplace.json both still read the base value `0.1.87`, and they agree with each other (root CLAUDE.md:107-112; CI job 'Version-bump gate' flags both version-touched and version-mirror-desync).
- [ ] The `hooks` keyword decision is applied symmetrically: `hooks` is present in cogni-consult/.claude-plugin/plugin.json keywords[] (base line 20) AND in the cogni-consult keywords[] of the root .claude-plugin/marketplace.json, or removed from both. Never one without the other — the two manifests must always agree (root CLAUDE.md:109).
- [ ] The test reads the matcher regex out of cogni-consult/hooks/hooks.json at runtime (grep the test for the hooks.json path plus a python3 json load of it) rather than hardcoding a copy of the regex string. This is the single most load-bearing property of the test: a hardcoded copy passes green while the shipped hook stays broken, which is precisely the silent no-op the issue is about.
- [ ] The test asserts a POSITIVE match for the plugin-qualified form (e.g. `cogni-consult:consult-dashboard-refresher`) and, separately, for the bare form (`consult-dashboard-refresher`) — both name forms pinned, so the regression is two-sided and a future revert to a bare-only matcher fails the test.
- [ ] Both name forms are pinned for all four agent names, not just one probe agent: the test's positive case set covers 8 strings (4 bare + 4 qualified). Count them in the test source; fewer means three of the four agents stay unpinned.
- [ ] The test carries NEGATIVE cases proving the matcher did not become over-wide: at least one foreign-plugin qualified name (e.g. `other-plugin:consult-empathy-mapper`) and at least one near-miss affix (e.g. `consult-empathy-mapper-x` and/or `xconsult-empathy-mapper`) are asserted NOT to match. Over-widening would inject cogni-consult's register contract into unrelated plugins' subagents.
- [ ] The shipped matcher in hooks.json is still fully anchored with the four names kept inside their own group — read the regex and confirm it starts with `^`, ends with `$`, and that any new optional prefix group does not leave the alternation unanchored (the classic defect `^(cogni-consult:)?a|b|c|d$` matches ANY string containing `b`). The negative cases of the previous item must fail against that broken form.
- [ ] The new matcher is a strict superset of the base regex `^(consult-persona-challenger|consult-empathy-mapper|consult-framework-adherence-reviewer|consult-dashboard-refresher)$` — every string the base matched still matches. Confirmed by the 4 bare-form positives; no existing acceptance is traded away to gain the qualified form.
- [ ] Running `bash cogni-consult/tests/test_subagent_start_hook.sh` from the repo root exits 0 and prints no line beginning `FAIL`. (Plugin tests are not wired into .github/workflows/lint.yml, which runs only the breadcrumb, external-dispatch, and version-bump gates — so this test is a manually-run regression test by design; do not grade the absence of CI wiring as a defect.)
- [ ] cogni-consult/CLAUDE.md gains, adjacent to the existing 'Language reaches subagents only through the hook' bullet (base line 143), a note recording the EMPIRICAL result: what was actually observed on a real dispatch of one of the four consult-* agents (whether the injected language/register block arrived), plus the concrete re-check a future reviewer can run. Grep CLAUDE.md for that new text within the same bullet block. This covers AC1 and AC3.
- [ ] That note separates verified from unverified: if the post-fix delivery could not be confirmed live in the authoring session (hooks load at session start from the installed plugin cache, so a branch-local matcher change is not live), the note SAYS SO explicitly rather than asserting confirmed delivery. A note claiming evidence the session could not obtain is worse than the prior silence and re-creates the exact defect the issue reports.
- [ ] The note is durable mechanism prose, not a ticket reference: it states which agent-name form the matcher must accept and why, so the finding stops recurring in plugin-validator runs (AC3). It must not consist only of 'a validator complained and we checked' — and it must not contradict the existing doctrine at cogni-consult/CLAUDE.md:139-141 that version-pinned reasons are not the durable thing to record.
- [ ] Doctrine placement is unchanged: cogni-consult/references/subagent-output-contract.md remains the home of the register rules, cogni-consult/hooks/on-subagent-start.sh still reads it from `$PLUGIN_ROOT/references/subagent-output-contract.md` (base line 46), and no register rule text is inlined into the hook script or into any of the four cogni-consult/agents/consult-*.md bodies. Verify by grepping the four agent bodies for `## Register` and `Executive register` — both must be absent (issue's explicit constraint: keep the doctrine in the reference file).
- [ ] The Architecture tree in cogni-consult/CLAUDE.md (base lines 60-64) still describes hooks/hooks.json truthfully after the change — if the matcher's meaning widened from bare names to bare-or-qualified, the line 'matched on the four consult-* agent types' is either still accurate or updated. Note the base tree already omits `tests/` entirely (grep CLAUDE.md for `tests/` returns nothing on base), so a missing tree entry for the new test file is PRE-EXISTING convention, not drift introduced by this PR — do not grade it as a violation.
- [ ] No new maintainer breadcrumbs: the diff introduces no `#NNNN` issue/PR reference and no `vX.Y.Z` version tag in any `*/skills/*/SKILL.md` or `*/agents/*.md` (the guard's globs, scripts/check-breadcrumbs.py:40; CI 'Maintainer-breadcrumb guard'). cogni-consult/CLAUDE.md is OUTSIDE that guard's scope, so a provenance reference there is a style call and not a gate violation.
- [ ] Test hygiene per the plugin's stdlib-only rule (cogni-consult/CLAUDE.md:175, 'bash + python3, no pip dependencies'): the test uses python3 — not `jq` — to read hooks.json, and if it needs scratch space it uses `mktemp -d` with a cleanup `trap` and writes nothing into the repo working tree (test_register_generator.sh:35-36).
- [ ] Scope containment: the change is confined to cogni-consult/hooks/hooks.json, cogni-consult/CLAUDE.md, and the added cogni-consult/tests/test_subagent_start_hook.sh. If instead the fix migrated off the hook to a skill-level seam, then the ~7 dispatch call sites in cogni-consult/skills/*/SKILL.md and references/orchestration/*.md that dispatch the four agents must ALL carry the new injection, and the delivery-path sentence at CLAUDE.md:143 ('registers a SubagentStart hook') must be rewritten in the same commit — a half-migration that updates some call sites is a worse silent no-op than the bug being fixed.

**Risk watchlist:**
- The headline failure mode: a matcher widened by pasting an optional prefix in front of an ungrouped alternation (`^(cogni-consult:)?consult-a|consult-b$`), which silently becomes a substring match on the tail alternatives and fires the hook for unrelated plugins' subagents.
- A regression test that hardcodes its own copy of the regex instead of reading hooks.json — permanently green, zero protection, and it makes the next reviewer believe the seam is pinned.
- Over-correction to a "*" or ".*" matcher: makes the hook demonstrably fire (satisfying the letter of AC1/AC4) while injecting cogni-consult's German/register contract into every other plugin's dispatched agents.
- A doc-only resolution that records 'SubagentStart is supported' as the finding and stops there. The event being supported is not the same claim as the block reaching the four agents; closing on the former leaves the actual defect shipped and the plugin still advertising the `hooks` keyword.
- Asserting post-fix delivery as empirically confirmed when it structurally cannot be, because hooks load at session start from the installed plugin cache rather than from the feature branch. This is the specific dishonesty the issue is a reaction to.
- Dropping consult-dashboard-refresher (the cheap live probe) or any other agent from the matcher while restructuring the regex — a silent narrowing disguised as a widening.
- Recording the verification with a Claude Code build number as the load-bearing evidence, which goes stale and re-opens the finding — and which cuts against the existing doctrine at cogni-consult/CLAUDE.md:139-141.
- Touching the version in plugin.json or marketplace.json 'to ship the fix', which trips the Version-bump gate, or bumping one manifest and not the mirror.
- Inlining the register rules into on-subagent-start.sh or the agent bodies to route around the delivery problem — fixes the symptom, violates the issue's explicit constraint, and creates the drift that references/subagent-output-contract.md exists to prevent.
- JUDGMENT-ONLY (no gate can decide it): whether the recorded note is actually persuasive enough to stop plugin-dev:plugin-validator from re-raising the finding on the next run. The validator is external and its VALID_EVENTS whitelist cannot be fixed from this repo, so the note is a mitigation, not a guarantee — a human should read the note and decide whether it discharges AC3.
- JUDGMENT-ONLY: whether widening the matcher (rather than migrating to a seam that is verifiable in-session) is the right durability trade. The matcher fix is smaller but leaves the plugin's language delivery dependent on an agent-name-resolution detail of the host runtime that this repo cannot test.

- Acceptance criterion 4 ('cogni-consult keeps the hooks keyword in both plugin.json and the marketplace.json mirror only if the hook demonstrably fires') conditions a manifest decision on evidence no PR can produce: hooks load at session start from the installed plugin cache, so a delivery fix living on a feature branch cannot be observed firing until after the merge and a cache refresh. Criterion 1 is likewise only satisfiable for the PRE-fix state. A human must decide whether the keyword may be retained on a fix that is verified by a regex regression test plus a deferred post-merge observation, or whether the keyword must be dropped until a live post-merge dispatch confirms delivery.
- The issue's plan-time acceptance criteria list 4 outcome-level criteria and ask for no regression test at all; a base-grounded regeneration derives ~22 gating items, including matcher anchoring and over-width safety, the strict-superset property, event-key/hookEventName consistency, the read-the-regex-from-hooks.json test property, the stdlib-only test convention, and version-gate compliance. The acceptance bar was materially under-specified relative to what a correct fix here has to get right. (recorded — does not hold the PR)

sdh07:
## Verdict: **pass** — merging

The merger held this PR at contract formation (Step 1.5) and **never graded the diff** — there is a review-contract comment but no verdict. So this is not a bounce; the diff had simply never been reviewed. Grading it now against that same committed contract, **by execution rather than by reading**: **20 of 22 items satisfied on arrival**, two minor findings, both fixed in `462cb681`. `hooks.json` is unchanged — the one-line widening is correct as shipped.

### The held escalation is discharged on evidence, not judgment

The hold was on issue acceptance criterion 4 — keep the `hooks` keyword *"only if the hook demonstrably fires"* — which the contract correctly noted no branch-local PR can satisfy. The PR was honest that its root cause was an **inference**.

It is no longer an inference. Read out of the Claude Code 2.1.221 binary:

- **Plugin agent registration** — the loader computes `agentType = [pluginName, ...subdirs, agentName].join(":")`. A plugin agent is registered as `cogni-consult:consult-dashboard-refresher`, never bare. (The subdirectory clause is observable in the live registry: `cogni-service:references:ship-captain-playbook`.)
- **Dispatch** — `executeSubagentStartHooks(agentId, agentType, …)` passes that string verbatim as both the `agent_type` payload field and the match query.
- **Selection** — `getMatchingHooks` does `case "SubagentStart": query = input.agent_type`.
- **Matching** — `new RegExp(matcher).test(agent_type)`. The two fallback expansions resolve tool-rename aliases only and never apply to agent names.

So the shipped bare-anchored `^(consult-…)$` provably could never match, and `^(cogni-consult:)?(consult-…)$` provably does. That is a stronger and more durable form of "demonstrably fires" than a single post-merge dispatch, and it does not go stale with the next build. **Keyword retained in both manifests; merging on this basis.** The post-merge live dispatch remains worth doing as confirmation, but it is not a gate.

### Findings fixed

**1 — the test's over-width guard was inert (minor).** `assert_match` used Python `re.fullmatch`; the host searches. For the shipped anchored matcher the two agree, so nothing was wrong in practice — but against the exact regression the contract names as the headline risk, an unanchored alternation, they diverge:

| candidate | test (`fullmatch`) | runtime (`search`) |
|---|---|---|
| `other-plugin:consult-empathy-mapper` | no match | **MATCHES** |
| `xconsult-empathy-mapper` | no match | **MATCHES** |

The suite did still go red on that form, but via the qualified *positives* — not via the three negative assertions that advertise over-width protection. Fixed: `re.search`, plus a structural assertion that every top-level alternative carries its own anchors (`^…$` at the string ends is not sufficient — `^(cogni-consult:)?a|b|c|d$` has both and still substring-matches), plus the missing prefix-affix rejection. Measured before/after on the unanchored form: **0 near-miss assertions fired, now 3 do.**

**2 — the `CLAUDE.md` note overclaimed and pinned its evidence to a build number (minor).** It read *"the mechanism here was verified live on build 2.1.220."* What was verified live is the **pre-fix failure**, not post-fix delivery. Contract item 16 requires the note to say explicitly that post-fix delivery was not confirmable in-session; it did not — the PR body says so at length, but the PR body is not the durable artifact. Build-number-as-evidence also contradicts this plugin's own doctrine three bullets above (`CLAUDE.md:139-141`: *"independent of Claude Code's version and is the only one to record"*), which the contract's risk watchlist flagged. Rewritten to the structural rule, what was observed, and what was not — with the re-check named.

### Contract items verified by execution

- Test run: **17 assertions, exit 0**, no `FAIL` line (15 on arrival + 2 added).
- Negative control A — restore the bare-only regex: exactly the **4 qualified-form assertions** fail, exit 1.
- Negative control B — the unanchored form: the anchoring assertion, the prefix-affix rejection and the drift guard fail, exit 1.
- All **7 sibling** `cogni-consult` suites pass.
- Test reads the regex out of `hooks.json` at runtime (contract's most load-bearing property) — confirmed, and it derives the agent list from `agents/` so a fifth agent cannot drift past the loops.
- Matcher is a **strict superset** of the base regex; all four agents retained; `consult-dashboard-refresher` not dropped.
- `python3 -m json.tool cogni-consult/hooks/hooks.json` → exit 0. Event key still equals the `hookEventName` emitted at `on-subagent-start.sh:116`. Script path and `timeout: 15` unchanged.
- Doctrine placement unchanged: `references/subagent-output-contract.md` still the home, still read at `on-subagent-start.sh:46`; grep of the four agent bodies for `## Register` / `Executive register` → **none**.
- Test hygiene: `set -u` and not `set -e` (matching `test_register_generator.sh:24`), `mktemp -d` + `trap`, pass/fail counters, python3 not `jq`, nothing written into the working tree.
- Version gate: `check-version-bump.py` → *14 plugins scanned, 1 touched, 0 violations*. Both manifests still read `0.1.87` and agree. `check-breadcrumbs.py`, `check-external-dispatch.py` → exit 0.
- Scope contained to the three intended files.

### The contract's two judgment-only items

- *Is the note persuasive enough to stop `plugin-dev:plugin-validator` re-raising this?* Better than it was: it now states the mechanism rather than a version-pinned anecdote. The validator is external and its `VALID_EVENTS` whitelist still omits the event, so this stays a mitigation, not a guarantee.
- *Is widening the matcher the right durability trade vs. migrating to an in-session-verifiable seam?* Yes. The concern was that the matcher leaves delivery dependent on an untestable host detail — but that detail is now pinned in the test and written down in `CLAUDE.md`, and a seam migration would have touched ~7 dispatch call sites across five SKILL.md files to fix a one-line defect.

### Recorded, not blocking

The contract's second escalation stands: the issue's 4 outcome-level criteria under-specified what a correct fix here had to get right, against ~22 derivable gating items. It also noted the issue's "Proposed change" posed a false binary that omitted the actual diagnosis — the diff follows neither literal branch and is still correct. Neither is an author defect. #1217 already carries the two separable follow-ups (CI discovery of plugin `tests/*.sh`; the plugin-qualified-matcher rule as a root-`CLAUDE.md` cross-plugin convention).

Issue #1221, opened 2026-08-04
cogni-trends: same bash-3.2 bare-argument eval bug silently zeroes the scout dimension counters


claude-cogni-work:
**Plugin:** cogni-trends (`cogni-trends/scripts/project-status.sh`)
**Marketplace:** insight-wave

## Current behavior

`cogni-trends/scripts/project-status.sh` opens a `python3` block at line 84 as a **bare argument** to `eval`:

```bash
eval "$(python3 -c "
...
            counts[dim] = {'act': 0, 'plan': 0, 'observe': 0, 'total': 0}
...
" 2>/dev/null)"
```

Under GNU bash 3.2.57 — macOS's system bash, which `cogni-knowledge/tests/README.md` explicitly targets — a command substitution carrying an inline multi-key dict literal in **argument position** is brace-expanded and word-split, re-invoking `python3` once per comma-delimited fragment with the outer braces stripped. Every fragment is invalid Python, so `python3 -c` fails to *compile* and emits nothing at all — not even output from `print()` calls earlier in the same source. `eval` then receives an empty string and the scout dimension counters silently keep their defaults.

Both the python's `except Exception: pass` and the shell's `2>/dev/null` mask the failure, so nothing reports it.

## Desired behavior

Apply the same assignment-then-eval split already used to fix the value-modeler block:

```bash
__SCOUT_COUNTS_OUT="$(python3 -c "
...
" 2>/dev/null)"
eval "$__SCOUT_COUNTS_OUT"
```

The fork is assignment-vs-argument: `VAR="$(…)"` followed by `eval "$VAR"` is unaffected; only the single-statement bare-argument form corrupts.

## Motivation

This is the same defect class as the value-modeler block, found while fixing it. It was left out of that change because no currently-red test suite covers it and it produces a distinct symptom (scout dimension counts rather than value-modeler counters), so bundling it would have widened a CI-plumbing PR into unrelated territory.

It is filed rather than dropped because the first-pass plan had asserted this block was safe, and re-reading `main` disproved that. Leaving it behind a false all-clear is how a latent bug survives the next audit.

**Platform asymmetry:** this is bash-3.2-specific. `ubuntu-latest` ships bash 5.x and most likely does not exhibit it, so CI will probably stay green while every macOS developer sees wrong counts. Do not treat a green CI run as evidence the block is fine.

## Acceptance criteria

[...]

PR #1222, opened 2026-08-04
CI: run plugin + root tests/*.sh; record plugin-qualified matcher rule

### LINKED ISSUE #1217: CI: run plugin tests/*.sh, and record the plugin-qualified hook-matcher rule as a cross-plugin convention

### LINKED ISSUE #1220: cogni-consult: point the SubagentStart test caveat at the new Plugin test suites CI job
**Plugin:** cogni-consult (`cogni-consult/CLAUDE.md`)
**Marketplace:** insight-wave

## Current behavior

`cogni-consult/CLAUDE.md` carries a caveat stating that the SubagentStart hook regression test is not CI-enforced:

> That test is not CI-enforced (no job discovers `tests/*.sh`), so run it by hand when touching this hook.

That sentence does not exist on `main` — it is introduced by the branch behind the hook fix. Once the `Plugin test suites` CI job lands, the caveat becomes false.

## Desired behavior

Update the caveat to point at the `Plugin test suites (discover and run tests/*.sh)` job in `.github/workflows/lint.yml`, so the guide states that the suite now runs in CI on every PR rather than telling the reader to run it by hand.

## Motivation

This is acceptance criterion 3 of the parent issue, deferred because it was not actionable at the time. Two blockers, both of which resolve on merge:

1. **The anchor text did not exist yet.** It arrives with the hook-fix branch; there was nothing to edit.
2. **The file was contested.** Two open draft PRs modified `cogni-consult/CLAUDE.md` concurrently, so an edit from the CI branch would have collided with both.

## Acceptance criteria

- [ ] The "not CI-enforced … run it by hand" caveat in `cogni-consult/CLAUDE.md` names the `Plugin test suites` job instead
- [ ] The surrounding paragraph still reads correctly — the point that the hook's failure mode is silent by construction must survive the edit
- [ ] No other plugin `CLAUDE.md` carries a now-stale "not CI-enforced" claim about `tests/*.sh` (grep before closing)

## Depends on

- The hook-fix PR that introduces the caveat must merge first — there is no text to edit until it does.

claude-cogni-work:
Closes #1217.
Closes #1220.

## Summary

Nothing in CI discovered or ran any of the repo's bash test suites. `lint.yml` ran three `python3 scripts/*.py` guards and stopped there, while 98 suites across six plugins and the repo root ran only when a human remembered. Two of them had quietly drifted out of agreement with their code.

- **`scripts/run-plugin-tests.py`** — stdlib-only discovery runner. Two non-recursive globs (`tests/*.sh`, `*/tests/*.sh`), invokes each suite as `bash <path>` with zero args, gates on **exit status only**, per-suite timeout. Repo-standard `{success,data,error}` JSON on stdout, progress on stderr.
- **`tests/test_run_plugin_tests.sh`** — the runner's own self-test, matching the four existing root suites that test `scripts/check-*.py`. 11 assertions across 7 cases.
- **`.github/workflows/lint.yml`** — a fourth job, `Plugin test suites`.
- **`cogni-trends/tests/test-stale-detection.sh`** — stale test, code was correct.
- **`cogni-trends/scripts/project-status.sh`** — live bug, test was correct.
- **`CLAUDE.md`** — the plugin-qualified `plugin:agent` hook-matcher rule as a Cross-Plugin Convention, plus a Contributing bullet.
- **`CONTRIBUTING.md`** — a `### Plugin Test Suites` section with the suite contract and copy-pasteable local-run commands, alongside the two existing guard sections.

Result: **100/100 suites green** on bash 3.2.57, full sweep ~3m20s.

## The two red suites split, and that changed the plan

Triage recorded both as stale expectations. Only one was.

**`test-stale-detection.sh` — stale test.** `trend-report` was split into `trend-research` / `trend-synthesis` / `trend-booklet`; that commit updated `project-status.sh` but never the test, and no `trend-report` skill exists today. Fixed by updating the expectations — and by asserting **both** `/trend-research` and `/trend-synthesis` appear in the reason text, so a half-updated injector can't pass.

**`test-project-status.sh` — live production bug.** `project-status.sh` invoked its value-modeler block as a bare argument:

```bash
eval "$(python3 -c "…")"      # line 319
```

Under bash 3.2.57 — macOS's system bash, which `cogni-knowledge/tests/README.md` explicitly targets — the multi-key dict literals in that block are brace-expanded and word-split in argument position, re-invoking `python3` once per comma fragment with the braces stripped. Every fragment fails to **compile**, so python emits nothing at all — not even `print()` output from earlier in the same source. `eval` receives an empty string and six counters silently keep their `0` defaults. The `except Exception: pass` and the `2>/dev/null` each mask half of it. Fixed by assigning first, then eval'ing: command substitution on the right of an assignment isn't subject to that expansion.

**Why this one is fixed here rather than deferred.** It is a bash-3.2-specific quirk; `ubuntu-latest` ships bash 5.x and most likely does not exhibit it. So the CI job this issue asks for would probably have shown this suite **green** while it fails on every macOS dev machine. Landing the runner without the fix would have converted an honest local failure into false CI confidence — the exact anti-pattern the issue was filed against.

## Design decisions

**Exit status is the only signal.** The issue describes a shared `set -u` + pass/fail-counter convention. Measured: only 11 of 93 suites declare a `set -` line early, and counter naming isn't uniform. Exit status is the one property all of them genuinely share, so the runner never parses stdout. Seven suites also lack the executable bit while carrying a valid shebang — hence `bash "$suite"`, never `./$suite`.

**Discovery is two non-recursive globs.** Matching exactly one directory level is what keeps the dead `cogni-knowledge/_archive/` suite and the sourced-only `tests/fixtures/*.sh` helper libraries out without enumerating either — both sit one level deeper than the globs reach. The second glob is `*/tests/*.sh`, not `cogni-*/tests/*.sh`: both match the identical 99 today, but the narrower form would silently skip a future non-`cogni-` component that ships suites, reproducing this PR's own failure class one level up.

**Empty discovery is a failure, not a pass.** If the globs ever stop matching, the runner exits non-zero rather than reporting a vacuous green.

**The shallow checkout is load-bearing — opposite to `version-bump-gate`.** `tests/test_check_version_bump.sh` case 10 runs the real version gate against the checked-out tree. With `fetch-depth: 0` that case grades for real, and a genuine `version-touched` finding would fail this sweep *in addition to* the dedicated gate — one defect reported twice, mislabelled here as a test-suite failure. Shallow, the gate reports `status: degraded` / `clean: true` / exit 0. The checkout step carries this reasoning inline, since the neighbouring job needs the exact opposite.

## Scope

**Landed here — AC3 (closes #1220).** The deferral premise was true only at the stale branch point. `cogni-consult/CLAUDE.md`'s "not CI-enforced … run it by hand" caveat is absent at this branch's merge-base (`4f8ad433c`) but **present on current `main`** — #1218 merged meanwhile. Leaving it would ship a repo whose own docs assert something false about its CI, on the exact subject this PR changes. The branch is updated from `main` and the sentence corrected in place, so this now delivers #1217 in full.

`cogni-knowledge/tests/README.md` carried the same claim ("CI integration is deferred (cogni-knowledge has no CI today)") and is corrected with it — same class, made false by the same change.

**Withdrawn — #1221.** I filed it during planning, claiming `project-status.sh` line 126 carried the same bug in the block opening at line 84. It does not: that block closes at line 109, and line 126 is inside `DIMENSION_COUNTS=$(python3 -c "` at line 115 — assignment position, the safe fork. Closed as not reproducible with the evidence recorded.

## Review notes

Three things surfaced in review that I deliberately did **not** fold in:

1. **The sweep is sequential (~204s).** Suites are independent and self-isolate via `mktemp -d`, so a `ThreadPoolExecutor` would cut it to roughly 60–70s. Left sequential: this is a brand-new job whose entire value is being trusted, and adding concurrency in the same change trades that for latency the 20-minute ceiling doesn't need. Worth revisiting once it has some track record.
2. **`test_check_version_bump.sh` case 10 is the one non-hermetic assertion in the corpus** — it grades the live tree, which is *why* the checkout comment above has to exist. Making it hermetic would remove the caveat entirely, but it means editing the version gate's own suite, well outside this diff.
3. **The hook-matcher rule ships as prose only.** A `scripts/check-hook-matchers.py` guard would fit the house pattern — the repo already has three guards with a lint job and a root suite each, for exactly this silent-violation class — and would flag the one live instance. Not done here because AC2 asks for a convention entry, and a guard is a materially larger deliverable that deserves its own issue. Flagging rather than silently dropping. Note the sole agent-name matcher in the tree (`cogni-consult/hooks/hooks.json`) is still bare-name on `main`; #1218 widens it.

4. **The diff is over the 400-line review bound and this revision grows it further** (seven files now, two new to the branch). Not split, deliberately: the merger has already graded this diff item-by-item against a committed review contract, and re-cutting it into two PRs would discard that grading and restart review from cold. The change also isn't cleanly separable — the runner, its self-test, the CI job, the two suites the job exposed as red, and the docs describing the job are one contract. Recorded rather than hidden.

No repo-wide all-clear is claimed for the bare-argument `eval` shape. It isn't cleanly regex-detectable — a naive brace-with-comma scan flags 36 sites, nearly all f-string groups that demonstrably work. What is verified: the two literals that actually broke are both in the one block this PR fixes, and all 99 suites pass.

## Test plan

- [x] `python3 scripts/run-plugin-tests.py --list` → 100 suites, zero `_archive` / `fixtures` paths
- [x] `bash tests/test_run_plugin_tests.sh` → 13/13 assertions pass
- [x] case 8: an empty root exits 1 with `success:false`, `total:0`, non-empty `error` — the zero-discovery guard is now pinned by a test, not only by production code
- [x] `python3 scripts/run-plugin-tests.py --filter knowledge` → 76 suites, exit 0 (the command the corrected cogni-knowledge README now documents)
- [x] `python3 scripts/run-plugin-tests.py` → 100/100 pass, exit 0, on macOS bash 3.2.57
- [x] stdout is valid JSON and carries no per-suite `output` blobs
- [x] `bash cogni-trends/tests/test-stale-detection.sh` passes
- [x] `bash cogni-trends/tests/test-project-status.sh` passes, stages 4–8 report `done`
- [x] `check-breadcrumbs.py`, `check-external-dispatch.py`, `check-version-bump.py` all clean (`status=ok clean=True`)
- [x] `lint.yml` parses; four jobs; new job's checkout carries no `with:` block
- [ ] The `Plugin test suites` job is green on this PR


claude-cogni-work:
## Ship-ready

Draft PR #1222 is authored and handed off for independent review. The author instance does not self-review; the merger instance forms its own acceptance checklist from this issue before reading the diff.

**Scope:** AC1 and AC2 land. AC3 is deferred to #1220 (its anchor text does not exist on `main` — PR #1218 introduces it), so the PR links `Refs #1217`, not `Closes`. This issue stays open until #1220 resolves.

**State:** 99/99 suites green on bash 3.2.57; all three existing CI guards clean.

**Worth the reviewer's attention:**

1. The PR fixes a live `cogni-trends` production bug alongside the CI plumbing. That widening is deliberate and argued in the PR body — the bug is bash-3.2-specific, so the new `ubuntu-latest` job would likely have reported the suite green while it fails on every macOS machine. Landing the runner without it would ship false confidence. Reasonable people could scope that differently.
2. The new job's shallow checkout is load-bearing in the *opposite* direction from the neighbouring `version-bump-gate`. The reasoning is inline in the workflow, and it rests on `tests/test_check_version_bump.sh` case 10 being non-hermetic.
3. Three review findings were deliberately not folded in (sequential sweep, case-10 hermeticity, a machine guard for the hook-matcher rule). Each is listed under "Review notes" in the PR body with its reason.
4. Follow-up #1221 was filed during planning and then **closed as not reproducible** — the premise was wrong and the review pass caught it. Evidence is on that issue.

claude-cogni-work:
## The job caught a third drifted suite on its very first run

Worth recording, because it is the argument for the PR making itself.

The first CI run was **green locally, red on `ubuntu-latest`** — `tests/test_check_version_bump.sh` case 4 failed. Cause: `check-version-bump.py` resolves the branch for its `^bump/` exemption from `VERSION_BUMP_BRANCH`, then `GITHUB_HEAD_REF`, and only then the checked-out branch. Under GitHub Actions `GITHUB_HEAD_REF` is set to the PR's own head branch, so case 4's fixture — deliberately sitting on `bump/auto-abc123` — was judged against `fix/issue-1217-ci-run-plugin-tests` instead, lost the exemption it exists to assert, and failed.

The suite was never hermetic against ambient CI environment variables. Nothing had ever noticed, because nothing had ever run it anywhere but a developer's shell.

Fixed by clearing both variables at the top of the suite, so every fixture case exercises the checked-out-branch fallback identically in both environments. Reproducible locally in one command, before and after:

```bash
GITHUB_HEAD_REF=anything bash tests/test_check_version_bump.sh
```

**Tally: three suites had drifted, in three different ways.** A stale expectation after a skill split, a live production bug behind a silent `eval`, and a hidden environment dependency. One per plugin-shaped failure mode, all invisible for the same single reason.

Final state: **99/99 green**, 1m18s on `ubuntu-latest` (~3m20s locally on bash 3.2), all five checks passing.

sdh07:

_The merger formed this expectation from issue #1217 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds one new job to `.github/workflows/lint.yml` (which today has exactly three jobs, lines 12-91, each `python3 scripts/*.py` and each carrying an inline comment justifying its checkout depth) backed by a new stdlib-only `scripts/run-plugin-tests.py` that follows the repo's script contract — `--root` flag, `{success, data, error}` envelope on stdout, exit 0/1/2 — exactly as `scripts/check-version-bump.py:60-63,164-172` and `scripts/check-external-dispatch.py:52-53` do, plus its own root self-test `tests/test_run_plugin_tests.sh` shaped like the four existing root suites (`tests/test_check_version_bump.sh:25-45,190-196`: `set -eu`, red/green helpers, `mktemp -d` + `trap`, FAILED counter, non-zero exit). The single hardest requirement is that the job cannot be silently vacuous: the issue exists because a matcher that never matched raised nothing anywhere, so a sweep that discovers zero suites, or whose non-zero exit is swallowed by `continue-on-error`/`|| true`, reproduces the exact failure mode it was written to end — `lint.yml:79-89` already encodes this discipline for the version-bump gate, and the new job must match it. Discovery must be non-recursive from the repo root (`tests/*.sh`, `cogni-*/tests/*.sh`) so it excludes `cogni-knowledge/_archive/tests/test_read_project_config_bare.sh`, the sourced-only helpers `cogni-knowledge/tests/fixtures/_cycle_guard_lib.sh` and `test_helpers.sh`, and the ~2000 `.sh` copies under `.git/cogni-worktrees/**`, and must invoke `bash "<path>"` because several tracked suites carry a shebang without the exec bit. For item 2, one entry under root `CLAUDE.md` `## Cross-Plugin Conventions` (line 31, before `## Plugin Data Flow` at line 82) recording the rule that today lives only mid-paragraph at `cogni-consult/CLAUDE.md:146` and is implemented at `cogni-consult/hooks/hooks.json:5` as `^(cogni-consult:)?(...)$`, plus a `## Contributing` bullet naming the job the way line 166 names the breadcrumb guard. The collateral edits to `cogni-trends/tests/test-stale-detection.sh` (whose case-2 literals `cogni-trends:trend-report` / `/trend-report` are stale against production's `cogni-trends:trend-research` at `cogni-trends/scripts/project-status.sh:1075-1081`), `cogni-trends/scripts/project-status.sh` (12 identical `eval "$(python3 -c "` sites at lines 37,53,84,197,221,256,269,288,304,319,374,395), and `tests/test_check_version_bump.sh` must make the sweep green by fixing code, never by weakening an assertion.

**Acceptance checklist:**
- [ ] `scripts/run-plugin-tests.py` exists, imports only Python stdlib modules (read the import block; no third-party name, no requirements file added), and prints a single `{"success": bool, "data": {...}, "error": "..."}` object on stdout — the repo contract at root `CLAUDE.md` line 35 and the shape at `scripts/check-version-bump.py:164-172`.
- [ ] `.github/workflows/lint.yml` gains exactly one new key under `jobs:`, at the same indentation as `version-bump-gate` (line 40), carrying a `name:` and `runs-on: ubuntu-latest`, whose run step invokes `python3 scripts/run-plugin-tests.py`; the three existing jobs `breadcrumb-guard`, `external-dispatch-guard`, `version-bump-gate` are unchanged in the diff.
- [ ] The new job's `actions/checkout@v4` step carries an inline comment stating why its fetch depth is what it is — the convention every existing job in `lint.yml` follows (lines 18-19, 32-33, 46-58). A checkout step with no depth rationale fails this item.
- [ ] Root `CLAUDE.md` gains the plugin-qualified matcher rule positioned between `## Cross-Plugin Conventions` (line 31) and `## Plugin Data Flow` (line 82) — not in a section elsewhere in the file.
- [ ] That new rule text names all three load-bearing facts, checkable as literal content: (a) a plugin-supplied agent is dispatched as `plugin:agent` because the host registers it as `pluginName:agentName`; (b) an agent-name matcher must therefore accept both the bare and the qualified form; (c) the regex stays anchored at both ends so it does not fire for another plugin's subagents. Compare the source at `cogni-consult/CLAUDE.md:146` and the live matcher `cogni-consult/hooks/hooks.json:5`.
- [ ] Root `CLAUDE.md` `## Contributing` gains a bullet naming the new sweep the way line 166 names the breadcrumb guard: the script path plus the job's `name:` string exactly as it appears in `.github/workflows/lint.yml`.
- [ ] `tests/test_run_plugin_tests.sh` exists and matches the four sibling root suites' shape: `#!/usr/bin/env bash`, an early `set -eu`, `red()`/`green()` helpers, a `FAILED` counter incremented by a `check` helper, `WORK="$(mktemp -d)"` with `trap 'rm -rf "$WORK"' EXIT`, and a non-zero exit when `FAILED` > 0 (pattern: `tests/test_check_version_bump.sh:25-45,190-196`).
- [ ] Non-zero propagation is intact end to end: the runner exits non-zero when any discovered suite exits non-zero, AND the new workflow job contains none of `continue-on-error`, `|| true`, `|| echo`, or a trailing unconditional `exit 0` on the sweep step. Any of those makes the job advisory and fails this item — an advisory sweep is the documentation-not-a-guard state the issue was filed to end.
- [ ] VACUITY GUARD: discovery finding zero suites is a hard failure (non-zero exit and `success: false`), not a silent green. Grep the runner for an explicit empty-set branch. A glob that matches nothing and reports success is the same silent-no-op class as the matcher that never fired; `lint.yml:79-89` already enforces exactly this shape for the version-bump gate, so the convention exists and the item is not a novel demand.
- [ ] `tests/test_run_plugin_tests.sh` contains a NEGATIVE CONTROL: a fixture suite that exits non-zero, asserting the runner surfaces it and itself exits non-zero — plus a case asserting the empty-discovery path from the previous item fails. Without both, a runner that executes nothing would pass its own self-test.
- [ ] The runner's `data` names every failed suite's path and its exit code (grep the JSON assembly for a per-suite record carrying both fields), so a red CI job is diagnosable from the log without re-running the sweep locally.
- [ ] Suites are invoked as `bash "<path>"` — never `./<path>` — because tracked suites exist that carry a valid shebang without the executable bit, which `./` would fail with ENOEXEC. Confirm the invocation string in the runner and a self-test fixture whose file mode is non-executable.
- [ ] Discovery is non-recursive relative to `--root`, so it excludes `cogni-knowledge/_archive/tests/test_read_project_config_bare.sh`, the sourced-only helper libraries `cogni-knowledge/tests/fixtures/_cycle_guard_lib.sh` and `cogni-knowledge/tests/fixtures/test_helpers.sh`, and the ~2000 `.sh` copies under `.git/cogni-worktrees/**` that a `**/tests/*.sh` recursion picks up in a dev clone. At least one of these exclusions is pinned by a self-test case.
- [ ] No self-recursion: `tests/test_run_plugin_tests.sh` is itself discovered by the sweep, so it must never invoke the runner in execute mode against the real repo root. Any real-tree invocation in the self-test is discovery/`--list` only — grep the self-test for every real-root invocation and confirm each is list-mode.
- [ ] Runtime is bounded and the bound is stated — the issue explicitly asks how long the full sweep may take. The new job declares `timeout-minutes:` AND the runner enforces a per-suite timeout with an explicit default value, with both numbers named in a comment or in the docs added above. An unbounded sweep fails this item.
- [ ] `tests/test_check_version_bump.sh` is in the changed files although the author's own plan record did not list it as a file to modify — so verify it was not WEAKENED: every `check "case N: …"` label present at base (cases 1-10, lines 118-188) still exists, and no assertion is loosened (an exact `[ "$(run_gate …)" = "0|ok|EMPTY" ]` equality must not become a `grep -q`, a substring test, or an unconditional pass). Any relaxation of case 10 (`real tree -> clean, exit 0`, lines 181-188) to make the new sweep go green is a hard fail: that is bending the guard to fit CI rather than CI to fit the guard.
- [ ] `cogni-trends/tests/test-stale-detection.sh` case 2 keeps full assertion strength through the rename: it still asserts `stale_count >= 1`, `subtype == "scout_drift"`, `"t-003" in added`, `"t-001" in changed`, that the first injected action's `skill` equals the literal production actually emits (`cogni-trends:trend-research`, per `cogni-trends/scripts/project-status.sh:1075`), and that its `reason` still contains the `"1 added"` and `"1 changed"` diff fragments (base lines 233-246). Deleting or softening any of these instead of renaming the literal fails.
- [ ] `cogni-trends/scripts/project-status.sh` carries 12 structurally identical `eval "$(python3 -c "` sites at base (lines 37, 53, 84, 197, 221, 256, 269, 288, 304, 319, 374, 395). If the diff rewrites only a subset, the PR body or an in-file comment names which sites were fixed and why the remainder are safe or deferred — a silent partial fix of a bug CLASS leaves the identical latent defect at the untouched call sites behind a newly green sweep.
- [ ] The `project-status.sh` production edit is test-covered: `cogni-trends/tests/test-project-status.sh` or `test-stale-detection.sh` still asserts the counters the rewritten block produces (base coverage: `cogni-trends/tests/test-project-status.sh:52-60`). A production-script change landing under a nominally CI-only issue with no assertion behind it fails this item.
- [ ] AC3, measured against base: `cogni-consult/CLAUDE.md:146` currently states "That test is not CI-enforced (no job discovers `tests/*.sh`), so run it by hand when touching this hook." `cogni-consult/CLAUDE.md` is NOT in the changed files, so once the sweep lands the merged repo asserts something false about its own CI. Either that sentence is corrected in the merge result, or the PR records a reason grounded in the ACTUAL base text — the plan record's stated premise, that the caveat does not exist on `origin/main`, is contradicted by the base clone at that line.
- [ ] Verify by acting on the merged tree, all three deterministic gates: `python3 scripts/check-breadcrumbs.py` exits 0 (no `#NNN`, `vX.Y.Z`, or milestone/slice code introduced in the new files); `python3 scripts/check-version-bump.py` exits 0 with `data.clean` true (no plugin version or `marketplace.json` entry touched — none appear in the changed files, so this must hold); and `.github/workflows/lint.yml` still parses as valid YAML with the new job as a top-level sibling under `jobs:`.
- [ ] `CONTRIBUTING.md` convention: every existing CI guard has its own `###` section with a copy-pasteable local-run command (`### Breadcrumb Guard` line 59 with commands at 69-72; `### Version-Bump Gate` line 80 with commands at 93-96). `CONTRIBUTING.md` is not in the changed files — either it gains an equivalent section for the sweep (including how to run it locally), or the PR states why the new job is exempt from a convention the two other guards both follow.

**Risk watchlist:**
- The sweep lands ADVISORY — `continue-on-error: true`, a trailing `|| true`, or a runner that always exits 0 — producing a green job that reports nothing. This is the exact failure mode the issue describes (a hook that exits 0 on every path), reintroduced one layer up.
- Discovery silently matches zero suites (wrong `--root` default, wrong glob relative to the workflow's working directory) and the runner reports success. Nothing anywhere would report this; the CI badge would be green forever while no plugin test ever ran.
- The sweep goes green by WEAKENING tests rather than fixing code — most plausibly by relaxing `tests/test_check_version_bump.sh` case 10 or by deleting the stale case-2 assertions in `cogni-trends/tests/test-stale-detection.sh` instead of renaming the literals. `tests/test_check_version_bump.sh` appears in the changed files despite being absent from the author's own declared file list, which is precisely where this would hide.
- Partial fix of a bug class: `cogni-trends/scripts/project-status.sh` has 12 identical `eval "$(python3 -c "` sites; fixing only the one a test happens to exercise leaves 10+ latent instances while the sweep certifies the file as green.
- Platform asymmetry hides real failures: the suites have only ever been run on macOS/bash 3.2, and the job runs ubuntu-latest/bash 5.x. A suite that fails locally but passes in CI (or vice versa) converts an honest local failure into false CI confidence — and the bash-3.2-specific `eval` bug is exactly this shape.
- Recursive discovery in a dev clone: `**/tests/*.sh` from the repo root matches ~2000 files under `.git/cogni-worktrees/**` plus the dead `_archive/` suite and the three sourced-only fixture helpers. CI would not see these (fresh checkout), so the defect would only ever bite a contributor running the runner locally — an untested-path asymmetry.
- Self-recursion: the sweep discovers `tests/test_run_plugin_tests.sh`; if that self-test invokes the runner in execute mode on the real root, the whole 90+-suite sweep nests inside one of its own suites.
- Unbounded runtime: no `timeout-minutes` on the job and/or no per-suite timeout means one hanging suite pins a runner until the platform's default kill, and the issue's explicit "how long may the sweep take" decision goes unrecorded.
- Undiagnosable red: the runner reports only an aggregate pass/fail count without naming which suite failed or its output, so every CI failure requires a local re-run to interpret.
- Two root suites (`test_check_breadcrumbs.sh` case 4, `test_check_version_bump.sh` case 10) grade the REAL working tree, so the sweep's verdict now partly duplicates and partly entangles with the dedicated `breadcrumb-guard` and `version-bump-gate` jobs — a single repo-state problem can turn two jobs red, and checkout-depth choices in the new job silently change what case 10 measures.
- The convention entry in root `CLAUDE.md` records the `plugin:agent` form but drops the anchoring half of the rule — an unanchored `cogni-consult:` prefix matcher fires for other plugins' subagents, which is the mirror-image bug and is equally silent.
- Documentation goes stale in the opposite direction: `cogni-consult/CLAUDE.md:146` keeps telling readers no job discovers `tests/*.sh` after a job does, so the next maintainer trusts a false statement in the very guide the issue cites as the reason the lesson needed lifting.
- JUDGMENT-ONLY, not gated: whether gating on exit status alone (rather than on the `set -u` + pass/fail-counter convention the issue body asserts) is the right contract for the long run. The issue asks for a decision, and a human should read whether the recorded rationale is honest about how many suites actually share the stated convention — my base reads found all 8 cogni-consult suites do declare `set -u`, which cuts against a blanket 'only a minority do' framing.
- JUDGMENT-ONLY, not gated: whether a hard-coded real-tree suite count anywhere in `tests/test_run_plugin_tests.sh` is an acceptable maintenance coupling — any PR in any plugin that adds a test file would turn an unrelated build red. A `>=` lower bound is the non-brittle form; a human should decide if an exact count was chosen deliberately.

- The plan-time acceptance criteria list 3 items, but a base-grounded regeneration derives 21 gating items — including the two properties that decide whether this change achieves anything at all (the sweep's non-zero exit must not be swallowed, and zero-suite discovery must fail rather than pass silently), plus the runtime bound and test-contract decision that the issue BODY explicitly demands but the AC section never pinned. The acceptance bar was materially under-specified relative to the issue's own text. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1222
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-validator (cogni-trends)

**Summary:** Contract grading surfaces four uncontested major gaps (items 6, 10, 20, 22) that keep this PR from shipping as-is; item 18 is resolved in plugin-validator's favor (bash 3.2.57 reproduction + full 11-site static sweep + confirmed PR-body disclosure outweighs the contract reviewer's assertion) and is downgraded to a waived minor rather than gating. plugin-validator's own residual finding — the in-file comment should name the unprotected-brace-comma discriminator, not just 'multi-key dict literals' — survives as a legitimate minor. The oversized-diff flag is informational only.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-validator (cogni-trends) | ✅ pass |

### Findings (11)

- **[major]** (auto-fixable) `review-contract:item-6` — Root CLAUDE.md's Contributing bullet (line 178) quotes the job as "Plugin test suites", but the job's actual name: in .github/workflows/lint.yml:93 is "Plugin test suites (discover and run tests/*.sh)" — the bullet does not use the name string exactly as it appears in the workflow, unlike the breadcrumb-guard bullet it is modeled on.
  - *Suggested fix:* Quote the full job name verbatim.

- **[major]** (auto-fixable) `review-contract:item-10` — tests/test_run_plugin_tests.sh has a negative-control fixture (case 2, test_broken.sh exiting 1) but no case running the runner against a zero-discovery root asserting non-zero exit + success:false. Only the production code (run-plugin-tests.py:169-178) implements the guard, unpinned by any self-test assertion.
  - *Suggested fix:* Add an empty-fixture-root case asserting exit 1 + success:false.

- **[major]** `review-contract:item-20` — cogni-consult/CLAUDE.md line 146's caveat ('That test is not CI-enforced... run it by hand') is unchanged by this PR. Independently confirmed absent at the PR's merge-base (4f8ad433c) but present on current origin/main (added by an intervening commit). Once this PR's job lands, the merged repo asserts something false about its own CI; the author's deferral premise was true only at the stale branch point.
  - *Suggested fix:* Correct the sentence in this PR, or record a reason grounded in the actual current base text.

- **[major]** (auto-fixable) `review-contract:item-22` — CONTRIBUTING.md is not in the changed files. Both existing CI guards have their own ### section with copy-pasteable local-run commands (Breadcrumb Guard, Version-Bump Gate); no equivalent section exists for the new Plugin Test Suites sweep, and no stated exemption is recorded.
  - *Suggested fix:* Add a ### Plugin Test Suites section with local-run commands, or state the exemption.

- **[minor]** `PR diff` — Diff is 515 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (waived) `cogni-trends/agents/` — All 12 agents carry valid frontmatter but none embeds <example> blocks in description.
  - *Justification:* Uniform across all 12 and untouched by this diff — a repo-wide style position, not a defect this PR introduced.
  - *Raised by:* plugin-validator (cogni-trends)

- **[minor]** (waived) `cogni-trends/scripts/initialize-trend-project.sh:276,280,284` — Bootstrap script still instructs users to run the nonexistent /trend-report skill.
  - *Justification:* Verified pre-existing on base; file not in this PR's changed set — pre-existing drift, recorded not gated.
  - *Raised by:* plugin-validator (cogni-trends)

- **[minor]** (auto-fixable) `cogni-trends/scripts/project-status.sh:319-327` — The in-file comment names the symptom but not the discriminator that keeps the other 11 eval sites safe. It says the trigger is 'the multi-key dict literals below', which reads as if any brace-with-comma is dangerous — yet six other sites contain brace-comma groups and are provably fine because escaped double quotes re-protect them. The actual rule is: an UNPROTECTED brace-comma group in eval argument position. Without that sentence a maintainer adding a bare dict literal to any of the 11 untouched sites reintroduces the identical silent-zero defect.
  - *Suggested fix:* Add one sentence naming the discriminator.
  - *Raised by:* plugin-validator (cogni-trends)

- **[minor]** (auto-fixable) `cogni-trends/tests/test-stale-detection.sh:5,91,133` — Three sibling comments still name the retired /trend-report as the producer of the drift anchor, though every assertion was renamed. Same stale-literal class the PR set out to clear, left in the very file that clears it.
  - *Suggested fix:* Update the three comments to name /trend-synthesis Phase 4.1.
  - *Raised by:* plugin-validator (cogni-trends)

- **[minor]** (waived) `cogni-trends/tests/test-stale-detection.sh:219,227` — refresh_actions is computed and threaded into SUMMARY JSON but never asserted; the rename kept a dead field alive.
  - *Justification:* Dead at base as well — an enhancement, not a regression.
  - *Raised by:* plugin-validator (cogni-trends)

- **[minor]** (waived) `review-contract:item-18` — Contract-grading-reviewer's original assessment: partial fix of the eval bug class in cogni-trends/scripts/project-status.sh with no disclosure of which sites were fixed.
  - *Justification:* Adjudicated in favor of plugin-validator's materially stronger evidence: the bug was reproduced under real bash 3.2.57 (base emits 0/0/0/0/0/0 on a populated fixture, HEAD emits 5/12/12/12/12/1.0), the discriminator was isolated to an unprotected brace-comma group in eval argument position (the escaped-quote f-string form is provably unaffected), and a static sweep of all 11 remaining eval sites found zero unprotected brace-comma groups. The PR body's 'Review notes' section discloses this exact scope. The original 'no disclosure' finding does not hold against this evidence and does not gate the verdict.

### Positive observations

- Items 1,2,3,4,5,7,8,9,11,12,13,14,15,16,17,19,21 all SATISFIED with cited file:line evidence, including stdlib-only runner with {success,data,error} envelope, exactly one new lint.yml job with the three existing jobs unchanged, non-zero propagation intact end to end, and explicit zero-discovery failure branch. (cogni-service:contract-grading-reviewer)
- Item 17 satisfied line by line with assertion strength ADDED under the rename; item 18's central risk disproven by independent bash 3.2.57 reproduction and a static sweep of all 11 remaining eval sites; item 19 satisfied with direct execution evidence (base emits all six counters 0, HEAD emits all six correct). (plugin-validator (cogni-trends))
- Fix shape is minimal-blast-radius (assignment-position command substitution, zero Python payload lines touched) with an unusually good nine-line explanatory comment naming mechanism, shell version, silence mechanism, and both maskers; both suites pass under forced bash 3.2.57 with byte-identical output vs bash 5.3.9. (plugin-validator (cogni-trends))
- Plugin structure clean: valid manifest, 9 skills with well-formed frontmatter, 12 agents complete, no version or marketplace file touched; diff is exactly the two declared files (+25/-9). (plugin-validator (cogni-trends))

### Recommendation

⚠️ **Revise before merge.** 7 finding(s) need attention.

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1222
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator

**Summary:** Revise round at head 5b7c4f4e: both routed perspectives passed (deterministic pre-gates also clean, 0 offenders / 0 violations). The 7 non-waived findings from the prior NEEDS_REVISION verdict (head 0d5d5f83) are gone from this round's output. The only two remaining findings are minor and non-gating — one advisory oversized-PR flag and one pre-existing agent name/filename mismatch in cogni-trends that plugin-validator explicitly scoped out of this PR and waived with justification. No major or critical findings; nothing blocks merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** `PR diff` — Diff is 576 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (waived) `cogni-trends/agents/trend-report-investment-theme-writer.md:2` — PRE-EXISTING, NOT INTRODUCED BY THIS PR — the agent's frontmatter declares `name: trend-investment-theme-writer` while the filename is `trend-report-investment-theme-writer.md`, and every dispatch site uses the filename form (skills/trend-synthesis/SKILL.md:341,345 and skills/trend-synthesis/references/synthesis-skeleton.md:10,137,147). 100 of 102 agents repo-wide have name == filename. If the host resolves the plugin-qualified agent id from the frontmatter name, trend-synthesis Step 2.1 cannot resolve the writer at all.
  - *Justification:* Outside this PR's diff — the PR touches only project-status.sh and test-stale-detection.sh in cogni-trends, and this defect predates it and is unrelated to the CI sweep. Raised for visibility only; it must not gate this PR.

### Positive observations

- The eval fix is a genuine silent-failure repair, not a style change — verified with set -x under /bin/bash 3.2 that the original form re-invokes python3 once per comma fragment with braces stripped, leaving every value-model counter at its 0 default. (plugin-dev:plugin-validator)
- The revise round's discriminator sentence is empirically correct, not a plausible-sounding guess: all four distinguished cases were tested — unprotected brace-comma group in eval argument position breaks; assign-then-eval works; braces inside escaped double quotes work; inside single quotes work. (plugin-dev:plugin-validator)
- The stale-detection test rename is a repair with a tightening, not a loosening — the revised test requires BOTH /trend-research and /trend-synthesis fragments, pinning the post-split two-stage refresh path. (plugin-dev:plugin-validator)
- Both cogni-trends suites are discovered by the new runner and pass; tests/fixtures/ is correctly excluded by the deliberately non-recursive */tests/*.sh glob, exactly as CONTRIBUTING.md documents. (plugin-dev:plugin-validator)
- The zero-discovery guard works — a filter matching nothing returns rc=1 with 'no test suites discovered', so a future glob regression fails loudly instead of passing an empty sweep. The CI annotation path was confirmed end-to-end with a forced failing suite. (plugin-dev:plugin-validator)
- The CI job's shallow-checkout comment documents a real trap: copying the version-bump gate's fetch-depth 0 would make test_check_version_bump.sh case 10 grade the live tree and double-report a version-touched finding as a test failure. (plugin-dev:plugin-validator)
- Root CLAUDE.md's new example regex is an exact structural match for the live matcher in cogni-consult/hooks/hooks.json, including the optional qualifier and both anchors. (plugin-dev:plugin-validator)
- No plugin-surface risk: no SKILL.md, agent, manifest, marketplace entry, or version was touched in cogni-trends. check-skill-names.sh, check-breadcrumbs.py and the version mirror all verify clean. (plugin-dev:plugin-validator)

### Recommendation

---

claude-cogni-work:
## Revision posted

**Commit:** `5b7c4f4e3b49bc08ba4888d5b67840032c48d77c`
**Addressed:** all 7 non-waived findings from the NEEDS_REVISION verdict — contract items 6, 10, 20, 22 and the two plugin-validator minors, plus the informational diff-size flag.
**New verdict:** ✅ PASS — see https://github.com/cogni-work/insight-wave/pull/1222#issuecomment-5183222965

The branch was updated from `origin/main` first: item 20's target sentence did not exist on this branch at all (it was added to `main` after the merge-base), so there was nothing to edit until the branch was refreshed. That merge is why the file count grew.

### Findings resolved (11)

- **[major]** `review-contract:item-6` — root `CLAUDE.md` quoted the job as "Plugin test suites" while `lint.yml:93` reads `Plugin test suites (discover and run tests/*.sh)`.
  - *Resolution:* the Contributing bullet now quotes the job name verbatim, matching how the neighbouring breadcrumb bullet quotes its own job.
  - *Commit:* `5b7c4f4e`

- **[major]** `review-contract:item-10` — no self-test pinned the runner's zero-discovery guard; only production code implemented it.
  - *Resolution:* added case 8 to `tests/test_run_plugin_tests.sh` — an empty fixture root asserting exit 1, `success:false`, `total:0`, and a non-empty `error`. Assertion count 11 → 13, all passing.
  - *Commit:* `5b7c4f4e`

- **[major]** `review-contract:item-20` — the "not CI-enforced … run it by hand" caveat would have shipped false once this job landed.
  - *Resolution:* fixed here rather than deferred — the finding was right that the deferral premise held only at the stale branch point. Confirmed absent at merge-base `4f8ad433c` and present on current `main`. The branch was merged from `main` and the sentence corrected in place. Follow-up #1220 is now closed by this PR (`Closes #1220` added), and the PR body's Scope paragraph records why the premise changed.
  - *Commit:* `5b7c4f4e`

- **[major]** `review-contract:item-22` — no CONTRIBUTING section for the new sweep, and no recorded exemption.
  - *Resolution:* added `### Plugin Test Suites` after the Version-Bump Gate section, modeled on the two existing guard sections: what the job runs and how non-recursive discovery excludes `_archive/` and `tests/fixtures/`, the three-property suite contract, and a bash block with the full sweep, `--list`, `--filter`, and the runner's self-test.
  - *Commit:* `5b7c4f4e`

- **[minor]** `cogni-trends/scripts/project-status.sh:319-327` — the comment named the symptom but not the discriminator.
  - *Resolution:* the comment now names the actual rule — an UNPROTECTED brace-comma group in eval ARGUMENT position, not the dict literal. It cites *any* quoting context rather than only escaped double quotes, because `project-status.sh:199` is protected by single quotes; naming one form would have misclassified that site. plugin-validator independently reproduced all four cases under bash 3.2 and confirmed the sentence is accurate.
  - *Commit:* `5b7c4f4e`

- **[minor]** `cogni-trends/tests/test-stale-detection.sh:5,91,133` — three comments still named the retired `/trend-report`.
  - *Resolution:* all three repointed at `/trend-synthesis` **Step 3.1**, not the suggested "Phase 4.1" — Phase 4.1 was the retired skill's own numbering, and `trend-synthesis/SKILL.md:430` is where `report_tier` / `content_hash_at_report` are actually written. Using the suggested label would have swapped one stale literal for another.
  - *Commit:* `5b7c4f4e`

- **[minor]** (waived) `PR diff` — diff is over the 400-line review bound, and this revision grows it further (576 lines now).
  - *Justification:* the verdict itself flagged this as informational only, and splitting is now net-negative: the merger has already graded this diff item-by-item against a committed review contract, and re-cutting it would discard that grading and restart review from cold. The change is also not cleanly separable — the runner, its self-test, the CI job, the two suites the job exposed as red, and the docs describing it are one contract. Stated rather than hidden: the PR body gained a Review-notes entry recording the decision.

- **[minor]** (waived) `cogni-trends/agents/` — all 12 agents carry valid frontmatter but none embeds `<example>` blocks in `description`.
  - *Justification:* carried forward from the prior verdict — uniform across all 12 and untouched by this diff; a repo-wide style position, not a defect this PR introduced.

- **[minor]** (waived) `cogni-trends/scripts/initialize-trend-project.sh:276,280,284` — bootstrap script still instructs users to run the nonexistent `/trend-report` skill.
  - *Justification:* carried forward from the prior verdict — verified pre-existing on base, file not in this PR's changed set.

- **[minor]** (waived) `cogni-trends/tests/test-stale-detection.sh:219,227` — `refresh_actions` is computed and threaded into SUMMARY JSON but never asserted.
  - *Justification:* carried forward from the prior verdict — dead at base as well; an enhancement, not a regression.

- **[minor]** (waived) `review-contract:item-18` — original assertion of undisclosed partial fix scope in `project-status.sh`.
  - *Justification:* carried forward from the prior verdict — adjudicated in favor of plugin-validator's bash 3.2.57 reproduction, the 11-site static sweep, and the PR body's Review-notes disclosure.

Auto-revise triggered by bot-authored verdict findings — no human feedback since the last automated review. See the linked verdict comment above for the original findings.

### Two things beyond the findings

The `/simplify` pass on the revise delta surfaced two items worth naming explicitly, since neither was in the verdict:

1. **A factual error on the very line item 6 asked me to touch.** Root `CLAUDE.md` documented discovery as `cogni-*/tests/*.sh`, but the runner globs `*/tests/*.sh` — and its own docstring explicitly rejects the narrower form, because it "would silently skip a future non-`cogni-` component that ships suites". The same wrong glob was in `tests/test_run_plugin_tests.sh`'s header. Both corrected. Note this would not have been caught by any assertion: case 7 asserts `any(r.startswith('cogni-'))`, which passes under either glob.

2. **`cogni-knowledge/tests/README.md` carried the same now-false claim as item 20** ("CI integration is deferred (cogni-knowledge has no CI today)"), made false by this PR's own change. Corrected on the same reasoning that put item 20 in scope, and the documented `--filter knowledge` command was verified (76 suites, exit 0).

**Verification:** 100/100 suites pass on macOS bash 3.2.57; `tests/test_run_plugin_tests.sh` 13/13; `check-breadcrumbs.py`, `check-external-dispatch.py`, `check-version-bump.py` all clean; `check-frontmatter.sh` clean across all three touched plugins.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1222
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator

**Summary:** All three perspectives passed and the 22-item review contract graded fully SATISFIED; the only un-waived findings are three advisory minors (oversized-diff flag, a failure-counter style nit, and a missing timeout clause in docs) surfaced for the human but non-gating - two further minors are explicitly waived with author-facing justifications pending confirmation.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (5)

- **[minor]** (waived) (auto-fixable) `.github/workflows/lint.yml:95-97` — The comment states 'The full sequential sweep measures ~3.5 minutes locally'. Measured on the merged tree, the 100 discovered suites sum to 59.0s of per-suite runtime (slowest 4.59s). The recorded number overstates by ~3.5x. The 20-minute bound is still correct and generous, so this is a stale figure in a comment, not a gating defect.
  - *Justification:* The author's measurement may have been taken on a cold cache or a busier machine; the direction of the error is conservative.

- **[minor]** (auto-fixable) `CONTRIBUTING.md:100-112` — The per-suite runtime bound (DEFAULT_TIMEOUT = 300 in scripts/run-plugin-tests.py) is surfaced only via --timeout's argparse help text - it is not stated in the new CONTRIBUTING.md 'Plugin Test Suites' section or in the root CLAUDE.md bullet, both of which otherwise document the suite contract. A contributor reading the docs learns a suite must exit non-zero and need no network, but not that it must finish inside 300 seconds.
  - *Suggested fix:* Add one clause to the CONTRIBUTING.md suite-contract paragraph: each suite must finish within the runner's 300-second per-suite default (override with --timeout), and the whole job is capped at 20 minutes.

- **[minor]** `PR diff` — Diff is 576 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

- **[minor]** (waived) `cogni-trends/scripts/project-status.sh:316-383` — The rewrite's stated premise is not reproducible with the repo's own fixture. Running cogni-trends/tests/test-project-status.sh against the BASE copy under both bash 3.2.57 and bash 5.3.9 passed all eleven assertions, including the exact counters the rewritten block emits. The in-file comment's claim that the base form 'emits nothing at all' did not hold in this environment on either bash. The change is behaviour-neutral and harmless and item 18 is satisfied because a rationale IS recorded in-file - but the rationale itself is unverified, which matters because it is the sole justification for leaving 11 of 12 sites of the alleged bug class untouched.
  - *Justification:* The author may have reproduced this on a platform or fixture not tested here (e.g. a real project workspace rather than the committed fixture); if so the comment stands as written and only the reproduction recipe is missing.

- **[minor]** (auto-fixable) `tests/test_run_plugin_tests.sh:41-48` — The `check` helper sets FAILED=1 as a flag rather than incrementing a counter. The four sibling root suites the contract names as the shape to match all use FAILED=$((FAILED + 1)) (tests/test_check_version_bump.sh:48). Behaviour is equivalent for the exit code, but the final summary loses the failure count, so a red run says 'Some run-plugin-tests runner tests failed' without saying how many.
  - *Suggested fix:* Change FAILED=1 to FAILED=$((FAILED + 1)) and report the count in the closing red() message, matching tests/test_check_version_bump.sh.

### Positive observations

- Contract item 8 satisfied: the sweep step uses `set -uo pipefail`, captures `rc=$?` immediately after the runner call, and the final line is `exit "$rc"` - no continue-on-error, || true, || echo, or trailing unconditional exit 0 anywhere in the job (lint.yml:92-131). (contract-grading-reviewer)
- Contract item 9 satisfied: run-plugin-tests.py has an explicit empty-set branch (lines 431-440) printing success:false and returning 1 when discovery finds zero suites. (contract-grading-reviewer)
- Contract item 16 satisfied: tests/test_check_version_bump.sh's diff only adds an `unset VERSION_BUMP_BRANCH GITHUB_HEAD_REF` guard; all ten `check "case N"` labels and their exact string-equality assertions, including case 10's real-tree check, are byte-identical to base. (contract-grading-reviewer)
- Contract item 17 satisfied: test-stale-detection.sh case 2 still asserts stale_count>=1, subtype==scout_drift, t-003 in added, t-001 in changed, and the reason string - now additionally requiring both /trend-research and /trend-synthesis fragments (a strengthening, not a loosening). (contract-grading-reviewer)
- Contract item 18 satisfied: only the value-modeler eval site was rewritten; the added in-file comment explicitly states why the remaining 11 eval sites are safe, confirmed by reading each remaining site. (contract-grading-reviewer)
- Contract item 20 satisfied: cogni-consult/CLAUDE.md:146's caveat is corrected in the diff from 'not CI-enforced ... run it by hand' to naming the new job - AC3 is landed, not deferred. (contract-grading-reviewer)
- Contract item 22 satisfied: CONTRIBUTING.md gains a `### Plugin Test Suites` section with copy-pasteable local-run commands, matching the existing guard sections' format. (contract-grading-reviewer)
- Vacuity guard is real and pinned from both ends: the runner has an explicit empty-set branch returning success:false / exit 1, and test case 8 asserts all three properties against an empty fixture root - the exact silent-zero class the issue was filed to end. (plugin-validator)
- Non-zero propagation verified by execution, not by reading: a failing fixture suite makes the runner exit 1 and names it in failed_suites, and the workflow step terminates in exit "$rc". (plugin-validator)
- Discovery correctness proven against the hostile case: --list against the real dev clone (whose .git/cogni-worktrees holds thousands of .sh files) returned 99 suites with zero dot-directory, _archive/, or fixtures/ paths. (plugin-validator)
- No self-recursion: the only real-root invocation in the self-test is --list, and case 7 asserts structure rather than an exact suite count, so adding a plugin suite will not turn an unrelated build red. (plugin-validator)
- The checkout-depth comment reasons in the OPPOSITE direction from version-bump-gate and explains why - shallow keeps case 10 from double-reporting a version violation as a test-suite failure. (plugin-validator)
- Test strength went up, not down, in the collateral edits: test-stale-detection.sh case 2 keeps every base assertion and adds a second required reason fragment. (plugin-validator)
- AC3 is landed rather than deferred; cogni-knowledge/tests/README.md's 'CI integration is deferred' paragraph was caught and corrected too, though nothing in the checklist demanded it. (plugin-validator)
- The root CLAUDE.md convention entry carries all three load-bearing facts including the anchoring half - the mirror-image bug the risk watchlist warned would get dropped. (plugin-validator)
- Full sweep verified green on the merged tree: 100/100 suites passed. The three touched plugins' plugin.json files are absent from the diff, parse as valid JSON, and match their marketplace.json entries exactly. (plugin-validator)

### Recommendation

---

Issue #1205, opened 2026-08-04
cogni-consult: trim consult-design-thinking and consult-resume back under the skill body-size cap


sdh07:
## Plugin
cogni-consult

## Current behavior

Three `cogni-consult` skills sit over their complexity-scaled body-size cap
(`cogni-service/scripts/measure-skill-length.sh`, `metric_version: 2`). This is
standing debt that predates the Step 0 work, not a regression from it:

| Skill | Before Step 0 | After Step 0 | Cap at head |
|---|---|---|---|
| `consult-design-thinking` | 31748 / 25500 (**+6248**) | 32332 / 27000 (**+5332**) | cx 14 |
| `consult-resume` | 20626 / 16500 (**+4126**) | 21211 / 18000 (**+3211**) | cx 8 |
| `consult-action-fields` | 20314 / 19500 (**+814**) | 21173 / 21000 (**+173**) | cx 10 |

The Step 0 change moved every margin **toward** budget — adding a step raises
complexity by one, which raises the cap by 1500, while the inserted block costs
only ~580–860 chars. `consult-action-fields` is now effectively at budget and
needs no work. The two that remain materially over are `consult-design-thinking`
and `consult-resume`.

Body size is not a CI gate today (`lint.yml` runs breadcrumb-guard,
external-dispatch-guard and version-bump-gate only), so this surfaces only when a
PR touching one of these files gets graded against a review contract — where it
reads as a blocking finding on work that improved the number.

## Desired behavior

- `consult-design-thinking/SKILL.md` and `consult-resume/SKILL.md` are at or under
  their measured cap, by relocating orchestration prose into
  `cogni-consult/references/` rather than by deleting instruction.
- Every relocated block keeps a pointer from the SKILL.md body, so no behavior is
  lost — the reduction is progressive disclosure, not removal.
- `consult-action-fields` is explicitly out of scope; verify it is still ≤ cap and
  leave it alone.

## Motivation

Two effects, both real. At runtime the whole body is loaded on every invocation,
so the excess is paid on every `consult-design-thinking` and `consult-resume`
call. At review time the over-cap condition fires as a finding against whichever
PR happens to touch the file next, regardless of whether that PR caused it —
which is exactly what happened to the Step 0 change, where the natural remedy was
forbidden by the same contract's scope-containment item.

## Acceptance criteria

- [ ] `measure-skill-length.sh` reports `over_cap: false` for
[...]

PR #1238, opened 2026-08-10
cogni-consult: precompute engagement last_activity in discovery

### LINKED ISSUE #1158: cogni-consult: precompute engagement last_activity in _discover_extractor.py instead of a per-engagement read loop
## Context

Deferred from #1140 during service-resolve. Optional performance follow-up — **not** required
for epic #1147's Definition of Done.

**Deferred item:** Precompute last activity at discovery time by extending
`cogni-consult/scripts/_discover_extractor.py` (and its host contract in
`cogni-workspace/scripts/discover-plugin-projects.sh`) to surface the newest
`.metadata/execution-log.json` `transitions[].timestamp` as an additive `last_activity` key —
replacing the per-engagement read loop that #1140 specifies in SKILL.md prose.

## Why it was deferred rather than shipped

#1140 needs an engagement list **sorted by last activity, descending**, but no pre-selection
data source carries last activity today (verified against `origin/main`):

- `scripts/engagement-status.sh` never opens `.metadata/execution-log.json`, and runs only
  *after* an engagement is selected.
- `scripts/_discover_extractor.py` `extract()` returns `path, slug, name, language,
  key_question, action_fields, scope_state, plugin_refs, updated` — no last-activity field.

#1140 therefore instructs the skill to read each engagement's execution log itself before the
sort. That works (discovery already returns `path`, and the skill's frontmatter already grants
`Read`), but it is N reads per invocation expressed as prose the model must honour.

Moving the computation into the extractor would make the sort data-backed instead of
prose-backed. It was deferred because the extractor's host was believed to be a **cogni-workspace**
script, so the change would cross a plugin boundary for what is, in #1140, a display
concern.

**That deferral rationale was disproved during grooming.**
`cogni-workspace/scripts/discover-plugin-projects.sh:182-191` passes the extractor dict
straight into `json.dumps` with **no key allowlist**, so an additive `last_activity` key
flows through to every consumer with no host change at all. The plugin boundary is never
crossed. The cost that justified splitting this out does not exist — which makes the work
cheaper than described above, not less warranted.

## Acceptance criteria

_Ratified via `/cogni-service:service-groom` — a base-grounded derivation against `origin/main`, replacing the informal sketch above. Two things changed in the process:_

- _**The stated premise is over-stated.** `cogni-workspace/scripts/discover-plugin-projects.sh:182-191` passes the extractor dict straight into `json.dumps` with **no key allowlist**, so an additive key flows through untouched. No host change is required, and the plugin-boundary cost that caused the deferral does not actually exist._
- _**Maintainer decision on timestamp shape:** keep `last_activity` raw (heterogeneous datetime/date) and render date-part-only at the consumer, so intra-day ordering survives in the data._

**Scope and gates**

- [ ] `cogni-consult/scripts/_discover_extractor.py` is modified and its `extract()` still returns a plain `dict` with the same signature `extract(d: str) -> dict` — confirm the `def` line is unchanged in the diff.
- [ ] The file imports **stdlib modules only** — grep the diff's import lines; `json`/`os` are already present, any addition must be stdlib.
- [ ] `cogni-portfolio/scripts/_discover_extractor.py` and `cogni-trends/scripts/_discover_extractor.py` are **not** in the changed-file set — the field is consult-specific and must not leak cross-plugin.
- [ ] `cogni-workspace/scripts/discover-plugin-projects.sh` is unmodified. Given the no-allowlist finding above, a host change is not needed; if the diff modifies it anyway, the PR body must state why and show no change to the JSON-emitting python block (`:182-191`) that alters the envelope shape for cogni-portfolio or cogni-trends.
- [ ] No plugin version is touched: `cogni-consult/.claude-plugin/plugin.json` and the repo-root `.claude-plugin/marketplace.json` version lines are unchanged on the branch (the CI Version-bump gate must pass, not report `version-touched`).
- [ ] No maintainer **breadcrumb** (issue/PR ref, version tag, slice code) is introduced into `consult-resume/SKILL.md` or any SKILL/agent file — the breadcrumb guard runs clean against the head.
- [ ] `cogni-consult/skills/consult-resume/SKILL.md` **frontmatter** remains valid with its required keys (`name`, `description`, `allowed-tools`); `allowed-tools` still includes `Read` and `Bash`, since Step 4 still reads `field.json`.
- [ ] `consult-resume/SKILL.md` stays within its **skill length** / body cap after the prose edit — the loop removal should reduce, never grow, the body.
- [ ] The plugin **structure** / **manifest** set stays valid (plugin.json + marketplace.json parse and agree).

**The extractor change**

- [ ] The returned dict carries a `last_activity` key on **every** code path, including when `consult-project.json` is absent. Confirm the assignment sits outside/after the `consult-project.json` `try` block, not nested inside it — today `:20-36` is one try/except, so a `last_activity` assigned only inside it would be missing on the fallback path.
- [ ] Every pre-existing key (`path`, `slug`, `name`, `language`, `key_question`, `action_fields`, `scope_state`, `plugin_refs`, `updated`) keeps its current name and value expression — the diff shows no edit to `:18` and `:25-34` other than additions.
- [ ] `last_activity` is the **newest** transition, computed as a max over all `transitions[].timestamp` values — not `transitions[-1]` or `transitions[0]`. Evidence: the diff uses `max(...)`/sorted-max over the collected timestamps, **and** the new test has a fixture whose `transitions[]` is stored out of chronological order (newest not last) and asserts the older-positioned newest timestamp is returned.
- [ ] The fallback chain is exactly three-tier and each tier is tested separately: (a) log present with non-empty valid transitions → newest timestamp; (b) log absent, unreadable, or `transitions[]` empty/missing → the root `updated` value; (c) neither `consult-project.json` nor a usable log → empty string `""` (never `None`, never a missing key).
- [ ] The fallback reads `project.get("updated", "")`, not `project['updated']` — the latter raises `KeyError` on exactly the fixture the extractor's `path`/`slug`-only path exists for.
- [ ] A malformed execution-log never breaks discovery: the new test covers, as distinct fixtures, at minimum invalid JSON bytes, `transitions` present but not a list, a `transitions` entry that is not a dict, and an entry missing `timestamp` or carrying a non-string `timestamp`. Each asserts `extract()` returns normally (no traceback) with `last_activity` falling back per the tier chain.
- [ ] A malformed execution-log does **not** suppress the `consult-project.json` fields: a fixture with a valid `consult-project.json` plus a corrupt `execution-log.json` asserts `slug` / `name` / `key_question` / `scope_state` still carry their file-derived values — confirming the new read has its own `try/except` rather than being folded into the existing one.
- [ ] The read is one targeted `os.path.join(d, '.metadata', 'execution-log.json')`, not a filesystem walk or a stat of every engagement subtree — trading N skill-level reads for N+ script-level reads is not the improvement asked for.

**Timestamp shape (ratified)**

- [ ] `last_activity` values stay **raw**: a transition-sourced value keeps its full ISO datetime (`2026-06-11T09:00:00Z`) and an `updated`-sourced value keeps its bare ISO date (`2026-06-11`). The extractor does not normalize either into the other.
- [ ] The heterogeneity is handled deterministically at the sort: the new test asserts that a datetime-sourced and a date-sourced value falling on the same day order deterministically, and `consult-resume/SKILL.md` states that the `Zuletzt bearbeitet` column renders the **date part only** for both shapes.

**The consumer change**

- [ ] `consult-resume/SKILL.md` Step 2 no longer instructs a per-engagement execution-log read: the "Read the log once per discovered engagement here, after discovery and before the sort" sentence (`:96-97`) and the `<path>/.metadata/execution-log.json` read instruction (`:89-90`) are removed or replaced, and the prose names the `last_activity` field from discovery instead. Grep the head SKILL.md for `.metadata/execution-log.json` — it should no longer appear in Step 2's sort instruction.
- [ ] The Step-2 rendering contract survives the rewrite verbatim in effect — grep the head SKILL.md for each of these four rules still being stated: sort by last activity **descending**; ties broken on engagement **name ascending** (`:73`); `Zuletzt bearbeitet` rendered **date-part-only**; an engagement with no last activity sorted **last** and rendered `—` (`:94`).
- [ ] Step 4's `zuletzt bearbeitet` line (`:153-155`, "resolves exactly as in step 2") still points at a rule that exists after the Step-2 rewrite — no dangling reference to a removed procedure.
- [ ] The five-row cap, the `Weitere N` overflow rule, the list-and-stop obligation, and the English/German parallel table all survive the loop removal unchanged. The edit should be surgical.
- [ ] `cogni-consult/CLAUDE.md`'s scripts table row for `_discover_extractor.py` ("Per-engagement JSON field extractor consumed by the discovery wrapper (reads the flat consult-project.json schema)") is updated to state that it also reads `.metadata/execution-log.json` and surfaces `last_activity` — the doc must not claim a single-file read after the change.

**Test discipline**

- [ ] A new regression test exists at `cogni-consult/tests/test_discover_extractor.sh` (or an equivalently named file under `cogni-consult/tests/`) following the existing convention: `set -u` (not `set -e`), `mktemp -d` fixture root with `trap … EXIT`, per-fixture `pass`/`fail` counter, non-zero exit on any failure — as in `tests/test_generate_engagement_readme.sh:34-52`.
- [ ] The end-to-end envelope is verified, not just the function: the test (or the PR's stated verification) runs `bash cogni-consult/scripts/discover-projects.sh --root <tmp> --json` against a fixture workspace and asserts each discovered entry contains `last_activity` — proving the additive key survives the host's `json.dumps` path.
- [ ] **Mutation check**: the test genuinely fails if the new `try/except` is removed. A test that constructs fixtures but asserts only the happy path claims the "defensive, never breaks discovery" requirement rather than demonstrating it.

**Watch at review (not gated)**

- Taking `transitions[-1]` as "newest" — the log is append-ordered in practice, so this passes casual manual testing and then silently mis-sorts any engagement whose log was hand-edited or merged. The single most likely wrong fix.
- Folding the execution-log read into the existing `try` at `:20-36`, so a corrupt log strips `slug`/`name`/`key_question` from the envelope and every downstream consult skill degrades — a cross-skill regression from a display-only change.
- Emitting `None` or omitting `last_activity` on the no-data path, so the sort compares `None` against strings and the "sort last, render —" branch cannot be implemented uniformly.
- Shipping the extractor field but leaving `consult-resume/SKILL.md`'s N-read loop in place, so the purpose is unmet and the cost is paid twice.

**Parent issue:** #1140
**Plan record:** https://github.com/cogni-work/insight-wave/issues/1140#issuecomment-5090140190

claude-cogni-work:
## Summary

`consult-resume` re-derived each engagement's freshness by opening every engagement's `.metadata/execution-log.json` itself, after discovery had already walked those same directories. This precomputes it once, in the place that already holds the path.

- `_discover_extractor.py` gains an additive `last_activity` key: the newest `transitions[].timestamp`, chosen with `max(...)` rather than `transitions[-1]`, so a log whose newest entry is not last still resolves correctly.
- Three-tier fallback: newest timestamp → root `updated` → `""`. The read sits at **function scope under its own `try/except`**, so a corrupt log degrades only this field, and the key is present as a `str` on every path — including the one where `consult-project.json` is absent entirely and a `project["updated"]` lookup would `KeyError`.
- Values stay **raw and heterogeneous** (full ISO datetime vs bare ISO date), per the ratified decision, so intra-day ordering survives in the data; date-part-only rendering stays a presentation rule in the skill.
- `consult-resume` steps 2 and 4 consume the field instead of prescribing the read loop, and now state an explicit total ordering for the two shapes (compare on the date part, then engagement name ascending).
- New `tests/test_discover_extractor.sh`: 14 cases over the tier ladder, the malformed-log surface, raw-shape preservation, and an end-to-end run through the real cogni-workspace host.

## Scope decisions

- **No host change.** `cogni-workspace/scripts/discover-plugin-projects.sh` runs `projects = [mod.extract(d) for d in dirs]` straight into `json.dumps` with no key allowlist, so an additive key flows through untouched. This is the grooming finding that disproved the original cross-plugin-boundary deferral rationale, re-verified here against `origin/main`.
- **Consult-only by design.** The sibling `cogni-portfolio` and `cogni-trends` extractors already carry divergent key sets, and nothing in code, test, doc, or CI asserts parity between the three — so this creates no unshipped obligation on them. Neither sibling has an equivalent transitions log to generalize over: cogni-portfolio has no `.metadata/execution-log.json` at all, and cogni-trends' execution log is plain text. Hoisting the computation into the shared host would force it to hardcode one plugin's log schema — pushing plugin knowledge *up* into shared infrastructure, the opposite of the right altitude.
- **No version bump.** The post-merge workflow owns the patch increment for every plugin the merge touches.
- **SKILL.md net line delta is −1** (12 added, 13 removed). The five-row cap, the `Weitere N` overflow rule (incl. never `Weitere 0`), the „alle" re-render rule, the list-and-stop obligation, the ` · noch nicht geschärft` suffix and the English parallel are all untouched.

## Test plan

- [x] `bash cogni-consult/tests/test_discover_extractor.sh` — 14/14 pass, no network, all writes under a `mktemp -d` root removed by an `EXIT` trap.
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-consult` — 10/10 suites pass (the new one is auto-discovered by the `*/tests/*.sh` glob, no registration step).
- [x] `python3 scripts/check-breadcrumbs.py` — clean, 0 violations.
- [x] Version-bump gate — 14 plugins scanned, **1 touched** (`cogni-consult`), **0 violations** vs `origin/main`; `clean: true`, `violations: []`. (The earlier "0 touched" reading in this body was wrong — the gate counts plugins the diff touches, not version lines. No version line is touched, which is what the gate passes on.)
- [x] `check-frontmatter.sh cogni-consult` — `success: true`; `allowed-tools: Read, Bash, Skill` unchanged.
- [x] `git diff --numstat origin/main -- cogni-consult/skills/consult-resume/SKILL.md` → `12 13` (deletions ≥ additions).

The end-to-end case (M13) runs the real `scripts/discover-projects.sh` and is hermetic on three axes, each load-bearing:

| Pin | Why it is required |
|---|---|
| `--root <tmp>` | confines the find walk to the fixture tree |
| `--registry <tmp>/empty-registry.json` | the helper's flag loop takes the **last** `--registry`, so the caller's value overrides the wrapper's hardcoded `$HOME/.claude/cogni-consult-projects.json`. Without it the developer's real engagements join the result and the count is unstable. Verified empirically by observing the helper seed the temp file with `{"projects":[]}` |
| `WORKSPACE_PLUGIN_ROOT` | the wrapper otherwise prefers an installed `$HOME/.claude/plugins/cache/...` copy over the in-repo sibling — on this machine it silently resolved to cogni-workspace **v0.6.46**, so a dev with a stale cache and CI would exercise different code |

The fixture is laid out as `<root>/cogni-consult/<slug>/` because the wrapper pins `--find "consult-project.json:*/cogni-consult/*:1"`; a flatter layout discovers **zero** engagements, which would make "every entry carries `last_activity`" trivially true. The case therefore asserts the engagement count *before* asserting on the field.

## Mutation recipe

Two guards are proven by falsifier rather than asserted. Both were run; both returned `"verdict": "guard_verified"` (mutated red, restored green).

```bash
  --root . \
  --file cogni-consult/scripts/_discover_extractor.py \
  --expr 's/max\(stamps\)/stamps[-1]/' \
  --test 'bash cogni-consult/tests/test_discover_extractor.sh' \
  --case M1
```

```bash
  --root . \
  --file cogni-consult/scripts/_discover_extractor.py \
  --expr 's/project\.get\("updated", ""\)/project["updated"]/' \
  --test 'bash cogni-consult/tests/test_discover_extractor.sh' \
  --case M4
```

M1's fixture deliberately places the newest transition in a non-final position, so `stamps[-1]` must turn it red. M4's fixture has no `consult-project.json`, so `project["updated"]` must `KeyError`. Run both from the repo root, so `--file` resolves inside `--root` and `--test` (which runs with `cwd=--root`) resolves too.

The `--expr` is a `perl -0pi -e` expression, not sed, and an expression matching nothing is a hard `expr_no_op` error rather than a silent pass — so `max(stamps)` must stay on one line with that exact spelling.

## Review notes

**The suite's `PASS:` / `FAIL:` label prefixes deviate from the plugin's inherited `OK   <name>` / `FAIL <name>:` shape on purpose.** `mutation-check.sh:412-413` classifies on `^[ \t]*FAIL:[ \t]+<case>(?:[ \t]|$)` and `^[ \t]*(?:ok|PASS):[ \t]+<case>(?:[ \t]|$)`; the inherited shape matches neither and would return `case_not_found`. The detail is separated from the case-id by a **space, not a colon**, for the same reason — the first draft used a colon, and the mutation check caught it by reporting `case_not_found` on a case that was genuinely red. `fail()` still writes to stderr as the sibling suites do; the harness captures the test command with `2>&1`.

**Pre-existing over-cap, annotate-only.** `consult-resume/SKILL.md`'s body is 21,595 chars against an 18,000-char cap. This is pre-existing and **this diff shrank it** — the base measured 21,624. Per the net-growth budget that lands in the annotate-only cell: no coupled rewrite here, and no follow-up filed, because the trim is already tracked by open issue #1205. A dedup sweep in this PR would also break the acceptance contract, which pins several surrounding clauses byte-identical. `skill-quality-reviewer` returned this as its only finding, tagged `introduced_by_change: false`, category `preference`.

**One behavioural difference worth naming.** The removed prose claimed the single-engagement and named-engagement branches "skip that cost entirely." Discovery now computes `last_activity` for all N engagements even when one is targeted. Measured at 0.246 ms per engagement on a deliberately fat fixture (200 transitions, 42 KB log), inside a process that is already opening all N `consult-project.json` files in that same branch — so this is not a regression worth acting on, but it is a real difference from what the old prose promised.

Closes #1158.

---

## Resolution of the held review (head `8aa51be2`)

Both majors were held for human adjudication. The maintainer directed the
resolution below; nothing here is a self-waiver.

### Item 19 — mutation falsifier: genuine gap, fixed

The two `--expr` recipes above were real and both returned `guard_verified`, but
they run against `cogni-service/scripts/mutation-check.sh`, which lives in the
managed-service marketplace and **not in this repo**. Nobody but the author on
that machine could reproduce them. The in-repo precedent item 19 names —
`cogni-portfolio/scripts/mutation-check.sh` — is a committed, no-argument,
self-contained harness with its mutations hardcoded, and has **no `--expr`
surface at all**. Item 19's two halves are therefore inconsistent in form; both
are now satisfied rather than one traded against the other: the `--expr` recipes
stay stated above, and the runnable harness is committed.

New `cogni-consult/scripts/mutation-check.sh` — no arguments, no network, kept
under `scripts/` (not `tests/`) so `run-plugin-tests.py`'s `tests/*.sh` /
`*/tests/*.sh` globs do not run it as an ordinary suite. Verified: `--list`
does not discover it. Each mutation copies the extractor to a `mktemp -d`,
applies the mutation there, and asserts the suite goes red; the working tree is
never mutated in place, and a mutation expression that matches anything other
than exactly once is a hard error, not a silent pass.

| Mutation | Cases that go red |
|---|---|
| `max(stamps)` → `stamps[-1]` | M1 — returns `2026-06-28T17:00:00Z`, the last-positioned entry, instead of the newest `2026-08-01T06:15:00Z` |
| the log read's `except Exception` narrowed to `except ZeroDivisionError` | M3, M4, M5, M10, M12 — `extract()` raises on an absent or corrupt log |

The second mutation narrows the guard rather than excising the block. Deleting
the block outright is the weaker mutant: `stamps` stays `[]` and the fallback is
still reached, so `last_activity` is still assigned and only M1/M11 red — which
would prove the sort, not the guard.

The suite gains one overridable assignment,
`SCRIPT="${CONSULT_DISCOVER_EXTRACTOR:-…}"`, so the harness can aim the
`assert_extract` cases at the mutant. This mirrors cogni-portfolio's
`PROJECT_STATUS_SCRIPT="$mutated"` precedent. M13 is deliberately unaffected —
it drives the real discovery wrapper, which loads the extractor by its own path,
so it stays green under mutation; the reds come from the `assert_extract` cases,
which is what these falsifiers target.

**Contract item 1 note.** Item 1 pins the changed-file set to "exactly these
four cogni-consult paths". The harness is a fifth path. It sits **inside**
cogni-consult, so item 1's stated concern — a cross-plugin leak — is untouched.
Flagging it rather than letting a grader discover it.

### Item 18 — e2e exact count: kept, rationale moved to the assertion

The item forbids an exact `count` *because* the wrapper merges the developer's
real `$HOME/.claude/cogni-consult-projects.json` registry into the result. The
`--registry` pin removes that cause — as `plugin-validator` verified
empirically, with 15 real registry entries the pin still yields exactly the two
fixture engagements. The exact assertion is strictly stronger than the item
required, and it is also what guards the vacuity case the contract's own risk
watchlist names: an empty `projects[]` satisfies every `all()` below it.

Taking the reviewer's second suggested fix: the assertion keeps `len(ps) == 2`
and now carries that rationale in a comment at the assertion site, not twenty
lines above it. Not weakened to `>= 2` — that would discard a deterministic,
stronger assertion to satisfy a prohibition whose cause the test already removed.

### Minor — CAPS command register

Applied at maintainer direction: `you MUST output … and STOP` → `output … and
stop`, and `you can regenerate and open it` → `regenerate and open it`. Both
line-neutral.

**Residual tension, stated rather than absorbed.** Contract item 24 requires the
list-and-stop obligation "with its git-signals prohibition" to *survive
unchanged*, and a register change is literally a change. The edit alters only the
imperative register — no clause added, removed, or reordered, and the git-signals
prohibition is byte-identical. The merger should take its own view on whether
that clears item 24.

### Minor — not applied

The intra-day-precision finding on "sort on the date part alone" stays as
shipped: contract item 22 names "compare on the date part, then break the tie on
name ascending" as its own acceptable exemplar.

### Verification on `8aa51be2`

- `bash cogni-consult/tests/test_discover_extractor.sh` → 14/14, exit 0.
- `bash cogni-consult/scripts/mutation-check.sh` → "All mutations caught.", exit 0;
  `git diff --quiet` on the extractor confirms the working tree is untouched after.
- `python3 scripts/run-plugin-tests.py --filter cogni-consult` → 10/10, exit 0;
  `--list` does not discover `mutation-check.sh`.
- `python3 scripts/check-breadcrumbs.py` → `success: true`, 0 violations.
- Version-bump gate → `ok`, 14 scanned, 1 touched, **0 violations**.
- `consult-resume/SKILL.md` frontmatter intact; `allowed-tools: Read, Bash, Skill`
  unchanged. Body now 21,580 chars — down from 21,595 at the reviewed head and
  21,624 at the base, so the diff still shrinks it.
- `git diff --numstat origin/main` for that file → `15 16`, deletions ≥ additions.
- Step 2, newline-normalised: `.metadata/execution-log.json` absent,
  `Read the log once per discovered engagement` absent, `last_activity` present.


claude-cogni-work:

Authored by the `--role author` instance. This PR is **draft** and carries `svc:needs-review`; no ready flip, no verdict label, no `reviewed-head-sha` marker was applied, so the merger's independent review is not short-circuited. Issue #1158 is now `svc:resolved`.

### Handoff preflight

`check-preflight.sh` against a checkout of the PR head: **clean, exit 0** — 4 passed, 3 skipped, 1 advisory.

| Instrument | Result |
|---|---|
| frontmatter | pass — clean for cogni-consult |
| breadcrumbs | pass — clean for cogni-consult |
| version-bump | pass — **no version-line touch** |
| skill-length | pass — 1 of 1 measured, 1 over cap (advisory) |
| run-tests | **skipped** — `no_suites_at_base`: the base ref carried no cogni-consult suites, so there is nothing to regress from (this PR adds the first) |
| agent-length | **skipped** — `not_applicable`, no touched `agents/*.md` |
| mutation-recipe | **skipped** — `not_applicable`: the instrument matches `*/tests/test-*.sh`, and this plugin's convention is underscores (`test_discover_extractor.sh`), so no touched path matched. The recipes are stated explicitly in the PR body and were both run — see below |

One process note worth recording, since it would otherwise look like a flapping gate: the first preflight run returned exit 1 on `version-bump`. That was an artifact of pointing `--root` at the shared clone, which is checked out on an unrelated branch (`feat/issue-1116-projects-dashboard`) and does not even contain `scripts/check-version-bump.py`. Re-run against a checkout of this branch, the instrument reports `no version-line touch`. Independently confirmed: `git diff origin/main...origin/feat/issue-1158-precompute-last-activity` touches no `plugin.json` or `marketplace.json`, and contains no added or removed `"version"` line.

### Author self-grade vs the issue-time contract

Graded by `contract-grading-reviewer` against the 25-item contract committed on #1158 at triage time (`--target issue --origin triage`), before any branch was cut.

**Verdict: pass — 25 satisfied, 0 violated, 0 unclear. No residual.**

The grader verified the two items the contract itself flagged as vacuously-passable, rather than taking the PR body on trust:

- **Item 20** (the read instruction is gone from Step 2) was checked with **newline normalisation** (`sed -n '63,128p' | tr '\n' ' '`), because the clause is line-wrapped in the base and a naive single-line grep returns zero even against an untouched file.
- **Item 18** (e2e hermeticity) — the grader read the host's flag loop and the consult wrapper's `exec` line itself and confirmed the test's `--registry` genuinely overrides the wrapper's hardcoded `$HOME` registry, because the caller's flag arrives later via `"$@"` and the case-statement is last-wins.

**One judgment the grader made explicit rather than resolving silently, flagged here so the merger can take its own view:** item 18's literal text says the test "does not assert on an exact `count`". This suite *does* assert `count == 2`. The grader graded it satisfied on the ground that the item's stated *rationale* — registry entries merging in and destabilising the count — is affirmatively neutralised by the `--registry` pin, so the exact-count assertion is deterministic here and strictly stronger than the item's letter required. A stricter reading could call that a literal violation. It is the only place the grade departs from the contract's literal text.

### Verification run on the head

- 10/10 cogni-consult suites pass, including the new 14-case `test_discover_extractor.sh`.
- Two mutation checks returned `guard_verified` (mutated red, restored green): `max(stamps)` → `stamps[-1]` reds M1, and `project.get("updated", "")` → `project["updated"]` reds M4.
- The mutation harness also caught a real defect in the first draft of the test itself: `fail()` printed `FAIL: <case>: <detail>`, but the harness matches case-id followed by whitespace-or-EOL, so a genuinely red case was misreported as `case_not_found`. Fixed to a space separator.

### Carried for the reviewer

- **Pre-existing over-cap, annotate-only.** `consult-resume/SKILL.md` is 21,595 chars against an 18,000 cap. Pre-existing — the base was 21,624, so this diff *shrank* it by 29 chars. Per the net-growth budget that is the annotate-only cell: no coupled rewrite, and no follow-up filed because open issue #1205 already tracks the trim. `skill-quality-reviewer` returned this as its only finding, tagged `introduced_by_change: false`, category `preference`.
- **A behavioural difference the removed prose promised.** The old text said the single-engagement and named-engagement branches "skip that cost entirely". Discovery now computes `last_activity` for all N engagements even when one is targeted — measured at 0.246 ms per engagement, inside a process already opening all N `consult-project.json` files in that same branch. Judged not worth acting on, but it is a real difference and the reviewer should see it stated rather than discover it.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1238
**Perspectives consulted:** cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:resume-audit-reviewer, cogni-service:loop-structure-reviewer

**Summary:** Verdict held at needs_revision on two major review-contract findings (items 18 and 19) that contract-grading-reviewer scores VIOLATED on literal text while plugin-validator scores SATISFIED on independent empirical verification -- a genuine contract-vs-diff divergence flagged for human adjudication via human_judgment rather than resolved here; all other checklist items pass across all six perspectives, with only pre-existing/advisory minor findings remaining.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| cogni-service:resume-audit-reviewer | ✅ pass |
| cogni-service:loop-structure-reviewer | ✅ pass |

### Findings (8)

- **[major]** (auto-fixable) `review-contract:item-18` — Item 18's hermeticity condition (b) requires the e2e test to pin WORKSPACE_PLUGIN_ROOT AND 'does not assert on an exact count'. test_discover_extractor.sh's M13 case asserts `len(ps) == 2` — an exact count — contradicting the contract's explicit prohibition. The reviewer confirms the test's trailing `--registry "$TMPROOT/empty-registry.json"` flag does genuinely override the wrapper's hardcoded registry (last --registry wins), so the count IS deterministic in this configuration — but that mechanism is not what the item's text anticipated or permitted, and the diff's literal action is the opposite of the item's literal instruction.
  - *Suggested fix:* Either loosen the M13 assertion to a non-exact form (len(ps) >= 2), or, if the exact-count assertion is intentionally kept because the --registry override makes it safe, note the rationale in a comment near the assertion.

- **[major]** `review-contract:item-19` — Item 19 requires the diff to supply a mutation falsifier — a recipe in the `--expr` form (rule 5), in-repo precedent cogni-portfolio/scripts/mutation-check.sh — that mutates the new try/except or swaps max() for transitions[-1] and demonstrates the suite going RED. Grep of the changed-file set finds no cogni-consult/scripts/mutation-check.sh and no --expr recipe in the diff itself. M1's inline assertion would catch such a mutation but is a passive fixture, not the demonstrated falsifier.
  - *Suggested fix:* Add cogni-consult/scripts/mutation-check.sh modeled on cogni-portfolio/scripts/mutation-check.sh: apply a perl -0pi mutation to _discover_extractor.py, run the M1 case, assert RED, restore.

- **[minor]** (waived) (advisory-addition) `cogni-consult/skills/consult-resume/SKILL.md` — Skill body is 21,595 chars against an 18,000-char cap (measure-skill-length.sh metric_version 2, complexity 8, over_cap true).
  - *Justification:* Pre-existing WITH a base measurement: origin/main measures 21,624 chars, so this diff SHRANK the body by 29 chars (net line delta -1). Grading it as a violation would grade pre-existing drift against a PR that improved it. Tracked by open issue #1205.
  - *Raised by:* plugin-dev:plugin-validator, also flagged by cogni-service:skill-quality-reviewer

- **[minor]** (waived) `cogni-consult/skills/consult-resume/SKILL.md:113-119` — The rewritten Step 2 says 'render and sort on the date part alone', which changes the sort key, not just the render shape. Two engagements worked on today at 09:00 and 17:00 now tie on the compared key and fall through to name-ascending, discarding the intra-day precision the extractor deliberately preserves.
  - *Justification:* item-22 names 'compare on the date part, then break the tie on name ascending' as its own acceptable exemplar, so the shipped wording is explicitly sanctioned by the authoritative rubric and this is not a checklist violation. Raised only because it is a user-observable ordering change that no checklist item measures.

- **[minor]** (waived) (advisory-addition) `cogni-consult/skills/consult-resume/SKILL.md:111-118` — The mixed-shape total ordering required by item 22 is assembled from two paragraphs ~20 lines apart: the comparison key sits in the rewritten paragraph, the tie-break in the untouched sort paragraph. The item is still satisfied on a newline-normalised whole-Step-2 read.
  - *Justification:* Item 24 requires deletions >= additions and item 22's stated achievable bar is the newline-normalised grep, which the current wording clears.

- **[minor]** (auto-fixable) (advisory-addition) `body §2 (the 'Multiple, and the user named none' bullet); secondary occurrence in §4` — The list-and-stop rule is written as a CAPS command stack in second person ('you MUST output the full engagement list ... and STOP') rather than reasoned imperative prose. The rule's rationale does follow, so the CAPS carries no load the prose doesn't already carry.
  - *Suggested fix:* Length-neutral replacement of the bullet in imperative form (4 source lines in, 4 out) so the net-shrink requirement is unaffected; in §4, 'you can regenerate and open it' -> 'regenerate and open it'. (ref: skill-creator §Writing Style; plugin-dev:skill-development §Writing Style Requirements)

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-consult/tests/test_discover_extractor.sh:242-243` — The M13 guard degrades to a printed SKIP that does not increment `failures` when the cogni-workspace host script is missing. Since M13 is the suite's only end-to-end assertion, a skip path means it can disappear while the suite still exits 0.
  - *Justification:* Unreachable in practice and in CI: run-plugin-tests.py runs from the repo root where cogni-workspace is committed. Hardening only.

- **[minor]** (waived) (advisory-addition) `PR body — Test plan, Version-bump gate line` — The PR body claims the version-bump gate reports '14 plugins scanned, 0 touched'; the gate actually reports plugins_touched: ['cogni-consult'] / '1 touched'. The gate is nonetheless clean (clean:true, violations:[], status:ok).
  - *Justification:* Prose accuracy in the secondary self-declared summary only; item-3's substance is satisfied and verified directly from the gate JSON.

### Positive observations

- Items 1-17 and 20-25 all satisfied with cited evidence from the head worktree; suite passes 14/14 and run-plugin-tests.py --filter cogni-consult exits 0. (contract-grading-reviewer)
- Extractor diff is 31 additions / 0 deletions, discharging items 6 and 7 mechanically; a delete-the-try/except mutation turned M3, M4, M5, M10, M12, M13 red. (plugin-validator)
- M13's hermeticity holds under adversarial conditions: with 15 real entries in the developer's registry, the --registry pin still yielded exactly the two fixture engagements. (plugin-validator)
- Items 20-24 (the consult-resume/SKILL.md half) all SATISFIED with cited evidence; net delta -1, no maintainer breadcrumbs. (skill-reviewer)
- All five governing review_contract items (20-24) satisfied on a content basis; frontmatter healthy and description triggers well. (skill-quality-reviewer)
- consult-resume passes all three structural checks after the last_activity refactor; audit-resume.sh --scope cogni-consult returned status audited with 1 pass / 0 warn / 0 fail. (resume-audit-reviewer)
- No anti-pattern found; duplicated inline log-read logic in two SKILL.md steps collapsed into one structured discovery field, consumed consistently by both. (loop-structure-reviewer)

### Recommendation

---

sdh07:
## Revised — head moved to `8aa51be2`

The two majors held for human adjudication were adjudicated by the maintainer and are resolved on the branch. Full record in the PR body under **Resolution of the held review**; the short version:

- **Item 19 (mutation falsifier)** — accepted as a genuine gap. The `--expr` recipes ran against a harness outside this repo, so they were not reproducible by anyone else. `cogni-consult/scripts/mutation-check.sh` is now committed on the `cogni-portfolio` precedent — no arguments, no network, under `scripts/` so `run-plugin-tests.py` does not pick it up. Two mutations, verified to red exactly the intended cases: `max(stamps)` → `stamps[-1]` reds M1; narrowing the log read's `except` reds M3/M4/M5/M10/M12. Neither touches the working tree.
- **Item 18 (e2e exact count)** — the exact `len(ps) == 2` is kept and the rationale now sits at the assertion. The item's prohibition exists because the wrapper merges the developer's `$HOME` registry in; the `--registry` pin removes that cause, so the exact count is deterministic here and also guards the vacuity case the contract's own risk watchlist names. Weakening it to `>= 2` would have traded a stronger assertion for the item's letter.

Two things the merger should take its own view on rather than inherit:

1. **Contract item 1** pins the changed-file set to exactly four paths; the harness is a fifth. It is inside `cogni-consult`, so item 1's stated concern — a cross-plugin leak — is untouched.
2. **Contract item 24** requires the list-and-stop obligation to survive *unchanged*, and the CAPS-register minor was applied at maintainer direction. The edit changes only the imperative register — no clause added, removed, or reordered, git-signals prohibition byte-identical — but it is literally a change to that text.

`svc:review-needs-revision` and `svc:needs-human-review` removed; `svc:needs-review` stands. The prior verdict's `reviewed-head-sha` marker (`b81f6e66`) is stale — it predates this commit.

Re-verified on the new head: suite 14/14; mutation harness "All mutations caught."; `run-plugin-tests.py --filter cogni-consult` 10/10; breadcrumb guard clean; version-bump gate 0 violations; `consult-resume/SKILL.md` numstat `15 16` (deletions ≥ additions), body down to 21,580 chars.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1238
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:resume-audit-reviewer, cogni-service:loop-structure-reviewer, cogni-service:contract-grading-reviewer

**Summary:** 24/25 committed review-contract items grade satisfied with cited empirical evidence (test_discover_extractor.sh 14/14, run-plugin-tests.py --filter cogni-consult 10/10, mutation-check.sh both falsifiers caught, version-bump/breadcrumbs/frontmatter/structure all clean at head 8aa51be2) and 0 are graded outright violated by the dedicated contract-grading reviewer, but the merge is not clean: two contract items (1 and 24) are independently flagged INTENT-UNCLEAR and need a human ruling rather than a diff-only resolution, and one un-waived major (item 18, the e2e exact-count assertion) is raised by resume-audit-reviewer as a violation. Item 18 is itself disputed rather than a settled defect: plugin-validator empirically verified that the item's stated prohibition-cause (the developer's real registry merging in) is eliminated by the test's --registry <tmp> pin and opted the point out as safe, and contract-grading-reviewer independently graded item 18 satisfied — so this reads as a genuine disagreement about what the contract bar requires, not a demonstrated author defect. Several minor findings were normalized to advisory (pre-existing body-length drift, a design-tension observation blessed by contract item 22, a pre-existing blockquote placement, and the oversized-PR notice) and do not gate.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| cogni-service:resume-audit-reviewer | ⚠️ needs_revision |
| cogni-service:loop-structure-reviewer | ✅ pass |

### Findings (9)

- **[major]** `review-contract:item-1` — INTENT-UNCLEAR: item 1 (SCOPE) caps the changed-file set at exactly four cogni-consult paths, but item 19 (MUTATION CHECK HAS TEETH) requires an in-diff mutation falsifier following the cogni-portfolio/scripts/mutation-check.sh precedent — the diff supplies exactly that as a fifth tracked file, cogni-consult/scripts/mutation-check.sh (under scripts/, not tests/, so run-plugin-tests.py does not run it as an ordinary suite, confirmed empirically). The two items are jointly unsatisfiable as literally written; the harness itself has real teeth (both falsifiers empirically caught, exit 0).
  - *Suggested fix:* Resolve at the contract level, not by editing the PR: either relax item 1's four-path cap to admit the scripts/mutation-check.sh companion item 19 anticipates, or clarify that item 19's falsifier evidence may live outside the tracked file set. A human should adjudicate which item yields.

- **[major]** `review-contract:item-18` — Contract item 18 (END-TO-END ENVELOPE) states the M13 e2e case must not assert an exact count, because the wrapper merges the developer's real $HOME/.claude registry into the result — yet test_discover_extractor.sh:274 asserts len(ps) == 2. The PR body claims a maintainer-directed re-verification, but that lives only in prose that does not amend the committed contract.json, the authoritative grading target. DISPUTED: plugin-validator independently verified the prohibition's stated cause is eliminated by the test's --registry <tmp> pin (load-bearing, since discover-projects.sh appends "$@" after its own --registry and the helper's arg parser is last-wins) and opted this out as safe; contract-grading-reviewer separately graded item 18 satisfied. Treat as a dispute about the contract's bar, not a settled defect, pending resolution.
  - *Suggested fix:* Relax the M13 assertion (e.g. >= 2 / superset-over-slugs) to comply with item 18's literal wording, or have the contract's author explicitly amend item 18 to accept the --registry-pinned exact-count assertion as-is before merging.
  - *Raised by:* cogni-service:resume-audit-reviewer

- **[major]** `review-contract:item-24` — INTENT-UNCLEAR: item 24 requires the list-and-stop obligation with its git-signals prohibition to survive unchanged, and the diff changes it. SKILL.md:69-70 went from 'you MUST output the full engagement list … and STOP for the user's explicit choice' to 'output the full engagement list … and stop for the user's explicit choice.' The immediately-following git-signals prohibition and back-references are byte-identical, so the obligation's substance is preserved but its imperative/CAPS emphasis is not. Unadjudicable from the diff alone: the edit is out of scope for the underlying issue (a review-round register change the PR body says was applied at maintainer direction, unverifiable from the diff), and the PR body explicitly declines to self-clear it ('The merger should take its own view on whether that clears item 24').
  - *Suggested fix:* Merger/maintainer to adjudicate item 24's strictness. If 'unchanged' means byte-identical, restore the 'you MUST … STOP' phrasing (line-neutral either way). If the register change is confirmed maintainer-directed and 'unchanged in effect' is intended, record that ruling and clear the item; no code change needed.
  - *Raised by:* plugin-dev:plugin-validator, also flagged by cogni-service:resume-audit-reviewer

- **[minor]** (waived) (advisory-addition) `cogni-consult/scripts/mutation-check.sh` — Item 1 pins the changed-file set to exactly four cogni-consult paths; the diff carries five, the fifth being this new mutation-falsifier script that item 19 requires in-repo. Internal contract inconsistency, not author overreach — item 1's substantive concern (nothing outside cogni-consult) is untouched: no sibling extractor or host file appears in git diff --name-status, and run-plugin-tests.py --filter cogni-consult ran exactly 10 suites with this script absent.
  - *Justification:* Satisfying item 19 in-repo necessarily adds this path, so the two items cannot both be met literally. The substantive half of item 1 (nothing outside cogni-consult; siblings and host untouched) is verified satisfied, and the added file is the artifact the contract asked for.

- **[minor]** (waived) (advisory-addition) `cogni-consult/skills/consult-resume/SKILL.md (whole file; 377 lines; chars_body=21578 vs cap_chars=18000 at complexity=8, metric_version=2; densest region Step 4 :144-264)` — Body exceeds its complexity-scaled character budget by ~20%. Progressive disclosure is in use for several references, but two blocks of variant-specific detail remain inline (bilingual rendering contracts :77-127/:158-188; four elect-only offer branches :319-361), so everything loads on every re-entry. PRE-EXISTING (introduced_by_change: false): the diff is net-shrinking (base 21624 chars -> head 21578); item 24's net-shrink test passes and the committed contract carries no length item.
  - *Justification:* Step 3.5 pre-existing-drift policy: structural_issues entry tagged introduced_by_change: false (base 21624 chars vs head 21578). Annotates, never gates.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-consult/skills/consult-resume/SKILL.md:116-118 (prose added by this change)` — The mixed-shape ordering rule is derivable but not stated in one place: the comparison key (date part) is at :117 while the tie-break (name ascending) is at :95-96, ~20 lines earlier. Since the determinism rule is carried entirely in model-executed prose with no executable sort to fall back on, the composition is worth making explicit at the point of use. Not a contract violation — item 22's own worked example blesses date-part-only comparison with a name-ascending tie-break, and its risk note forbids re-litigating the mechanism as an author defect.
  - *Suggested fix:* At :116-117 replace 'so render and sort on the date part alone;' with 'so render and sort on the date part alone — a full-timestamp value and a bare-date value falling on the same day are therefore a tie and resolve by the name-ascending rule above;'.

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-consult/skills/consult-resume/SKILL.md:264 (pre-existing)` — A blockquote about enabling the plugin's Strategy Advisor output styles sits at the end of Step 4, after four consecutive re-entry actions — session-configuration guidance, not part of presenting the dashboard. Pre-existing, untouched by this diff.
  - *Justification:* Pre-existing placement the diff never touched; Step 3.5 pre-existing-drift policy waives it to an annotation.

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-consult/tests/test_discover_extractor.sh:273-278` — Item 18 condition (b) says the e2e test 'does not assert on an exact count'; the test asserts len(ps) == 2. Deviation verified safe rather than accepted on the author's argument: the prohibition's stated cause (the developer's real registry merging in) is eliminated by --registry "$TMPROOT/empty-registry.json", and the override is load-bearing (discover-projects.sh appends "$@" after its own --registry; the helper's arg loop is last-wins). See the dedicated review-contract:item-18 major finding above for the disagreement this creates with resume-audit-reviewer's grading of the same item.
  - *Justification:* The prohibition's stated cause is provably eliminated by the --registry pin, verified against the wrapper's arg order and the helper's last-wins flag loop. The deviation is stronger than the item requires and its failure mode is a red test, not a vacuous pass.

- **[minor]** (advisory-addition) `PR diff` — Diff is 475 changed lines; the reviewable bound is 400. PRs this large are harder to review thoroughly. Note 411 of the 475 lines are new test/mutation-harness files (test_discover_extractor.sh 296, mutation-check.sh 115).

### Positive observations

- 24 of 25 committed review-contract items grade satisfied with cited evidence; 0 graded violated by the dedicated contract-grading reviewer — only items 1 and 24 flagged INTENT-UNCLEAR pending human adjudication. (cogni-service:contract-grading-reviewer)
- Every behavioral claim verified empirically by at least two perspectives: test_discover_extractor.sh 14/14 exit 0, run-plugin-tests.py --filter cogni-consult 10/10 exit 0, mutation-check.sh both falsifiers caught exit 0. (plugin-dev:plugin-validator, cogni-service:resume-audit-reviewer, cogni-service:contract-grading-reviewer)
- Version-bump, breadcrumbs, frontmatter, and structure gates all clean at head. (plugin-dev:plugin-validator, cogni-service:resume-audit-reviewer, cogni-service:contract-grading-reviewer)
- consult-resume passes all three audit-resume.sh criteria (C1 discovery+setup fallback, C2 phase-branching next actions, C3 7/7 skills referenced) with no pre-existing resume drift. (cogni-service:resume-audit-reviewer)
- The loop-bearing SKILL.md file moves last-activity computation from a per-engagement prose read-loop into a discovery-provided field with no anti-pattern introduced; discovery still re-scans live state every session. (cogni-service:loop-structure-reviewer)
- SKILL.md diff is a genuine net shrink (numstat 15/16); the five-row cap, Weitere-N overflow, alle re-render rule, list-and-stop's git-signals prohibition, and the bilingual parallel table all survive byte-identical. (plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:resume-audit-reviewer)
- Consumer-side rewrite (contract items 20-23, 25) is clean and substantively improved: the stale per-engagement log-read clause is gone, replaced by consumption of a discovery-supplied last_activity field with an explicit 'Never read that log here' prohibition, and rule (4) is correctly re-keyed onto the empty-string tier-(c) value. (plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer)
- Extractor placement, log-read isolation, and max(stamps) ordering all verified against actively-falsifying fixtures (M1, M4, M10, M13); the mutation harness gates on a green baseline and hard-errors on ambiguous mutation matches. (plugin-dev:plugin-validator)

### Recommendation

---

sdh07:
## Human adjudication — the three held majors, ruled

Head is now `733141f6`. All three round-2 majors were adjudication questions, not
author defects. Each was re-verified in-session against the committed 25-item
contract on #1158 — the rulings below do not inherit any reviewer's grade.

| Item | Ruling | Change |
|---|---|---|
| 1 (SCOPE) vs 19 | Contract defect — item 1 yields | none |
| 18 (END-TO-END ENVELOPE) | Cleared — prohibition's cause is eliminated | none |
| 24 (SURGICAL EDIT) | **Upheld** — reverted | 2 clauses |

### Item 1 yields to item 19 — contract defect, not author overreach

Item 1 caps the changed-file set at exactly four paths; item 19 requires an
in-repo mutation falsifier and names `cogni-portfolio/scripts/mutation-check.sh`
as the harness precedent. Satisfying item 19 in-repo *necessarily* adds a fifth
file. The two items are jointly unsatisfiable as written.

Item 1's substantive concern is stated in the issue's own acceptance criteria —
"the field is consult-specific and must not leak cross-plugin". All five paths
are under `cogni-consult/`; both sibling extractors and the cogni-workspace host
are untouched. The enumeration predates item 19's harness being contemplated as
a file. Item 1 yields; what it protects is intact.

### Item 18 cleared — the prohibition's stated cause no longer exists

Item 18 forbids an exact `count`, and states its cause with a "since": the
wrapper hardcodes the developer's real `$HOME/.claude/cogni-consult-projects.json`
and those entries merge in. Read directly rather than taken from either
perspective:

- `cogni-consult/scripts/discover-projects.sh` `exec`s the host with its own
  `--registry "$HOME/…"` **first**, then `"$@"`.
- `cogni-workspace/scripts/discover-plugin-projects.sh:48-55` parses flags in a
  plain `while`/`case` loop — `--registry) REGISTRY_FILE="$2"` — with no
  first-wins guard. Last wins.

So the test's trailing `--registry "$TMPROOT/empty-registry.json"` genuinely
overrides, and the cause is provably gone. The assertion is also stronger than a
count — it pins the exact values `['2026-02-09', '2026-08-01T06:15:00Z']` — and
it discharges the vacuity risk the contract's own watchlist names ("an empty
`projects[]` satisfies every `all()`"), for which the watchlist proposed only the
weaker `count > 0`.

Weakening to `>= 2` would trade a deterministic, stronger assertion for the
letter of a prohibition whose reason has been removed. Kept as shipped.

**Both of the above are cleared as contract defects. A future grader should not
re-raise them against this diff.**

### Item 24 upheld — the register edit is reverted

The one finding with teeth. Item 24 requires the list-and-stop obligation to
survive **unchanged**, and the diff changed it. The git-signals prohibition was
byte-identical and no clause was added, removed or reordered — but the edit is
unrelated to #1158's refactor. It entered as a round-1 style *preference* minor
(never a gate) and is precisely the collateral prose churn item 24 exists to
prevent. That the replacement reads better does not make it in scope here.

`733141f6` restores both clauses to their `origin/main` text byte-identically:
`you MUST output … and STOP`, and `you can regenerate and open it`. Line-neutral.

The register cleanup is a legitimate improvement on its own merits and moves to
the open `consult-resume` body-trim issue rather than riding an extractor
refactor.

Item 24's net-shrink half was re-measured, not assumed: numstat `12 13`
(deletions ≥ additions), body 22606 chars against a 22635-char base — still a
net shrink.

### Re-verified on `733141f6`

`test_discover_extractor.sh` 14/14 · `mutation-check.sh` all mutations caught,
extractor tree clean afterwards · `run-plugin-tests.py --filter cogni-consult`
10/10 (`mutation-check.sh` confirmed absent from discovery) · breadcrumb guard
clean · version-bump gate 0 violations · `last_activity` still consumed in
Steps 2 and 4.

### Note for whoever authors these contracts

The #1158 contract is internally inconsistent: an item enumerating an exact
file set cannot coexist with another item mandating an additional artifact.
Worth fixing at the triage template rather than per-PR.

Proceeding to ready-flip and merge.

sdh07 (approved):
Approved. The three held majors are adjudicated in the comment above: items 1 and 18 cleared as contract defects (each verified against the code, not inherited from a perspective), item 24 upheld and reverted in `733141f6`. All gates green on the new head.

PR #1252, opened 2026-08-10
cogni-projects: projects-backfill roll-off recommender skill

### LINKED ISSUE #1115: cogni-projects: backfilling — recommend next assignment when a consultant rolls off
## Plugin
cogni-projects

## Epic context
Phase 3 (portfolio steering — deferred) of "Roadmap: cogni-projects — partner project-portfolio steering (staffing MVP first)". Deferred follow-up that reuses the Phase 2 staffing scorer; it must not block the staffing MVP.

## Use case
When a consultant rolls off a project (or a project ends), the partner needs to find the best next assignment for them and to fill the vacated role — steering people between projects with minimal bench time.

## Proposed solution
Add a domain-prefixed `projects-backfill` skill that, given a rolling-off consultant or a newly-open role, reuses `staffing-score.py` to (a) rank open roles for the freed consultant and (b) rank replacement consultants for the vacated role. Sketch AC: (1) for a given consultant, output ranks open roles by the same availability/profile/impact scoring; (2) for a vacated role, output ranks replacement candidates excluding the departing consultant; (3) respects availability windows so it never proposes an already-fully-allocated consultant.

**Two claims in this paragraph were disproved during grooming — the ratified `## Acceptance criteria` below supersede them:**

- **Sketch AC-3 is not satisfied by reuse.** `_availability_score()` (`staffing-score.py:198-212`) returns the exclusion signal only on **zero temporal overlap**, so a consultant at `allocation_pct: 100` is still emitted as a ranked candidate. Excluding a fully-allocated consultant needs an explicit `allocation_pct >= 100` filter in the new skill, and the departing consultant must be exempted from it or the feature filters out the very person it exists to place.
- **"~1 skill, minimal new script code" understates it.** Beyond the filter above, direction (a) has to decide how `strategic_impact` orders roles (it is constant within a role but varies across them), and true capacity netting would require reading `assignments/`, which `staffing-score.py` does not do at all (`_load_entities` covers only consultants and projects, `:319-324`). The ratified scope keeps this out by exempting the departing consultant and stating the limitation.

## Alternatives considered
A standalone re-implementation of the scoring logic — rejected: reuse the Phase 2 scorer to keep ranking consistent. Folding backfill into projects-staff — rejected: distinct trigger and output, cleaner as its own skill.

## Acceptance criteria

_Ratified via `/cogni-service:service-groom` — a base-grounded derivation against `origin/main`, with two maintainer decisions folded in: "fully allocated" means `allocation_pct >= 100` read from the raw field, and the departing consultant is **exempt** from that filter with the limitation stated in prose rather than netting capacity from assignment records._

**Structure and gates**

- [ ] A new skill file exists at exactly `cogni-projects/skills/projects-backfill/SKILL.md`; the directory name is domain-prefixed `projects-backfill`, matching the sibling convention in `cogni-projects/CLAUDE.md` Conventions.
- [ ] The new SKILL.md has valid **frontmatter** with the required keys (`name`, `description`, `allowed-tools`) and `name:` equal to the directory name `projects-backfill` — confirm via the frontmatter gate, not by eye.
- [ ] The new SKILL.md's `description` is written in the sibling trigger-phrase style (a "This skill should be used when…" sentence plus an explicit `Trigger on:` phrase list, as in `projects-staff/SKILL.md:3-12`) and names backfill/roll-off triggers such as "rolls off", "backfill", "next assignment", "replace on".
- [ ] The new SKILL.md is within the **skill length** / body cap budget — confirm by running the skill-length measurement gate, not by eye.
- [ ] No **breadcrumb** is introduced: the new SKILL.md and any edited SKILL.md / CLAUDE.md / README carry no issue or PR ref, version tag, or milestone/slice code. This issue's dependency reference is a live copy-paste risk — confirm with the breadcrumb gate.
- [ ] Plugin **structure** and the **manifest**s stay valid, and no version is bumped in the branch: `cogni-projects/.claude-plugin/plugin.json` still reads `"version": "0.0.13"` and the repo-root `.claude-plugin/marketplace.json` mirror is unchanged (the post-merge workflow owns the bump).
- [ ] `cogni-projects/README.md` lists the new skill in both auto-generated sections: a `- **projects-backfill** — …` bullet under `## What it does`, and a `| Skill | projects-backfill | … |` row in the `## Components` table. The intro sentence at `:21` ("Backfilling builds on top of these.") is updated so it no longer describes backfilling as not-yet-existing.
- [ ] `cogni-projects/CLAUDE.md` is updated in all three places that currently declare backfilling unbuilt: the header sentence "the backfilling recommender arrives in later roadmap children", the `skills/` tree block (a `projects-backfill/SKILL.md` line is added), and the "Planned (not yet scaffolded …): `skills/` for the backfilling recommender" item is removed or narrowed.

**Reuse, not re-implementation**

- [ ] The scoring logic is reused: the new SKILL.md invokes `scripts/staffing-score.py` (grep the skill body for `staffing-score.py`) using the same `python3 "${CLAUDE_PLUGIN_ROOT:-…}/scripts/staffing-score.py"` idiom as `projects-staff/SKILL.md:79`, and the diff introduces no second copy of the scoring constants — grep the whole diff for `W_AVAILABILITY`, `W_PROFILE_FIT`, `SENIORITY_NORM`, `TOKEN_OVERLAP_DISCOUNT`; they must appear only in `cogni-projects/scripts/staffing-score.py`.
- [ ] `projects-staff` is not regressed: if `staffing-score.py` is modified at all, the one-argument invocation `python3 staffing-score.py <portfolio-dir>` still returns the same `data` shape (`portfolio`, `weights`, `project_order`, `consultant_count`, `project_count`, `ranked_candidate_count`, `projects[]`) and the same exit codes (0 ok / 1 domain failure / 2 usage) — in particular the `len(argv) != 1` usage guard at `:412` must not have been loosened. If the script is unmodified, this item passes trivially; say so in the grade.

**Behavior — the two directions**

- [ ] Direction (a), freed consultant → ranked open roles, is specified concretely: the SKILL.md states that it walks `data.projects[].open_roles[].candidates[]` and keeps the entries whose `consultant` equals the rolling-off consultant's slug, producing one ranked row per open role. Grep for `open_roles` and `candidates` in the new skill body.
- [ ] Direction (a)'s cross-role ordering handles `strategic_impact` explicitly. Because `strategic_impact` is a per-project constant *within* a role (`staffing-score.py:55-64`, `:358-363`) but varies *across* roles, the SKILL.md either orders the consultant's role list using `candidates[].scores.strategic_impact` (or the project-level `strategic_impact_norm`) in addition to `scores.combined`, or explicitly states and justifies that role ranking uses `combined` alone. A skill that sorts by `combined` with no mention of strategic impact fails this item — grep the new skill body for `strategic_impact`.
- [ ] Direction (b), vacated role → replacement candidates, excludes the departing consultant by slug: the SKILL.md states that the departing consultant's entry is dropped from the role's `candidates[]` before ranking, and names the field it keys on (`candidates[].consultant`).

**AC-3 — never propose an already-fully-allocated consultant**

- [ ] The exclusion is an **explicit filter**, not the scorer's existing behavior: the SKILL.md excludes consultants whose `allocation_pct` is `>= 100`, read from the **raw field**, from both directions' recommendations. The bar is that the skill does NOT rely on the scorer for this — `_availability_score()` (`staffing-score.py:198-212`) returns the exclusion signal only on **zero temporal overlap**, so a consultant at `allocation_pct: 100` is still emitted with a positive availability sub-score. A diff whose only exclusion mechanism is the scorer's `excluded_count` fails this item.
- [ ] The filter keys on the raw `allocation_pct` field, **not** on `_headroom()`: `_headroom()` (`staffing-score.py:181-195`) also returns `0.0` for a present-but-non-numeric `allocation_pct`, which would silently drop data-quality cases as if they were fully booked. Grep the skill body to confirm it names `allocation_pct` and not headroom as the filter input.
- [ ] The **departing consultant is exempt** from that filter in direction (a), and the SKILL.md says so in a sentence naming the departing consultant and `allocation_pct` — otherwise a 100%-allocated departing consultant is filtered out of their own recommendation, which is self-defeating.
- [ ] That exemption's limitation is **stated, not hidden**: the SKILL.md records that the departing consultant's freed capacity is *assumed* rather than computed, because `staffing-score.py` never reads `assignments/` at all (`_load_entities` is called only for consultants and projects, `:319-324`). A skill that claims to net out freed capacity without reading assignment records fails this item.
- [ ] Availability-window respect is stated for the reused path too: the SKILL.md tells the reader that consultants with no overlap between their `available_from` / `available_until` window and the project window are already excluded by the scorer and surface only as the per-role `excluded_count`, and are never proposed. Grep for `excluded_count` or `available_from` in the new skill body.

**State, edge cases, determinism**

- [ ] The backfill run does not clobber `projects-staff`'s state: the SKILL.md writes its own artifact under `cogni-projects/<portfolio-slug>/` with a filename distinct from `staffing-recommendations.md`, and does not overwrite `.metadata/staffing-recommendations.json`, which `projects-staff/SKILL.md:96-100` owns. Grep the new skill for `staffing-recommendations` and confirm any mention is read-only.
- [ ] If the skill appends a run record, it uses an envelope `portfolio-init.sh` already seeds (`:87-89` — `execution-log.json` `{"transitions": []}`, `staffing-log.json` `{"matches": []}`, `decision-log.json` `{"decisions": []}`) and preserves that object envelope rather than overwriting it with a bare array, as `projects-staff/SKILL.md:144-148` instructs. If a new log filename is introduced instead, the skill must instruct creating it defensively with its exact seed shape, since `portfolio-init.sh` will not have seeded it.
- [ ] Degenerate and failure inputs are handled in prose, each findable as a distinct sentence: (i) the named consultant / project / role does not exist in the portfolio; (ii) the vacated role is not present in the project's `open_roles` — the common case immediately after a roll-off, so the skill must say to add it via `/cogni-projects:projects-entities` or handle the not-yet-open role; (iii) a role whose candidate list is empty after exclusions; (iv) the scorer returning `success: false`, including that an `error` naming `validate-entities.py` means a broken install, not a portfolio-data problem (`projects-staff/SKILL.md:82-88`).
- [ ] Test demand matches where the mechanism lives. If the diff adds or modifies executable code, it adds a stdlib-only regression test under `cogni-projects/tests/` runnable as `bash cogni-projects/tests/test-<name>.sh`, matching the house convention in `tests/test-render-dashboard.sh` (no pytest/pip, `{success,data,error}` envelope assertions), asserting at minimum the fully-allocated exclusion and the departing-consultant exemption. If the mechanism is carried entirely in SKILL.md prose with no executable change, a behavioral test is structurally impossible: the achievable bar is the code-shape-anchored greps above, and this item passes — do not re-litigate test coverage for prose.
- [ ] Determinism is preserved: no wall-clock value (`datetime.now`, `date.today`, `time.time`) is introduced into `staffing-score.py` or any new script, and any new ordering the skill specifies names a deterministic tiebreak (e.g. project or consultant slug ascending), matching the guarantee at `staffing-score.py:28-31`.

**Watch at review (not gated)**

- The likeliest wrong fix is pure reuse — transposing the scorer JSON and calling AC-3 satisfied because "the scorer already excludes unavailable consultants". It does not; it excludes only on zero temporal overlap.
- Direction (a) ranked by `scores.combined` alone quietly drops `strategic_impact` from the exact ranking where it finally discriminates.
- A half-measure that ships only direction (b) — the near-free slice of the existing output — and leaves direction (a), the actual "minimal bench time" value, as a sentence.
- `plugin.json`'s description still ends "…so the backfilling recommender builds on the same portfolio", which reads as forward-looking once backfilling exists. Description alignment is `cogni-docs:doc-sync` territory, not gated here.

_Affected components: cogni-projects/skills/projects-backfill/SKILL.md, cogni-projects/scripts/staffing-score.py_

_Complexity hint: medium_

## Depends on
- #1114 cogni-projects: staffing match engine (availability + profile + strategic impact) — MVP

claude-cogni-work:
Closes #1115.

## Summary

When a consultant rolls off a project, two questions open together: **where they go next**, and **who takes the role they left**. This adds `cogni-projects/skills/projects-backfill/SKILL.md`, which answers both from a single `staffing-score.py` run.

- **Direction (a)** walks `data.projects[].open_roles[].candidates[]` and keeps the rows whose `consultant` matches the departing consultant's slug — one ranked row per open role.
- **Direction (b)** takes the vacated role's `candidates[]` and drops the departing consultant, keyed on `candidates[].consultant`.

## The two non-obvious decisions

**Direction (a) leads on `strategic_impact`, not `scores.combined`.** Strategic impact is a per-project constant *within* a role, so it cannot discriminate inside one role's shortlist — but this table compares roles *across* projects, which is the one ranking where it does. Sorting on `combined` alone would silently drop the firm's own priority ordering at the exact moment it becomes meaningful. Ordering is `strategic_impact` desc → `scores.combined` desc → project slug → role label. A null raw `strategic_impact` renders an em dash and is keyed on the **raw** field, since the normalized value is `0.0` for both an absent value and a raw `1`.

**AC-3 is an explicit filter, not scorer behavior.** The scorer excludes a consultant only on **zero availability-window overlap** — a consultant at `allocation_pct: 100` still overlaps the window and is still ranked, merely with a lower availability sub-score. So the skill applies its own filter, and reads `allocation_pct` from the **raw consultant record**, because a scorer candidate carries only `consultant`, `name`, and `scores`. It deliberately does **not** key on the scorer's headroom: `_headroom()` returns `0.0` for a present-but-non-numeric `allocation_pct` as well as a genuinely full one, so filtering on it would discard a typo as though it were a booked consultant. Malformed values surface as data-quality findings instead.

The departing consultant is **exempt** from that filter in direction (a) — they are the one person the recommendation exists to place. That exemption's limitation is stated in the skill and in its output rather than hidden: their freed capacity is **assumed, not computed**, because `staffing-score.py` never reads `assignments/` (`_load_entities` is called only for consultants and projects).

## Scope decisions

- **Prose-only; no executable change.** The ratified acceptance criteria sanction either, and state that when the mechanism is carried entirely in SKILL.md prose the test item passes on its explicit prose branch. `scripts/staffing-score.py` is byte-identical to base, so the *projects-staff not regressed* criterion passes trivially — verify with `git diff --stat origin/main -- cogni-projects/scripts/ cogni-projects/tests/` (empty).
- **No scoring constant is duplicated.** `git grep -nE 'W_AVAILABILITY|W_PROFILE_FIT|SENIORITY_NORM|TOKEN_OVERLAP_DISCOUNT' cogni-projects` hits only `scripts/staffing-score.py`.
- **Own artifacts and own log.** `backfill-recommendations.md` + `.metadata/backfill-recommendations.json` (fixed shape, documented), plus `.metadata/backfill-log.json` with a create-if-absent instruction and its exact `{"backfills": []}` seed — the scaffolder seeds only the execution, staffing, and decision logs. The `projects-staff`-owned `staffing-recommendations*` files are referenced read-only and never written.
- **Doc scope is wider than the criteria enumerate.** Beyond the listed surfaces this also corrects the opening paragraph, the "more skills land later" italic, the MEANS paragraph, the `## Architecture` tree count and entries, the standalone-scope sentence in `README.md`, and the `.metadata/` bullet in `CLAUDE.md`. Each is a one-sentence truth edit inside a file already being touched; a surviving "backfilling is not built yet" sentence would just be a review finding.
- **No version bump.** `plugin.json` and `marketplace.json` both stay at `0.0.24`; the post-merge workflow owns the bump. (The issue text asserting `0.0.13` is stale — the binding intent is do-not-touch.)
- Nothing deferred; the issue ships whole.

## Verification

All run against the branch:

| Gate | Result |
|---|---|
| `run-plugin-tests.py --filter cogni-projects` | 5/5 suites pass |
| `check-frontmatter.sh cogni-projects` | clean, 0 offenders, `files_scanned` 4 to 5 |
| `check-breadcrumbs.py` | 0 violations |
| `check-version-bump.py` | 0 plugins touched |
| `check-skill-names.sh` | pass — domain-prefixed `projects-backfill` |
| `measure-skill-length.sh` | `chars_body` 13041 vs `cap_chars` 24000, `over_cap: false` |
| `check-token-scope.sh --strict` | exit 0 |

Criterion-anchored greps against the new body, all hitting: `open_roles`, `candidates`, `candidates[].consultant`, `strategic_impact`, `allocation_pct`, `excluded_count`, `available_from`, `staffing-score.py`. Every `staffing-recommendations` mention is read-only phrasing.

**Not applicable: `## Mutation recipe`.** The touched set contains no `*/tests/test-*.sh` and no `*/scripts/check-*.sh`, so `check-preflight.sh`'s only gating instrument records `not_applicable`.

## Review notes

A `skill-quality-reviewer` pass returned `needs_revision` with **zero structural issues** and four auto-fixable polish items; all four were applied before commit — a fixed shape for the `.metadata` JSON artifact, the null-`strategic_impact` rendering rule, a fuller determinism statement naming both tiebreaks, and `"roll-off"` added to the trigger list.

The prose-simplifier pass was deliberately skipped: the body sits at 54% of its cap so there is no net-growth pressure, and several acceptance criteria require specific sentences to remain individually findable (the exemption sentence, the assumed-capacity limitation, the headroom rule-out, and four distinct degenerate-case sentences). A functionality-preserving prose editor cannot reliably distinguish those from ordinary redundancy.

### DIFF
--- a/cogni-projects/CLAUDE.md
+++ b/cogni-projects/CLAUDE.md
@@ -3,8 +3,9 @@
 Partner-facing project-portfolio steering for consulting firms. Models
 consultants, projects, and staffing so partners can match people to work by
 availability, profile fit, and strategic impact. The data model, entity
-authoring, the staffing match engine, and the read-only portfolio dashboard
-have landed; the backfilling recommender arrives in later roadmap children.
+authoring, the staffing match engine, the read-only portfolio dashboard, and the
+backfilling recommender have landed; cross-plugin bridges arrive in later
+roadmap children.
 
 ## Plugin Architecture
 
@@ -17,7 +18,8 @@ cogni-projects/
 │   ├── projects-setup/SKILL.md         Initialize a portfolio directory (entry point)
 │   ├── projects-entities/SKILL.md      Author + register one consultant/project/assignment
 │   ├── projects-staff/SKILL.md         Rank candidate consultants per open project role
-│   └── projects-dashboard/SKILL.md     Render a partner-meeting portfolio dashboard (read-only)
+│   ├── projects-dashboard/SKILL.md     Render a partner-meeting portfolio dashboard (read-only)
+│   └── projects-backfill/SKILL.md      Rank next roles for a rolling-off consultant + replacements
 ├── scripts/
 │   ├── _projects_lib.py                Shared loader for the hyphen-named validate-entities.py
 │   ├── portfolio-init.sh               Idempotent portfolio scaffolder (stdlib-only)
@@ -55,7 +57,7 @@ bash-level assertions, so they do not source it.
 
 Planned (not yet scaffolded — see the roadmap epic):
 
-- `skills/` for the backfilling recommender.
+- Cross-plugin bridges to cogni-consult and cogni-portfolio.
 
 ## Data Model
 
@@ -72,6 +74,12 @@ A portfolio is one `cogni-projects/<portfolio-slug>/` directory:
   array. Also holds `staffing-recommendations.json` — the last-run staffing
   scorer output snapshot (overwritten each run, **not** an append-only log) that
   the backfilling recommender and dashboard read.
+- `backfill-log.json` — `projects-backfill`'s own append-only log, an object
+  `{"backfills": [...]}`. `portfolio-init.sh` does **not** seed it, so that skill
+  creates it on demand with that exact envelope. It also writes
+  `backfill-recommendations.json`, its last-run snapshot, kept distinct from the
+  `projects-staff`-owned `staffing-recommendations.json` so neither skill
+  clobbers the other.
 
 ## Conventions
 
--- a/cogni-projects/README.md
+++ b/cogni-projects/README.md
@@ -4,7 +4,7 @@
 
 > **insight-wave readiness (Claude Code desktop recommended)** — Claude Code desktop is the recommended interface for insight-wave today. Cowork is a secondary path and is not yet production-ready for insight-wave workflows because of context-window and Pencil-MCP fidelity gaps — see the [deployment guide](../docs/deployment-guide.md) for detail. This guidance will flip when those gaps close upstream.
 
-Partner project-portfolio steering for consulting firms on Claude Code. cogni-projects gives partners one place to model consultants, projects, and staffing — so they can match the right people to the right work by availability, profile fit, and strategic impact. The portfolio scaffold, entity authoring, the staffing match engine, and a read-only partner-meeting dashboard have shipped; the backfilling recommender arrives in a later release.
+Partner project-portfolio steering for consulting firms on Claude Code. cogni-projects gives partners one place to model consultants, projects, and staffing — so they can match the right people to the right work by availability, profile fit, and strategic impact. The portfolio scaffold, entity authoring, the staffing match engine, a read-only partner-meeting dashboard, and the backfilling recommender have shipped; cross-plugin bridges arrive in a later release.
 
 ## Why this exists
 
@@ -20,7 +20,7 @@ Consulting partners steer a portfolio of projects and people with spreadsheets a
 
 cogni-projects is a Claude Code plugin that holds a **self-contained project portfolio**: a directory of consultants, projects, and assignments rooted by a single manifest. It is the 14th plugin in the insight-wave ecosystem and the home for partner-facing portfolio steering.
 
-It provides the `projects-setup` skill that scaffolds a portfolio directory, the `projects-entities` skill that authors and registers validated consultant/project/assignment records, the `projects-staff` staffing match engine that ranks candidate consultants per open role, and the read-only `projects-dashboard` skill that renders a partner-meeting portfolio snapshot. Backfilling builds on top of these.
+It provides the `projects-setup` skill that scaffolds a portfolio directory, the `projects-entities` skill that authors and registers validated consultant/project/assignment records, the `projects-staff` staffing match engine that ranks candidate consultants per open role, the read-only `projects-dashboard` skill that renders a partner-meeting portfolio snapshot, and the `projects-backfill` skill that steers a rolling-off consultant into their next role while ranking replacements for the one they vacate. Backfilling builds on the same scorer as staffing.
 
 ## What it does
 
@@ -28,12 +28,13 @@ It provides the `projects-setup` skill that scaffolds a portfolio directory, the
 - **projects-entities** — author one consultant, project, or assignment record and register it in the portfolio manifest, with structural validation.
 - **projects-staff** — rank candidate consultants for every open project role on availability, profile fit, and strategic impact, and write a staffing-recommendations artifact.
 - **projects-dashboard** — render a read-only partner-meeting snapshot of the portfolio: staffing coverage per project, at-risk projects, and portfolio value by strategic impact.
+- **projects-backfill** — when a consultant rolls off, rank the open roles that best fit them next and rank replacement candidates for the role they vacate, reusing the staffing scorer.
 
-_More skills (backfilling recommender) land in later roadmap releases._
+_Cross-plugin bridges to cogni-consult and cogni-portfolio land in later roadmap releases._
 
 ## What it means for you
 
-You get one durable, browsable home for your project portfolio — the foundation every staffing and steering capability writes into. Author your consultants and projects, then run the staffing engine to get a defensible ranked shortlist for each open role, then render a partner-meeting dashboard over it; later backfilling plugs into the same directory as it ships.
+You get one durable, browsable home for your project portfolio — the foundation every staffing and steering capability writes into. Author your consultants and projects, then run the staffing engine to get a defensible ranked shortlist for each open role, then render a partner-meeting dashboard over it, and when someone rolls off, backfill from the same directory — their next role and their replacement, ranked together.
 
 ## Installation
 
@@ -84,6 +85,7 @@ The manifest holds portfolio identity (`slug`, `name`, `language`, timestamps) p
 | Skill | `projects-entities` | Author + register a consultant/project/assignment record |
 | Skill | `projects-staff` | Rank candidate consultants per open project role |
 | Skill | `projects-dashboard` | Render a read-only partner-meeting portfolio snapshot |
+| Skill | `projects-backfill` | Rank next roles for a rolling-off consultant and replacements for the vacated role |
 | Script | `_projects_lib.py` | Shared loader for the sibling hyphen-named validate-entities.py |
 | Script | `portfolio-init.sh` | Idempotent portfolio scaffolder |
 | Script | `validate-entities.py` | Entity validator (frontmatter + assignment refs) |
@@ -96,11 +98,12 @@ The manifest holds portfolio identity (`slug`, `name`, `language`, timestamps) p
 ```
 cogni-projects/
 ├── .claude-plugin/plugin.json          Plugin manifest
-├── skills/                             4 portfolio skills
+├── skills/                             5 portfolio skills
 │   ├── projects-setup/                 Initialize a portfolio directory (entry point)
 │   ├── projects-entities/              Author + register one consultant/project/assignment
 │   ├── projects-staff/                 Rank candidate consultants per open project role
-│   └── projects-dashboard/             Render a partner-meeting portfolio dashboard (read-only)
+│   ├── projects-dashboard/             Render a partner-meeting portfolio dashboard (read-only)
+│   └── projects-backfill/              Rank next roles for a rolling-off consultant + replacements
 ├── scripts/                            6 stdlib-only utilities
 │   ├── _projects_lib.py                Shared loader for the hyphen-named validate-entities.py
 │   ├── portfolio-init.sh               Idempotent portfolio scaffolder
@@ -113,7 +116,7 @@ cogni-projects/
 └── references/                         Consultant / project / assignment entity schemas
 ```
 
-cogni-projects is standalone at this stage. Its portfolio directory is the shared substrate for its skills (entity authoring, staffing match, dashboard) and for later ones (backfilling), and for planned cross-plugin bridges to cogni-consult and cogni-portfolio.
+cogni-projects is standalone at this stage. Its portfolio directory is the shared substrate for its skills (entity authoring, staffing match, dashboard, backfilling) and for planned cross-plugin bridges to cogni-consult and cogni-portfolio.
 
 ## Dependencies
 
--- a/cogni-projects/skills/projects-backfill/SKILL.md
+++ b/cogni-projects/skills/projects-backfill/SKILL.md
@@ -1,0 +1,268 @@
+---
+name: projects-backfill
+description: |
+  This skill should be used when a consultant is leaving a project and the
+  partner needs both halves of the resulting gap answered: which open roles the
+  freed consultant should move to next, and who replaces them on the role they
+  vacated. Trigger on: "roll-off", "rolls off", "rolling off", "is coming off",
+  "backfill", "backfill this role", "who replaces", "find a replacement for",
+  "next assignment", "what should they work on next", "where do we put them
+  next", "fill the vacated role", "re-staff this role", "minimize bench time",
+  or any request to move a consultant between projects on a cogni-projects
+  portfolio — even if the user does not name the portfolio.
+allowed-tools: Read, Write, Edit, Bash, Glob, Grep
+---
+
+# Projects Backfill
+
+Answer a roll-off from both sides at once. When a consultant comes off a project,
+two questions open together: **where does that consultant go next**, and **who
+takes the role they left**. This skill answers both from one scorer run, so a
+partner can move people between projects with minimal bench time and defend each
+move with the same visible sub-scores the staffing engine produces.
+
+The ranking itself is not recomputed here. `scripts/staffing-score.py` — the same
+deterministic scorer `projects-staff` uses — produces every score, and this skill
+selects, filters, and reorders its output. Reusing the scorer is deliberate: a
+second ranking implementation would drift from the first, and two contradictory
+shortlists for the same portfolio is worse than none.
+
+## Core concept
+
+One scorer run yields a candidate x role matrix for the whole portfolio. A
+roll-off is two different reads of that same matrix:
+
+- **Direction (a) — the freed consultant.** Hold the consultant fixed and vary
+  the role: walk every open role in the portfolio and keep the rows where the
+  departing consultant appears as a candidate. This produces their ranked
+  next-assignment options.
+- **Direction (b) — the vacated role.** Hold the role fixed and vary the
+  consultant: take the vacated role's candidate list and drop the departing
+  consultant. This produces the replacement shortlist.
+
+On top of the scorer's own ranking this skill applies **one exclusion the scorer
+does not make**: a consultant whose `allocation_pct` is `100` is already fully
+committed and must never be proposed. Read the section below before assuming the
+scorer covers this — it does not.
+
+### What the scorer excludes, and what it does not
+
+The scorer drops a consultant from a project's ranking **only** when their
+`available_from` / `available_until` window has zero overlap with the project
+window. Those consultants never appear in `candidates[]` at all; they surface
+only as the per-role `excluded_count`, and are never proposed.
+
+Full allocation is a different matter. A consultant at `allocation_pct: 100`
+still overlaps the project window, so the scorer keeps ranking them — their
+capacity merely lowers the availability sub-score. Proposing them would be wrong.
+Apply the filter in Step 3 explicitly; do not treat `excluded_count` as if it
+already covered full allocation.
+
+The field contract for the records these read lives in
+[`../../references/data-model.md`](../../references/data-model.md).
+
+## Workflow
+
+### Step 1: Locate the portfolio and pin the roll-off
+
+Find the portfolio directory the user means. If they name a slug, use
+`cogni-projects/<portfolio-slug>/`. Otherwise glob
+`cogni-projects/*/projects-portfolio.json` and, when more than one portfolio
+exists, ask which one. If none exists, direct the user to
+`/cogni-projects:projects-setup` — this skill steers an existing portfolio, it
+does not scaffold one.
+
+Then pin three inputs: the **departing consultant** slug, the **project** they
+are leaving, and the **role** they vacate. Ask for whichever the user did not
+supply rather than guessing.
+
+Resolve each against the portfolio before running anything. When a named
+consultant, project, or role does not exist in the portfolio, report the
+unresolved name as a portfolio-data gap and stop — the fix is to author or
+correct the record with `/cogni-projects:projects-entities`, not to proceed
+against a guess.
+
+The vacated role is frequently **not** present in the project's `open_roles` at
+this point: a role only becomes open once someone declares it open, and a
+roll-off is usually what makes it open. Treat this as the expected case, not an
+error. Direction (b) needs the role to be open in order to have a candidate list,
+so instruct the user to add it to the project's `open_roles` via
+`/cogni-projects:projects-entities` and re-run; alternatively, deliver direction
+(a) alone and say plainly that the replacement half is pending that edit.
+
+### Step 2: Run the scorer
+
+Run `staffing-score.py` against the portfolio directory and capture its JSON:
+
+```bash
+python3 "${CLAUDE_PLUGIN_ROOT:-$(ls -td "$HOME"/.claude/plugins/cache/*/cogni-projects/*/ 2>/dev/null | head -1)}/scripts/staffing-score.py" "cogni-projects/<portfolio-slug>"
+```
+
+It returns `{"success", "data", "error"}` (exit 0 ok / 1 domain failure / 2 usage
+or unexpected failure). On `success: false`, surface `error` to the user and
+stop. A domain failure (exit 1) usually means the portfolio directory or its
+manifest is unreadable — but an `error` naming `validate-entities.py` instead
+means a broken or partial plugin install, not a portfolio-data problem; say so
+rather than pointing the user at their entities.
+
+On success, `data.projects[]` holds the non-closed projects, each with an
+`open_roles[]` array whose entries carry a `candidates[]` list already ranked by
+combined score and a per-role `excluded_count`.
+
+Read the scorer output only. Both `staffing-recommendations.md` and
+`.metadata/staffing-recommendations.json` belong to `projects-staff`; consult
+them for context if useful, but never write or overwrite either from this skill.
+
+### Step 3: Apply the full-allocation filter
+
+A scorer candidate carries `consultant`, `name`, and `scores` — it does **not**
+carry `allocation_pct`. Read that field from the consultant's own record at
+`cogni-projects/<portfolio-slug>/consultants/<slug>.md`.
+
+Drop every candidate whose `allocation_pct` is a number `>= 100` from both
+directions' output.
+
+Key the filter on the **raw** `allocation_pct` value, never on the scorer's
+internal headroom calculation. Headroom returns zero for a present-but-non-numeric
+`allocation_pct` as well as for a genuinely full one, so filtering on it would
+silently discard a typo as though it were a booked consultant. Instead, treat a
+present-but-non-numeric value as a data-quality finding: keep the candidate,
+and report the malformed field so someone can fix the record. An absent
+`allocation_pct` means no committed allocation is recorded — do not filter it.
+
+**The departing consultant is exempt from this filter in direction (a).** They
+are rolling off, so their recorded `allocation_pct` still reflects the assignment
+they are leaving; filtering them out would remove the one person the
+recommendation exists to place.
+
+State that exemption's limitation in the output rather than hiding it: the
+departing consultant's freed capacity is **assumed, not computed**. The scorer
+never reads `assignments/` at all, so nothing nets the vacated assignment out of
+their recorded allocation. A partner reading the recommendation should know the
+freed capacity is taken on faith.
+
+### Step 4: Direction (a) — rank the freed consultant's next roles
+
+Walk `data.projects[].open_roles[].candidates[]` across every project and keep
+the entries whose `consultant` equals the departing consultant's slug. Each
+surviving entry becomes one row: the project, the role, and that candidate's
+sub-scores.
+
+Order the rows by **`candidates[].scores.strategic_impact` descending first**
+(equivalently the project-level `strategic_impact_norm` — they carry the same
+normalized value), then `scores.combined` descending, then project slug ascending
+and role label ascending as the deterministic tiebreak.
+
+Lead with strategic impact deliberately. It is a project attribute, identical for
+every candidate within a role, so it cannot discriminate inside one role's
+shortlist — but this table compares roles *across* projects, which is the one
+ranking where it does discriminate. Sorting on `scores.combined` alone would
+silently discard the firm's own priority ordering at the exact moment it becomes
+meaningful. Render both columns so a partner can re-read the table by fit alone.
+
+Key the missing-value decision on the **raw** `data.projects[].strategic_impact`
+being null, not on the normalized score — the normalized value is `0.0` for both
+an absent value and a raw `1`, so it cannot distinguish "missing" from
+"tactical". Render an em dash (`—`) in the Strategic impact column for those
+rows; they tie at `0.0` with genuinely tactical projects, and the project-slug
+tiebreak settles the order.
+
+Exclude the role the consultant is vacating from their own next-assignment list.
+
+When the list is empty after filtering, say so and name the reason — usually no
+open role anywhere overlaps the consultant's availability window. That is a
+portfolio-data gap, not a scoring failure.
+
+### Step 5: Direction (b) — rank replacements for the vacated role
+
+Take the vacated role's `candidates[]` list. Drop the departing consultant's own
+entry, keyed on `candidates[].consultant` matching their slug. Apply the
+full-allocation filter from Step 3 — with **no** exemption here, since a fully
+committed consultant genuinely cannot take the role.
+
+Preserve the scorer's within-role ordering for the survivors; that ordering is
+already the combined-score ranking and re-sorting it would diverge from
+`projects-staff` for the same role.
+
+When the candidate list is empty after exclusions, report that explicitly rather
+than rendering an empty table, and give the count that was excluded for no
+availability overlap alongside the count dropped for full allocation — the two
+have different fixes.
+
+### Step 6: Write the backfill artifact
+
+Render both directions into
+`cogni-projects/<portfolio-slug>/backfill-recommendations.md`, and write the
+selected raw JSON alongside it at
+`cogni-projects/<portfolio-slug>/.metadata/backfill-recommendations.json`.
+
+Give that file a fixed shape so a later dashboard can read it without sniffing:
+an object with `portfolio`, `departing_consultant`, `vacated_project`,
+`vacated_role`, `next_assignments[]` (the direction (a) rows, in rendered order)
+and `replacements[]` (the direction (b) rows, in rendered order), each row
+carrying the candidate's `consultant`, `name`, and `scores` object copied
+unchanged from the scorer plus the row's `project` and `role`. Copying `scores`
+verbatim keeps the JSON and the markdown provably the same numbers.
+
+```markdown
+# Backfill — <consultant name> off <project name>
+
+## Next assignments for <consultant name> (`<slug>`)
+
+| Rank | Project | Role | Strategic impact | Availability | Profile fit | Combined |
+|------|---------|------|------------------|--------------|-------------|----------|
+| 1 | <project> | `<role>` | 0.75 | 0.84 | 0.53 | 0.70 |
+
+_Freed capacity is assumed, not computed: the recorded allocation still includes
+the assignment being left._
+
+## Replacements for `<role>` on <project name>
+
+| Rank | Consultant | Availability | Profile fit | Strategic impact | Combined |
+|------|-----------|--------------|-------------|------------------|----------|
+| 1 | <name> (`<slug>`) | 0.80 | 0.61 | 0.75 | 0.71 |
+
+_<n> excluded for no availability overlap; <m> dropped as fully allocated._
+```
+
+Show the sub-scores separately in both tables. The per-factor breakdown is what a
+partner defends the move with — never collapse it to one opaque number.
+
+### Step 7: Append a run record
+
+Record the run in `cogni-projects/<portfolio-slug>/.metadata/backfill-log.json`,
+an object with a `backfills` array. The portfolio scaffolder seeds only the
+execution, staffing, and decision logs, so this file will usually be absent on
+first run: create it with exactly `{"backfills": []}`, then read-append-write.
+Preserve that envelope on every subsequent run — never overwrite it with a bare
+array.
+
+Each record captures `portfolio`, `departing_consultant`, `vacated_project`,
+`vacated_role`, `ranked_role_count`, `replacement_candidate_count`, and
+`artifact_path`. Record no wall-clock value: the log stays reproducible.
+
+### Step 8: Summarize
+
+Report the artifact path, the top next-assignment for the departing consultant,
+and the top replacement for the vacated role. Keep it short. Call out any empty
+list and its reason, and repeat the assumed-capacity caveat once.
+
+## Notes
+
+- **The scorer is the source of truth for scoring.** This skill selects, filters,
+  and reorders; it never re-weights or recomputes. To change how candidates are
+  scored, change `scripts/staffing-score.py`, not this skill. No scoring weight
+  or constant is redeclared here.
+- **The full-allocation filter belongs to this skill, not the scorer.** The
+  scorer excludes only on zero availability-window overlap. Keeping the filter
+  here leaves `projects-staff` output unchanged.
+- **Freed capacity is assumed.** Nothing reads `assignments/` to net a departing
+  consultant's vacated allocation out of their recorded `allocation_pct`. Netting
+  live assignments is a separate capability.
+- **Deterministic output.** Identical portfolio inputs always yield the same
+  recommendation: ties break deterministically — consultant slug ascending
+  within a role (the scorer's own tiebreak, preserved in direction (b)), and
+  project slug then role label ascending across roles in direction (a) — and no
+  wall-clock value enters the artifact or the log.
+- **`projects-staff` artifacts are read-only here.** This skill writes only its
+  own `backfill-*` files.


claude-cogni-work:
## Ship-ready

Author instance done. This PR is handed off for independent review — it stays **draft** with `svc:needs-review`, and carries no `svc:review-*` label and no verdict, because the review belongs to the merger instance.

**Issue:** #1115 · **Branch:** `feat/issue-1115-projects-backfill-skill` · **Scope:** full, nothing deferred

### Handoff preflight

`check-preflight.sh` exits **0 — clean, 0 violations**. 4 passed, 3 skipped, 1 advisory.

| Instrument | Status | Reason |
|---|---|---|
| `run-tests` | pass | suites green for cogni-projects |
| `frontmatter` | pass | clean for cogni-projects |
| `breadcrumbs` | pass | clean for cogni-projects |
| `mutation-recipe` | skip — `not_applicable` | no touched path matches `*/tests/test-*.sh` or `*/scripts/check-*.sh` |
| `agent-length` | skip — `not_applicable` | no touched `agents/*.md` |
| `skill-length` | skip — **`instrument_error`** | see below |

**One skip is worth a reviewer's attention.** `skill-length` did not measure anything: the branch worktree was removed before the preflight ran, so the instrument found no file on disk (`file not found`). It is advisory, not gating, so it did not affect the exit code — but it means the length signal in this run came from nowhere. It was measured directly against the file before commit: `chars_body` **13041** vs `cap_chars` **24000**, `over_cap: false`. Re-measuring from a fresh checkout is cheap if the merger wants an independent read.

### Contract grading

No issue-time review contract exists for #1115 (`read-review-contract --target issue --origin triage` returns `found: false`), so no self-grade ran. The merger's independent review is ungraded-by-author by default here.

### Residual for the merger

- The `skill-length` instrument_error above — signal supplied out-of-band rather than by the instrument.
- The issue's acceptance criteria assert `plugin.json` "still reads `0.0.13`". That text is **stale** — it reads `0.0.24`, and both manifests agree. The binding intent (do not touch the version in a feature branch) is satisfied; the literal value in the criterion is not checkable as written.
- The prose-simplifier pass was deliberately skipped, with reasoning recorded in the PR body's Review notes.

sdh07:

_The merger formed this expectation from issue #1115 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds a prose-carried skill at cogni-projects/skills/projects-backfill/SKILL.md that orchestrates the UNMODIFIED cogni-projects/scripts/staffing-score.py (invoked with the `python3 "${CLAUDE_PLUGIN_ROOT:-...}/scripts/staffing-score.py" <portfolio-dir>` idiom copied from projects-staff/SKILL.md:79) in two directions: (a) walk data.projects[].open_roles[].candidates[] keeping entries whose `consultant` equals the rolling-off consultant's slug to produce one ranked row per open role, and (b) drop that same slug from a vacated role's candidates[] to rank replacements. The load-bearing correctness point is AC-3: the scorer does NOT exclude a fully-booked consultant - _availability_score() (staffing-score.py:186-200) returns its exclusion signal only when _overlap_fraction() is 0.0, so a consultant at allocation_pct 100 is still emitted with a positive availability sub-score (_headroom() at :169-183 merely scales it) - so the skill must apply its own explicit `allocation_pct >= 100` filter, and it must read that value from the raw consultant entity frontmatter, because the scorer's candidate objects carry only `consultant`, `name` and `scores{availability,profile_fit,strategic_impact,combined}` (:354-363) and never expose allocation_pct at all. The departing consultant must be exempt from that filter with the limitation stated in prose (freed capacity is assumed, not computed, because _load_entities is called only for consultants and projects at :309-314 and never reads assignments/). Finally the change must de-stale the docs that still declare backfilling unbuilt - CLAUDE.md:7, the skills/ tree at :16-20 and the 'Planned (not yet scaffolded' item at :56-58; README.md:7, :23, :32, :36, the '## What it does' list, the '## Components' table, the '## Architecture' tree at :99-103 and :116 - while leaving cogni-projects/.claude-plugin/plugin.json (base "version": "0.0.24") and the repo-root .claude-plugin/marketplace.json mirror (also 0.0.24) untouched.

**Acceptance checklist:**
- [ ] A new skill file exists at exactly cogni-projects/skills/projects-backfill/SKILL.md - the directory is domain-prefixed `projects-backfill`, satisfying the prefix convention CLAUDE.md:86-88 attributes to cogni-workspace/scripts/check-skill-names.sh.
- [ ] The new SKILL.md has valid frontmatter with the required keys (name, description, allowed-tools) and `name: projects-backfill` equal to its directory name - confirm by running the frontmatter gate against cogni-projects, not by eye.
- [ ] The new SKILL.md's `allowed-tools` list includes Bash (it must shell out to staffing-score.py) and Write (it must emit a recommendation artifact) - read the frontmatter line; a Read/Glob/Grep-only tool list makes the documented workflow unexecutable.
- [ ] The new SKILL.md's `description` follows the sibling trigger-phrase style - a 'This skill should be used when...' sentence plus an explicit 'Trigger on:' phrase list, as in projects-staff/SKILL.md:3-12 and projects-dashboard/SKILL.md:3-11 - and names roll-off/backfill triggers; grep the description block for at least three of: "rolls off", "backfill", "next assignment", "replace", "vacated".
- [ ] The new SKILL.md is within the skill length / body cap budget - confirm by running the skill-length measurement gate against the new file, not by eye.
- [ ] No breadcrumb is introduced: the new SKILL.md and the edited CLAUDE.md / README.md carry no issue or PR ref, version tag, or milestone/slice code - confirm with the breadcrumb gate. (This issue's '## Depends on #1114' line is a live copy-paste risk.)
- [ ] Plugin structure and the manifests stay valid and no version is bumped: confirm cogni-projects/.claude-plugin/plugin.json still reads its BASE value "version": "0.0.24" and the repo-root .claude-plugin/marketplace.json cogni-projects entry still reads "0.0.24" - both unchanged from base. (The issue AC's "0.0.13" is a stale anchor; the binding requirement is do-not-touch, so 0.0.24 unchanged PASSES this item.) Also run the structure/manifest gate.
- [ ] cogni-projects/README.md registers the new skill in both generated sections: a `- **projects-backfill** - ...` bullet under `## What it does` (alongside the four existing bullets at :27-30) and a `| Skill | projects-backfill | ... |` row in the `## Components` table (alongside :83-86).
- [ ] cogni-projects/CLAUDE.md is corrected in all three places that declare backfilling unbuilt: the header sentence at :7 ("the backfilling recommender arrives in later roadmap children") no longer says it is unbuilt; a `projects-backfill/SKILL.md` line is added to the `skills/` tree block at :16-20; and the `Planned (not yet scaffolded ...)` item at :56-58 listing "`skills/` for the backfilling recommender" is removed or narrowed to something still true.
- [ ] No surviving 'backfilling is not built yet' claim in either touched doc: grep the post-change cogni-projects/README.md and cogni-projects/CLAUDE.md for `later roadmap|later release|not yet scaffolded|arrives in|land in later|builds on top of these` and confirm no surviving hit describes backfilling as future work. Base sites that must be resolved: README.md:7, :23, :32, :36, :116 and CLAUDE.md:7, :56-58. Counts in structural blocks must also be true - README.md:99 currently reads `skills/  4 portfolio skills` and its tree lists four entries at :100-103; with a fifth skill both must be updated. (Untouched files such as projects-staff/SKILL.md:100 are pre-existing drift and are NOT graded here.)
- [ ] Scoring is reused, not re-implemented: grep the new SKILL.md body for `staffing-score.py` and confirm it invokes the scorer with the same `python3 "${CLAUDE_PLUGIN_ROOT:-...}/scripts/staffing-score.py" "cogni-projects/<portfolio-slug>"` idiom as projects-staff/SKILL.md:79.
- [ ] No second copy of the scoring constants enters the diff: grep every changed file for `W_AVAILABILITY`, `W_PROFILE_FIT`, `W_OVERLAP`, `W_HEADROOM`, `SENIORITY_NORM`, `TOKEN_OVERLAP_DISCOUNT`, `SCORE_PRECISION` - they must appear only in cogni-projects/scripts/staffing-score.py, and the new SKILL.md must not restate their numeric values as its own ranking rule.
- [ ] projects-staff is not regressed: if cogni-projects/scripts/staffing-score.py appears in the diff at all, the one-argument invocation still returns the same data shape (portfolio, weights, project_order, consultant_count, project_count, ranked_candidate_count, projects[]) and the same exit codes (0 ok / 1 domain failure / 2 usage), with the `len(argv) != 1` usage guard at staffing-score.py:402 unloosened, and the five existing suites under cogni-projects/tests/ still pass. If the script is absent from the diff this item passes trivially - say so in the grade.
- [ ] Direction (a) - freed consultant to ranked open roles - is specified concretely: grep the new SKILL.md for `open_roles` AND `candidates` AND `consultant`, and confirm it states that the skill walks data.projects[].open_roles[].candidates[] and keeps the entries whose `consultant` equals the rolling-off consultant's slug, producing one ranked row per open role. A skill that only describes direction (b) fails this item.
- [ ] Direction (a)'s cross-role ordering handles strategic_impact explicitly: grep the new SKILL.md for `strategic_impact`. Because strategic_impact is a per-project constant WITHIN a role (staffing-score.py:335, :360) but varies ACROSS roles, the skill must either order the consultant's role list using candidates[].scores.strategic_impact or projects[].strategic_impact_norm in addition to scores.combined, OR explicitly state and justify that role ranking uses combined alone (it may also legitimately state that it preserves the scorer's own project order, `strategic_impact desc, slug asc`, staffing-score.py:327 and :393). Sorting by combined with no mention of strategic impact FAILS.
- [ ] Direction (b) - vacated role to replacement candidates - excludes the departing consultant by slug: the SKILL.md states the departing consultant's entry is dropped from that role's candidates[] before ranking and names the field it keys on, `candidates[].consultant`. Grep for that field name near the exclusion sentence.
- [ ] AC-3 is an EXPLICIT filter, not the scorer's existing behavior: the SKILL.md states that consultants whose `allocation_pct` is >= 100 are excluded from both directions' recommendations. Grep the body for `allocation_pct` and `100`. A diff whose only exclusion mechanism is the scorer's per-role `excluded_count` FAILS - _availability_score() (staffing-score.py:186-200) returns its exclusion signal only when _overlap_fraction() is 0.0, so a 100%-allocated consultant is still emitted with a positive availability sub-score.
- [ ] The filter's input is named correctly and is actually obtainable: the SKILL.md says it reads the RAW `allocation_pct` from the consultant entity records (consultants/<slug>.md frontmatter), NOT from _headroom() and NOT from the scorer's candidate objects. Two independent reasons the item must show this - (i) _headroom() (staffing-score.py:169-183) also returns 0.0 for a present-but-non-numeric allocation_pct, which would silently drop data-quality cases as if fully booked; (ii) a scorer candidate carries only `consultant`, `name` and `scores{}` (staffing-score.py:354-363) and no allocation_pct at all, so an instruction to 'filter the scorer's candidates where allocation_pct >= 100' is a silent no-op. Grep the body: it must name allocation_pct as read from the consultant records, and must not name headroom as the filter input.
- [ ] The departing consultant is EXEMPT from that >= 100 filter in direction (a): the SKILL.md contains a sentence naming the departing/rolling-off consultant and allocation_pct together and saying the filter does not apply to them - otherwise a 100%-allocated departing consultant is filtered out of their own recommendation.
- [ ] That exemption's limitation is stated, not hidden: the SKILL.md records that the departing consultant's freed capacity is ASSUMED rather than computed, because staffing-score.py never reads assignments/ (_load_entities is called only for consultants and projects, staffing-score.py:309-314). Grep for `assignments` in the new skill body. A skill claiming it nets out freed capacity from assignment records FAILS.
- [ ] Availability-window respect is stated for the reused path too: grep the new SKILL.md for `excluded_count` or `available_from`, and confirm it tells the reader that consultants whose available_from/available_until window has zero overlap with the project window are already excluded by the scorer, surface only as the per-role excluded_count, and are never proposed - and (for direction (a)) that this is why a role may be missing from the freed consultant's list rather than an error.
- [ ] The backfill run does not clobber projects-staff's state: the SKILL.md writes its own artifact under cogni-projects/<portfolio-slug>/ with a filename distinct from staffing-recommendations.md, and does not write cogni-projects/<portfolio-slug>/.metadata/staffing-recommendations.json, which projects-staff/SKILL.md:96-101 owns. Grep the new skill for `staffing-recommendations` and confirm every occurrence is read-only.
- [ ] Any run record the skill appends respects the log envelope: if it writes one of the three logs portfolio-init.sh already seeds (portfolio-init.sh:86-90 - execution-log.json {"transitions": []}, staffing-log.json {"matches": []}, decision-log.json {"decisions": []}), it must instruct reading, appending to the named array, and writing back, preserving the object envelope rather than overwriting it with a bare array, as projects-staff/SKILL.md:144-149 does. If it introduces a NEW .metadata/ log filename instead, the skill must instruct creating it defensively with its exact seed shape, because portfolio-init.sh will not have seeded it.
- [ ] Degenerate and failure inputs are each handled as a distinct findable sentence: (i) the named consultant / project / role does not exist in the portfolio; (ii) the vacated role is not present in the project's open_roles - the common case immediately after a roll-off, so the skill must say to add it via /cogni-projects:projects-entities or otherwise handle the not-yet-open role; (iii) a role whose candidate list is empty after exclusions; (iv) the scorer returning success: false, including that an `error` naming validate-entities.py means a broken install rather than a portfolio-data problem (mirroring projects-staff/SKILL.md:82-88 and staffing-score.py:417-420).
- [ ] Test demand matches where the mechanism lives. If the diff adds or modifies executable code under cogni-projects/scripts/, it must add a stdlib-only bash regression suite under cogni-projects/tests/ runnable as `bash cogni-projects/tests/test-<name>.sh` in the house style of tests/test-render-dashboard.sh (no pytest, no pip, {success,data,error} envelope assertions), asserting at minimum the fully-allocated exclusion and the departing-consultant exemption. If the mechanism is carried entirely in SKILL.md prose with no executable change, a behavioral test is structurally impossible and the achievable bar is the code-shape-anchored greps above - this item PASSES, and the grader must not re-litigate test coverage for prose. Note cogni-projects has no scripts/mutation-check.sh at base, so no mutation recipe is demanded here.
- [ ] Determinism is preserved: grep the whole diff for `datetime.now`, `date.today`, `time.time` - none may be introduced into staffing-score.py or any new script - and any new ordering the SKILL.md specifies must name a deterministic tiebreak (e.g. project slug or consultant slug ascending), matching the guarantee documented at staffing-score.py:28-31 and the existing sorts at :327 and :365-367.

**Risk watchlist:**
- The likeliest wrong fix is pure reuse: transpose the scorer JSON, then declare AC-3 satisfied because 'the scorer already excludes unavailable consultants'. It does not - _availability_score() excludes only on ZERO temporal overlap (staffing-score.py:197-198); a consultant at allocation_pct 100 with an overlapping window is still ranked, merely with headroom 0.0 damping 40% of the availability sub-score.
- A subtler wrong fix writes the >= 100 filter but points it at the scorer's candidates[] objects, which carry no allocation_pct field (staffing-score.py:354-363). The instruction then reads correct and silently never excludes anyone. Watch for whether the skill says where allocation_pct is read FROM.
- Keying the filter on _headroom() == 0.0 instead of the raw field looks equivalent and is not: _headroom() also returns 0.0 for a present-but-non-numeric allocation_pct (staffing-score.py:180-181), so a typo'd record would be silently reported as 'fully booked' rather than as a data-quality problem.
- Direction (a) ranked by scores.combined alone quietly drops strategic_impact from the exact ranking where it finally discriminates - it is constant within a role, so it can only ever matter across roles, which is precisely direction (a).
- A half-measure shipping only direction (b) - the near-free slice of the existing per-role output - and leaving direction (a), the actual 'minimal bench time' value, as a single sentence.
- Forgetting the departing-consultant exemption inverts the feature: a 100%-allocated consultant rolling off is filtered out of their own recommendation and the skill returns nothing for the one person it exists to place.
- Claiming the skill nets out the departing consultant's freed capacity. It cannot: staffing-score.py never opens assignments/ (_load_entities is called only for consultants and projects, :309-314). An overclaiming sentence is worse than the stated limitation.
- Doc de-staling that stops at the two sections the AC enumerates. Base carries at least seven stale sites (README.md:7, :23, :32, :36, :99 tree count, :100-103 tree entries, :116; CLAUDE.md:7, :16-20, :56-58); a surviving 'backfilling arrives later' sentence in a file the PR already touches is a guaranteed reviewer finding.
- A new .metadata/ log file introduced without a create-if-absent instruction: portfolio-init.sh seeds exactly three logs (:86-90), so a fourth will never exist on an existing portfolio and a naive read-append-write will fail on first run.
- Overwriting .metadata/staffing-recommendations.json, which projects-staff owns as its last-run snapshot (projects-staff/SKILL.md:96-101, CLAUDE.md:69-74). Two skills writing one snapshot file is a silent data-loss regression in projects-staff, not in the new skill.
- JUDGMENT-ONLY, not gated: plugin.json's description still ends '...so the backfilling recommender builds on the same portfolio', which reads as forward-looking once backfilling exists. Description alignment is cogni-docs:doc-sync territory and the manifest must not be touched on this branch anyway (version-bump gate). A human may want to queue that separately.
- JUDGMENT-ONLY, not gated: projects-staff/SKILL.md:100 and CLAUDE.md:69-74 both describe the backfilling recommender as a future reader of staffing-recommendations.json. projects-staff/SKILL.md is not in scope for this PR; grading it would be a pre-existing-drift false violation.
- JUDGMENT-ONLY, not gated: whether a prose-only skill is the right vehicle at all versus a small filtering script. The AC sanctions either, and a prose skill is genuinely cheaper - but it means the AC-3 filter is enforced only by the model reading the instruction, with no regression test able to pin it. A human should decide whether that is acceptable for the one rule the issue calls out as the hard part.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1252

**Summary:** All four perspectives (contract-grading-reviewer, plugin-validator, skill-reviewer, skill-quality-reviewer) confirm all 26 committed review-contract items are satisfied, including the load-bearing AC-3 filter/sourcing/headroom chain; every remaining finding is a minor doc-path or prose-precision observation with no un-waived major/critical findings - safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (8)

- **[minor]** (auto-fixable) `cogni-projects/CLAUDE.md:77-82` — The backfill-log.json bullet is added as a sibling of projects-portfolio.json, consultants/, and .metadata/, so it reads as a portfolio-ROOT file. The skill actually writes it at .metadata/backfill-log.json (SKILL.md:233) and its snapshot at .metadata/backfill-recommendations.json (SKILL.md:196-197). A developer following CLAUDE.md would look in the wrong place, and the bullet is the only doc that names the log's seed envelope.
  - *Suggested fix:* Prefix the bullet path (`- `.metadata/backfill-log.json` - ...`) or fold it into the existing `.metadata/` bullet, and name `.metadata/backfill-recommendations.json` with its path too.
  - *Raised by:* plugin-dev:plugin-validator, also flagged by cogni-service:skill-quality-reviewer

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/README.md:71` — The Data-model tree annotation still reads `.metadata/  Append-only logs (execution, staffing, decisions)`. With projects-backfill shipping a fourth log plus a second snapshot in that directory the parenthetical enumeration is now incomplete. Not a gated item (checklist item 10 enumerates :7, :23, :32, :36, :99-103, :116, all of which are correctly resolved), but it is the same class of count/enumeration staleness the PR set out to fix.
  - *Suggested fix:* Extend to `(execution, staffing, decisions, backfill)` or drop the parenthetical enumeration so it cannot drift again.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/skills/projects-backfill/SKILL.md body Notes (lines 250-268)` — Four of the five Notes bullets restate points the workflow already makes in full. The rendered-caveat and summary repetitions are justified because that text must reach the artifact, but the Notes recap adds body weight without adding instruction.
  - *Suggested fix:* Trim Notes to the two bullets that state something new (Deterministic output; projects-staff artifacts are read-only here) and drop the three recaps. Leave every workflow-step statement intact - they are where the contract greps land. (ref: skill-creator SS Progressive Disclosure / SS Improving the skill)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/skills/projects-backfill/SKILL.md body Step 3 (lines 118-120)` — Step 3 names the file the raw allocation_pct is read from (consultants/<slug>.md) but not that the field lives in that record's YAML frontmatter. The skill's own correctness argument rests on reading the raw field, so the reader benefits from knowing exactly where to look rather than scanning the record body.
  - *Suggested fix:* Say 'the YAML frontmatter of the consultant's own record at cogni-projects/<portfolio-slug>/consultants/<slug>.md'. (ref: plugin-dev:skill-development SS Writing Style Requirements (objective, instructional language))

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/skills/projects-backfill/SKILL.md body Step 3 (lines 118-124)` — The filter requires an allocation_pct lookup per candidate in the vacated role's list, which on a sizable portfolio is one file read per consultant, with no guidance on doing it in a single pass.
  - *Suggested fix:* Add after the read instruction: 'Collect the values in one pass rather than one read per candidate - a single grep for allocation_pct across consultants/*.md gives every value at once, and only the candidates actually in a shortlist need one.' (ref: skill-creator SS Improving the skill (look for repeated work; keep the prompt lean))

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/skills/projects-backfill/SKILL.md:118` — Step 3 correctly sends the reader to consultants/<slug>.md for the raw allocation_pct (the item-18 requirement), but gives no guidance on scope: taken literally against a portfolio with many consultants it implies one Read per candidate per role before either table can be rendered. The instruction is correct but unbounded.
  - *Suggested fix:* Add one sentence scoping the reads, e.g. 'Read each distinct candidate slug's record once and reuse the value across roles; a single Grep for allocation_pct across consultants/ is the cheaper path when the shortlist is long.'

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/skills/projects-backfill/SKILL.md:122` — Step 3 instructs dropping every candidate whose allocation_pct is >= 100 'from both directions' output', but direction (a)'s row set consists solely of the departing consultant, who is exempted eleven lines later. The filter is therefore inert in direction (a); the 'both directions' phrasing reads as if it binds twice when only Step 5 can ever drop a row. Contract item 17 is still satisfied (it asks for exactly this both-directions statement), so this is prose precision, not a fault.
  - *Suggested fix:* Rephrase Step 3 to state where the filter actually binds, e.g. 'Apply this filter to every candidate that reaches either table. In practice it only ever drops rows in direction (b): direction (a) contains a single consultant, the departing one, who is exempt (below).'

- **[minor]** (auto-fixable) (advisory-addition) `cogni-projects/skills/projects-backfill/SKILL.md:172-174` — Checklist item 21's direction-(a) clause ('this is why a role may be missing from the freed consultant's list rather than an error') is covered only for the FULLY empty list. The partial case - the consultant appears for three of five open roles because their availability window misses the other two projects - is not framed as expected rather than a bug, even though SKILL.md:50-53 states the underlying mechanism correctly. Item graded satisfied; this is a prose-completeness observation.
  - *Suggested fix:* Add a clause to the Step 4 empty-list paragraph noting that a role absent from a non-empty list means the same thing: the scorer already excluded the consultant from that project for zero window overlap, visible as that role's excluded_count.

### Positive observations

- items 1-7 satisfied: new file at exactly cogni-projects/skills/projects-backfill/SKILL.md with domain-prefixed dir; frontmatter gate clean with name==dir; allowed-tools includes Bash and Write; description hits all five required trigger terms; length 13041/24000 chars under cap; breadcrumb grep zero hits across all three files; version-bump gate clean with plugin.json and the marketplace mirror both unchanged at base 0.0.24. (contract-grading-reviewer)
- items 8-10 satisfied: README adds both the `## What it does` bullet and the `## Components` row; CLAUDE.md is corrected in all three unbuilt-declaring places; and the post-change staleness grep shows every remaining hit describes cross-plugin bridges (genuinely still future), none describes backfilling as unbuilt - the README architecture tree count is updated from 4 to 5 skills with a matching entry. (contract-grading-reviewer)
- items 11-13 satisfied: the scorer invocation at SKILL.md:98 is byte-identical to the projects-staff idiom; no scoring constant is restated anywhere in the diff; and staffing-score.py is absent from the diff entirely, so the projects-staff non-regression item passes trivially. (contract-grading-reviewer)
- items 14-16 satisfied: direction (a) walks data.projects[].open_roles[].candidates[] keeping the departing slug; its cross-role sort leads on scores.strategic_impact with explicit reasoning; direction (b) drops the departing consultant keyed on candidates[].consultant. (contract-grading-reviewer)
- item 17 satisfied (load-bearing): SKILL.md:43-59 and :122 make the >= 100 exclusion the skill's OWN explicit filter and explicitly warn against treating the scorer's excluded_count as if it already covered full allocation. (contract-grading-reviewer)
- item 18 satisfied (load-bearing): SKILL.md:119 states the scorer's candidate objects carry no allocation_pct and instructs reading the raw field from the consultant's own record; :125-126 explicitly rejects keying on headroom. (contract-grading-reviewer)
- items 19-21 satisfied: the departing consultant is exempted by name alongside allocation_pct; the freed capacity is recorded as assumed-not-computed because the scorer never reads assignments/; and the availability-window exclusion is stated as surfacing only via excluded_count. (contract-grading-reviewer)
- items 22-23 satisfied: staffing-staff artifacts are referenced read-only and the skill writes distinctly-named backfill-recommendations.md / .metadata/backfill-recommendations.json; the new .metadata/backfill-log.json carries a create-if-absent instruction with its exact {"backfills": []} seed shape, correct because portfolio-init.sh seeds only three logs. (contract-grading-reviewer)
- items 24-26 satisfied: all four degenerate/failure inputs are distinct findable sentences including the validate-entities.py broken-install case; the prose-only mechanism makes the code-shape greps the correct bar with no behavioral test demanded; and the wall-clock grep is clean with explicit deterministic tiebreaks named for both directions. (contract-grading-reviewer)
- Item 17/18 - the load-bearing AC-3 correctness point is handled exactly right and defensively: SKILL.md:118-131 states that a scorer candidate carries only consultant, name, scores and does NOT carry allocation_pct, points the >= 100 filter at the raw field in consultants/<slug>.md, and explicitly forbids keying on headroom, spelling out the present-but-non-numeric trap and routing it to a data-quality finding that KEEPS the candidate. (plugin-validator)
- Item 19/20 - the departing-consultant exemption and its limitation are both present, and the skill states freed capacity is 'assumed, not computed' because nothing reads assignments/, rather than overclaiming capacity netting. The caveat is carried into the rendered artifact and the Step 8 summary. (plugin-validator)
- Item 15 - the cross-role strategic_impact ordering is not just mentioned but justified, with a correct explanation that the value is constant within a role and only discriminates across them. The null-vs-tactical distinction is keyed on the raw field - verified accurate against _strategic_impact_norm (staffing-score.py:249-260). (plugin-validator)
- Items 22/23 - state isolation is clean: every staffing-recommendations mention is explicitly read-only, and the new .metadata/backfill-log.json carries a create-if-absent instruction with its exact seed shape plus a preserve-the-envelope rule - the exact failure portfolio-init.sh's three-log seeding would otherwise cause on first run. (plugin-validator)
- Items 6/7/13 - gates run for real and pass: check-breadcrumbs.py 0 violations; check-skill-names.sh OK; check-version-bump.py status `ok` (NOT degraded, so non-vacuous), 14 scanned, 1 touched, 0 violations. plugin.json and the marketplace mirror both still read 0.0.24, and staffing-score.py is absent from the diff entirely. (plugin-validator)
- Item 11/12 - the scorer invocation at SKILL.md:98 is byte-identical to the projects-staff/SKILL.md:79 idiom, and no scoring constant appears anywhere in the three changed files. (plugin-validator)
- Structure and frontmatter are clean: frontmatter parses with precisely name/description/allowed-tools, name equals the directory name, the description is 664 chars (well under the 1024 cap) in the sibling style hitting all five required trigger phrases, and the body is 254 lines - the longest in the plugin but comfortably inside the cap. (plugin-validator)
- The manifest was correctly left untouched despite plugin.json's description still reading forward-looking - the contract flags that as judgment-only and the version-bump gate forbids touching it on this branch, so leaving it is the right call, not an omission. (plugin-validator)
- Item 24 - all four degenerate/failure paths are distinct findable sentences, including the vacated role not yet in open_roles framed as the EXPECTED post-roll-off case with a /cogni-projects:projects-entities remedy and a deliver-direction-(a)-alone fallback. (plugin-validator)
- Item 11/12: the scorer invocation at :98 is byte-identical to the projects-staff/SKILL.md:79 idiom, Notes at :252-255 explicitly forbid re-weighting, and no scoring constant is restated anywhere in the body. (skill-reviewer)
- Items 17-18: the AC-3 trap is defused on both axes the risk watchlist names - the candidate object carries no allocation_pct (verified against staffing-score.py:354-363), the consultant record is named as the read source, and keying on headroom is explicitly forbidden with the correct reason (headroom returns 0.0 for a present-but-non-numeric value, verified at :169-183). (skill-reviewer)
- Items 19-20: :133-142 exempts the departing consultant by name alongside allocation_pct and immediately states the limitation. Confirmed against _load_entities call sites at staffing-score.py:309-314, which cover only consultants and projects. The caveat reaches the rendered artifact and the summary step, so it reaches the partner, not just the skill author. (skill-reviewer)
- Item 15: :151-168 orders direction (a) by scores.strategic_impact desc first with an explicit justification, and correctly asserts the equivalence with strategic_impact_norm (both are the same rounded strat value, staffing-score.py:335/:360/:378). (skill-reviewer)
- Items 14/16/21: both directions are specified concretely, and :48-59 draws the sharp line between what the scorer excludes (zero window overlap, surfacing only as excluded_count) and what it does not (full allocation). (skill-reviewer)
- Items 22-23: writes only its own artifacts, marks both projects-staff artifacts read-only, and the new .metadata/backfill-log.json carries the create-if-absent instruction with envelope preservation, matching projects-staff/SKILL.md:144-149. (skill-reviewer)
- Item 24: all four degenerate cases are distinct, findable sentences, with the two exclusion counts separated because they have different fixes. (skill-reviewer)
- Item 26: every new ordering names a deterministic tiebreak, and :242 explicitly bars any wall-clock value from the log. (skill-reviewer)
- Items 3-5 and 8-10: allowed-tools includes Bash and Write; the description follows the sibling style with all the required triggers; body is 13041 chars against a 24000 cap; and the doc de-staling went past the two enumerated sections, with plugin.json and the marketplace mirror both untouched at 0.0.24. (skill-reviewer)
- Item 13/25: staffing-score.py in the worktree matches the contract's re-anchored base line numbers exactly, so the scorer is unmodified - projects-staff is not regressed and the prose-only mechanism makes the code-shape greps the correct bar. (skill-reviewer)
- Every one of the 26 review-contract checklist items is satisfied, including the four the contract's risk_watchlist flags as the likeliest wrong fixes: the filter is explicit rather than delegated to the scorer's excluded_count, sourced from the raw consultant record rather than the scorer's candidate objects, keyed off the raw field rather than headroom with the non-numeric-typo reason given, and direction (a) is fully specified rather than reduced to a sentence. (skill-quality-reviewer)
- Load-bearing factual claims about the scorer check out against the source: candidates[].scores.strategic_impact and projects[].strategic_impact_norm are assigned the same strat value (staffing-score.py:335, :360, :378); candidate objects genuinely carry only consultant/name/scores (:354-363); and the preserved within-role tiebreak is the scorer's own consultant-slug ascending (:365-367). (skill-quality-reviewer)
- The description follows the sibling house style exactly and closes with the pushy 'even if the user does not name the portfolio' clause that skill-creator recommends against under-triggering. (skill-quality-reviewer)
- The body explains the why behind every rule instead of stacking MUSTs - why the scorer is reused, why strategic impact leads direction (a)'s sort but cannot discriminate within a role, why the departing consultant is exempt, and why the exemption's limitation is surfaced rather than hidden. No CAPS-heavy imperatives anywhere. (skill-quality-reviewer)
- Progressive disclosure is handled correctly: 13041 chars against a 24000 cap, and rather than inlining the entity schemas the skill points at the existing plugin-level ../../references/data-model.md, confirmed present in the worktree. (skill-quality-reviewer)
- Output is templated per skill-creator's format pattern, and the fixed JSON shape is justified by a stated reason - a later dashboard can read it without sniffing, and copying scores verbatim keeps the JSON and markdown provably the same numbers. (skill-quality-reviewer)
- No maintainer-meta breadcrumbs in the new SKILL.md or the doc edits: the issue's '## Depends on #1114' line was correctly not copied through. (skill-quality-reviewer)
- The doc de-staling went past the enumerated AC surfaces as the plan promised; no surviving 'backfilling is not built yet' claim remains in either touched file. (skill-quality-reviewer)

### Recommendation

---

PR #1266, opened 2026-08-10
cogni-consult: read-only Swiss-ss orthography drift scan

### LINKED ISSUE #1261: cogni-consult: the ß/ss register rule loses to an engagement's own stored corpus
## Plugin
cogni-consult

## Current behavior
`references/user-facing-output.md` (a) states the German orthography rule: ß after a
long vowel or diphthong, ss after a short vowel, and *"kein schweizerisches ss an
ß-Stellen"*. It is stated as an imperative and nothing defends it.

A live run on the `it-saar-leadership-development` engagement produced **12 Swiss-ss
occurrences and zero ß** in freshly generated prose — `Messgrösse`, `heisst`,
`schliesslich`. The engagement's own stored corpus is uniformly Swiss: the files
carrying it (`scope/key-question.md`, `action-fields/diagnostic-as-is/befragungsbefund.md`,
`action-fields/wirkungsmodell/field.json`, `.metadata/decision-log.json`) all pre-date
the run. The fresh turns mirrored the stored text rather than the rule.

The failure is engagement-scoped, not global: other engagements in the same workspace
are orthographically correct — `cogni-work-model-sharpening` carries 257 ß against 13
Swiss forms.

## Why this is worth fixing
In-context engagement text beats a reference paragraph. Once an engagement's stored
corpus is wrong, every subsequent turn reinforces it, because the model reads the
corpus at Step 0 and matches its register. A stated rule that loses to its own data
is not enforced — and the register exists precisely to be enforceable on the default
path.

This is independent of the output-style collapse: the same rule was stated in the same
imperative form in the deleted `strategy-advisor-de.md`, so the defect pre-dates it.

## Desired behavior
Options, in ascending cost:

1. **Detect and report.** A check script that scans an engagement's stored files for
   Swiss forms in ß positions and reports them, run from `consult-resume` Step 0 or as
   a standalone lint. Cheapest; surfaces the drift without touching generation.
2. **Detect and offer repair.** The same scan plus a one-shot rewrite pass over the
   offending files, so a contaminated corpus stops reinforcing itself.
3. **Strengthen the rule's placement.** State the orthography rule where it is read
   *after* the corpus rather than before — currently (a) is read at Step 0 and the
   engagement files are read later in the same turn.

Option 1 alone would have caught this instance. Options 1+2 are the durable fix; 3 is
speculative until 1 shows how often it fires.

## Motivation
Found while grading #1146 (PR #1258) with the collapsed register live on the plugin
path. Recorded there as a non-blocker because it pre-dates the collapse.

_Affected components: cogni-consult/references/user-facing-output.md, cogni-consult/scripts/_

_Complexity hint: small_

## Acceptance criteria

_Derived by `review-contractor` against the base at grooming time and ratified via `/cogni-service:service-groom`. Scope is **Option 1 (detect and report)** as this issue's own cost ordering proposes: Option 1 alone would have caught the observed instance, Option 2's repair pass is a separate consultant-elected invocation, and Option 3 is explicitly deferred as speculative until Option 1 shows how often the drift fires._

**The framing that shapes the bar.** The ß/ss rule is currently stated three times and enforced zero times — `references/user-facing-output.md:21-27`, `hooks/on-subagent-start.sh:63-70`, and `cogni-workspace/hooks/on-session-start-language.sh:58-65` all assert it, while nothing ever reads an engagement's stored corpus back. **The hard part is precision, not recall.** "ss where ß belongs" is undecidable from orthography alone: `müssen`, `Schlüssel`, `Ausschuss` are correct ss, while `Messgrösse` is a compound whose Swiss form sits at the tail and whole-word matching misses it. A noisy detector the consultant learns to ignore is a worse outcome than the unenforced rule.

**The detector**

- [ ] A new check script exists under `cogni-consult/scripts/` whose header states purpose, a `Usage:` line, and `Output: JSON {"success": bool, "data": {...}, "error": "string"}` — the shape `scripts/assumption-change-frequency.sh` already models.
- [ ] It is stdlib-only and offline: no non-stdlib import, no pip dependency, no network call.
- [ ] All three envelope keys are emitted on **every** exit path, including the error path.
- [ ] It takes an engagement path argument and scans only beneath it. The suite includes a sibling engagement directory carrying Swiss text and asserts it contributes zero findings — the issue's central observation is that the failure is engagement-scoped.
- [ ] It is read-only: the suite asserts fixture contents (or a checksum) are unchanged after a scan.
- [ ] Findings are data, not an error: a scan that finds drift returns `success: true` with findings under `data` and an empty `error`. The exit-code contract is stated in the header, and whichever contract the header declares is what the code and any caller implement.
- [ ] Fail-loud on a bad target: a missing, non-directory or unreadable engagement path returns `success: false` with a non-empty `error` — never a silent empty result that reads as "no drift".
- [ ] Files are opened with an explicit `encoding="utf-8"`, and the lexicon/pattern constant uses literal umlauts and ß, never ASCII substitutes.

**Precision — the load-bearing half**

- [ ] The detection vocabulary is explicit and reviewable in **one named place**: a module-level lexicon list or pattern constant with a comment stating the false-positive policy — why these forms and not a bare `[aeiouäöü]ss` regex. Detection logic scattered across ad-hoc inline branches fails.
- [ ] True positives: the suite proves the scan flags all three observed forms — `Messgrösse`, `heisst`, `schliesslich`. **`Messgrösse` is load-bearing**: it is a compound whose Swiss form sits at the tail, so a whole-word-only lexicon fails this item.
- [ ] Precision: a fixture containing at least `dass`, `muss`, `Prozess`, `müssen`, `Schlüssel` and `Ausschuss` (the last a morpheme boundary after a diphthong) yields **zero** findings, asserted as an explicit empty result rather than as the absence of a crash.
- [ ] The scanned-set boundary is documented in the script header: which engagement subtrees are scanned and which are excluded and why, explicitly covering `sources/` (verbatim third-party material that may legitimately be Swiss) and generated artifacts (`output/`, `README.md`, `assumptions.md`). Either decision passes; silence on the boundary fails.

**Carrier coverage and report content**

- [ ] The suite proves a finding is reported from a `.md` carrier (e.g. `scope/key-question.md`) **and** from a `.json` carrier (`action-fields/<slug>/field.json` or `.metadata/decision-log.json`) — the four file classes this issue names as already contaminated. A markdown-only scanner fails.
- [ ] Each finding carries the engagement-relative path, a line number, the offending token, and the corrected ß form; `data` carries a total count, asserted against a fixture with a known expected count — not merely `success: true`.

**Registration and suite contract**

- [ ] The script is registered in **all four** documentation surfaces: a row in the `## Scripts` table of `cogni-consult/CLAUDE.md`, a `| ... | Script | ... |` row in the `## Components` table of `cogni-consult/README.md`, and a line under the `scripts/` block of each file's Architecture tree. A missing one fails — this is how the script inventory drifts.
- [ ] A suite `cogni-consult/tests/test_<name>.sh` satisfies the runner's three-property contract: exits non-zero on any assertion failure, runs as `bash <path>` with no arguments from any cwd (self-locating via `TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"`), needs no network or credentials, and writes only under its own `mktemp -d`.
- [ ] It sits exactly one level under `cogni-consult/tests/` so the non-recursive glob discovers it — `python3 scripts/run-plugin-tests.py --list` returns it in `data.suites[]`.
- [ ] **Mutation proof.** Because this mechanism lives in an executable script, a behavioural falsifier is demanded rather than a grep: `cogni-consult/scripts/mutation-check.sh` gains a `run_mutation <label> <description> <python-re-program>` entry that neuters the detector's core (e.g. the compound/substring match) and the run reports `OK: mutation caught`. Note for the resolver: this repo's harness takes a python `re.subn` program that must report exactly one substitution — it has **no** `--expr` / `perl -0pi` form, so do not write one.

**Rule statements stay in step**

- [ ] `references/user-facing-output.md` section (a) still states the rule verbatim — grep for `Kein schweizerisches ss an ß-Stellen`. The file may gain a pointer to the new check, but the rule paragraph must not be deleted or thinned.
- [ ] If the ß/ss sentence's wording is touched anywhere, **all three copies** stay in step: `references/user-facing-output.md:21-27`, `hooks/on-subagent-start.sh:63-70` (`LANGUAGE_RULES["de"]`), and `cogni-workspace/hooks/on-session-start-language.sh:58-65` (`RULES["de"]`). A one-sided edit fails.

**Boundaries and gates**

- [ ] If `skills/consult-resume/SKILL.md` is touched, its **read-only contract** (base lines 368-371) is either unchanged or amended to match — never silently contradicted — and `bash cogni-consult/tests/test_step0_register_block.sh` still exits 0. Any new Bash call's `description` obeys the five constraints of section (f): resolved interaction language, outcome-shaped, at most 6 words, no script/file/skill names, not derived from the filename.
- [ ] Option 2 is **not** implemented as an auto-firing pass. If a repair capability lands at all, it is a separate consultant-elected invocation that never fires on re-entry — an unreviewed mass edit of the consultant's stored corpus is the failure mode to avoid.
- [ ] Option 3 (relocating or re-emphasising the rule) is **not** substituted for Option 1. A reference-prose-only diff does not clear this bar.
- [ ] `python3 scripts/run-plugin-tests.py --filter consult` exits 0 with `data.failed: 0` — the new suite passes and none of the existing `cogni-consult/tests/*.sh` suites regresses.
- [ ] No new maintainer breadcrumb (no `#1261`, version tag, milestone or finding code) in any touched `*/skills/*/SKILL.md` or `*/agents/*.md`; `python3 scripts/check-breadcrumbs.py` exits 0 with no new baseline entry.
- [ ] No plugin version is touched — `python3 scripts/check-version-bump.py` reports `data.clean: true`.

**Known open risk, recorded deliberately**

- [ ] This issue's own theory is that in-context corpus beats a stated rule. Detection alone therefore may not change generation behaviour — a report the model never reads at generation time does not fix the register. Option 1 is scoped as the **evidence-gathering** step that shows how often the drift fires; whether it also changes the outcome is not a criterion here and must not be graded as one.

claude-cogni-work:
Closes #1261.

A stated rule that loses to its own data is not enforced. The German orthography rule in `references/user-facing-output.md` §(a) is imperative prose, and on a live engagement whose stored corpus was uniformly Swiss, fresh turns mirrored the corpus (12 Swiss-ss forms, zero ß). Stored prose arrives as in-context evidence while the rule is read once, so the drift reinforces itself silently. This ships the Option-1 half of the fix — detect and report — so a wrong corpus becomes visible instead of self-perpetuating.

## Summary

- Adds `cogni-consult/scripts/orthography-drift-scan.py`: read-only, stdlib-only, engagement directory as the first and only positional argument, single-line JSON envelope on every exit path. Reports Swiss-ss spellings sitting in ß positions with engagement-relative path, line number, matched form and suggested ß replacement, plus an aggregate total and a per-file breakdown.
- Detection is a **curated-list heuristic with bounded recall**, stated in those words in the module docstring, in the emitted `detection` field, and in the CLAUDE.md row. `SWISS_PAIRS` is the single machine-readable home of the pair list (12 entries incl. `Grösse→Größe`, `heisst→heißt`, `schliesslich→schließlich`, `Strasse→Straße`, `weiss→weiß`). Matching is stem/substring and **not** whole-word-anchored, so `Messgrösse` reports from the bare `Grösse` entry; each entry also matches its sentence-initial variant, so `heisst` and `Heisst` both report.
- Correct short-vowel ss cannot match — `dass`, `muss`, `Prozess` are simply never entries — and the ambiguous homographs `Masse` and `Busse` are deliberately excluded, because spelling alone cannot tell `Masse` (mass) from `Maße` (measurements).
- Markdown is scanned whole. JSON is reached only through named prose keys (`title`, `framing`, `name`, `rationale`, `decision`, `question`, `summary`, `intent`, `key_question`), so engine tokens (`state`, `dt_stage`, slugs, ids, `kind`, `verdict`) are never read. Recursion covers `.md` and `.json` including dot-directories — only `.git` is skipped — so `.metadata/decision-log.json` is covered.
- Adds `cogni-consult/tests/test_orthography_drift_scan.sh` — the plugin's 11th suite, house format (`set -u` only, `OK   %s` / `FAIL %s: %s`, an `assert_envelope` helper), with heredoc fixtures because no engagement corpus is committed to the repo.
- Adds one row to the `cogni-consult/CLAUDE.md` Scripts table and one pointer paragraph in `references/user-facing-output.md` §(a) directly after the orthography paragraph. **The rule's own sentences are byte-unchanged** and stay in §(a); the diff to that file is additions only.

## Two defects caught in review, before commit

Both are silent-recall failures — the worst shape for a lint, because they hide inside a successful zero-finding report rather than erroring.

**Escaped JSON lost a third of the pair list.** The scan validated JSON with `json.loads` and then matched against the *raw* text, so a `field.json` written the way `json.dump` writes it by default (`ensure_ascii=True`) carried `Grösse` and matched nothing. Reproduced: a value reading `"Wir gehen schliesslich vor, gemäss der Grösse."` reported only `schliesslich`, silently dropping all four non-ASCII entries. Each captured value is now JSON-unescaped before matching. Line numbers still resolve, anchored on where the value starts — exact, because a JSON string cannot contain a literal newline. Pinned by case 8b.

**The generated-echo exclusion was inverted.** It began as three hardcoded engagement-relative paths, which excluded a `README.md` *whether or not it was generated*. But `generate-engagement-readme.py:39-41` states the opposite semantics outright — "a root README.md without this marker is hand-authored and is never overwritten" — and both generators key their overwrite guards on that footer sentinel. So the path list blinded the scan to precisely the file a consultant owns and would edit. Exclusion now tests for the sentinel, matching what the generators already treat as authoritative, and case 8 asserts a marker-less file is still scanned.

## Envelope contract

- Clean corpus → `success: true`, `data.total_findings: 0`.
- Drift found → `success: true`, non-zero `data.total_findings`, exit 0. **Drift is a successful scan, never `success: false`** — consumers branch on the count, never the exit code.
- Missing path → `success: false`, non-empty `error`, `data.failed_check: "engagement_missing"`. Not a directory → `not_a_directory`. No argument → `usage`. Any extra argument, `--fix` included → `unexpected_argument`.
- `argv` is hand-parsed rather than handed to `argparse`, so no exit path can print usage text to stdout instead of the envelope, and `main()` is wrapped so an unexpected exception still emits one JSON object (`unexpected_error`).
- `ensure_ascii=False` on emission, mirroring `assumption-change-frequency.py:60-65`, so reported strings are literally `Grösse` / `Größe` — no ASCII substitutes, no mojibake.

## Strictly read-only

No `--fix` / `--repair` / `--rewrite` / `--write` / `--apply` flag exists, every file is opened in `"r"` mode only, and nothing is written under the engagement root. A suite case digests the whole fixture tree before and after a drift-reporting scan and asserts byte-identity — with the baseline captured before *any* scan runs, so the assertion is actually load-bearing rather than comparing two post-scan states.

## Recorded teeth

The suite carries its own mutate-and-assert-red block (prior art: `tests/test_step0_register_block.sh:178-204`). It copies the script and mutates the copy two ways, requiring for each that the mutation actually applied (a no-op regex fails the case rather than passing it) and that the mutant still emits a valid `success: true` envelope, so a crash is never mistaken for detection:

1. `SWISS_PAIRS` emptied between its sentinel comments → the drift fixture drops to `total_findings: 0` while the pristine script reports non-zero.
2. The matcher anchored whole-word → `Messgrösse` is no longer reported from the bare `Grösse` entry.

Both were **run**, not merely written: `bash cogni-consult/tests/test_orthography_drift_scan.sh` exits 0 with all 21 assertions `OK`, the two goes-red cases included, so CI records the teeth on every run.

No `scripts/mutation-check.sh` recipe is offered, and that is deliberate rather than an omission: that harness hardwires `TARGET=_discover_extractor.py` / `SUITE=test_discover_extractor.sh` at lines 32-33 and reads them as globals, so adding a second target means restructuring a shared harness this issue does not scope.

## Scope decisions

- **In:** the scan script, its suite, the CLAUDE.md row, the reference pointer.
- **Out — Option 2** (detect *and* repair). It introduces the first write path over an engagement's stored prose and needs its own idempotency, confirmation and rollback review; the issue itself gates it on what Option 1's frequency data shows.
- **Out — Option 3** (relocating or restating the rule text). Its original target was wrong anyway: `consult-resume` Step 0 resolves the interaction language and reads no engagement file. The corpus enters context via `consult-design-thinking`, `consult-personas`, `consult-publish` and `consult-scope`.
- **No SKILL.md is wired.** All nine `consult-*` skills already carry `Bash`, so the standalone lint is invocable today without a behaviour change to any skill body, and no honest justification for a Step-0 call site exists.
- **The rule's twin delivery path is untouched.** `hooks/on-subagent-start.sh:64-68` keeps its `LANGUAGE_RULES["de"]` copy byte-for-byte, because the rule text does not change — the added paragraph is operational guidance, not doctrine, so the one-contract-two-delivery-paths rule at `CLAUDE.md:147` stays satisfied.
- **No version touched** in `plugin.json` or `marketplace.json` — the post-merge workflow owns the bump.

## Known limitations

- **Not language-gated.** `consult-project.json.language` records the engagement's output language, and the scan does not read it, so on an `fr`/`it`/`pl`/`en` engagement it would apply German pair matching to non-German prose (`Strasse` and `weiss` can appear as proper nouns). Gating on that field is a behaviour change beyond the triaged two deliverables and is left for a follow-up; the motivating engagement is German.
- **JSON recall is bounded twice** — by `SWISS_PAIRS` and again by `PROSE_KEYS`. A prose key added to the data model later is silently unscanned. The emitted `detection` string names both bounds so a reader is not misled about what a clean report proves.
- **No committed corpus to regress against.** Nothing in CI can test against the real engagement text, so the suite's synthetic fixtures are the only evidence detection works.
- `cogni-consult/README.md`'s Components table would take a row for this script, but that section is `cogni-docs` `doc-generate`-owned; hand-editing it creates drift the generator would overwrite, so `doc-audit` should pick it up as ordinary drift.

## Test plan

- [x] `bash cogni-consult/tests/test_orthography_drift_scan.sh` — 21/21 `OK`, exit 0, including both goes-red cases and the read-only byte-identity case.
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-consult` exits 0; discovered suite count rises 10 → 11 and all 11 pass.
- [x] `check-frontmatter.sh cogni-consult` → `clean: true`, 0 offenders (clean at base too, so nothing introduced).
- [x] Breadcrumb guard → clean, 0 newly introduced refs.
- [x] `python3 scripts/check-version-bump.py` → 14 plugins scanned, 0 touched, 0 violations vs `origin/main`.
- [x] `git diff origin/main -- '*/.claude-plugin/plugin.json' '.claude-plugin/marketplace.json'` is empty.
- [x] `git diff origin/main -- cogni-consult/references/user-facing-output.md` shows additions only; lines 21-27 byte-unchanged, new paragraph before `## (b)`.
- [x] `git diff origin/main -- cogni-consult/hooks/on-subagent-start.sh` is empty.
- [x] `bash cogni-consult/tests/test_step0_register_block.sh` still green after the reference edit — its `owner-present` assertion greps three literal needles in that file, all still present.
- [x] Every exit path pipes cleanly into `json.load`: drift, clean, missing-path, not-a-directory, no-argument and `--fix` each emit exactly one JSON line.
- [x] `grep -nE "open\(.*['\"][wa]"` on the script finds nothing.


claude-cogni-work:

Author instance done. Draft PR carrying `svc:needs-review`; no ready flip and no verdict label, so the merger's independent review is the first grading of this diff.

**Handoff preflight:** exit 0, clean — 1 passed, 0 failed, 6 skipped.

|---|---|---|
| `run-tests` | skip | `no_suites_discovered` for cogni-consult — **see the instrument note below** |
| `frontmatter` | skip | no touched SKILL.md or `agents/*.md` |
| `breadcrumbs` | skip | no touched SKILL.md or `agents/*.md` |
| `skill-length` | skip | no touched SKILL.md |
| `agent-length` | skip | no touched `agents/*.md` |
| `mutation-recipe` | skip | no touched `*/tests/test-*.sh` or `*/scripts/check-*.sh` |

**Instrument note — the `run-tests` skip is a blind spot, not evidence.** `run-tests.sh` discovers `tests/test-*.sh` (hyphen), and all 11 cogni-consult suites are `test_*.sh` (underscore), so it structurally cannot see them and reports `no_suites_discovered` — including the claim that "the base ref carried no suites", which is false: the base carries 10. The repo's own CI runner globs `tests/*.sh` and does run them, so the substance is covered there, not here. Verified independently on this head: `python3 scripts/run-plugin-tests.py --filter cogni-consult` exits 0 with 11 suites discovered (base 10) and all 11 passing. Please do not read that skip as a green regression check.

**A first preflight run false-failed** `version-bump` with 1 violation. Cause was `--root` pointing at the shared clone, which is checked out on `feat/issue-1116-projects-dashboard` — an unrelated in-flight branch that bumps four plugin versions. Re-run against a root on this branch: clean. This diff touches 4 files and zero version manifests.

Graded against the 22-item acceptance contract posted to #1261 at triage, before any branch was cut. **21 satisfied, 1 violated, 0 unclear.**

The violation was **item 16**: `data.by_file` was produced by the script but never read by any test case — an aggregate total cannot express "12 occurrences across 4 named files", which is the shape the evidence takes. Fixed in `08b5bedb` with an assertion that the breakdown matches the findings list exactly and sums to `total_findings`. Suite re-run: 22/22 `OK`, exit 0. Preflight re-run on the new head: still clean.

**Residual: none.** No `unclear` items, no non-mechanical violations, nothing a fix pass failed to land.

Two items where the implementation deviates from the contract's letter, both graded satisfied on intent — flagging them so the merger reviews the reasoning rather than inheriting it:

- **item 20 (recorded teeth)** offered a `mutation-check.sh` recipe or a documented side-command. The goes-red mutations instead live *inside* the new suite as case 13, following this plugin's own prior art at `tests/test_step0_register_block.sh:178-204`, so they run on every CI invocation rather than as a command someone must remember. `mutation-check.sh` was left untouched because it hardwires `TARGET`/`SUITE` at lines 32-33 and is not parameterizable; generalizing a shared harness is unscoped here.
- **item 15 ("recursive over both .md and .json")** is satisfied as directory-walk recursion. JSON *content* is deliberately reached only through named prose keys, because a whole-file JSON text scan false-positives on engine tokens (`state`, `dt_stage`, slugs).

## Two defects the review caught before commit

Both are silent-recall failures that hide inside a successful zero-finding report:

1. **Escaped JSON lost a third of the pair list.** The scan matched raw text after validating with `json.loads`, so a `field.json` written with the default `ensure_ascii=True` carried `Grösse` and matched nothing. Reproduced: a value reading `"Wir gehen schliesslich vor, gemäss der Grösse."` reported only `schliesslich`. Values are now JSON-unescaped before matching; pinned by case 8b.
2. **The generated-echo exclusion was inverted.** A hardcoded path list excluded `README.md` whether or not it was generated, while `generate-engagement-readme.py:39-41` states the opposite — a marker-less README "is hand-authored". So the list blinded the scan to the file a consultant actually owns. Exclusion now tests the generators' own footer sentinel.

## Known limitations, stated rather than hidden

- **Not language-gated.** `consult-project.json.language` exists and the scan does not read it, so on a non-German engagement it would apply German pair matching to non-German prose. Gating is a behaviour change beyond the triaged two deliverables.
- **JSON recall is bounded twice** — by `SWISS_PAIRS` and again by `PROSE_KEYS`. The emitted `detection` string names both so a clean report is not over-read.
- **No committed corpus** to regress against; the synthetic fixtures are the only evidence detection works.
- `cogni-consult/README.md`'s Components table would take a row, but that section is `doc-generate`-owned — hand-editing it creates drift the generator overwrites, so `doc-audit` should pick it up as ordinary drift.

sdh07:

_The merger formed this expectation from issue #1261 before checking out the PR head — the review grades the diff against it._

**Expected approach:** Option 1 (detect and report) only: one new read-only, stdlib-only scan script under `cogni-consult/scripts/` (base dir holds 14 files and no `check-*` scanner) plus one new suite under `cogni-consult/tests/` (base holds exactly 10 `test_*.sh` suites). The script must take an engagement directory as an argument, walk both `.md` and `.json` carriers beneath it — including dot-directories, since `.metadata/` is a first-class engagement subtree per `cogni-consult/references/data-model.md:15-36` — match a single named, reviewable curated Swiss→ß lexicon (a bare `ss` regex is all false positives: `dass`, `muss`, `Prozess` are the rule's own correct-ss examples at `cogni-consult/references/user-facing-output.md:21-27`), and emit the house single-line envelope `{"success", "data", "error"}` on every exit path, mirroring the docstring shape of `cogni-consult/scripts/assumption-change-frequency.py:2-33`. The suite must follow the house template of `cogni-consult/tests/test_assumption_change_frequency.sh:22-58` (`set -u`, self-locating `TESTS_DIR`, `mktemp -d` + `trap ... EXIT`, per-case counter, non-zero exit) and synthesize its own fixtures, since no engagement corpus is committed. Registration must land on all four documentation surfaces the AC names — the `## Scripts` table at `cogni-consult/CLAUDE.md:167-178`, the Architecture tree `scripts/` block at `cogni-consult/CLAUDE.md:80-104`, the `## Components` table at `cogni-consult/README.md:142-168`, and the Architecture tree `scripts/` block at `cogni-consult/README.md:214-217` — and a falsifier must be added to `cogni-consult/scripts/mutation-check.sh`, whose `run_mutation` currently hardwires `TARGET`/`SUITE` at lines 32-33 and takes a python `re.subn` program (header at :37-42), not an `--expr`. The rule text itself does not move (Option 3 out of scope), nothing writes to the corpus (Option 2 out of scope), and the three existing rule restatements — `references/user-facing-output.md:21-27`, `cogni-consult/hooks/on-subagent-start.sh:63-70`, `cogni-workspace/hooks/on-session-start-language.sh:58-65` — stay in step.

**Acceptance checklist:**
- [ ] A new scan script exists under `cogni-consult/scripts/`; its header/docstring states the purpose, a `Usage:` line, and an `Output:` line naming the envelope `{"success": bool, "data": {...}, "error": "string"}` — the two-liner shape `cogni-consult/scripts/assumption-change-frequency.py:2-33` models. Grep the added file for `Usage:` and `Output:`; both present or the item fails.
- [ ] The script is stdlib-only and offline: every `import` in the added script resolves to the Python standard library (no `requests`, `yaml`, `regex`, `chardet`), the diff adds no pip/requirements step, and the file contains no network call (`urllib.request`, `socket`, `http.client`, `curl`, `wget`).
- [ ] The script header documents the scanned-set boundary in words: which engagement subtrees are scanned and which are excluded and why, explicitly naming `sources/` (the verbatim third-party inbox at `cogni-consult/references/data-model.md:35-36`, which may legitimately be Swiss) and the generated artifacts (`output/`, the engagement-root `README.md`, `assumptions.md`). Either decision passes; silence on the boundary fails.
- [ ] The script header states in words that detection is a curated-list heuristic with bounded recall — a zero-finding report is not proof the corpus is ß-correct. Grep the header for the recall caveat; absent means a caller can read a clean scan as a guarantee.
- [ ] All file reads pass an explicit `encoding="utf-8"`, and the lexicon/pattern constant is written with literal umlauts and ß — grep the added script for ASCII substitutes in the pair list and for any replacement value spelled without `ß`; any ASCII substitute in the pair list fails.
- [ ] The detection vocabulary lives in exactly ONE named module-level constant (a lexicon list or pattern constant), carrying a comment stating the false-positive policy — why these forms and not a bare `[aeiouäöü]ss` regex. Evidence: grepping a distinctive member such as `schliesslich` across all added non-test lines returns hits in exactly one file; detection logic scattered across ad-hoc inline branches, or a second human-readable copy of the list in `CLAUDE.md` / `references/user-facing-output.md`, fails.
- [ ] A suite `cogni-consult/tests/test_<name>.sh` exists exactly one level under `cogni-consult/tests/` so the runner's non-recursive glob (`scripts/run-plugin-tests.py` module docstring, lines 33-48) discovers it, and it satisfies the three-property contract: self-locating via `TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"`, runs as `bash <path>` with no arguments from any cwd, exits non-zero on any assertion failure, needs no network or credentials, and writes only under its own `mktemp -d`. Evidence: `python3 scripts/run-plugin-tests.py --list` returns the new path in `data.suites[]` and cogni-consult's suite count rises from 10 to 11.
- [ ] REGISTRATION SURFACE 1 of 4 — the `## Scripts` table of `cogni-consult/CLAUDE.md` (base rows at :167-178) gains exactly one new row naming the new script. Scoped to the new script only: the pre-existing absence of rows for `dt-stage-advance.sh`, `schedule-edit.py`, `mutation-check.sh` and `generate-dashboard.py` is base drift and must NOT be graded as a violation.
- [ ] REGISTRATION SURFACE 2 of 4 — the Architecture tree `scripts/` block of `cogni-consult/CLAUDE.md` (base :80-104) gains a line naming the new script in the same two-column tree style as the surrounding entries.
- [ ] REGISTRATION SURFACE 3 of 4 — the `## Components` table of `cogni-consult/README.md` (base rows at :142-168) gains one `| <script-name> | Script | <description> |` row for the new script. Verified against base: `cogni-consult/README.md` carries no generated-file marker and no generator in this repo owns that section (no `cogni-docs` plugin exists here — `*/.claude-plugin/plugin.json` enumerates 14 plugins, none of them cogni-docs), so a hand edit is not overwritten and the AC's demand stands.
- [ ] REGISTRATION SURFACE 4 of 4 — the Architecture tree `scripts/` block of `cogni-consult/README.md` (base :214-217) names the new script. That block is a compressed summary rather than a per-script listing, so either extending the summary text or adding a line satisfies it; leaving the block silent about the new script fails.
- [ ] MUTATION PROOF — because this mechanism lives in an executable script, a behavioural falsifier is demanded rather than a grep: `cogni-consult/scripts/mutation-check.sh` gains a `run_mutation <label> <description> <python-re-program>` entry that neuters the detector's core (e.g. empties the lexicon or anchors the matcher whole-word), and the run prints `OK: mutation caught`. The expression form is the one that harness's own header documents at `cogni-consult/scripts/mutation-check.sh:37-42` — a python `re.subn` program that must report exactly one substitution (`n != 1` fails loud at :65-67); this repo-local harness has NO `--expr` form. Because `TARGET`/`SUITE` are hardwired at :32-33, a second target requires parameterizing `run_mutation`.
- [ ] `bash cogni-consult/scripts/mutation-check.sh` still exits 0 and still prints `All mutations caught.` — the two existing `_discover_extractor.py` falsifiers must not regress when the harness is parameterized.
- [ ] The script writes exactly one JSON object to stdout on EVERY exit path with no human-readable preamble, and all three envelope keys are present on every path including the error path. Evidence: `<invocation> | python3 -c 'import json,sys; json.load(sys.stdin)'` succeeds for the drift-found run, the clean run, and the bad-target run alike, and each parsed object has `success`, `data` and `error` keys.
- [ ] Findings are data, not an error: a scan that finds drift returns `success: true` with the findings under `data` and an empty `error` — pinned by a test case. The exit-code contract (whether drift sets a non-zero exit or only a count) is stated in the script header, and whatever the header declares is what the code and the suite implement; a header/behaviour mismatch fails.
- [ ] Fail-loud on a bad target: a missing path, a non-directory path and an unreadable path each return `success: false` with a non-empty `error` (or the house `data.failed_check` used at `cogni-consult/tests/test_assumption_change_frequency.sh:47-58`) — never a silent empty result that a caller reads as "no drift". Pinned by test cases, at minimum the missing-path and non-directory cases.
- [ ] The script takes an engagement path argument and scans only beneath it: the suite builds a SIBLING engagement directory carrying Swiss text and asserts it contributes ZERO findings — the issue's central observation is that the failure is engagement-scoped, not global.
- [ ] The script is strictly read-only over the scanned tree: it exposes no `--fix`, `--repair`, `--rewrite`, `--write` or `--apply` flag, contains no write/replace/rename call targeting a path under the engagement root, and the suite asserts the fixture tree is byte-identical (per-file checksum or `diff -r` against a copy) after a scan that reported findings.
- [ ] TRUE POSITIVES — the suite proves the scan flags all three live-observed forms `Messgrösse`, `heisst` and `schliesslich`. `Messgrösse` is load-bearing: it is a compound whose Swiss form sits at the tail, so a whole-word/`\b`-anchored lexicon fails this item; the matcher must be stem/substring-based and the test case must assert the compound is reported.
- [ ] PRECISION — a fixture containing at least `dass`, `muss`, `Prozess`, `müssen`, `Schlüssel` and `Ausschuss` (the last a morpheme boundary after a diphthong) yields ZERO findings, asserted as an explicit empty/zero result read out of the envelope, not merely as the absence of a crash.
- [ ] CARRIER COVERAGE — the suite proves a finding is reported from a `.md` carrier (e.g. `scope/key-question.md`) AND from a `.json` carrier (`action-fields/<slug>/field.json` or `.metadata/decision-log.json`), the file classes the issue names as already contaminated. A markdown-only scanner, or a walk that skips dot-directories and so never opens `.metadata/`, fails.
- [ ] REPORT CONTENT — each finding carries the engagement-relative path, a line number, the offending token and the corrected ß form, and `data` carries a total count asserted against a fixture with a known expected count. Evidence: a test that reads those exact fields (and the count) out of the envelope, not merely `success: true`.
- [ ] `cogni-consult/references/user-facing-output.md` section (a) still states the rule verbatim: `grep -F 'Kein schweizerisches ss an ß-Stellen'` hits, and the rule paragraph at base lines 21-27 (ß after long vowel/diphthong; ss after short vowel; the Swiss-ss prohibition) is neither deleted nor thinned. The file may gain a pointer to the new check.
- [ ] If the ß/ss sentence's wording is touched anywhere, ALL THREE copies stay in step: `cogni-consult/references/user-facing-output.md:21-27`, `cogni-consult/hooks/on-subagent-start.sh:63-70` (`LANGUAGE_RULES["de"]`), and `cogni-workspace/hooks/on-session-start-language.sh:58-65` (`RULES["de"]`). A one-sided edit fails; leaving all three unchanged satisfies this vacuously.
- [ ] Option 2 is NOT implemented as an auto-firing pass: the diff contains no code path that rewrites engagement files on re-entry or on any scheduled/milestone hook. If a repair capability lands at all it is a separate consultant-elected invocation — an unreviewed mass edit of the stored corpus is the failure mode to avoid.
- [ ] Option 3 is NOT substituted for Option 1: the diff must contain an executable detector, not merely relocated or re-emphasised reference prose. A reference-prose-only diff (no new script + suite) does not clear this bar.
- [ ] CONDITIONAL — if `cogni-consult/skills/consult-resume/SKILL.md` is touched, its read-only contract at base lines 369-371 is either unchanged or amended to match (never silently contradicted), `bash cogni-consult/tests/test_step0_register_block.sh` still exits 0, and any new Bash call's `description` obeys the five constraints of section (f) of `references/user-facing-output.md`. Vacuously satisfied if no SKILL.md is touched.
- [ ] No new maintainer breadcrumb is introduced — no `#1261`, version tag, milestone or finding code in any touched `*/skills/*/SKILL.md` or `*/agents/*.md`; `python3 scripts/check-breadcrumbs.py` exits 0 and `scripts/baselines/breadcrumb-baseline.json` gains no new entry.
- [ ] No plugin version is touched — `python3 scripts/check-version-bump.py` reports `data.clean: true`, and the `version` in `cogni-consult/.claude-plugin/plugin.json` plus cogni-consult's entry in the root `.claude-plugin/marketplace.json` are unchanged (the post-merge workflow owns the bump).
- [ ] `python3 scripts/run-plugin-tests.py --filter consult` exits 0 with `data.failed: 0` — the new suite passes and none of the 10 pre-existing `cogni-consult/tests/*.sh` suites regresses.

**Risk watchlist:**
- A naive `grep -c 'ss'` or `[aeiouäöü]ss` implementation that reports `dass`/`muss`/`Prozess`/`müssen`/`Schlüssel`/`Ausschuss` as drift — it would flag the rule's own examples and be abandoned on first use. A noisy detector is a worse outcome than the unenforced rule.
- The mirror failure: a matcher so tightly anchored (whole-word, case-sensitive) that it reports 0 findings on the very corpus the issue documents as carrying 12. Only the `Messgrösse` compound case and a case-variance case (`heisst` vs sentence-initial `Heisst`) separate this from a correct fix.
- Registration drift: landing 1, 2 or 3 of the four documentation surfaces. `cogni-consult/README.md` in particular is the surface most likely to be skipped on a generator-ownership argument — verified against base, no `cogni-docs` plugin exists in this repo and README.md carries no generated marker, so the argument does not hold here. A human should still confirm no out-of-repo generator owns `README.md:140-168`.
- The mutation-check obligation being satisfied on paper rather than by a run: `run_mutation` hardwires `TARGET`/`SUITE` at `mutation-check.sh:32-33`, so a second target requires real parameterization. Writing a recipe in the wrong harness dialect would silently no-op.
- A file walk that skips dot-directories, so `.metadata/decision-log.json` — one of the four evidence files the issue names — is never scanned and the tool under-reports on exactly the corpus that motivated it.
- Duplicating the Swiss→ß pair list into `references/user-facing-output.md` or `CLAUDE.md` as a human-readable table AND into the script — a new drift class with nothing to detect it.
- Over-claiming recall: wording the script or its docs as if it enforces the orthography rule when it can only sample a curated list. A caller that treats a clean report as a guarantee re-creates the original defect one level up.
- Scope creep into Option 2 (a `--fix` / in-place rewrite / auto-firing repair on re-entry) or into Option 3 (moving or duplicating the rule text).
- Ambiguous homographs whose `ss` spelling is itself valid standard German (`Masse` = mass, `Busse` = buses) entering the lexicon and producing confident wrong corrections. No gate is written for this because the AC does not name it; a human should eyeball the lexicon for such entries.
- Suggestion casing (whether sentence-initial `Heisst` yields `Heißt` or `heißt`) is cosmetic and left to the author, but a human should glance at it.
- Scanning generated artifacts (`output/`, engagement-root `README.md`, `assumptions.md`) and reporting derived noise the consultant cannot fix at the source. Either decision is acceptable; only silence on the boundary is gated.
- No engagement corpus is committed to this repo, so nothing in CI can regression-test against the real `it-saar-leadership-development` text — the suite's synthetic fixtures are the only evidence the detection works, which raises the bar on their fidelity.
- NOT A GATE, recorded per the issue's own instruction: the issue's theory is that in-context corpus beats a stated rule, so detection alone may not change generation behaviour. Option 1 is scoped as the evidence-gathering step; whether it also changes the register outcome is explicitly not a criterion and must not be graded as one.
- Pre-existing base drift must not be graded as a violation: `cogni-consult/CLAUDE.md`'s Scripts table already lacks rows for `dt-stage-advance.sh`, `schedule-edit.py`, `mutation-check.sh` and `generate-dashboard.py`, and `README.md`'s Components table already lacks `mutation-check.sh` and `_discover_extractor.py`. Only the new script's rows are in scope.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1266
**Perspectives consulted:** plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer, cogni-service:service-review

**Summary:** Two independent perspectives both return needs_revision on concordant evidence. The detector itself is sound and demonstrably works — all 11 cogni-consult suites pass, the new suite is runner-discoverable, the read-only posture and the load-bearing Messgroesse/heisst/schliesslich cases are proven, and both pre-gates are clean. The blockers are registration and two behavioural gaps: 3 of the 4 documentation surfaces the committed contract names were never touched (items 9, 10, 11), the mutation-check.sh falsifier item 12 demands was declined, the script's header is silent on the sources/ and output/ scanned-set boundary (item 3), the precision fixture omits three of the six named tokens and the drift total is never pinned to an exact count (items 20, 22), an unreadable engagement directory returns success:true with zero findings (item 16), and the bare `weiss` lexicon entry false-positives on correct German compounds (Beweissicherung, Beweisstueck, Ausweisstelle).

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (13)

- **[major]** (auto-fixable) `cogni-consult/README.md:142-168` — Contract item 10 (REGISTRATION SURFACE 3 of 4) violated — flagged independently by both perspectives. `cogni-consult/README.md` is not in the diff at all, so the `## Components` table gains no `| orthography-drift-scan.py | Script | ... |` row. The PR justifies skipping it as `cogni-docs` `doc-generate`-owned; both reviewers verified that claim against the tree and it does not hold here: no `cogni-docs` plugin exists among the 14 manifests, no `doc-*` skill exists in any plugin, and README.md carries no generated-file marker (its footer is a hand-authored 'Built by cogni-work' line). A hand edit is not overwritten by anything in this repo, so the AC's demand stands.
  - *Suggested fix:* Add one row after the `assumption-change-frequency.sh` row: `| `orthography-drift-scan.py` | Script | Read-only scan reporting Swiss-`ss` spellings sitting in `ss`/eszett positions across one engagement's stored corpus; curated-list heuristic with bounded recall, no repair mode |`. If an out-of-repo generator genuinely owns that section, amend the AC in issue #1261 rather than settling it in a PR comment.

- **[major]** (auto-fixable) `cogni-consult/CLAUDE.md:80-105` — Contract item 9 (REGISTRATION SURFACE 2 of 4) violated — flagged independently by both perspectives. The Architecture tree `scripts/` block gains no line for `orthography-drift-scan.py`. The whole CLAUDE.md diff is one line (+1, the Scripts table row at :177); the tree block still ends `assumption-change-frequency.sh` -> `discover-projects.sh` -> `_discover_extractor.py`. This is not pre-existing base drift: every other script listed in the Scripts table also has a tree entry, so the new script is the only omission.
  - *Suggested fix:* Insert a two-column tree entry after the `assumption-change-frequency.sh` block, matching the surrounding continuation-line indentation.

- **[major]** (auto-fixable) `cogni-consult/README.md:214-217` — Contract item 11 (REGISTRATION SURFACE 4 of 4) violated — flagged independently by both perspectives. The Architecture tree `scripts/` block is silent about the new script and is unchanged by the diff. The contract accepts either extending that summary sentence or adding a line; silence fails.
  - *Suggested fix:* Extend the summary text to include the scan, e.g. `... README front-door generator, plus a read-only Swiss-ss orthography drift scan over an engagement corpus (stdlib-only)`.

- **[major]** `cogni-consult/scripts/mutation-check.sh:32-33` — Contract item 12 (MUTATION PROOF) violated — flagged independently by both perspectives. `mutation-check.sh` is not in the diff, so it gains no `run_mutation <label> <description> <python-re-program>` entry for the new detector. The PR declines on the ground that `TARGET`/`SUITE` are hardwired globals and parameterizing them is out of scope, but the committed rubric anticipated exactly that and made the parameterization part of the ask. Mitigating and verified by a reviewer rather than taken on trust: the suite's own goes-red block (test lines 417-485) is a genuine falsifier — each mutation asserts `n != 1` fails loud, and the anchored-matcher case additionally asserts the mutant still emits a valid `success: true` envelope so a crash is not mistaken for detection. Item 13 is separately satisfied: `bash cogni-consult/scripts/mutation-check.sh` still exits 0 and still prints `All mutations caught.`
  - *Suggested fix:* Parameterize `run_mutation` to take `<target> <suite>` and register one recipe against `orthography-drift-scan.py` / `test_orthography_drift_scan.sh` — the empty-`SWISS_PAIRS` sentinel substitution from the suite transplants directly as a `re.subn` program with `count=1`. Then re-run the harness and confirm it still prints `All mutations caught.`

- **[major]** (auto-fixable) `cogni-consult/scripts/orthography-drift-scan.py:221-234` — Contract item 16 (fail-loud on a bad target) violated for the unreadable-path case. `os.walk(root)` is called with no `onerror` handler, so a permission error on the engagement root is swallowed and the walk yields nothing. Reproduced from the worktree: `chmod 000` on an engagement directory returns `{"success": true, ..., "total_findings": 0, "skipped": [], "error": null}` with exit 0 — the exact 'silent empty result that a caller reads as no drift' the item forbids. An unreadable *file* beneath a readable root is handled correctly (it lands in `skipped[]`). The missing-path and non-directory cases are correct and pinned by tests, but the item's behavioural demand names all three.
  - *Suggested fix:* Pass `onerror` to `os.walk` and either fail the envelope or record every unreadable directory in `skipped[]`; a pre-flight `os.access(root, os.R_OK | os.X_OK)` returning `success: false` with `failed_check: "engagement_unreadable"` is the smaller change. Pin it with a suite case (guarded to skip when running as root).

- **[major]** (auto-fixable) `cogni-consult/scripts/orthography-drift-scan.py:52` — The bare `weiss` entry produces false positives on correct standard German compounds where `s` + `s` meet at a morpheme boundary. Verified by running the committed script on a fixture: `Die Beweissicherung und das Beweisstueck liegen bei der Ausweisstelle.` returns `total_findings: 3`, each `{"form": "weiss", "suggestion": "weiss/eszett"}`. All three are correct German (Beweis+sicherung, Beweis+stueck, Ausweis+stelle) and the suggested form is wrong in every case. The `_ENTRY_GUARDS` mechanism at :68 already guards `weiss`'s right-extension (`weissagen`) but nothing guards the left side, and the script's own comment at :60-61 shows the author reasoned about exactly this class for `gross`/`fuss` while missing it for `weiss`. This is the failure mode the AC's framing paragraph calls out as worse than the unenforced rule ('a noisy detector the consultant learns to ignore').
  - *Suggested fix:* Add a left-context guard for `weiss`: skip the match when the preceding characters complete a `-weis` stem (`beweis`, `ausweis`, `nachweis`, `hinweis`, `verweis`, `reis`), e.g. by extending `_ENTRY_GUARDS` with a left-stem form checked against `text[:start]`. Pin `Beweissicherung` as a zero-finding suite case alongside the existing `Masse`/`Busse` homograph case.

- **[major]** (auto-fixable) `cogni-consult/tests/test_orthography_drift_scan.sh:205` — Contract item 22 (REPORT CONTENT) violated on the aggregate-count half. The suite's only assertion on the drift fixture's total is `[ "$(total_findings "$env_drift")" -gt 0 ]` — a non-zero check, not a pin to a specific known expected count. No case compares the count against a literal expected number for a fixture with findings (only the clean fixture's expected `0` is pinned). The per-finding fields (path, line, form, suggestion) and the `by_file` breakdown ARE asserted, and the escaped-JSON case pins an exact three-element expected result, so this is the aggregate-total half alone.
  - *Suggested fix:* Add an assertion that the drift fixture's `data.total_findings` equals its exact known count — count the Swiss forms actually seeded in `build_drift` and assert that literal number.

- **[major]** (auto-fixable) `cogni-consult/scripts/orthography-drift-scan.py:2-29` — Contract item 3 (scanned-set boundary in the header) violated — flagged independently by both perspectives. The module docstring covers purpose, the recall caveat, read-only posture, the exit-code contract, Usage and Output, but says nothing about which engagement subtrees are scanned and which are excluded. Grepping the file for `sources` returns zero hits, and `output/` is never named either — the two surfaces the item requires to be addressed by name. The exclusion policy that does exist (the `_Auto-generated` sentinel comment at :85-91) is a module comment rather than the header, and it names generator scripts rather than subtrees. This became load-bearing when the design moved from a hardcoded path list to a sentinel test: `sources/`, `output/` and a marker-less engagement-root `README.md` are now all scanned — a defensible decision, but the item says either decision passes and only silence fails.
  - *Suggested fix:* Add a short docstring paragraph: everything under the engagement root with a `.md` or `.json` extension is scanned, including dot-directories such as `.metadata/`; only `.git` is skipped by path; `sources/` (verbatim third-party inbox, may legitimately be Swiss) and `output/` ARE scanned deliberately, and generated echoes (`README.md`, `assumptions.md`) are excluded only when they carry the generators' own footer sentinel.

- **[major]** (auto-fixable) `cogni-consult/tests/test_orthography_drift_scan.sh:126-131` — Contract item 20 (PRECISION) violated — flagged independently by both perspectives. The clean fixture carries `dass`, `muss`, `Prozess` (plus the `Masse`/`Busse` homographs and `Einfluss`), but three of the six tokens the item names by name are absent: `muessen`, `Schluessel` and `Ausschuss`. `Ausschuss` is the one the AC calls out specifically as the morpheme-boundary-after-a-diphthong case. Behaviour is correct today — a reviewer ran the committed script on a fixture containing all three and it returned `total_findings: 0` — but nothing in CI pins it, so a later lexicon addition could regress it silently. The rest of the item is satisfied: the zero is read explicitly out of the envelope via `total_findings`, not inferred from the absence of a crash.
  - *Suggested fix:* Extend the `build_clean` heredoc to include `muessen`, `Schluessel` and `Ausschuss` (e.g. `Wir muessen den Schluessel finden, bevor der Ausschuss tagt.`, with the proper umlauts). No assertion change needed — the case already asserts `total_findings == 0`.

- **[minor]** (auto-fixable) `cogni-consult/scripts/orthography-drift-scan.py:293` — The success path emits `"error": null`, while the header at :28 declares the envelope as `{"success": bool, "data": {...}, "error": str}` and contract item 15 words the requirement as 'success: true with the findings under data and an empty error'. The house convention in this plugin is the empty string: `resolve-assumptions.py:356,417` and `assumption-change-frequency.py:199-206` all pass `""` on the success path, as do the repo's own `run-plugin-tests.py` / `check-version-bump.py`. All three keys are present on every exit path (item 14 is satisfied), so this is a type mismatch against the script's own declared contract rather than a missing key — but a consumer doing `if envelope["error"]:` on a `str` assumption gets `None`.
  - *Suggested fix:* Change `_emit(True, _scan_engagement(root), None)` to pass `""`, matching the declared `"error": str` and the three sibling scripts.

- **[minor]** (auto-fixable) `cogni-consult/tests/test_orthography_drift_scan.sh:184-201` — Contract item 17 (engagement-scoping) is satisfied only incidentally. `CLEAN`, `DRIFT` and `ESCAPED` are all built as siblings under `TMPROOT` before any scan runs, so when the case asserts `total_findings == 0` for `CLEAN`, a sibling engagement full of Swiss text is genuinely present and genuinely contributes nothing — the item's substance holds. But nothing in the suite names or asserts that relationship, so a later refactor to per-case `mktemp -d` fixtures would silently stop testing scoping while every assertion still passed. Given that engagement-scoping is the issue's own central observation, the property deserves an explicit anchor.
  - *Suggested fix:* Rename or split the assertion so the sibling relationship is stated, e.g. a `scoping: a sibling engagement's Swiss text contributes zero findings` case that asserts `$DRIFT` is a sibling of `$CLEAN` under `$TMPROOT` and that the `CLEAN` scan's `by_file` is empty.

- **[minor]** (auto-fixable) `cogni-consult/scripts/orthography-drift-scan.py:164` — Findings carry `line` but no column or offset, so multiple occurrences of the same form on one line are emitted as byte-identical objects (a `Beweissicherung` probe returned three indistinguishable findings), and inside JSON carriers the line is deliberately pinned to the value's start, so several hits in one long `framing` string collapse to the same coordinate. Contract item 22's per-finding half is satisfied as written, but a consultant reading the report cannot navigate to the second occurrence.
  - *Suggested fix:* Carry the match offset (and for markdown a column derived from it) on each finding; `_scan_text` already returns the offset and it is discarded at :164 and :196.

- **[minor]** `PR diff` — Diff is 805 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- Executed from the worktree, not taken on trust: `bash cogni-consult/tests/test_orthography_drift_scan.sh` exits 0 with 22 OK assertions, and `python3 scripts/run-plugin-tests.py --filter consult` returns total: 11, passed: 11, failed: 0 — the new suite passes and none of the 10 pre-existing cogni-consult suites regresses (item 30).
- Runner discoverability confirmed (item 7): `--list` returns the new suite in `data.suites[]` and cogni-consult's count rises 10 -> 11. The suite follows the house template of `test_assumption_change_frequency.sh` faithfully (set -u, self-locating TESTS_DIR, mktemp -d + trap, per-case counter, non-zero exit).
- Detection quality on the load-bearing cases is genuinely proven (items 19-21): `Messgroesse` reports from the bare `Groesse` entry via stem matching with no word anchors, `heisst` and sentence-initial `Heisst` both report through the first-character case flip, `schliesslich` reports from a nested `field.json`, and `.metadata/decision-log.json` reports `Strasse` — proving the walk does not skip dot-directories.
- The `_flip_first` design reasoning (two variants rather than case-insensitive matching, because uppercase German renders the eszett as SS so an all-caps heading is correct orthography) is a genuinely sharp piece of design.
- The escaped-JSON defect the PR body reports catching pre-commit is real and its fix is verified: a `field.json` written with the default `ensure_ascii=True` carries a unicode escape that no pair-list entry would match, and the suite asserts the decoded value yields exactly the expected three forms. That is precisely the silent-recall-loss-inside-a-clean-report failure the risk watchlist names.
- The in-suite goes-red block is a real falsifier and was verified by running it: each mutation asserts `n != 1` fails loud so a no-op regex cannot pass as a defended detector, and the anchored-matcher case additionally asserts the mutant still emits a valid envelope so a crash is never mistaken for detection.
- Read-only posture holds under inspection (item 18): no --fix/--repair/--rewrite/--write/--apply flag exists, `--fix` is actively rejected as `unexpected_argument`, no write-mode call targets the engagement root, and the byte-identity case captures its digest baseline BEFORE any scan runs, so the assertion is load-bearing rather than comparing two post-scan states.
- Stdlib-only and offline (item 2): imports are exactly json, os, re, sys; no network call and no pip step. Every read passes explicit encoding=utf-8 and the lexicon is written with literal umlauts and the eszett (item 5).
- The lexicon has exactly one machine-readable home (item 6) — no second human-readable copy in CLAUDE.md or references/user-facing-output.md, avoiding the duplication drift class the watchlist names — and carries an explicit false-positive-policy comment explaining why the `Masse`/`Busse` homographs and the bare `gross`/`fuss` stems are deliberately absent.
- Envelope discipline is thorough (items 14-15): argv is hand-parsed rather than handed to argparse so no exit path can print usage text instead of the envelope, main() is wrapped so an unexpected exception still emits one JSON object, and the exit-code contract is stated in the header, implemented in the code and pinned by the suite with no mismatch.
- The recall caveat (item 4) is stated in the header, repeated in the emitted `detection` field AND in the CLAUDE.md row — so a caller who only ever sees the JSON is still told what a clean report does not prove.
- Rule text is intact (items 23-24): the diff to references/user-facing-output.md is additions only (9 lines, zero deletions), the rule paragraph is byte-unchanged and still in section (a), and all three restatements are untouched, so the one-contract-two-delivery-paths rule stays satisfied vacuously.
- Scope discipline is clean: no Option 2 write path or auto-firing repair anywhere in the diff (item 25), an executable detector rather than relocated prose (item 26), no SKILL.md touched so the consult-resume conditional is vacuous (item 27).
- Plugin manifest valid and untouched: kebab-case name, semver 0.1.95, well-formed author/license/keywords; all 14 plugin manifests and the root marketplace.json parse. Both deterministic pre-gates clean — check-frontmatter.sh reports 0 offenders across cogni-consult, and check-version-bump.sh reports status ok / data.clean true (item 29). check-breadcrumbs.py exits 0 with no baseline change (item 28).
- The mutation harness is not collaterally damaged (item 13): `bash cogni-consult/scripts/mutation-check.sh` still exits 0 and still prints `All mutations caught.` with both pre-existing falsifiers intact.

### Recommendation


---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1266

**Summary:** All three perspectives pass at head 933cc284: the plugin-validator's prior four advisory minors are down to two non-gating ones, contract-grading-reviewer confirms the previously-needs_revision item 3 (scanned-set boundary) is now satisfied with zero remaining findings, and the only other finding is the informational oversized-diff notice from service-review — all un-waived findings are minor, so Rule 5b lands the merge as pass.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (auto-fixable) `cogni-consult/CLAUDE.md:180` — The Scripts row re-spells the full PROSE_KEYS enumeration as a second human-readable copy of a module-level constant. The gated duplication check (item 6) is scoped to the Swiss pair list and passes cleanly, but adding a prose key to the script would leave CLAUDE.md silently stale with no gate to notice.
  - *Suggested fix:* Replace the parenthetical enumeration with a pointer to the constant, mirroring how the same row already points at SWISS_PAIRS rather than restating its members.

- **[minor]** `cogni-consult/scripts/orthography-drift-scan.py:303` — An unreadable SUBDIRECTORY is silently omitted from the report, while an unreadable FILE is recorded in data.skipped. Probed: a chmod 000 subdirectory holding Swiss text returns success true with skipped:[] and no trace a subtree was never opened. This is the same swallow the root guard closes, one level down. Not gated by any acceptance item (item 16 requires missing, non-directory and unreadable-ROOT, all three of which pass), and the reporting channel already exists.
  - *Suggested fix:* Pass an onerror callback to os.walk so the existing skipped[] ledger covers directories too, and extend the 12c case with a nested-unreadable-directory assertion.

- **[minor]** `PR diff` — Diff is 1072 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- The merger's recorded escalation is resolved rather than argued: README.md was edited after all, landing both the Components row and an extension of the Architecture tree summary, so all four registration surfaces are covered (items 8-11).
- The mutation obligation is met behaviourally, not on paper: run_mutation was genuinely parameterized to <target> <suite> <env-var> with DISCOVER/ORTHOGRAPHY triples, the suite gained a CONSULT_ORTHOGRAPHY_SCAN override, and a live run prints all three 'OK: mutation caught' lines with the two pre-existing falsifiers not regressing.
- run_mutation asserts the suite is GREEN at baseline before mutating and the mutation program must report exactly one substitution, so a moved sentinel fails loud instead of yielding a silent no-op falsifier.
- Precision and recall are both implemented and pinned: the clean fixture (dass, muss, Prozess, müssen, Schlüssel, Ausschuss plus the -weis compounds) asserts an explicit zero out of the envelope, while the drift fixture is pinned to an exact 8 rather than merely -gt 0, so a regression losing most findings still goes red.
- The suite's goes-red cases verify each mutant applies, still emits a valid success envelope so a crash cannot be mistaken for detection, and loses the specific finding it targets.
- The unreadable-root guard plus its 12c case closes the one zero a caller could not distinguish from a real one, and is correctly skipped when running as root.
- The reference-prose addition leaves the rule paragraph byte-identical, so the three rule restatements stay in step and Option 3 is not substituted for Option 1.
- item 3 satisfied (fixed from prior needs_revision): the SCANNED SET paragraph names the generated artifacts explicitly, and states that naming them is not excluding them by path — a marker-less README.md is hand-authored and still scanned.
- items 12-13 satisfied: the parameterized harness registers the empty-SWISS_PAIRS falsifier against the new detector and still exits 0 printing 'All mutations caught.'
- item 17 satisfied: the 12b scoping case asserts CLEAN and DRIFT are siblings under one mktemp -d root and that the clean engagement's by_file is {} despite the Swiss-laden sibling.
- items 28-30 satisfied: breadcrumbs clean, version-bump clean (no version line touched), and 11/11 cogni-consult suites pass with none of the 10 pre-existing suites regressing.

### Recommendation

---

claude-cogni-work:
## Revision posted

Commits `a4053454` and `933cc284` address all 13 findings from the previous verdict. Twelve are resolved by edit; one is waived with rationale below. Fresh verdict: ✅ **PASS**.

| # | Finding | Resolution | Commit |
|---|---|---|---|
| 1 | **[major]** `README.md:142-168` — item 10, Components row missing | Row added after `assumption-change-frequency.sh`. The `cogni-docs`-owned justification was wrong and is dropped — no such plugin or `doc-*` skill exists in this repo and README.md carries no generated marker, so a hand edit stands. | `a4053454` |
| 2 | **[major]** `CLAUDE.md:80-105` — item 9, Architecture tree missing | Three-line tree entry inserted between `assumption-change-frequency.sh` and `discover-projects.sh`, matching the surrounding continuation-line indentation. | `a4053454` |
| 3 | **[major]** `README.md:214-217` — item 11, tree silent | Summary sentence extended: "plus a read-only Swiss-ss orthography drift scan over an engagement corpus". | `a4053454` |
| 4 | **[major]** `mutation-check.sh:32-33` — item 12, mutation proof declined | Done rather than declined. `run_mutation` now takes `<target> <suite> <env-var> <label> <description> <program>`; each subject is named once as a triple (`DISCOVER`, `ORTHOGRAPHY`). The suite gained a `CONSULT_ORTHOGRAPHY_SCAN` override mirroring `test_discover_extractor.sh:49`. Falsifier 3 = the empty-`SWISS_PAIRS` sentinel substitution. **Run:** all three print `OK: mutation caught`, exit 0, `All mutations caught.` | `a4053454` |
| 5 | **[major]** `orthography-drift-scan.py:221-234` — item 16, unreadable root | Pre-flight `os.access(root, R_OK\|X_OK)` → `success:false`, `failed_check: "engagement_unreadable"`, exit 1. Reproduced the old silent-zero first, then confirmed the fix. Pinned by case 12c, skipped when `id -u` is 0. | `a4053454` |
| 6 | **[major]** `orthography-drift-scan.py:52` — bare `weiss` false-positives | `_ENTRY_LEFT_GUARDS`, keyed identically to `_ENTRY_GUARDS`. Stems are the bound forms of the `-weisen` verbs (12 of them), so the list is extensible rather than anecdotal. Verified: `Beweissicherung`/`Beweisstück`/`Ausweisstelle` now report nothing; `Er weiss`, `Messgrösse`, `heisst` and `schneeweiss` still report. | `a4053454` |
| 7 | **[major]** `test_...sh:205` — item 22, aggregate count | `-gt 0` replaced with `= 8`, the exact seeded count — **measured** by running the scan against the fixture, not guessed, with the eight enumerated in a comment. | `a4053454` |
| 8 | **[major]** `orthography-drift-scan.py:2-29` — item 3, scanned-set boundary | `SCANNED SET` paragraph added naming `sources/` and `output/` as deliberately scanned. Re-review found it still did not name the generated artifacts literally; `933cc284` names the engagement-root `README.md` and `assumptions.md`, and states that naming them is not excluding them by path. | `a4053454`, `933cc284` |
| 9 | **[major]** `test_...sh:126-131` — item 20, precision | `build_clean` extended with `müssen`, `Schlüssel`, `Ausschuss` (proper umlauts) plus the `-weis` compound sentence. Case 1's existing zero-assertion now covers them — and goes red without the guard from finding 6. | `a4053454` |
| 10 | **[minor]** `orthography-drift-scan.py:293` — `error: null` | Now `""`, matching the declared `"error": str` and the three sibling scripts. Pinned by a new envelope-shape assertion. | `a4053454` |
| 11 | **[minor]** `test_...sh:184-201` — item 17, incidental scoping | New named case 12b asserts `dirname CLEAN == dirname DRIFT == TMPROOT` and that CLEAN's `by_file` is `{}`, so a refactor to per-case `mktemp -d` fails here instead of silently ceasing to test scoping. | `a4053454` |
| 12 | **[minor]** `orthography-drift-scan.py:164` — no coordinate | Markdown findings carry a 1-based `column`; JSON findings carry `value_offset` into the decoded value (a file offset cannot survive the decode). Sort key extended for determinism. Asserted in cases 2 and 6. | `a4053454` |
| 13 | **[minor]** `PR diff` — 805 lines vs the 400 bound | **Waived** — see below. | — |

### Why finding 13 is waived

Splitting a merged-scope PR mid-review is not a delta on a branch: it means closing this PR, re-filing the work as two issues and re-authoring both from a fresh base, discarding the review history and the twelve fixes above. It is also self-defeating here — the four registration surfaces, the harness parameterization and the six suite additions are all mandated by *other* findings in the same verdict, so no ordering satisfies both the split and the remaining twelve. The diff is now larger, not smaller. If the bound is to be enforced, the instrument is a decompose at triage, not a mid-review split.

### Two things the cleanup pass caught, worth flagging

- **A performance regression I introduced and then fixed.** The first cut of the `weiss` guard hoisted `text.lower()` above the guard lookup, making matching O(matches × filesize) — measured at 34s on a 400KB corpus where the pre-change code took under a second. Only one entry carries guards, so it now bails before lowering. Behaviour verified identical across the guard cases; cost back to parity.
- **One of my own claims was wrong.** A comment justified omitting the `reis` stem by saying `Reissverschluss` is "a genuine Swiss form worth reporting". It is not reported — no `SWISS_PAIRS` entry matches it. The comment now states the real reason (`weiss` cannot match inside a Reis/Reiß word at all).

### Gates, all run rather than asserted

`test_orthography_drift_scan.sh` 25/25 · `run-plugin-tests.py --filter cogni-consult` 11/11 · `mutation-check.sh` exit 0, `All mutations caught.` · `check-frontmatter.sh` 0 offenders · `check-breadcrumbs.py` 0 violations · `check-version-bump.sh` `status: ok`, `clean: true`, no version line touched.

> Note on the base: `origin/main` advanced mid-revision (PR #1267 merged, auto-bumping cogni-marketing to 0.0.5). A raw `git diff origin/main` on the manifests therefore shows this branch "reverting" that bump — a base artifact, not a version touch. `git status` on the manifests is empty and the gate, which anchors on the fork point, reports clean.

### Known, not actioned

Two advisory minors from `plugin-validator` remain on the fresh verdict and are recorded there rather than fixed here: an unreadable *subdirectory* is silently omitted from `skipped[]` (the same swallow as finding 5, one level down — real, but no acceptance item covers it and the fix needs its own test), and the `PROSE_KEYS` enumeration is restated in `CLAUDE.md:180` where nothing would catch its drift.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1266

**Summary:** All 30 committed contract items are satisfied with live-executed corroborating evidence (contract-grading-reviewer and plugin-validator both pass, no critical/major findings); the only findings are an informational oversized-diff note and four merger-added preference notes, so the PR is safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (5)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-consult/CLAUDE.md:180` — The new Scripts-table row re-enumerates all nine PROSE_KEYS verbatim, creating a second human-readable copy of a module-level constant with nothing detecting drift between them. Checklist item 6's gate is on the Swiss-ss pair list and that is clean (grepping schliesslich returns exactly one file), but the watchlist's 'a new drift class with nothing to detect it' shape applies to this enumeration too.
  - *Suggested fix:* Replace the nine-key enumeration with a pointer, keeping the constant the single home.

- **[minor]** (advisory-addition) `cogni-consult/scripts/orthography-drift-scan.py:22-24` — The proper-noun false-positive class the header self-discloses reproduces: 'Herr Weiss und Frau Strasser' yields two confident findings suggesting Weiss->Weiß and Strasser->Straße on stakeholder names. No acceptance item gates this and the header names the limitation explicitly, so it is a disclosed tradeoff; recorded because the contract's risk watchlist asks for a human eyeball on this class.
  - *Suggested fix:* No code change required. If it becomes noisy, a name-particle suppression or a per-engagement ignore list would narrow it.

- **[minor]** (advisory-addition) `cogni-consult/scripts/orthography-drift-scan.py:67-80` — Recall observation, not a contract violation: SWISS_PAIRS omits several high-frequency Swiss forms (schliessen/beschliessen, Massstab, massgeblich, ausserdem, äusserst, fliessen). Confirmed live: a probe containing Ausserdem reported nothing. The header's bounded-recall caveat discloses this and item 19 gates only the three live-observed forms, all covered.
  - *Suggested fix:* Consider adding the -schliessen family, Massstab/massgeblich and ausserdem/äusserst to SWISS_PAIRS; each is unambiguously Swiss, so it costs no precision.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-consult/tests/test_orthography_drift_scan.sh:41` — TESTS_DIR is derived from BASH_SOURCE[0] while all ten pre-existing cogni-consult suites - and the literal spelling in checklist item 7 and the issue AC - use the dirname $0 form. Functionally equivalent under `bash <path>` (verified: the suite exits 0 by absolute path from an unrelated cwd), so the item's stated property holds; house-style deviation only.
  - *Suggested fix:* Use the dirname $0 form to match the other ten suites and the AC's literal spelling.

- **[minor]** `PR diff` — Diff is 1072 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

### Positive observations

- All 30 contract items graded satisfied with live-executed evidence; no item unclear, no INTENT-UNCLEAR divergence. (contract-grading-reviewer)
- All four registration surfaces landed, resolving the contract's recorded escalation: the plan record declared README.md would deliberately NOT be edited on a cogni-docs-ownership argument, but the diff adds both the Components row (README.md:167) and the Architecture-tree scripts/ extension (README.md:215-220), alongside the CLAUDE.md Scripts row (:180) and tree entry (:104-106). (plugin-validator)
- Mutation proof is real, not paper: run_mutation was genuinely parameterized to a (target, suite, env-var) triple, the two _discover_extractor.py falsifiers were rewired without regressing, and mutation-check.sh exits 0 printing three 'OK: mutation caught' lines. (plugin-validator)
- run-plugin-tests.py --filter consult exits 0 with total 11, passed 11, failed 0; the new suite sits exactly one level under tests/ and prints 25 OK lines. (plugin-validator)
- Envelope discipline holds on every exit path constructed - clean, drift, missing path, non-directory, no argument, two arguments, empty dir, unparseable JSON, undecodable non-UTF-8, LC_ALL=C - each exactly one valid JSON line with all three keys. (plugin-validator)
- Precision is the load-bearing half and it holds: the clean fixture reports an explicit zero, Masse/Busse are documented deliberate non-entries, and the morpheme guard suppresses Beweissicherung/Ausweisstelle/Hinweisschild/Reissverschluss while still reporting genuine compounds. (plugin-validator)
- Recall on the headline case is pinned behaviourally: matching is re.escape with no word-boundary anchor, case-variant per entry, and the suite's 13b goes-red mutation proves anchoring whole-word loses Messgrösse. (plugin-validator)
- Carrier coverage is genuine on both classes: scope/key-question.md, a nested action-fields/<slug>/field.json framing, and a .metadata/decision-log.json rationale, with the dot-directory walk asserted. (plugin-validator)
- Strictly read-only and stdlib-only, verified rather than asserted: the only open() is read-mode utf-8; imports are json/os/re/sys; --fix is rejected; and the suite compares a sha256 tree digest before and after a drift-reporting scan. (plugin-validator)
- Fail-loud coverage exceeds the AC's two-case minimum - the os.access(root, R_OK|X_OK) pre-flight closes the hole os.walk would silently produce, pinned by a chmod-000 case. (plugin-validator)
- The rule text stayed in step trivially: user-facing-output.md:21-27 byte-identical to origin/main, the grep still hits, and neither hook file is touched. (plugin-validator)
- Boundary gates all clean: no skills/ or agents/ file touched, check-breadcrumbs.py exits 0, check-version-bump reports clean, plugin.json 0.1.95 and the marketplace manifest unchanged. (plugin-validator)

### Recommendation

---

Issue #1309, opened 2026-08-11
cogni-visual: ASCII-ified German in the story-to-web image-prompt example body


claude-cogni-work:
## Plugin
cogni-visual

## Problem

`cogni-visual/skills/story-to-web/references/04-image-prompts.md:240` carries ASCII-ified German in a worked example body:

```
body: "Wearable-Sensoren ubermitteln Vitalwerte in Echtzeit an das Arzteteam..."
```

`ubermitteln` should be `übermitteln`; `Arzteteam` should be `Ärzteteam`.

Verified against `origin/main` @ `a58ae47f` — one word-boundary occurrence each, both on line 240.

## Why this is filed separately

Surfaced while resolving the German-fidelity guard issue, which asked for a CI guard and not a copy repair. It was deliberately kept out of that PR: a guard PR that also edits content is reviewed under two different lenses at once, and the guard's acceptance criteria never asked for the repair.

It is also **not** caught by that guard. The guard matches a curated 18-token vocabulary, and neither `ubermitteln` nor `Arzteteam` is on it — precisely why the guard can satisfy its own "exits 0 on current main" criterion. So this defect is live and stays live until repaired here.

## Note on internal inconsistency

The same file writes correctly-accented German 33 lines later:

```
body: "KI-gestützte Regelprüfung scannt Transaktionen in Echtzeit..."
```

so the reference is inconsistent within its own worked examples.

## Proposed change

Replace the two tokens on line 240 with their proper Unicode forms. Change nothing else in the file.

## Decision to make while here

Whether `ubermitteln` and `Arzteteam` should join the guard's vocabulary table once repaired. Adding them is only safe **after** the repair — adding them first would turn the guard red on `main`.

## Acceptance criteria

- [ ] `grep -nw 'ubermitteln\|Arzteteam' cogni-visual/skills/story-to-web/references/04-image-prompts.md` returns nothing
- [ ] Line 240 reads `übermitteln` and `Ärzteteam`; the rest of the file is unchanged
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-visual` exits 0
- [ ] A conscious decision is recorded on whether the two tokens join the guard vocabulary

claude-cogni-work:
## Triage Report

- **Severity:** S4-enhancement (unchanged — content-correctness defect in a reference example, no runtime impact)
- **Component:** cogni-visual / `skills/story-to-web` (references chain)
[...]

PR #1310, opened 2026-08-11
cogni-visual: CI guard for ASCII-ified German in generated copy

### LINKED ISSUE #1299: cogni-visual: no guard exists for ASCII-ified German in generated-copy content
## Plugin
cogni-visual

## Current behavior

No umlaut / German-fidelity checker exists anywhere in this repository. Confirmed twice by full-tree grep during the work on #1290 (once by a read-only diagnostician, once by the plan refiner): every hit for `umlaut|ascii.?ified|german.?fidelity` is either **reverse** transliteration (Unicode → ASCII, deliberately, for slugs and filenames — e.g. `cogni-trends/skills/trend-scout/scripts/generate-project-slug.sh`, `cogni-knowledge/scripts/_knowledge_lib.py`, `cogni-visual/agents/web.md:55-56`) or unrelated test commentary about umlaut-safe tokenization.

Nothing validates the opposite direction: that German-language *generated-copy* content preserves proper Unicode umlauts.

This is why the defect class stood undetected long enough to need two separate PRs (#1292 for the rule and `03-section-copywriting.md`, #1298 for the five-file example/reference chain, 78 substitutions). It will recur in the next German example anyone writes.

## Desired behavior

A CI-discovered `cogni-visual/tests/` suite that fails on ASCII-ified German in German-language content.

It must cover **both** corruption styles actually observed on `origin/main`:

| Style | Observed examples |
|---|---|
| dropped-vowel | `Stillstande`, `fur`, `uber`, `nachste`, `Verschleiss`, `Wettbewerbsfahigkeit`, `Krafte`, `Fuhrung`, `Erstgesprach`, `jahrlich`, `ausrusten`, `Uberblick` |
| ae/oe/ue digraph | `Qualitaetsdrift`, `hoehere`, `LOESUNG`, `FAEHIGKEITEN`, `Anlagenverfuegbarkeit`, `Qualitaet` |

The digraph style was absent from #1290's own risk note and was only found by direct file survey — a guard covering just the dropped-vowel style would have missed six sites in one file.

## Required exemptions — the hard part

A naive guard is worse than none, because these sites are **correct** and must never be flagged:

- **`cogni-visual/skills/story-to-web/references/05-validation.md:253-254`** — states the rule itself using the literal ASCII tokens `ae`/`oe`/`ue` to *name* the defect class ("German umlauts preserved where possible (a/o/u not ae/oe/ue)"). A guard that flags this would pressure someone into deleting the rule.
- **`cogni-visual/agents/web.md:55-56`** — deliberate ä→ae mapping for the `narration_slug` used in output filenames.
- **Skill frontmatter `description:` trigger phrases** — ASCII spellings there are deliberate, so a user on a non-German keyboard still triggers the skill. Respelling one silently drops German triggering coverage.
- **Slugs, filenames, ids, URLs, enum values, and English text** — including the English `image_prompt` bodies that sit interleaved with German copy in `story-to-web/references/04-image-prompts.md`.
- **Already-correct German that merely looks suspicious**: singular `Stillstand` and the `Stillstands*` compounds (`Stillstandstag`, `Stillstandskosten`) — 31 such sites in the two example briefs alone, against 10 genuine `Stillstande` plurals. `Vorwarnung`, `Prozesse` and `Ersatzteil-Lieferzeit` are likewise correct.

The word-boundary discipline matters: in #1298 every enumerated corrupt token was verified to be a standalone word-boundary match, which is what made the substitution safe. A substring-based guard would produce false positives on the compounds above.

## Acceptance criteria

- [ ] A suite under `cogni-visual/tests/` fails (exit non-zero) when ASCII-ified German is introduced into German-language generated-copy content, covering both corruption styles above
- [ ] It exits 0 on current `main` (which is clean after #1298 lands)
- [ ] Every exemption listed above is honoured — in particular the suite must NOT flag `05-validation.md:253-254`, `agents/web.md:55-56`, skill `description:` trigger phrases, or the singular `Stillstand` / `Stillstands*` compounds
- [ ] The suite runs as `bash <path>` with no arguments and touches no network, so `scripts/run-plugin-tests.py` discovers it
- [ ] It carries its own executed negative case (a mutation that provably turns the suite red), not merely a described one
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-visual` exits 0
- [ ] No `plugin.json` / `marketplace.json` version touched

## Related

- #1290 / PR #1298 — the five-file example/reference cleanup where this gap was measured
- #1270 / PR #1292 — the companion change that scoped the rule and cleared `03-section-copywriting.md`

## Note on scope

This is deliberately **not** filed as deferred scope of #1290. #1290's acceptance criteria are fully satisfied by the five-file fix, so it closes on PR #1298. This is separate, new work: a guard is a bash-harness change needing different review expertise than German copy fidelity.

claude-cogni-work:
Closes #1299.

## Summary

Adds `cogni-visual/tests/test-de-ascii-orthography.sh` — a self-contained, CI-discovered suite that fails when German-language copy in this plugin's markdown loses its Unicode diacritics. It covers both corruption styles the issue names: the style that drops the mark outright, and the style that expands it into a two-letter digraph.

Registered in `cogni-visual/CLAUDE.md` under the `tests/` block, matching the shape the existing suite established.

17 cases: 4 non-vacuity guards, 2 vocabulary-integrity checks, 2 positive-detection cases, 5 exemption cases, 1 correct-German-lookalike case, 2 tracked-corpus cases, and 1 executed negative case.

## Why a curated vocabulary rather than a general orthography pattern

This is forced by the acceptance criteria, not a style preference, and it is the single most important thing to understand when reviewing this PR.

The corpus **already contains** ASCII-ified German outside the issue's enumerated list — `skills/story-to-web/references/04-image-prompts.md:240` has carried it since before this work started. A general "this looks like a de-umlauted German word" pattern would flag it and go **red on `main`**, contradicting the criterion that the suite exit 0 on a clean base. Satisfying both would require either repairing that line inside a guard PR, or hand-exempting a genuine defect. Neither is defensible.

An enumerated vocabulary satisfies both cleanly: all 18 listed forms have zero word-boundary matches on the base tree.

The cost is real and is stated rather than hidden. The suite header and the `CLAUDE.md` entry both say plainly that **a green run means no enumerated form appeared — not that the corpus is orthographically correct.** Every run also emits `NOTE:` lines reporting how many files it scanned and how many carry German, so a green result self-reports its own scope.

## How the design defends itself

**The vocabulary cannot silently rot.** Case W1 folds each row's correct spelling under its declared style and requires the result to equal the declared ASCII form — so a mistyped row that could never match anything cannot be committed. Case W2 set-differences the vocabulary against the forms the positive fixtures actually exercise, in **both** directions: you cannot add a row without adding fixture coverage, and you cannot leave a fixture pinning a form nobody declared.

**Exemptions are content-anchored, never line-anchored.** Two files legitimately contain ASCII forms: one states the umlaut rule by naming the spellings it forbids, the other defines a deliberate diacritic-to-ASCII mapping for filename slugs. Pinning those by line number would rot on the next edit above them, and the failure mode would be a *false positive on correct content* — precisely the "a naive guard is worse than none" outcome the issue warns about. Each exemption keys on a substring of the line itself, and case C2 asserts every anchor still resolves, so a reword forces a human to re-confirm the carve-out instead of letting it become dead weight.

**Exemption cases cannot pass by the matcher being dead.** Each exemption fixture plants a *real* vocabulary form inside an exempt context and requires zero findings. P1/P2 are the paired controls: they plant the same forms in plain German content and require detection. Both must hold.

**Two forms carry an extra guard.** Two of the dropped-vowel forms are also ordinary English or brand words. They are marked `de-context` and count only on a line carrying an independent German signal. Without that, they would be the first to false-positive the moment English prose is added to the scanned tree.

**Non-vacuity is enforced before any real assertion.** V1–V4 check that the scan root exists, the vocabulary parsed with both styles present, at least one markdown file was actually scanned, and at least one scanned file carries real German characters. A scanner that silently scans nothing is indistinguishable from a clean tree and would report green forever — the same silent degradation this guard exists to catch, one level up.

## Scope decisions

**Kept self-contained inside cogni-visual.** A sibling plugin has a prose scanner for a *different* German defect class; it was used as a model, not a component. It is deliberately not called cross-plugin and not vendored: the guard-sharing mechanism (subprocess vs. vendored copy) is an explicitly undecided open question parked on another epic, and picking either option here would pre-empt a human decision.

**The live corruption is not repaired here.** `04-image-prompts.md:240` is filed as #1309. It is content repair under a different review lens, and no acceptance criterion asks for it. Note it is **not** caught by this guard either — neither form is in the vocabulary, which is exactly why the guard can be green on `main`. #1309 also records the decision to make: whether those forms join the vocabulary once repaired (safe only *after* the repair).

**No version bump** — the post-merge workflow owns it. `check-version-bump.py` reports 14 plugins scanned, 0 touched, 0 violations.

## A note for the reviewer on where to read this

The repo-root clone is frequently checked out on a feature branch whose `cogni-visual/` still carries the full corruption set. Reviewing there will suggest the base is dirty and that the "green on main" criterion is unsatisfiable. This PR was authored and verified in a worktree cut from `origin/main`; please review against the PR head or a clean base.

## Test plan — all executed

- [x] `bash cogni-visual/tests/test-de-ascii-orthography.sh` exits 0, emitting `ok:` for all 17 registered cases and `RESULT: all de-ascii-orthography cases passed.`
- [x] M1 (executed negative case) confirms the guard can fail: the mutant child exits 1 and reports `C1` red naming both injected forms.
- [x] Independent red-proof outside M1: injecting two forms into a scratch copy of a real tracked reference file turns `C1` red with exact `path:line:form` for each.
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-visual` exits 0 — `total: 2, passed: 2, failed: 0`, both suites listed (was `total: 1` before this change).
- [x] `git status` clean after a run — the suite writes only under its own `mktemp -d` and never touches the tracked tree.
- [x] The `CLAUDE.md` entry contains **no** vocabulary form. This is load-bearing: `CLAUDE.md` is inside the scanned tree, so naming a form there would turn this suite's own `C1` case red. Verified by word-boundary search over all 18.
- [x] `python3 scripts/check-breadcrumbs.py` exits 0; `python3 scripts/check-version-bump.py` exits 0 (0 plugins touched).

## Mutation recipe

Executed, not merely recorded — the returned verdict is quoted below.

```
  --root . \
  --file cogni-visual/libraries/EXAMPLE_WEB_BRIEF.md \
  --expr 's/Vorwarnung/Uberblick/g' \
  --test 'bash cogni-visual/tests/test-de-ascii-orthography.sh' \
  --case C1
```

The mutation replaces an already-correct German word — one deliberately included in the correct-German-lookalike case, so it must *not* be flagged as-is — with an enumerated dropped-vowel form, which must turn `C1` red. Result:

```
"mutated_result": "red", "restored_result": "green", "verdict": "guard_verified"
```

Run from a checkout of this PR's head (`--root .` is that checkout, not the shared clone, which is usually on another branch).

Case ids are bare single tokens (`C1`, never `C1:`) and verdicts use the colon form (`ok: <id>` / `FAIL: <id>`), matching the sibling suite. That shape is required by the harness's output-line classifier — the older no-colon shape returns `case_not_found` and silently disarms the recipe while the suite still looks green. The summary is prefixed `RESULT:` and advisory lines `NOTE:` so neither is ever parsed as a case verdict.


claude-cogni-work:

Author instance is done. This PR is a draft carrying `svc:needs-review`; it has **not** been self-approved and no review verdict has been posted. The independent review belongs to the merger instance.

### Handoff preflight — clean

`clean=true`, 0 violations, 4 instruments skipped.

|---|---|---|
| run-tests | pass | suites green for cogni-visual |
| mutation-recipe | pass | recorded recipe runnable as written — `C1` found in the suite |
| frontmatter | skip | not_applicable — no touched SKILL.md or agents/*.md |
| breadcrumbs | skip | not_applicable — no touched SKILL.md or agents/*.md |
| skill-length | skip | not_applicable — no touched SKILL.md |
| agent-length | skip | not_applicable — no touched agents/*.md |

All four skips are `not_applicable` on this diff's shape (one new `tests/*.sh`, one `CLAUDE.md` edit), not instrument errors.

Graded against the 22-item acceptance contract derived at triage, **before** this branch existed. Result: **22/22 satisfied, 0 violated, 0 unclear. No residual.**

The grade was taken by executing, not only by reading — the grader ran the suite (17/17 green) and the filtered gate (`total=2, failed=0`) itself, and did the set-difference on both token lists rather than confirming the tokens it could see.

This is an author self-grade and is **advisory**. It does not substitute for the merger's independent review, and no `svc:review-*` label or verdict marker has been written.

### What to look at first

1. **The vocabulary is curated, not pattern-based — and that is forced.** The corpus already contains ASCII-ified German outside the enumerated list (`skills/story-to-web/references/04-image-prompts.md:240`, tracked as #1309). A general orthography pattern would flag it and go red on `main`, contradicting the "exits 0 on a clean base" criterion. If you disagree with the curated approach, that pre-existing line is the fact to argue with.
2. **The limit is stated, not hidden.** A green run means no enumerated form appeared, not that the corpus is correct. Both the suite header and the `CLAUDE.md` entry say so.
3. **The registration prose is deliberately token-free.** `CLAUDE.md` is inside the scanned tree, so naming a vocabulary form there would turn this suite's own `C1` case red. An earlier candidate plan did exactly that — it spelled out six forms as illustrative examples and would have shipped a suite that failed on its own registration. Worth re-checking if the entry is ever reworded.
4. **Exemptions are content-anchored on purpose.** No line-number literals appear in the suite. Case `C2` asserts each anchor still resolves, so a reword surfaces as a failure rather than silent over-suppression.

### Evidence

- 17/17 cases green; `RESULT: all de-ascii-orthography cases passed.`
- M1 executed negative case: mutant child exits 1 and reports `C1` red naming both injected forms. Both corruption styles are exercised.
- Independent red-proof outside M1: injecting into a scratch copy of a real tracked reference file yields `C1` red with exact `path:line:form`.
- Mutation recipe **executed**: `"mutated_result": "red", "restored_result": "green", "verdict": "guard_verified"`.
- `run-plugin-tests.py --filter cogni-visual` → `total: 2, passed: 2, failed: 0` (was `total: 1`).
- Full-repo sweep → **all 107 suites passed**.
- `check-breadcrumbs.py` exit 0; `check-version-bump.py` exit 0, 14 plugins scanned, 0 touched.
- `git status` clean after a run — writes only under its own `mktemp -d`.

### Reviewer note on checkout

The repo-root clone is often on an unrelated feature branch whose `cogni-visual/` still carries the full corruption set (29 word-boundary hits at the time of writing). Reviewing there will suggest the base is dirty and the green-on-main criterion is unsatisfiable. Please review against this PR's head or a clean base.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1310

**Summary:** All 22 committed contract items (20 BINDING, 2 QUALITY) are satisfied, independently re-verified by executing the suite (exit 0, 19 green cases), the plugin-tests filter (exit 0, both suites discovered, 0 failures), and hand-derived set-differences on both token vocabularies. Both deterministic pre-gates are clean. All three perspectives passed. Seven minor findings remain, none gating: six are merger-added preference notes (advisory-addition, non-gating per the gatekeeper pre-step) covering a weak lookalike-fixture sentinel, an aggregate-only non-vacuity guard, an unattributed known-blind-spot disclosure, a dead-code readability nit, an oversized CLAUDE.md registration line, and the 733-line oversized-PR flag; the seventh is an un-demoted defect-classified minor — a hyphenated-compound false negative in the slug heuristic (test-de-ascii-orthography.sh:226-251) that the validator confirms is broader coverage than contract item 15 requires, not a violation.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (7)

- **[minor]** (advisory-addition) `PR diff` — Diff is 733 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier. Mitigating context: 732 of the 733 lines are a single new self-contained test suite with no production-code surface, and the suite was executed on the PR head (exit 0, 19 green cases) rather than only read.

- **[minor]** (advisory-addition) `cogni-visual/CLAUDE.md:82` — The registration line satisfies contract item 2's shape (two-space indent, filename, two spaces, single line) but runs ~1,700 characters against the precedent test-arc-taxonomy-sync.sh entry's ~480. The tests/ tree block is an at-a-glance index; an entry this long stops functioning as one and duplicates material the suite's own 78-line header already carries better.
  - *Suggested fix:* Trim to roughly the precedent's density — what it pins, the hard assert, and the fact that it carries an executed negative case (M1) — and leave the vocabulary-extension mechanics, green-run semantics and exemption rationale to the suite header.

- **[minor]** (advisory-addition) `cogni-visual/tests/test-de-ascii-orthography.sh:17-22` — The header honestly discloses that the corpus already contains ASCII-ified German outside the vocabulary and that the suite is deliberately green on it, but does not name the tracking issue. The instance is live on this tree: skills/story-to-web/references/04-image-prompts.md:240 carries ASCII-ified German inside a worked example in exactly the file class the suite scans. This is NOT a contract violation — adding those tokens would break binding item 11 (exit 0 on base) and item 7 fixes the vocabulary to exactly the 12+6 set — but an unattributed 'tracked separately' is how a known blind spot becomes a forgotten one.
  - *Suggested fix:* Cite the tracking issue number in the header block next to 'That defect is tracked separately', so the follow-up is reachable from the file that deliberately does not catch it.

- **[minor]** `cogni-visual/tests/test-de-ascii-orthography.sh:226-251` — is_slug_context() treats a hyphen as identifier glue unconditionally, so any corrupted token inside a hyphenated German compound is silently exempted as a 'slug'. Verified by probe: a synthetic corpus containing 'Die Prozess-Qualitaet der Anlage ist dokumentiert' and 'Der Uberblick-Bericht wird an den Betrieb geschickt' produced zero findings for either, while the bare 'Stillstande' on the next line was reported. Hyphenated compounds are routine in this corpus's German prose (Ersatzteil-Lieferzeit, Wearable-Sensoren), so this is a real coverage hole rather than a theoretical one. Contract item 15 requires slugs to be exempt; this implementation is simply broader than that requirement, so it is not a contract violation.
  - *Suggested fix:* Tighten the slug heuristic so a hyphen only counts as glue when the expanded span actually looks like a slug/path/filename — e.g. require the whole expanded span to be lowercase, or require a '/' or a file extension somewhere in the span. Then add a fixture pinning that 'Prozess-Qualitaet' in German prose IS flagged while 'predictive-maintenance-fur-alle' in the X4 fixture stays exempt.

- **[minor]** (advisory-addition) `cogni-visual/tests/test-de-ascii-orthography.sh:234-236` — Dead/unreachable condition in the left-expansion loop. `line[left - 1] in "."` is an obscure spelling of `== "."`, and the guard `if line[left - 1] == "." and not (left < len(line)): break` can never fire for any left < len(line), which is every left reachable here. The behaviour is right (the trailing-dot rule is correctly implemented on the right side at line 245); the code reads as if a rule exists where none does.
  - *Suggested fix:* Delete the unreachable `not (left < len(line))` break and write the glue test as `line[left - 1] in "-_/."`, or mirror the right-side rule explicitly.

- **[minor]** (advisory-addition) `cogni-visual/tests/test-de-ascii-orthography.sh:378-424` — The non-vacuity guards assert only aggregate counts (V3: scanned > 0, V4: German-bearing > 0). Contract item 10 names two specific surfaces that must be in scope — libraries/EXAMPLE_*BRIEF.md and skills/*/references/*.md — and today they are, because the walk is unfiltered. But if a directory filter is ever added, V3/V4 would remain green while the guard quietly lost the surfaces where three of the five real corruption sites live.
  - *Suggested fix:* Extend the V-block with an assertion that the scanned set contains at least one path matching libraries/EXAMPLE_*BRIEF.md and at least one matching skills/*/references/*.md, failing the run via bail_out otherwise.

- **[minor]** (advisory-addition) `cogni-visual/tests/test-de-ascii-orthography.sh:629-645` — Contract item 14 says the lookalike case 'must fail if word-boundary discipline is later removed'. G1's fixture cannot: none of Stillstand, Stillstandstag, Stillstandskosten, Vorwarnung, Prozesse, Ersatzteil-Lieferzeit contains any of the 18 enumerated tokens as a substring, so G1 stays green with or without the lookarounds. Verified by patching a copy of the suite to a bare substring matcher: G1 stayed green and only C1 went red, via an ambient corpus hit. The item's primary requirement — a named case asserting those five forms are not matched — IS met, so this is not a violation; but the fixture is a weaker sentinel than intended.
  - *Suggested fix:* Add one or two lookalikes to the G1 fixture that legitimately CONTAIN an enumerated token as a substring, e.g. 'Der Krafteinsatz wird geplant.' (contains Krafte) and 'Die Nachstellung der Anlage ist dokumentiert.' (contains nachste). Both are correct German, both stay green under the current lookarounds, and both turn G1 red the moment word-boundary anchoring is dropped.

### Positive observations

- Contract items 1/2/3 clean and independently verified: the path has exactly one segment under cogni-visual/tests/ so both of run-plugin-tests.py's non-recursive globs see it; the file carries mode 100755 matching the precedent suite; CLAUDE.md gains the registration line in the same tests/ tree block; the diff is exactly two files and touches neither plugin.json nor marketplace.json. (plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer)
- Contract items 11 and 20 re-verified by execution, not taken on trust: `bash cogni-visual/tests/test-de-ascii-orthography.sh` exits 0 with 19 green cases and a `RESULT:` summary, and `python3 scripts/run-plugin-tests.py --filter cogni-visual` exits 0 with data.total=2, data.failed=0, both suites listed passed=true. (both perspectives, independently)
- Contract item 4 verified behaviourally: the suite ran from the worktree root and `git status --porcelain` was empty afterwards, so M1's mutation genuinely never touches the tracked tree. No curl/wget/gh/nc/ssh anywhere, no positional arguments consumed, HERE self-location plus mktemp -d with an EXIT trap, and bash -n is clean. (plugin-dev:plugin-validator)
- Contract items 7 and 8 are exact set matches with no survivors in either direction — `comm -23` against the contract's exact 12-token dropped-vowel and 6-token digraph sets produced ZERO survivors; none missing, none spurious. (cogni-service:contract-grading-reviewer)
- Contract item 9 holds with real anchoring: scan.py builds PATTERNS using explicit `(?<![A-Za-z0-9_ÄÖÜäöüß])token(?![A-Za-z0-9_ÄÖÜäöüß])` lookaround guards, not a bare substring test. (cogni-service:contract-grading-reviewer)
- The vocabulary is self-policing beyond what the contract demanded: W1 folds each correct_form under its declared style and rejects a row whose ascii_token is not what it actually folds to, and W2 set-differences the vocabulary against the fixture-exercised tokens in BOTH directions — so the vocabulary cannot grow without its proof growing with it. (plugin-dev:plugin-validator)
- Contract items 12 and 13 are exemplary: both exemptions key on content substrings ('ae/oe/ue', 'Replace German umlauts'), grep confirms zero 253/254/56 line-number literals in the file, and C2 re-asserts on every run that each anchor still resolves — so a reword surfaces as a red case instead of rotting into a false positive. (both perspectives)
- Contract item 15 is handled the hard, correct way: 04-image-prompts.md is scanned line-by-line rather than excluded wholesale, so it stays a live corruption site while the image_prompt: block-scalar rule, the frontmatter description: rule, inline-code/URL blanking and the de-context guard on the two English-colliding tokens (fur, uber) protect the English prose beside it. (both perspectives)
- Contract items 16/17/18/19 are all met by a single well-built M1: it copies the corpus into its own mktemp -d, injects one dropped-vowel and one digraph token into the COPY, re-invokes the same file via VISUAL_DE_ASCII_ROOT behind a VISUAL_DE_ASCII_MUTANT recursion guard, and requires the child to exit non-zero AND emit `FAIL: C1` naming both injected tokens. It goes further and cross-checks the child's emitted case ids against ALL_CASES in both directions. (both perspectives)
- Contract items 5 and 6 hold: pass()/fail() emit the colon form `ok: <id>` / `FAIL: <id>` matching the shared harness classifiers, the summary line is prefixed `RESULT:` not `FAIL:`, and all 30 call sites use bare single-token case ids. (cogni-service:contract-grading-reviewer)
- Contract item 14's primary requirement is met by an explicit named fixture case (G1) rather than an appeal to ambient tree-greenness. (both perspectives)
- The bail_out design closes the failure mode most suites of this kind get wrong: when a non-vacuity guard trips, every downstream case is reported red under its own id in the harness's colon form, so an abandoned run can never be mistaken for coverage. (plugin-dev:plugin-validator)
- QUALITY item 21 substantially exceeds the bar: the 78-line header names the defect class, explains why prose did not hold, states precisely what a green run does and does NOT mean, and justifies the colon label shape against the harness classifiers. QUALITY item 22 is vacuous — the PR ships no mutation-check recipe. (both perspectives)
- Both deterministic pre-gates ran clean: check-frontmatter.sh reports 0 offenders on cogni-visual, and check-version-bump.sh reports status ok / 0 violations vs origin/main.

### Recommendation

---

PR #1341, opened 2026-08-12
cogni-portfolio: Agent-grant check counts handoff citations

### LINKED ISSUE #1333: cogni-portfolio: the Agent-grant check is blind to skills that cite the handoff reference without naming the agent
## Plugin
cogni-portfolio

## Current behavior

`test_dispatch_sites_grant_agent` in `cogni-portfolio/tests/test-dashboard-handoff.sh` exists to catch a skill that documents a `dashboard-refresher` dispatch without granting `Agent` in `allowed-tools` — a documented step that cannot run. It derives its scan surface from a purely literal grep:

```
grep -nF 'dashboard-refresher' "$f"
```

But the canonical end-of-step handoff **deliberately never names the agent**. The wording every wired skill carries is:

> Regenerate the dashboard and hand the user its link, every time, per `$CLAUDE_PLUGIN_ROOT/references/dashboard-handoff.md`. Print the link; do not open a browser.

That omission is load-bearing for two other cases (`test_handoff_not_duplicated` fails any handoff section naming the dispatch parameters, and keeping the name out also keeps the section clear of `test_open_browser_optin`'s 8-line window). The side effect is that **a skill whose only dispatch is the canonical citation is structurally invisible to the grant check.**

## Why this is not hypothetical

It has now recurred twice.

- #1320 filed / #1323 fixed exactly this class for three skills that *did* name the agent.
- The six pipeline and ops skills wired in #1316 (PR #1332) do **not** name it, and five of the six lacked `Agent` entirely. The grant check stayed green throughout. The only reason it was caught is that a human-directed triage pass audited the `allowed-tools` lines by hand.

PR #1332 closed the instance by folding a grant assertion into its own new `test_pipeline_skills_cite_handoff`, scoped to a fixed six-skill roster. That fixes those six. It does not fix the class: a seventh skill that cites the reference tomorrow is invisible again, and nothing fails.

## Desired behavior

Widen the dispatch-line derivation in `test_dispatch_sites_grant_agent` so a citation of the canonical reference counts as a documented dispatch site, alongside naming the agent directly. Roughly, replace the fixed-string grep with an ERE matching either `dashboard-refresher` or `references/dashboard-handoff\.md`.

Everything downstream already works unchanged: the governing-SKILL.md mapping (a dispatch in `skills/<x>/references/*.md` is governed by `skills/<x>/SKILL.md`) and the whole-token `,Agent,` match on the `^allowed-tools:` line both stay as they are. The change is the surface, not the assertion.

The payoff is that the grant becomes **derived rather than rostered** — skill fifteen is covered with no edit — and the roster in `test_pipeline_skills_cite_handoff` can then drop its grant arm and go back to asserting only the citation.

## Verification that this is verdict-neutral on today's tree

Checked while planning #1316, and worth re-checking at implementation time rather than trusting:

- All eight entity SKILL.md files already contain **both** literals and already grant `Agent`.
- No `skills/*/references/*.md` file cites the reference path; the only one carrying `dashboard-refresher` is `propositions/references/quality-gates.md`, already governed and already granted.
- `README.md` and `references/dashboard-handoff.md` both cite the path but sit **outside** the case's three-glob surface (`agents/*.md`, `skills/*/SKILL.md`, `skills/*/references/*.md`), so neither is pulled in.
- After PR #1332 the six pipeline/ops skills all grant `Agent`, so widening the surface should newly scan them and find them clean.

## Alternatives considered

- **Leave it; the new roster covers the six.** Rejected — that is the status quo this issue exists to change. Two recurrences of one class say the guard is the defect, not the instances.
- **Have the canonical section name the agent so the existing grep finds it.** Rejected outright: it would fail `test_handoff_not_duplicated`, which forbids a handoff section containing `dashboard-refresher`, `project_dir`, `plugin_root` or `open_browser`, and would risk pulling the section into `test_open_browser_optin`'s offender window. The omission is deliberate.
- **A separate case scanning only for reference citers.** Rejected as redundant — it would duplicate the governing-file and whole-token logic the existing case already has.

## Acceptance criteria

- [ ] A skill whose only documented dispatch is a `references/dashboard-handoff.md` citation, and whose `^allowed-tools:` line lacks a whole `Agent` token, makes `test_dispatch_sites_grant_agent` go RED; falsified by that case staying green on such a skill.
- [ ] The widened surface is proved by a registered mutation in `cogni-portfolio/scripts/mutation-check.sh`: strip `Agent` from a skill that cites the reference but does **not** name the agent (any of the six wired in #1316), and the case must go RED on the mutant and GREEN on HEAD. The existing `mutation_dispatch_agent_grant` targets `products/SKILL.md`, which names the agent, so it does not cover this arm.
- [ ] The whole-token `,Agent,` match stays scoped to the `^allowed-tools:` line — never a body-wide grep for `Agent`, which every dispatching skill's prose satisfies vacuously.
- [ ] `test_pipeline_skills_cite_handoff` drops its now-redundant grant arm and asserts only the citation and section presence, so the grant is asserted in exactly one place.
- [ ] The full suite and `bash cogni-portfolio/scripts/mutation-check.sh` stay green, and no currently-passing skill is newly flagged.

_Affected components: cogni-portfolio/tests/test-dashboard-handoff.sh, cogni-portfolio/scripts/mutation-check.sh_

_Complexity hint: low_

_Filed from the #1316 resolution as the rejected root-cause alternative — recorded rather than dropped because the class has now recurred twice._

claude-cogni-work:
Closes #1333.

## Summary

- `test_dispatch_sites_grant_agent` derives its dispatch lines with an ERE alternation (`dashboard-refresher|references/dashboard-handoff\.md`) instead of a fixed-string grep for the agent name, so a skill whose only documented dispatch is the canonical citation is now governed by the grant rule. Governed skills move **11 → 17** — the 11 name-the-agent skills plus exactly the six pipeline/ops skills. `propositions` matches via both its `SKILL.md` and its `references/quality-gates.md` and is collapsed by the existing `$seen` dedupe, so the total is 17, not 18. The pass message renders `$governed`, so there was no hardcoded count to edit.
- Everything downstream of the grep is untouched: the three `find` globs, the `[ "$f" = "$AGENT" ]` skip, the governing-`SKILL.md` `case`, the `$seen` dedupe, `grep -m1 '^allowed-tools:'`, and the whole-token match on the grant line.
- `test_pipeline_skills_cite_handoff` drops its grant arm and asserts only the citation and the handoff section. The rationale that claimed the grant "cannot live there" and both result messages are corrected in the same commit.
- `mutation-check.sh` gains `mutation_citation_only_agent_grant` (mutation 9) plus its registration line and header entry. It strips the **mid-list** `Agent` token from `portfolio-ingest/SKILL.md` — the existing mutation 6 targets `products/SKILL.md`, which names the agent, and its `, Agent$` anchor would match nothing on any of the six citation-only skills.

## Not in the issue, added deliberately: per-arm liveness

Widening a detector to two forms introduces a failure mode the issue does not mention, and this diff would otherwise make it worse rather than better.

The existing `dispatch_count -eq 0` check is a **whole-detector** floor. The name arm satisfies it forever on its own, so the citation arm can die — a renamed reference, a lost escape, a careless regex edit — and the case stays green while the six citation-only skills silently revert to ungoverned. That is the exact defect this PR exists to remove. And because the same diff deletes the rostered assertion that used to cover those six, the net trade without a floor is *a CI-enforced assertion for one that can fail silently* (`mutation-check.sh` is deliberately outside CI, per its own header).

So each arm now carries its own non-zero floor, matching `test_invocation_surface_intact` in `tests/test-bash-grant.sh`, whose header states the same rationale for the same reason. Verified both directions:

- Killing the citation arm → `FAIL: … a derivation arm went dead: name hits=12 citation hits=0`.
- On the pre-floor version of this branch the same edit left the **full suite exiting 0**.

The citation floor is built from `HANDOFF_PATH_LITERAL`, so a rename of the reference now fails loudly here instead of quietly narrowing the surface. Floors, never exact counts, so honest churn cannot trip them.

## Scope decisions

- **In:** `cogni-portfolio/tests/test-dashboard-handoff.sh` and `cogni-portfolio/scripts/mutation-check.sh` only. No `SKILL.md` or agent content changes, so `scripts/check-breadcrumbs.py` (which scans `*/skills/*/SKILL.md` and `*/agents/*.md`) is structurally unaffected and no new maintainer breadcrumb is introduced.
- **Deliberately not done — the DRY refactor.** `test_open_browser_optin`'s derivation block stays byte-identical to base. Hoisting both onto a shared helper would drag citation-only skills into the browser-opt-in scan, a different rule with a different surface.
- **Deliberately not done — widening the `find` globs.** The reference file itself has no governing `SKILL.md`, so admitting it to the surface is silently skipped: noise instead of a red case. The grep is the correct lever.
- **Considered and skipped — building the ERE from `HANDOFF_PATH_LITERAL`.** A review pass proposed `grep -nF -e 'dashboard-refresher' -e "$HANDOFF_PATH_LITERAL"`, which reuses the constant and makes the dot literal by construction. It was skipped because the issue and the acceptance contract both specify an ERE alternation with an escaped dot, and substituting the raw constant into an ERE would turn its `.` into a wildcard (`references/dashboard-handoffXmd` would start matching), falsifying the comment's own claim. The rename-fragility that motivated the proposal is instead covered by the liveness floor above, which *is* keyed on the constant. Flagging it so the choice is visible rather than accidental.
- **Left alone —** `test_resume_shows_dashboard_row` keeps its own `^allowed-tools:.*Agent` check on `portfolio-resume`. It is a narrower per-skill claim inside a row assertion, not a second roster; the new comment is worded to say so rather than to claim sole ownership.
- **Version pin in the recipe block.** The new suite-header recipe reuses the `cogni-service/0.0.383` path the five sibling recipes already use (verified present in the local cache, so the recipe replays as written). The installed cogni-service is newer, so that pin is already mildly stale for all six blocks; refreshing it repo-wide is a separate change, and introducing a third value here would be worse than matching the sibling convention.
- No version field touched in `cogni-portfolio/.claude-plugin/plugin.json` or the repo-root `.claude-plugin/marketplace.json` — the post-merge workflow owns the bump.

## Test plan

All run on this branch, against `origin/main` @ `2445fd1b`.

- [x] `bash cogni-portfolio/tests/test-dashboard-handoff.sh` exits 0, no `FAIL:`, and the grant case reads `all 17 dispatching skills grant Agent (26 dispatch lines scanned)`.
- [x] `bash cogni-portfolio/scripts/mutation-check.sh` exits 0, prints `All mutations caught.`, exactly nine `OK: mutation caught` lines and zero `FALSIFIER:`.
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` exits 0 — 6 suites, none failed.
- [x] **Red-proof, both directions.** Strip the mid-list `Agent` token from `portfolio-ingest/SKILL.md`, then run the single case: against the **base** suite it exits 0 (`all 11 dispatching skills grant Agent` — the mutant is invisible, which is the bug), and against **this** suite it exits 1 naming `skills/portfolio-ingest/SKILL.md:missing-Agent`. This is what proves the new mutation exercises the new arm rather than the pre-existing one.
- [x] **Arm-liveness proof.** Deleting the citation alternative makes the case exit 1 (`citation hits=0`); on the pre-floor version of this branch the same edit left the suite green.
- [x] Shape checks: `grep -cF "grep -nF 'dashboard-refresher'"` → 1 (2 on base); `grep -cF ',Agent,'` → 1 (2 on base, at 449 and 779; the survivor is the one inside `test_dispatch_sites_grant_agent`); `grep -c '|| overall=1'` → 9 (8 on base).

**One correction to the acceptance contract, not a deviation.** Contract item 14 asserts the suite prints "12 `ok:` lines". Both base and head print **13** — the number was wrong when the contract was derived. The case count is unchanged by this diff (`ALL_TESTS` untouched), no `FAIL:` is printed, and the grant case reads 17, so the substance of the item holds.

## Mutation recipe

Proves the newly added citation-only arm has teeth. Replayable as written from the repo root; the target names the agent nowhere, so it is invisible to the base-ref derivation and red only after this change.

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.383/scripts/mutation-check.sh --root . --file cogni-portfolio/skills/portfolio-ingest/SKILL.md --expr 's#Grep, Bash, Agent, Skill#Grep, Bash, Skill#' --test 'bash cogni-portfolio/tests/test-dashboard-handoff.sh test_dispatch_sites_grant_agent' --case test_dispatch_sites_grant_agent
```

The searched literal occurs exactly once in the target (the frontmatter grant line) and is metacharacter-free, which matters because the harness feeds the expr to `perl -0pi`. The replacement is deliberately **disjoint**: it carries no `Agent` token, so the grant is genuinely gone — a replacement that still contained it would leave the case green and the mutation would survive unnoticed. `portfolio-verify` carries a byte-identical `allowed-tools` tail, which is harmless because `--file` scopes the harness to one file. The in-repo equivalent is registered as `mutation_citation_only_agent_grant`.

## Noted, not filed — the same class one level up

Surfaced by the review pass, verified as **latent, not live**, and left out of this diff: both derived guards (`test_dispatch_sites_grant_agent` here and `test_invocation_sites_grant_bash` in `test-bash-grant.sh`) scan `skills/*/references/*.md` but **exclude plugin-level `cogni-portfolio/references/*.md`**. That directory holds the shared contracts — `references/data-model.md` names `scripts/register-solution-candidates.py`, and ten skills cite that file. A skill whose only documented invocation were "via the path described in `references/data-model.md`" would be ungoverned by the Bash-grant guard: the identical indirection class this PR fixes for the Agent grant.

Nothing is falsely green today — `portfolio-canvas`, `portfolio-ingest` and `portfolio-setup` each name the script path directly — so this is a follow-up, not a blocker. Recording it here rather than filing an issue from a single review pass; a maintainer may want it filed against both suites.

Also observed and left alone as pre-existing drift: `mutation-check.sh`'s in-file section banners are mis-numbered — line 267 and line 322 both read `Mutation 5`. The new block is banner-numbered 9 and refers to its sibling by function name rather than ordinal to avoid compounding the ambiguity.

### DIFF
--- a/cogni-portfolio/scripts/mutation-check.sh
+++ b/cogni-portfolio/scripts/mutation-check.sh
@@ -31,6 +31,10 @@
 #      allowed-tools line of skills/portfolio-canvas/SKILL.md, a skill that
 #      documents script invocations; expect test_invocation_sites_grant_bash
 #      to go RED.
+#   9. citation-only dispatch Agent grant: strip the mid-list Agent token from
+#      the allowed-tools line of skills/portfolio-ingest/SKILL.md, a skill whose
+#      only documented dispatch is the canonical handoff citation (it never
+#      names the agent); expect test_dispatch_sites_grant_agent to go RED.
 #
 # Kept under scripts/ (NOT tests/) deliberately: run-plugin-tests.py auto-discovers
 # tests/*.sh and would run this as a normal suite; this is a manual meta-check that
@@ -498,6 +502,72 @@ PY
 }
 
 
+# --- Mutation 9: citation-only dispatch Agent grant ------------------------
+# The citation-only counterpart of mutation_dispatch_agent_grant, which targets
+# products/SKILL.md and so exercises only the arm the case could already see.
+# portfolio-ingest reaches the scan surface solely by citing the canonical handoff
+# reference, and is not otherwise mutated here, so the two stay independent.
+mutation_citation_only_agent_grant() {
+  local target="$PLUGIN_DIR/skills/portfolio-ingest/SKILL.md"
+  local suite="$PLUGIN_DIR/tests/test-dashboard-handoff.sh"
+  local test_name="test_dispatch_sites_grant_agent"
+  for f in "$target" "$suite"; do
+    [ -f "$f" ] || { echo "FAIL: missing $f" >&2; return 1; }
+  done
+
+  # Baseline: the target test must be GREEN before we mutate, or a later red
+  # tells us nothing.
+  if ! bash "$suite" "$test_name" >/dev/null 2>&1; then
+    echo "FAIL: baseline $test_name is already red before mutation" >&2
+    return 1
+  fi
+
+  local backup; backup="$(mktemp)"
+  cp "$target" "$backup"
+
+  if ! python3 - "$target" <<'PY'
+import re, sys
+path = sys.argv[1]
+src = open(path, encoding="utf-8").read()
+# MID-LIST, deliberately not the trailing `, Agent$` anchor mutation 6 uses: none
+# of the citation-only skills ends its grant line with Agent, so that anchor would
+# match nothing here and the mutation could never run.
+mutated, n = re.subn(
+    r"^(allowed-tools: .*), Agent,(.*)$",
+    r"\1,\2",
+    src,
+    count=1,
+    flags=re.MULTILINE,
+)
+if n != 1:
+    sys.stderr.write("allowed-tools line with a mid-list Agent token not found — cannot mutate.\n")
+    sys.exit(10)
+# Disjointness is the whole point, so assert it rather than trusting the regex: a
+# replacement that left any whole Agent token would keep the grant, keep the case
+# green, and let this mutation survive unnoticed.
+line = next(l for l in mutated.splitlines() if l.startswith("allowed-tools:"))
+if "Agent" in [t.strip() for t in line.split(":", 1)[1].split(",")]:
+    sys.stderr.write("mutant still grants Agent — mutation is not disjoint.\n")
+    sys.exit(11)
+open(path, "w", encoding="utf-8").write(mutated)
+PY
+  then
+    echo "FAIL: citation-only Agent-grant mutation could not be applied" >&2
+    cp "$backup" "$target"; rm -f "$backup"; return 1
+  fi
+
+  local rc=0
+  if bash "$suite" "$test_name" >/dev/null 2>&1; then
+    echo "FALSIFIER: mutation survived — $test_name stayed GREEN with a citation-only skill's Agent grant stripped" >&2
+    rc=1
+  else
+    echo "OK: mutation caught — $test_name goes red when a skill whose only dispatch is the canonical citation loses its Agent grant"
+  fi
+  cp "$backup" "$target"; rm -f "$backup"   # always restore
+  return "$rc"
+}
+
+
 mutation_commercial_consolidation || overall=1
 mutation_solution_candidates || overall=1
 mutation_dashboard_no_skip_path || overall=1
@@ -506,6 +576,7 @@ mutation_resume_dashboard_row || overall=1
 mutation_dispatch_agent_grant || overall=1
 mutation_pipeline_handoff_citation || overall=1
 mutation_invocation_bash_grant || overall=1
+mutation_citation_only_agent_grant || overall=1
 
 if [ "$overall" -eq 0 ]; then
   echo "All mutations caught."
--- a/cogni-portfolio/tests/test-dashboard-handoff.sh
+++ b/cogni-portfolio/tests/test-dashboard-handoff.sh
@@ -19,7 +19,7 @@
 #   test_refresher_returns_url       a successful result carries a clickable url
 #   test_entity_skills_cite_handoff  all eight entity skills end by citing the canonical handoff
 #   test_pipeline_skills_cite_handoff
-#                                    the six pipeline and ops skills cite it and grant Agent
+#                                    the six pipeline and ops skills each cite it exactly once
 #   test_secondary_paths_reach_handoff
 #                                    alternate write paths point back to that terminal section
 #   test_handoff_not_duplicated      the handoff is authored once, and no skill restates it
@@ -28,7 +28,7 @@
 #   test_midflow_offers_survive      the pre-existing review offers still open on request
 #   test_resume_shows_dashboard_row  portfolio-resume surfaces a flag-driven dashboard row
 #   test_handoff_does_not_open       the resume generation offer dispatches without opening a browser
-#   test_dispatch_sites_grant_agent  every skill documenting a dispatch grants Agent in allowed-tools
+#   test_dispatch_sites_grant_agent  every skill dispatching by name or citation grants Agent
 #
 # Usage: bash cogni-portfolio/tests/test-dashboard-handoff.sh [test_name ...]
 #   No args -> run every test (the CI path). One or more names -> run only those
@@ -106,6 +106,31 @@
 #   `Agent` would leave the grant intact, the case green, and the mutation would
 #   survive unnoticed. A documentation file again, for the reason above.
 #   The in-repo equivalent is registered as mutation_dispatch_agent_grant in
+#   cogni-portfolio/scripts/mutation-check.sh.
+#
+# Mutation recipe — test_dispatch_sites_grant_agent (citation-only arm). Same
+#   SHARED harness, same classify-on-labels contract, replayable as written from
+#   the repo root.
+#   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.383/scripts/mutation-check.sh \
+#     --root . \
+#     --file cogni-portfolio/skills/portfolio-ingest/SKILL.md \
+#     --expr 's#Grep, Bash, Agent, Skill#Grep, Bash, Skill#' \
+#     --test 'bash cogni-portfolio/tests/test-dashboard-handoff.sh test_dispatch_sites_grant_agent' \
+#     --case test_dispatch_sites_grant_agent
+#   The recipe above covers the OTHER arm of the same case. Its sibling recipe
+#   mutates products/SKILL.md, which names the agent outright; portfolio-ingest
+#   names it nowhere — neither its SKILL.md nor any of its references/*.md carries
+#   the literal — so it enters the surface only through the canonical citation.
+#   Before this case derived both forms, the mutant here survived: the skill was
+#   invisible to the scan, so stripping its grant changed nothing.
+#   `Grep, Bash, Agent, Skill` occurs exactly once in that file and is
+#   metacharacter-free, which matters because the harness feeds the expr to
+#   perl -0pi. The replacement is deliberately DISJOINT from the searched string
+#   and retains no whole Agent token — one that still granted Agent would leave
+#   the case green and the mutation would survive unnoticed. portfolio-verify
+#   carries a byte-identical allowed-tools tail, which is harmless because --file
+#   scopes the harness to one file.
+#   The in-repo equivalent is registered as mutation_citation_only_agent_grant in
 #   cogni-portfolio/scripts/mutation-check.sh.
 #
 # Mutation recipe — test_secondary_paths_reach_handoff. Same SHARED harness, same
@@ -419,15 +444,17 @@ test_entity_skills_cite_handoff() {
 # do. Same exactly-one-citation discipline as the entity case, for the same reason:
 # the single-occurrence count keeps the registered mutation's anchor unambiguous.
 #
-# This case also carries the Agent-grant assertion for its roster, which looks like
-# it belongs in test_dispatch_sites_grant_agent and cannot live there. That case
-# derives its surface from files containing the literal `dashboard-refresher`, and
-# the canonical citation deliberately never names the agent — so these six are
-# invisible to it while still depending on the grant to execute the handoff at all.
+# The Agent grant is NOT asserted here. test_dispatch_sites_grant_agent derives its
+# dispatch lines from either form — naming the agent, or citing the canonical
+# reference — so these six sit inside that case's surface and it covers their grant
+# without a roster. Keeping a rostered copy too would mean the rostered one fails
+# first, and only ever for the skills someone remembered to list.
+# (test_resume_shows_dashboard_row still checks portfolio-resume's own grant as part
+# of its row assertion; that is a narrower claim about one skill, not this roster.)
 test_pipeline_skills_cite_handoff() {
   pipeline_surface_ok test_pipeline_skills_cite_handoff || return
 
-  local offenders="" skill f count grant
+  local offenders="" skill f count
   for skill in $PIPELINE_SKILLS; do
     f="$PLUGIN_DIR/skills/$skill/SKILL.md"
     count="$(grep -cF "$HANDOFF_PATH_LITERAL" "$f")"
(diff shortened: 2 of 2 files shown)


claude-cogni-work:
## Ship-ready

Author instance (`--role author`) has handed this off. Draft, carrying `svc:needs-review` — awaiting the merger instance's independent review. No verdict label, no `reviewed-head-sha` marker, and no draft→ready flip has been applied here by design.

**Handoff preflight:** exit 0, clean — 3 gating checks passed (`run-tests`, `version-bump`, `mutation-recipe`), 0 failed, 4 skipped. Every skip is `not_applicable` because the diff touches no `SKILL.md` and no `agents/*.md`: `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length`. The `mutation-recipe` instrument reports the recorded recipe is "runnable as written".

**Author self-grade against the issue-time contract** (`#1333`, `origin=triage`, derived before the branch was cut): **21/21 satisfied — pass.** No violated items, no unclear items, so no fix pass ran. **Residual: none.**

Two things the merger should look at rather than rediscover:

1. **Contract item 14 says the suite prints "12 `ok:` lines". Base and head both print 13.** The number was wrong when the contract was derived — it is not a regression this PR introduced. Both the author and the grader verified the base count independently. The substance of the item holds: head exits 0, prints zero `FAIL:`, the case count is unchanged, and the grant case reads `all 17 dispatching skills grant Agent (26 dispatch lines scanned)` — satisfying the item's own fallback clause (recomputed union, strictly greater than the pre-change 11).

2. **The diff adds one thing the contract never asked for: a per-arm liveness floor** in `test_dispatch_sites_grant_agent`. The grader graded it satisfied and explicitly flagged the tension rather than absorbing it silently, so it is stated here for a human:
   - *Why it is here:* contract item 8 removes the rostered grant assertion. Without a per-arm floor, the surviving `dispatch_count -eq 0` check is satisfied by the name arm forever, so the citation arm could die — a renamed reference, a lost escape — and the six citation-only skills would silently revert to ungoverned with CI green. That is the exact defect this PR exists to remove, and `mutation-check.sh` is deliberately outside CI, so nothing else would catch it.
   - *Why it does not breach item 6:* the four artifacts item 6 enumerates (the governing-file `case`, the `$seen` dedupe, `grep -m1 '^allowed-tools:'`, the whole-token `,Agent,` match) are byte-identical to base. The floor sits between the existing `dispatch_count` check and that block.
   - *The adversarial reading, stated plainly:* item 6 opens with a universal ("everything downstream of the grep is unchanged"), and this is new logic positioned downstream of the widened grep that no item requested. The grader resolved it toward the enumeration because this contract uses colon-then-list as its operative definition elsewhere (items 8, 9, 12). A merger who reads item 6's opening clause as controlling would be entitled to call it out-of-scope.

**Verification run on this branch** (base `origin/main` @ `2445fd1b`, head `b37b59db`):

| Check | Result |
|---|---|
| `bash cogni-portfolio/tests/test-dashboard-handoff.sh` | exit 0, no `FAIL:`, grant case reads 17 |
| `bash cogni-portfolio/scripts/mutation-check.sh` | exit 0, `All mutations caught.`, 9 `OK:`, 0 `FALSIFIER:` |
| `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` | exit 0, 6 suites, none failed |
| Red-proof (mutant vs base suite) | exit 0 — `all 11 dispatching skills grant Agent`, mutant **invisible** (the bug) |
| Red-proof (mutant vs head suite) | exit 1 — `skills/portfolio-ingest/SKILL.md:missing-Agent`, **caught** |
| Arm-liveness proof (citation arm deleted) | exit 1 — `a derivation arm went dead: name hits=12 citation hits=0` |

The two red-proofs are the load-bearing ones: they show the new mutation exercises the newly added arm rather than the pre-existing one.

**Out of scope, recorded not filed:** both derived guards exclude plugin-level `cogni-portfolio/references/*.md` from their scan surface — the same indirection class, latent rather than live. Detail is in the PR body under "Noted, not filed"; a maintainer may want it filed against both suites.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1341

**Summary:** All perspectives passed; contract-grading-reviewer found the diff satisfies all 21 review-contract items with live test execution, and plugin-validator's sole finding is a merger-added minor preference (advisory-addition) that does not gate.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (advisory-addition) `cogni-portfolio/tests/test-dashboard-handoff.sh:797-800` — The two new arm floors grep the whole `path:line:content` entry, not just the matched content, so the file path itself can satisfy an arm. Today no file under the scan surface has a path containing `dashboard-refresher` or `references/dashboard-handoff.md`, so both floors are honest right now. If such a references file is ever added, the name floor could read non-zero from the path alone while the name arm is actually dead — the exact vacuity the floors exist to prevent.
  - *Suggested fix:* Compute the floors over the matched content only, e.g. strip the `file:lineno:` prefix before counting (`sed 's/^.*:[0-9][0-9]*://'`) or derive the arms with the same per-file grep that built $dispatch_lines.

### Positive observations

- Items 1-6: the functional diff is exactly the two named files; the three find globs, the $AGENT skip, the governing-file mapping, the $seen dedupe, `grep -m1 '^allowed-tools:'` and the whole-token `,Agent,` match all carry no diff hunk; the widened line escapes the dot; test_open_browser_optin is byte-identical to base, so the 8-line open-browser window was not widened by a shared-helper refactor — the top risk in the watchlist is avoided. (both perspectives)
- Items 7-9: `grep -cF ',Agent,'` is 1 on head (2 on base); test_pipeline_skills_cite_handoff no longer reads `^allowed-tools:` at all, and the rationale comment, both result messages and file-header line 22 were corrected in the same commit. (both perspectives)
- Items 10-13: mutation-check.sh gains both `mutation_citation_only_agent_grant` and its registration line (`|| overall=1` count 8 -> 9); the target portfolio-ingest cites the reference and never names the agent; the strip regex is mid-list and disjoint from the trailing anchor at line 357; the sibling shape (existence check, GREEN baseline pre-check, mktemp backup, restore on every exit path, canonical OK/FALSIFIER lines) is followed. (both perspectives)
- Items 14-16 verified by live execution: the suite exits 0 with no FAIL and prints `all 17 dispatching skills grant Agent (26 dispatch lines scanned)` (base printed 11); mutation-check.sh exits 0 with 'All mutations caught.', nine OK lines and zero FALSIFIER, leaving the tree clean; `run-plugin-tests.py --filter cogni-portfolio` exits 0 across all 6 suites. (both perspectives)
- Item 14 caveat resolved, not waived: the contract's 'prints 12 ok: lines' sub-clause does not hold literally (head prints 13). Confirmed pre-existing and diff-unrelated by checking out base 2445fd1 in an isolated worktree and running the identical suite there, which also prints 13 — a plan-time miscount, not a regression this diff introduced. (cogni-service:contract-grading-reviewer)
- Item 17 red-proof reproduced by hand: stripping the Agent token from portfolio-ingest's allowed-tools line makes test_dispatch_sites_grant_agent exit 1 naming `skills/portfolio-ingest/SKILL.md:missing-Agent`, while the SAME mutation on the base suite stays GREEN — direct evidence the widened arm is what caught it. Restored; worktree clean. (both perspectives)
- Items 18-21: the suite header gains the citation-only mutation recipe naming its in-repo function, with a metacharacter-free single-occurrence anchor and a disjoint replacement; no SKILL.md or agents/*.md is touched, so no maintainer breadcrumb can have been introduced; neither plugin.json nor marketplace.json is modified; the unchanged $seen dedupe collapses propositions to one entry, evidenced by the governed total of 17 rather than 18. (both perspectives)

### Recommendation

---

Issue #1357, opened 2026-08-13
Repo-root test suites emit ANSI-wrapped PASS:/FAIL:, defeating mutation-harness parsing


claude-cogni-work:
## Problem

Every repo-root `tests/*.sh` suite wraps its result lines in ANSI colour:

```bash
red()   { printf '\033[31m%s\033[0m\n' "$1"; }
green() { printf '\033[32m%s\033[0m\n' "$1"; }
```

so an emitted line is `\033[31mFAIL: <case>\033[0m`, not `FAIL: <case>`.

Any harness that classifies a suite's captured output by matching
`^[[:space:]]*FAIL:[[:space:]]+<case>` therefore matches **neither** the red nor the
green line and reports `case_not_found` instead of a verdict. The practical effect is
that these suites cannot have their teeth verified: a mutation check against them
reports "case not found" rather than proving the guard under test is load-bearing.

## Scope

Five repo-root suites still carry the ANSI emitters:

- `tests/test_apply_version_bump.sh`
- `tests/test_check_breadcrumbs.sh`
- `tests/test_check_plugin_inventory.sh`
- `tests/test_check_version_bump.sh`
- `tests/test_run_plugin_tests.sh`

(Roughly seventeen more under `cogni-knowledge/tests/` share the pattern, plus the
plugin-scoped helper `cogni-knowledge/tests/fixtures/test_helpers.sh`, whose own header
documents the byte-identical inlining of `red()`/`green()` as the anti-pattern it was
created to end — it ships the ANSI-always variant, so sourcing it does not fix this.)

`tests/test_check_external_dispatch.sh` was fixed in passing while resolving #1346, and
is the reference shape: plain `printf '%s\n'` emitters, unconditionally.

## Proposed change

Emit plain, uncoloured `PASS:` / `FAIL:` labels in the remaining repo-root suites,
matching the convention already used by the harness-parsable suites elsewhere in the
repo (`cogni-consult/tests/test_reference_pointers.sh`,
`cogni-visual/tests/test-arc-taxonomy-sync.sh`).

Do **not** gate the choice on `[ -t 1 ]`. A TTY-conditional emitter makes parsability
environment-dependent: run the suite under a pty or `script(1)` and the colour — and
the failure — come back, through a channel nothing tests. That was considered and
rejected in #1346 for exactly this reason.

## Acceptance criteria

- [ ] Every repo-root `tests/*.sh` suite emits result lines beginning with a literal
[...]

Issue #1347, opened 2026-08-13
Add wiki-namespace-sync guard; remove ~8 stale pages for already-deleted plugins


sdh07:
## Problem

`cogni-workspace:ask` has a strict contract: answer only from the bundled wiki, never from memory, and cite every claim with a `[[wikilink]]`. That contract is only as good as the wiki's freshness — and nothing currently enforces it.

The evidence says it does not stay fresh. After the `cogni-research` / `cogni-wiki` / `cogni-consulting` absorptions, `cogni-workspace/wiki/` was cleaned but the root `wiki/` tree was **not**. It still contains stale pages for plugins that no longer exist:

- `wiki/wiki/pages/agent-cogni-research-*.md` (7 pages)
- `wiki/wiki/pages/agent-cogni-consulting-phase-analyst.md`

So `ask` can today confidently cite plugins that were deleted — a correctness regression in the flagship skill, with no test that would catch it.

`scripts/check-external-dispatch.py` cannot help: both wiki trees are excluded from it by `EXCLUDE_SEGMENTS`/`EXCLUDE_TOPLEVEL`, deliberately.

## Proposed change

Add `cogni-workspace/tests/test-wiki-namespace-sync.sh` that fails when any page filename in `wiki/wiki/pages/` or `cogni-workspace/wiki/wiki/pages/` encodes a plugin name absent from `.claude-plugin/marketplace.json`.

Page filenames encode the namespace (`skill-cogni-visual-enrich-report.md`, `agent-cogni-claims-claim-verifier.md`), so this is a filename-pattern check against the marketplace roster — cheap and deterministic.

Then delete the ~8 stale pages the new guard surfaces, and fix any `[[wikilink]]` left dangling by the deletions.

## Acceptance criteria

- [ ] New test fails on a planted page for a non-existent plugin
- [ ] New test passes on the tree after the stale pages are removed
- [ ] Zero pages in either wiki tree name a plugin absent from `marketplace.json`
- [ ] No dangling `[[wikilinks]]` remain after deletion
- [ ] Test is discovered by `scripts/run-plugin-tests.py` (lives under `*/tests/`)

## Use case

Makes wiki regeneration a **gate** rather than a follow-up. Each phase of the consolidation epic renames or removes plugin pages; without this guard, nothing tells you the wiki went stale, and `ask` degrades silently. Retro-fixes the existing drift as a side effect.

Ships independently and is worth doing whether or not the consolidation proceeds.

[...]

Issue #1345, opened 2026-08-13
Fix vacuously-passing test assertions (5 cogni-knowledge absence tests + verify-theme-backcompat skip-instead-of-fail, not in CI)


sdh07:
## Problem

Two classes of test assertion in this repo pass **vacuously** — they are satisfied by the thing they guard ceasing to exist. Both are latent today and both become actively dangerous during any plugin rename or absorption.

**1. Absence assertions with no positive control.** Five sites assert `assert_not_grep 'cogni-claims:'`:

- `cogni-knowledge/tests/test_verify_contract.sh:117`
- `cogni-knowledge/tests/test_finalize_contract.sh:103`
- `cogni-knowledge/tests/test_contradictor_contract.sh:162`
- `cogni-knowledge/tests/test_reviewer_contract.sh:149`
- `cogni-knowledge/tests/test_source_contradictor_contract.sh:108`

The moment `cogni-claims:` stops appearing anywhere, all five pass — silently losing the invariant they exist to protect. (By contrast `cogni-knowledge/tests/test_refresh_resweep_contract.sh:50,75` uses `assert_grep` and would fail loudly, which is the desired behaviour.)

**2. A skip-instead-of-fail in the theme back-compat harness.** `cogni-workspace/scripts/verify-theme-backcompat.sh:333-352` calls `c_skip` (a bare `printf`, line 45) rather than `fail` (line 50, which exits 1) when a consumer's `SKILL.md` is missing. Renaming any consumer skill silently degrades all six assertions to SKIP.

Worse: that harness lives in `scripts/`, not `tests/`, so `scripts/run-plugin-tests.py` (which globs `*/tests/*.sh`) **never discovers it — it is not in CI at all.**

## Proposed change

- Convert the five `assert_not_grep` sites to positive-control form: pair each absence assertion with a presence assertion that the intended replacement *is* dispatched, so the invariant cannot be satisfied by deletion.
- Change `c_skip` to `fail` at `verify-theme-backcompat.sh:333-352` for a missing `SKILL.md` on a listed consumer. Keep `c_info` for the genuinely-external `document-skills:*` case.
- Wire the harness into CI: either move it to `cogni-workspace/tests/` so `run-plugin-tests.py` discovers it, or add a dedicated job in `.github/workflows/lint.yml`.

## Acceptance criteria

- [ ] No `assert_not_grep` touching a plugin namespace prefix exists without a paired presence assertion
[...]

PR #1358, opened 2026-08-13
External-dispatch guard: retired prefixes as data, not regex

### LINKED ISSUE #1346: Generalize check-external-dispatch.py to a retired-prefix registry
## Problem

`scripts/check-external-dispatch.py` guards against dispatches to retired plugins, but hardcodes the retired set in a regex:

```python
DISPATCH_RE = re.compile(r"\bcogni-(wiki|research):")
```

It is wired into `.github/workflows/lint.yml:35` ("Assert zero cogni-wiki and cogni-research dispatches in live surfaces"). Every future absorption requires editing the regex and the job name — an error-prone edit in the one place that must not be got wrong.

## Proposed change

Replace the hardcoded regex with a small JSON registry of retired prefixes (e.g. `scripts/retired-plugins.json`), read at runtime and compiled into the same pattern. Adding a retired plugin becomes a one-line data change.

Preserve unchanged:
- the per-line `# external-dispatch-guard:allow` escape hatch for genuine lineage prose
- `EXCLUDE_PREFIXES` / `EXCLUDE_SEGMENTS` / `EXCLUDE_TOPLEVEL` semantics (the wiki trees and `references/` stay out of scope by design)
- the CI job's failure behaviour

Update the job name in `.github/workflows/lint.yml` to be registry-driven rather than naming specific plugins.

## Acceptance criteria

- [ ] Retired prefixes live in a JSON file, not in Python source
- [ ] `python3 scripts/check-external-dispatch.py` still returns clean-zero on the current tree
- [ ] Adding a prefix to the registry causes the guard to fire on a planted test dispatch
- [ ] The `# external-dispatch-guard:allow` escape hatch still works, and still requires a same-line rationale
- [ ] `tests/` coverage for the registry-loading path

## Use case

Every phase of the plugin consolidation epic ends by adding a retired prefix. Making that a data edit rather than a regex edit removes the highest-risk repeated manual step in the program.

Ships independently and is worth doing whether or not the consolidation proceeds.

claude-cogni-work:
Closes #1346.

## Summary

`scripts/check-external-dispatch.py` hardcoded the retired set in a single regex
(`DISPATCH_RE = re.compile(r"\bcogni-(wiki|research):")`), so every absorption meant
editing Python in the one place that must not be got wrong. This PR moves the retired
set into `scripts/retired-plugins.json` and compiles the same pattern from it at
runtime — retiring another plugin is now a one-line data edit.

- **The registry is mandatory config, not best-effort.** Missing, malformed, empty, or
  wrongly-typed all raise into the existing `{"success": false, "data": {}, "error": …}`
  envelope and exit **2**. There is no `except: return []`, no hardcoded fallback list,
  and no path on which the guard scans for nothing and exits 0. A guard that passes
  because it loaded nothing is worse than no guard — CI stays green while the invariant
  goes unprotected — so this follows the `check-plugin-inventory.py` mandatory-file
  precedent rather than `check-breadcrumbs.py`'s silent-empty-on-missing baseline load.
  (It does mirror `check-breadcrumbs.py` for the `__file__`-anchored default path and
  the override-flag shape.)
- **Token semantics are identical to base:** leading `\b`, `re.escape`d alternation,
  mandatory trailing colon. A bare `cogni-research` noun still does not match, and a
  hit's `match` is still the full `cogni-wiki:`.
- **Unchanged:** the per-line `# external-dispatch-guard:allow` escape hatch,
  `DEFAULT_GLOBS`, `EXCLUDE_PREFIXES` / `EXCLUDE_SEGMENTS` / `EXCLUDE_TOPLEVEL`, the JSON
  envelope, and the 0/1/2 exit contract.
- **CI behaviour unchanged.** Only the step name at `.github/workflows/lint.yml:35` is
  reworded; the job name (line 27) was already registry-neutral. The run body still uses
  `set -euo pipefail` + `python3 scripts/check-external-dispatch.py` with no `|| true`,
  no `continue-on-error` and no rc capture — so exit 2 still fails the build.
- The stderr FAIL banner is rendered from the loaded registry, so a newly-registered
  prefix names itself in CI output for free.

## Scope decisions

**Four files only:** `scripts/retired-plugins.json` (new), `scripts/check-external-dispatch.py`,
`tests/test_check_external_dispatch.sh`, `.github/workflows/lint.yml`. No `cogni-*/` tree
is touched, no other `lint.yml` job, no other `scripts/*.py`, and no `version` field
anywhere (the post-merge workflow owns the bump).

Repo-root `CLAUDE.md` / `README.md` are deliberately untouched: the guard has zero
mentions in either today, so documenting the registry there would widen the change. The
how-to-add-a-prefix instruction lives in the script docstring and the registry's own
`_comment`, which is where the next absorber already looks.

**Two scoping choices are documented in the docstring** because the sibling guard next
door made the opposite one, and a reader deserves to see that reconciled rather than
discover it:

- *Enumerated, not derived.* `check-plugin-inventory.py` explicitly avoids a
  hand-maintained list of retired names. That is right there, where the live set is the
  datum; here a derived `\bcogni-[a-z0-9-]+:` minus the installed plugins would also
  flag prose mentions of any plugin not installed in the reader's workspace. The cost of
  enumerating is honestly stated: the registry can go **stale**, and nothing detects
  that — the retiring change is expected to add its own entry, and that expectation is
  the control.
- *Plugin-granular.* An entry is a whole plugin prefix and the script appends the colon,
  so a retired *skill* on a live plugin cannot be expressed today. A colon-bearing entry
  is rejected loudly rather than compiling `cogni-wiki::`, which would silently match
  nothing. Widening to full dispatch tokens is a ~10-line change if that case arrives.

**One judgement call worth a reviewer's eye:** the suite's `PASS:`/`FAIL:` emitters
dropped ANSI colour. The base wrapped every result line in escape codes, which no
`^[[:space:]]*FAIL:` matcher can satisfy — so the suite could not have its teeth
verified at all. A `[ -t 1 ]` conditional was considered and rejected: it makes
parsability environment-dependent, and under a pty the colour and the failure both come
back through a channel nothing tests. Plain output matches the harness-parsable suites
already in the repo (`cogni-consult/tests/test_reference_pointers.sh`,
`cogni-visual/tests/test-arc-taxonomy-sync.sh`). Cases 1-5 keep their behaviour and
labels; the `FAILED` flag, `check()`, `assert_json()`, `mktemp -d` and cleanup trap are
all preserved.

## Test plan

`bash tests/test_check_external_dispatch.sh` — cases 1-5 unchanged, plus six registry
cases. Each new case is labelled `<id> <description>` — an id token followed by a space,
never a colon abutting the id — so one `--case` value matches both its `PASS:` and
`FAIL:` line.

R1-R5 stage a **copy** of the guard (`cp` from the suite's `$GUARD` at run time, so a
mutation to the real script reaches them) under `$WORK/<name>/` with an optional sibling
registry, and invoke it **without** `--registry`. That is what makes them honestly red
against the base implementation: base ignores the registry and exits 0 on these
invocations, so an exit-2 assertion fails. Passing `--registry` instead would have made
base argparse exit 2 and the cases *vacuously green*.

| Case | Setup | Expected | Against base |
|---|---|---|---|
| R1 | staged copy, no sibling registry | exit 2, `success:false`, error names the path (also proves the default path follows `__file__`, not `--root`) | RED — base exits 0 |
| R2 | sibling registry is malformed JSON | exit 2 | RED — base exits 0 |
| R3 | sibling registry is `{"retired_prefixes": []}` | exit 2, asserted on the exit code | RED — base exits 0 |
| R4 | 6 degenerate variants: wrong type, non-string entry, blank entry, **whitespace-padded entry**, colon-bearing entry, missing key | exit 2 each | RED — base exits 0 |
| R5 | registry lists `cogni-wiki`, `cogni-research`, `cogni-demo`; input plants a `cogni-demo:` dispatch | exit 1, `match == "cogni-demo:"` | RED — base's hardcoded regex ignores `cogni-demo` |
| R6 | sibling registry lists `cogni-demo`, but `--registry` points at a fixture omitting it | exit 0 — the override **replaces** rather than unions, and doubles as the not-over-broad control | red only because base argparse rejects `--registry`; stated rather than claimed as evidence |

R1-R3 are the anti-vacuity trio the acceptance contract requires red against base, and
all three are red for a behavioural reason. R3's scanned fixture deliberately carries an
ordinary word-colon token (`name: demo`) so that under the empty-registry mutant the
degenerate `\b(?:):` alternation demonstrably fires.

Verified on this branch:

- [x] `python3 scripts/check-external-dispatch.py` → exit 0, `success: true`, `total: 0` on the real tree
- [x] `bash tests/test_check_external_dispatch.sh` → exit 0, all 19 assertions pass (R4 now covers 6 variants)
- [x] `python3 scripts/run-plugin-tests.py` → **all 106 suites pass**
- [x] `check-breadcrumbs.py` and `check-plugin-inventory.py` → exit 0
- [x] Base implementation re-run against the R-case invocations → exit 0 in both, confirming R1-R5 are genuinely red against base
- [x] Mutation recipe below → `verdict: guard_verified`

## Mutation recipe

Proves the empty-registry guard is load-bearing rather than decorative: remove the
empty-registry clause and case **R3** must go red, then green again on restore.

```
bash <YOUR-MARKETPLACE>/cogni-service/scripts/mutation-check.sh \
  --root <ABSOLUTE-PATH-OF-YOUR-CHECKOUT-OF-THIS-BRANCH> \
  --file scripts/check-external-dispatch.py \
  --expr 's/ or not prefixes:/:/' \
  --test 'bash tests/test_check_external_dispatch.sh' \
  --case R3
```

Expected: `data.mutated_result: "red"`, `data.restored_result: "green"`,
`data.verdict: "guard_verified"` — confirmed on this branch.

**What to substitute — two paths, and only two.** `--root` takes the absolute path of
*your* checkout carrying this branch (`gh pr checkout`, then `pwd`); the harness never
infers it and runs `--test` with `cwd=--root`, so `--file` and `--test` are relative to
it and paste unchanged. `<YOUR-MARKETPLACE>` is your installed `cogni-service` path.
Run it on a clean tree — the harness edits in place and restores afterwards.

**Why the expression is safe.** It deletes only the empty-registry clause from the single
line `if not isinstance(prefixes, list) or not prefixes:`, leaving
`if not isinstance(prefixes, list):` — still valid Python, so the mutant imports and
every other case keeps running. (A mutant that failed to import would turn *every* case
red and launder a false "guard verified".) The replacement `:` does not contain the
literal it replaces, so the substitution cannot re-satisfy the guard.

**Two invariants a later edit must not break** — both are pinned as in-line comments at
their sites so this recipe stays executable:

- The literal `" or not prefixes:"` occurs **exactly once** in
  `scripts/check-external-dispatch.py`, on one physical line. `mutation-check.sh` applies
  `perl -0pi` with **no** `/g`, so a second occurrence would mutate the wrong site and
  the case would stay green — reported as "guard not load-bearing" when it is.
- Case ids are `R3`, not `R3:`. The harness classifies with
  `^[[:space:]]*FAIL:[[:space:]]+<case>([[:space:]]|$)` whole-token; since the suite emits
  `FAIL: R3 empty registry …`, passing `--case 'R3:'` matches neither line and returns
  `case_not_found` instead of a verdict.

## Follow-up issues

- #1357 — Repo-root test suites emit ANSI-wrapped `PASS:`/`FAIL:`, defeating
  mutation-harness parsing. The emitters were fixed here for the one suite this change
  needed to be mutation-testable; five other repo-root suites (and ~17 under
  `cogni-knowledge/tests/`) still carry the pattern. Widening that sweep into this PR
  would breach the acceptance contract's scope boundary, so it is tracked separately.

- #1366 — CI gate jobs can vanish silently: no suite parses the workflow YAML, and a
  workflow-load failure never reaches the PR check surface. Raised as a waived advisory
  on this PR's review; filed separately rather than swept in, since a real fix needs a
  stdlib-vs-PyYAML decision and the surfacing half is a config question, not a test.

Nothing in #1346's own acceptance criteria is deferred — this PR ships full scope.

---

## Review-fix pass (commit `85f66987`)

Two findings from the BLOCK verdict, both closed. Same four files — no scope widening.

**Critical — `.github/workflows/lint.yml:35` took all four gate jobs offline.** The step
name was an unquoted YAML plain scalar carrying a colon-space, so `yaml.safe_load` raised
`ScannerError` at line 35 col 74 and Actions could not load the workflow at all. Reworded
to `(registry scripts/retired-plugins.json)` rather than quoted, so a later edit cannot
re-break it, with a comment at the site explaining why. Confirmed on the new head: the
`Lint` run now materialises **all four jobs** where the previous run died in 0s.

This is also what re-grades review-contract item 6. The `run:` body was always correct —
`set -euo pipefail`, no `|| true`, no rc capture — but the job never executed, so the
build-failure guarantee was never exercised. It is now satisfied *on execution*, not on
reading.

**Minor — padded registry entries compiled to a pattern that never matches.**
`load_registry` tested emptiness with `entry.strip()` but fed the unstripped entry into
`re.escape`, so `{"retired_prefixes": ["  cogni-demo  "]}` returned `success:true`,
`total: 0`, exit 0 against a real `cogni-demo:` dispatch — the exact silent vacuity the
module docstring promises the registry prevents. Now rejected loudly, mirroring the
colon-bearing rejection rather than quietly repairing the entry, and pinned as R4's sixth
variant. The reviewer's reproduction returns exit 2.

Re-verified on the new head: guard clean-zero on the real tree · suite 19/19 with R4 at 6
variants · **all 106 plugin suites pass** · `check-breadcrumbs.py` and
`check-plugin-inventory.py` exit 0 · `mutation-check.sh --case R3` still returns
`guard_verified`. The mutation-recipe invariant holds — `" or not prefixes:"` still occurs
exactly once, on one physical line.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,7 +32,10 @@ jobs:
         # Default shallow checkout is sufficient: the guard reads the working
         # tree via `git ls-files`, not git history.
 
-      - name: Assert zero retired-plugin dispatches in live surfaces (set: scripts/retired-plugins.json)
+      # Keep this name free of ": " — an unquoted YAML plain scalar containing a
+      # colon-space makes the whole workflow unparseable, which silently takes
+      # every job in this file offline rather than failing one step.
+      - name: Assert zero retired-plugin dispatches in live surfaces (registry scripts/retired-plugins.json)
         run: |
           set -euo pipefail
           python3 scripts/check-external-dispatch.py
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,7 +32,7 @@ jobs:
         # Default shallow checkout is sufficient: the guard reads the working
         # tree via `git ls-files`, not git history.
 
-      - name: Assert zero cogni-wiki and cogni-research dispatches in live surfaces
+      - name: Assert zero retired-plugin dispatches in live surfaces (set: scripts/retired-plugins.json)
         run: |
           set -euo pipefail
           python3 scripts/check-external-dispatch.py
--- a/scripts/check-external-dispatch.py
+++ b/scripts/check-external-dispatch.py
@@ -13,9 +13,10 @@ file rather than a regex edit here:
 
     {"retired_prefixes": ["cogni-wiki", "cogni-research"]}
 
-Entries are BARE prefixes with no trailing colon — this script appends the colon that
-makes the token dispatch-shaped, and rejects a colon-bearing entry rather than
-compiling a `cogni-wiki::` pattern that would silently match nothing.
+Entries are BARE prefixes with no trailing colon and no surrounding whitespace — this
+script appends the colon that makes the token dispatch-shaped, and rejects both a
+colon-bearing entry (`cogni-wiki::`) and a padded one (`re.escape("  cogni-wiki  ")`)
+rather than compiling a pattern that would silently match nothing.
 
 The registry is **mandatory config, not an optional ratchet**: absent, unreadable,
 malformed, wrongly-typed or empty all exit 2. A guard that quietly loaded nothing and
@@ -149,6 +150,11 @@ def load_registry(path):
             raise RuntimeError(
                 "retired-plugin registry {}: every prefix must be a non-empty "
                 "string, got {!r}".format(path, entry))
+        if entry != entry.strip():
+            raise RuntimeError(
+                "retired-plugin registry {}: prefix {!r} must not carry surrounding "
+                "whitespace — re.escape would compile it into a pattern that never "
+                "matches".format(path, entry))
         if ":" in entry:
             raise RuntimeError(
                 "retired-plugin registry {}: prefix {!r} must not carry a colon — "
--- a/scripts/check-external-dispatch.py
+++ b/scripts/check-external-dispatch.py
@@ -1,11 +1,44 @@
 #!/usr/bin/env python3
 """check-external-dispatch.py — deterministic external-dispatch guard.
 
-Asserts that **no live-dispatch surface** dispatches the retired engines
-`cogni-wiki:` or `cogni-research:` anywhere in the repo. This is the machine
-proof behind the FMO archival decision: once cogni-research and cogni-wiki are
-archived (source kept, not installable), a live caller dispatching them would
-break at runtime. The guard makes that absence objective and regression-proof.
+Asserts that **no live-dispatch surface** dispatches a retired engine anywhere in
+the repo. This is the machine proof behind the archival decision: once a plugin is
+archived (source kept, not installable) or absorbed into another, a live caller
+dispatching it would break at runtime. The guard makes that absence objective and
+regression-proof.
+
+The retired set is **data, not source** — it lives in `scripts/retired-plugins.json`
+(override with `--registry`), so retiring another plugin is a one-line edit to that
+file rather than a regex edit here:
+
+    {"retired_prefixes": ["cogni-wiki", "cogni-research"]}
+
+Entries are BARE prefixes with no trailing colon — this script appends the colon that
+makes the token dispatch-shaped, and rejects a colon-bearing entry rather than
+compiling a `cogni-wiki::` pattern that would silently match nothing.
+
+The registry is **mandatory config, not an optional ratchet**: absent, unreadable,
+malformed, wrongly-typed or empty all exit 2. A guard that quietly loaded nothing and
+reported a clean zero would be worse than no guard, because CI would stay green while
+the invariant went unprotected.
+
+Two scoping choices, stated because the sibling guard next door made the opposite one:
+
+  - **Enumerated, not derived.** `check-plugin-inventory.py` deliberately avoids a
+    hand-maintained list of retired names, preferring a bijection that works without
+    being told any. That is the right shape *there*, where the live set is the datum.
+    Here the live set is not enough: `\bcogni-[a-z0-9-]+:` minus the installed plugins
+    would also flag every prose mention of a plugin that simply is not installed in
+    the reader's workspace, so the guard would fire on trees it should not. The cost
+    of enumerating is a registry that can go **stale** — a plugin retired without an
+    entry here is silently unguarded — and nothing detects that today. The retiring
+    change is expected to add its own entry; that expectation is the control.
+  - **Plugin-granular.** An entry is a whole plugin prefix, and this script appends
+    the colon. A retired *skill* on a live plugin (`cogni-x:some-skill`) therefore
+    cannot be expressed, and the loader rejects the colon-bearing entry a maintainer
+    would reach for rather than silently compiling `cogni-x::`, which would match
+    nothing. Widening to full dispatch tokens is a ~10-line change to `load_registry`
+    (validate the shape instead of manufacturing it) if that case ever arrives.
 
 Scope — the surfaces a running session actually dispatches FROM:
 
@@ -37,10 +70,10 @@ a lineage mention there ("modeled on the cogni-research verify-report skill") is
 not a caller.
 
 Unlike the breadcrumb guard, this guard targets a HARD clean-zero — there is no
-legitimate live `cogni-wiki:`/`cogni-research:` dispatch after archival, so the
-ratchet/baseline model is the wrong tool. The single escape hatch is a per-line
-marker for a genuine NON-dispatch prose mention that must keep the literal token
-(rare); state the rationale on the same line:
+legitimate live dispatch of a registered retired prefix, so the ratchet/baseline
+model is the wrong tool. The single escape hatch is a per-line marker for a genuine
+NON-dispatch prose mention that must keep the literal token (rare); state the
+rationale on the same line:
 
     ... see cogni-research:verify-report for the lineage  # external-dispatch-guard:allow
 
@@ -77,13 +110,60 @@ EXCLUDE_PREFIXES = ("cogni-knowledge/",)
 EXCLUDE_SEGMENTS = ("/wiki/", "/docs/")
 EXCLUDE_TOPLEVEL = ("wiki/", "docs/")
 
-# The dispatch-shaped token: `cogni-wiki:` / `cogni-research:` — the `plugin:`
-# half of a `plugin:skill` dispatch reference. A bare "cogni-research" (no
-# colon) is a plain noun and is NOT matched.
-DISPATCH_RE = re.compile(r"\bcogni-(wiki|research):")
+# Default registry location, resolved next to THIS script (never under --root, so
+# scanning a foreign tree still reads our own retired set). Override: --registry.
+REGISTRY_BASENAME = "retired-plugins.json"
 
 # Per-line escape hatch for a genuine non-dispatch prose mention.
 ALLOW_MARKER = "external-dispatch-guard:allow"
+
+
+def load_registry(path):
+    """Return the list of bare retired prefixes from the registry at `path`.
+
+    Every degenerate input raises RuntimeError, which main() renders as the exit-2
+    envelope — there is deliberately no empty-list fallback and no hardcoded default
+    set (see the module docstring for why).
+    """
+    if not os.path.exists(path):
+        raise RuntimeError("retired-plugin registry not found: {}".format(path))
+    try:
+        with open(path, "r", encoding="utf-8") as fh:
+            data = json.load(fh)
+    except (OSError, ValueError) as exc:
+        raise RuntimeError("bad retired-plugin registry {}: {}".format(path, exc))
+
+    prefixes = data.get("retired_prefixes") if isinstance(data, dict) else None
+
+    # Mutation-recipe invariant: keep the next line exactly as written, on ONE
+    # physical line, and do not repeat its text anywhere else in this file — the
+    # recorded recipe rewrites the FIRST match only (perl -0pi, no /g), so a second
+    # occurrence would mutate the wrong site and report the guard as decorative.
+    if not isinstance(prefixes, list) or not prefixes:
+        raise RuntimeError(
+            "retired-plugin registry {} must carry a non-empty "
+            "'retired_prefixes' list".format(path))
+
+    for entry in prefixes:
+        if not isinstance(entry, str) or not entry.strip():
+            raise RuntimeError(
+                "retired-plugin registry {}: every prefix must be a non-empty "
+                "string, got {!r}".format(path, entry))
+        if ":" in entry:
+            raise RuntimeError(
+                "retired-plugin registry {}: prefix {!r} must not carry a colon — "
+                "this guard appends it".format(path, entry))
+    return prefixes
+
+
+def compile_dispatch_re(prefixes):
+    r"""Compile `\b(?:prefix|…):` — the `plugin:` half of a dispatch reference.
+
+    The trailing colon is MANDATORY, so a bare "cogni-research" is a plain noun and
+    is NOT matched, and a hit's `match` is the full "cogni-research:".
+    """
+    return re.compile(
+        r"\b(?:" + "|".join(re.escape(p) for p in prefixes) + r"):")
 
 
 def discover_files(root):
@@ -109,7 +189,7 @@ def discover_files(root):
     return files
 
 
-def scan_file(abs_path, rel_path):
+def scan_file(abs_path, rel_path, dispatch_re):
     """Yield violation dicts for one file."""
     try:
         with open(abs_path, "r", encoding="utf-8") as fh:
@@ -120,7 +200,7 @@ def scan_file(abs_path, rel_path):
         line = raw.rstrip("\n")
         if ALLOW_MARKER in line:
             continue
-        for m in DISPATCH_RE.finditer(line):
+        for m in dispatch_re.finditer(line):
             yield {
                 "file": rel_path,
                 "line": lineno,
@@ -129,7 +209,7 @@ def scan_file(abs_path, rel_path):
             }
 
 
-def collect(root, explicit_files):
+def collect(root, explicit_files, dispatch_re):
     occ = []
     if explicit_files:
         pairs = []
@@ -143,7 +223,7 @@ def collect(root, explicit_files):
     else:
         pairs = [(os.path.join(root, rel), rel) for rel in discover_files(root)]
     for abs_path, rel in pairs:
-        occ.extend(scan_file(abs_path, rel))
+        occ.extend(scan_file(abs_path, rel, dispatch_re))
     return occ
 
 
@@ -154,13 +234,21 @@ def main(argv):
     ap.add_argument("--root", default=None,
                     help="repo root for discovery + relative paths "
                          "(default: parent of scripts/)")
+    ap.add_argument("--registry", default=None,
+                    help="retired-prefix registry JSON (default: {} next to this "
+                         "script). REPLACES the default file — it is not merged "
+                         "with it.".format(REGISTRY_BASENAME))
     args = ap.parse_args(argv)
 
     script_dir = os.path.dirname(os.path.abspath(__file__))
     root = os.path.abspath(args.root) if args.root else os.path.dirname(script_dir)
+    # Anchored on the script, never on --root: scanning a foreign tree must still
+    # read OUR retired set, not whatever that tree happens to ship.
+    registry_path = args.registry or os.path.join(script_dir, REGISTRY_BASENAME)
 
     try:
-        violations = collect(root, args.files)
+        retired = load_registry(registry_path)
+        violations = collect(root, args.files, compile_dispatch_re(retired))
     except RuntimeError as exc:
         print(json.dumps({"success": False, "data": {}, "error": str(exc)}))
         return 2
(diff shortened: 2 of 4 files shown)


claude-cogni-work:
## Ship-ready

Author instance handing off for independent review. Draft, carrying `svc:needs-review`.
Issue #1346 is on `svc:resolved`; the `Closes #1346` line closes it on merge.

Graded against the `--target issue --origin triage` contract derived at
`2026-08-13T06:36:25Z` — before any branch was cut, so it is diff-blind by chronology.

- **19 items graded, 19 satisfied, 0 violated, 0 unclear.** No residual.
- No fix pass was needed, so nothing was auto-fixed and nothing is carried forward.

This is an advisory self-grade only — no verdict label, no `post-verdict`, no
`reviewed-head-sha` marker. The independent review is yours.

### One thing the merger should know: the preflight was vacuously green

`check-preflight.sh` exited 0 with `clean: true` — but **all 7 instruments skipped**,
`gating_applicable: 0`. Two of those skips are wrong for this change rather than
genuinely not-applicable:

| Instrument | Reported skip reason | Reality |
|---|---|---|
| `run-tests` | "no touched `scripts/`, `tests/`, SKILL.md or agents/*.md path" | the diff touches **both** `scripts/check-external-dispatch.py` and `tests/test_check_external_dispatch.sh`; the matcher is plugin-scoped (`*/scripts/…`), so repo-root paths never match |
| `mutation-recipe` | "no path matching `*/tests/test-*.sh` or `*/scripts/check-*.sh`" | plugin-scoped **and** a hyphen glob — `test_check_external_dispatch.sh` is underscore-named, so it could not match even under a plugin |

The other five (`frontmatter`, `breadcrumbs`, `skill-length`, `agent-length`,
`version-bump`) are correctly not-applicable: no SKILL.md, no agents/*.md, no plugin.

Net effect: **a repo-root-only change gets zero automated preflight coverage**, and the
green result says nothing about this diff. Since the two applicable instruments could
not run, their equivalents were run by hand on the branch:

- `python3 scripts/run-plugin-tests.py` → **all 106 suites pass**
- `bash tests/test_check_external_dispatch.sh` → exit 0, 19/19 assertions
- `python3 scripts/check-external-dispatch.py` → exit 0, `success: true`, `total: 0`
- `check-breadcrumbs.py`, `check-plugin-inventory.py` → exit 0
- `mutation-check.sh --case R3` → `verdict: guard_verified` (`mutated: red`, `restored: green`)

This is a gap in the preflight instrument's path matchers, not in this PR. It belongs to
`cogni-service` (external repo), so nothing was filed against `insight-wave` for it —
flagging it here for whoever owns that classifier.

### Follow-up filed

- #1357 — repo-root suites emit ANSI-wrapped `PASS:`/`FAIL:`, defeating mutation-harness
  parsing. Fixed here for the one suite this change needed; five more repo-root suites
  (and ~17 under `cogni-knowledge/tests/`) still carry it. Out of the contract's scope
  boundary, so tracked separately rather than swept in.

### Concurrency

Authored by one of several concurrent author drains. Phase-1 siblings #1345 and #1347
were under fresh foreign claims at pick time and were left alone; this branch shares no
file with them, and notably does not touch repo-root `CLAUDE.md`.

sdh07:
## Automated Review Verdict: 🚫 BLOCK

**PR:** https://github.com/cogni-work/insight-wave/pull/1358

**Summary:** Blocked: a critical unquoted-colon YAML defect at .github/workflows/lint.yml:35 makes the workflow unparseable, taking all four CI gate jobs offline -- including the external-dispatch guard this PR hardens. Independently verified three ways. Because that job never runs, review-contract item 6 (the guard must still fail the build under set -euo pipefail with no exit-code masking) is violated, not satisfied, overriding contract-grading-reviewer's all-19-pass report. The Python and registry work itself is sound and every issue acceptance criterion grades satisfied.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[critical]** `.github/workflows/lint.yml:35` — The renamed step name is an unquoted YAML plain scalar containing a colon-space sequence -- `(set: scripts/retired-plugins.json)`. `yaml.safe_load` raises ScannerError 'mapping values are not allowed here, line 35, column 74'; the base commit parses and yields all four jobs. GitHub Actions cannot load a workflow it cannot parse, so breadcrumb-guard, external-dispatch-guard, version-bump-gate AND plugin-test-suites all stop running -- a PR hardening the guard against silent vacuity ships a silently vacuous CI. Independently verified three ways: local parse fails identically; the branch's `Lint` workflow run failed in 0s (workflow-load failure, run 31676737751); and `gh pr checks 1358` lists only `cla`, with all four gate jobs absent. This same parse failure is why review-contract item 6 must be treated as violated rather than satisfied: the step's pipefail behaviour is never exercised because the job never runs at all.
  - *Suggested fix:* Quote the step name, or drop the colon: `(registry scripts/retired-plugins.json)`. Verified to restore all four jobs and to make item 6's build-failure guarantee actually testable.

- **[minor]** `scripts/check-external-dispatch.py:147-156` — `load_registry` validates entries with `entry.strip()` for emptiness but feeds the UNSTRIPPED entry into `re.escape`, so a whitespace-padded prefix compiles to a pattern that can never match. Reproduced: `{"retired_prefixes": ["  cogni-demo  "]}` against a file containing `dispatches cogni-demo:demo-skill` returns success:true, total 0, exit 0 -- the exact silent-vacuity failure mode the module docstring says the registry is mandatory to prevent.
  - *Suggested fix:* Normalize with `entry.strip()` into the returned list, or reject when `entry != entry.strip()` using the same message shape as the colon check. Worth a sixth variant in the R4 table.

- **[minor]** (waived) (advisory-addition) `tests/` — No suite parses the workflow YAML, which is why the critical finding reached review undetected. Pre-existing repo-level gap, not a regression this PR introduces.
  - *Justification:* Pre-existing coverage gap, not introduced by this change.

- **[minor]** (waived) (advisory-addition) `scripts/check-external-dispatch.py:201` — Issue AC clause 'still requires a same-line rationale' holds only in the doc-convention sense it held at base: the marker must be on the offending line but no rationale text is asserted. Code path byte-identical to base, so graded satisfied on a non-regression reading.
  - *Justification:* Non-regression; identical to base behaviour.

### Positive observations

- All 19 committed contract checklist items graded with cited evidence, 18 satisfied -- notably item 8 (real tree exits 0), items 14/15/16 (missing/malformed/empty registry all exit 2), item 17 (anti-vacuity cases are red against base), item 19 (scope confined to the four declared files) (cogni-service:contract-grading-reviewer)
- All five issue acceptance criteria satisfied; the hardcoded DISPATCH_RE is genuinely gone with no fallback set anywhere (plugin-dev:plugin-validator)
- Registry resolution mirrors check-breadcrumbs.py's script-anchored convention, correctly kept off --root (plugin-dev:plugin-validator)
- R1-R4 are constructed to be red against the base implementation, and stage a runtime copy of the real guard rather than a vendored snippet (plugin-dev:plugin-validator)
- No plugin directory, plugin.json, or marketplace.json touched; check-plugin-inventory.py and check-breadcrumbs.py both exit 0 (plugin-dev:plugin-validator)
- R6 proves --registry replaces rather than unions the default set (plugin-dev:plugin-validator)

### Recommendation

🚫 **Blocked — escalate.** Fix .github/workflows/lint.yml:35 (quote the step name or drop the colon) to restore all four CI gate jobs before re-review. That is also what resolves review-contract item 6: contract-grading-reviewer graded it satisfied by reading the `run:` body in isolation, but the confirmed parse failure means the job -- and therefore the `set -euo pipefail` guarantee item 6 requires -- never executes at all. Treat item 6 as violated until the workflow parses and is confirmed to fail the build on a guard exit 2. Separately, the check-external-dispatch.py:147-156 whitespace-stripping defect (minor, not blocking on its own) is worth closing in the same pass since it undermines the same anti-vacuity goal this PR targets.

---

sdh07:
## Block cleared — both findings closed, all four gate jobs now green

Fixed in `85f66987`. Both findings independently reproduced before being fixed, and the fix verified against the reviewer's own reproductions rather than by inspection.

### Critical — `.github/workflows/lint.yml:35`

Confirmed exactly as reported. The step name was an unquoted YAML plain scalar carrying a colon-space, `(set: scripts/retired-plugins.json)`; `yaml.safe_load` raised `ScannerError: mapping values are not allowed here, line 35, column 74`, so Actions could not load the workflow and every job in the file went offline.

Reworded to `(registry scripts/retired-plugins.json)` rather than quoted. Quoting is equally valid YAML, but a colon-free scalar cannot be re-broken by a later edit that drops the quotes — and a comment now sits at the site saying why the name must stay free of `": "`.

**Verified on the new head — the same three ways the finding was raised:**

| Evidence | Before | After |
|---|---|---|
| `yaml.safe_load` on `lint.yml` | `ScannerError` line 35 col 74 | parses, yields all 4 jobs |
| `Lint` workflow run | run 31676737751 — **failure in 0s** (load failure) | run 31689562288 — **success**, all 4 jobs executed |
| `gh pr checks 1358` | only `cla` | all 4 gate jobs + `cla`, **5 pass / 0 fail** |

### Review-contract item 6, re-graded

Item 6 is now **satisfied on execution**, not on reading. Your call to treat it as violated was right: the `run:` body was always correct — `set -euo pipefail`, no `|| true`, no `continue-on-error`, no rc capture — but a body that never runs guarantees nothing, and contract-grading-reviewer graded it by reading the body in isolation. The External-dispatch guard job is now observed completing in 7s with conclusion `success`, so the pipefail path is genuinely exercised.

### Minor — padded registry entries (`check-external-dispatch.py:147-156`)

Confirmed, including the mechanism: emptiness was tested with `entry.strip()` while the **unstripped** list went into `re.escape`. Your reproduction reproduced exactly — `{"retired_prefixes": ["  cogni-demo  "]}` against a real `cogni-demo:demo-skill` dispatch returned `success:true`, `total: 0`, exit 0.

Closed by rejecting rather than normalizing, which is the choice consistent with the module's own philosophy: a colon-bearing entry is already rejected loudly instead of quietly repaired, and a padded entry fails for the same reason — it compiles to a pattern that can never match. Same reproduction now returns **exit 2** with a message naming the entry.

Pinned as R4's sixth variant, per your suggestion. R4's label self-reports the count, so it now reads `(6 variants)`.

### Waived advisories

Both left as waived, and the workflow-parse gap filed as **#1366** rather than swept in — it is a pre-existing repo-level gap, a real fix needs a stdlib-vs-PyYAML decision (PyYAML is not stdlib, and the repo convention is stdlib-only), and there is a self-reference trap worth writing down: a parse suite running *inside* `lint.yml` cannot catch `lint.yml` being unparseable.

That issue also captures what I think is the more important half of this incident. The failure signal **did** exist — the branch's `Lint` run went red in 0s — it just never reached the surface a reviewer reads, because the jobs that would have been the checks never materialised. A broken workflow removes its own failure signal from the PR. That is a surfacing problem, not a test-coverage problem.

### Re-verification on the new head

- `python3 scripts/check-external-dispatch.py` → exit 0, `success: true`, `total: 0`
- `bash tests/test_check_external_dispatch.sh` → exit 0, 19/19, R4 at 6 variants
- `python3 scripts/run-plugin-tests.py` → **106/106 suites pass**, 0 failed
- `check-breadcrumbs.py`, `check-plugin-inventory.py` → exit 0
- `mutation-check.sh --case R3` → `mutated: red`, `restored: green`, **`verdict: guard_verified`**
- Mutation-recipe invariant intact: `" or not prefixes:"` still occurs **exactly once**, on one physical line

### Still draft, deliberately

Left as a draft carrying no merge. `svc:review-blocked` was an escalation to a human, and clearing the finding that caused it does not clear the escalation — the merge call is the maintainer's.

Label swapped `svc:review-blocked` → **`svc:needs-review`**, not `svc:review-pass`. The findings are closed, but no fresh independent multi-perspective review has run against the new head — the prior verdict's `reviewed-head-sha` marker still points at `bfcbcfdd`, which is now two commits stale. Applying a pass label would assert a verdict nobody ran.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1358

**Summary:** Both perspectives pass with zero findings — the committed 19-item review contract for issue #1346 is fully satisfied with execution-verified evidence, the plugin marketplace/directory bijection and guard coverage are unaffected, both deterministic pre-gates (frontmatter schema, version-bump) came back clean/not-applicable, and CI is green across all 5 checks.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- Items 1-19 all satisfied with file:line evidence; anti-vacuity family (14/15/16/17) verified by execution, not by reading assertions. (cogni-service:contract-grading-reviewer)
- Registry carries exactly the base pair cogni-wiki/cogni-research; no third prefix introduced. (cogni-service:contract-grading-reviewer)
- Default registry path anchored on os.path.dirname(os.path.abspath(__file__)), never args.root; --registry override added and exercised by fixture test R6. (cogni-service:contract-grading-reviewer)
- No except-return-[] swallow anywhere in the load path; missing/malformed/empty registry each exit 2 with success:false. (cogni-service:contract-grading-reviewer)
- lint.yml run body unchanged: set -euo pipefail, no || true, no continue-on-error, no rc capture, so exit 2 still fails the build. (cogni-service:contract-grading-reviewer)
- FAIL banner rendered from the loaded registry; EXCLUDE_*/DEFAULT_GLOBS/ALLOW_MARKER byte-identical to base; JSON envelope and 0/1/2 exit contract preserved. (cogni-service:contract-grading-reviewer)
- Scope boundary held: no cogni-*/ tree, no other lint.yml job, no other scripts/*.py, no version field touched. (cogni-service:contract-grading-reviewer)
- Zero changes under any cogni-*/ directory, no plugin.json, no marketplace.json, no SKILL.md, no agents/commands/hooks files. (plugin-dev:plugin-validator)
- check-plugin-inventory.py exits 0 on the PR head: marketplace_plugins 13 == plugin_directories 13, violations []. (plugin-dev:plugin-validator)
- Effective retired set identical to base; guard returns success:true total:0 on the current tree with no plugin edit required as a side effect. (plugin-dev:plugin-validator)
- No self-masked CI: lint.yml parsed with yaml.safe_load yields all four jobs and eight steps intact; the step name was deliberately kept free of colon-space with an inline comment explaining the hazard. (plugin-dev:plugin-validator)
- No shared test-registry collision: run-plugin-tests.py discovers via tests/*.sh globs, no ALL_TESTS line to append to. (plugin-dev:plugin-validator)

### Recommendation

---

Issue #1364, opened 2026-08-13
cogni-knowledge: consolidate the inlined red/green emitters onto the shared test helper


claude-cogni-work:
## Problem

`cogni-knowledge/tests/fixtures/test_helpers.sh` exists specifically to end byte-identical
inlining of the `red`/`green` result emitters. It has not finished that job: some suites under
`cogni-knowledge/tests/` still carry their own inlined copies rather than sourcing the helper.

The emitter *bodies* are being made uniform separately (#1359), so this is no longer a
correctness gap — every copy says the same thing. It is a duplication gap: N copies of a
definition that exists precisely once for a reason, and N places a future change has to reach.

Verify the current split before starting, rather than trusting a count recorded here:

```bash
grep -lE '^(red|green)\(\)' cogni-knowledge/tests/*.sh | wc -l    # files inlining a copy
grep -lF  'test_helpers.sh'  cogni-knowledge/tests/*.sh | wc -l    # files sourcing the helper
```

## Proposed change

In each suite that still inlines the pair, delete the two definition lines and source the helper
instead, using the idiom the helper's own header documents and that the majority of suites
already use:

```bash
. "$(dirname "$0")/fixtures/test_helpers.sh"
```

Call sites do not change — the function names and signatures are identical either way.

## Evidence this is safe

Measured during #1359's cleanup review, on a scratch copy of the tree at the time:

- All 15 then-inlining suites were converted and run against the real scripts. All stayed green,
  including all 7 that call the emitters from an end-of-run summary block — the case that makes
  a careless rewrite dangerous, because a suite whose summary call is orphaned dies under
  `set -eu` *after* every case line has printed, so it still looks normal at a glance.
- None of the inlining suites defines a competing `assert_grep` / `assert_not_grep`, so sourcing
  the helper introduces no name collision. The locally-defined helpers found were `fail`,
  `has_line`, `count_pat`, `extract_resolver`, `run_resolve*` — no overlap.
- The diff gets *smaller*: each suite goes from two lines to one.

Re-confirm each of these against the tree as it stands before relying on them.

## Why this was not done in #1359

Scope. #1359's falsifier is "no escape sequence in a result line"; who owns the definition is a
different property with a different falsifier. #1359 converts the bodies in place and leaves
ownership alone.

`cogni-knowledge/tests/test_plain_result_emitters.sh` (added by #1359) is deliberately neutral on
this: its liveness floor counts files that *exercise* the contract — defining the emitters **or**
[...]

Issue #1362, opened 2026-08-13
CI guard: no escape sequence before a PASS:/FAIL: label in any discovered test suite


claude-cogni-work:
## Problem

No guard prevents a regression to ANSI-wrapped `PASS:`/`FAIL:` result lines in test suites.

The conversion itself is being done in pieces, because the pieces are owned by different PRs:

| Surface | Owner | State when this issue was filed |
|---|---|---|
| 5 repo-root `tests/*.sh` suites | #1357 / PR #1360 | PR open, unmerged |
| `tests/test_check_external_dispatch.sh` | #1346 / PR #1358 | PR open, unmerged |
| `cogni-knowledge/tests/` — shared helper + 15 inlining suites | #1359 | PR open, unmerged |

The existing acceptance grep from #1357 forbids *capability probes*; it does not match a raw
escape sequence, so nothing in the repo would catch a straight revert to the coloured form. The
most likely regression path is someone copying a `red()` out of an existing suite into a new one.

## Proposed change

Add one guard asserting that no `tests/*.sh` or `*/tests/*.sh` result line carries an escape
sequence before its `PASS:`/`FAIL:` label, and wire it into CI (`.github/workflows/lint.yml`,
alongside the existing `check-*.py` gates).

Scope the guard to the same two discovery globs `scripts/run-plugin-tests.py` uses —
`tests/*.sh` and `*/tests/*.sh`. That boundary is load-bearing, not incidental: it is what
naturally excludes `cogni-knowledge/_archive/tests/test_read_project_config_bare.sh`, which
carries the same escape pair but sits one directory level deeper and is parked outside CI by
design.

Do **not** gate any emitter on `[ -t 1 ]`, `tput`, `$TERM`, or any other environment probe —
that makes parsability environment-dependent, and the colour returns under a pty through a
channel nothing tests. Rejected in #1346, again in #1357, and again in #1359.

Do **not** ship an exclusion, allowlist, or skip marker for any suite. An exclusion someone must
remember to delete is precisely the bandaid the guard exists to replace.

## Blocked on

This guard is green only once **every** in-scope suite has been converted. Before starting,
verify on `main` that all three surfaces in the table above have actually landed:

```bash
grep -rn '\033' tests/ */tests/
```

That command returning nothing on `main` is the entry condition. If it still returns hits, the
guard would be red for reasons outside its own diff, and the remaining conversion PR should land
first. Do not treat the table above as current — it records the state at filing time, and the
[...]

PR #1360, opened 2026-08-13
repo-root tests: emit plain PASS:/FAIL: result lines

claude-cogni-work:
Closes #1357.

## Summary

The five repo-root suites defined their result emitters as `printf '\033[31m%s\033[0m\n'`, so an emitted line was `\033[31mFAIL: <case>\033[0m` rather than `FAIL: <case>`. A harness anchoring `^[[:space:]]*FAIL:[[:space:]]+<case>` matched neither the red nor the green arm and reported `case_not_found` instead of a verdict — meaning these suites could not have their teeth verified at all.

This rewrites the two function bodies to `printf '%s\n'` in each of the five suites, and adds a two-line comment recording why the format is plain and unconditional.

**The helper names and signatures are kept deliberately.** Each helper has **four** call sites per suite, not two: `check()` uses both, and the end-of-run summary calls both again (measured: every suite reports 4). Deleting the helpers and inlining `printf` at the `check()` sites would orphan the summary calls and abort each suite under `set -eu` — *after* every case line had already printed, so the output would still read as normal at a glance.

No output-format branching is introduced. The emitters print identical bytes to a pipe, a file, and a pty. A TTY-conditional emitter was rejected upstream because it makes parsability environment-dependent — the colour returns under `script(1)` through a channel nothing tests.

No `plugin.json` / `marketplace.json` version change; the post-merge workflow owns the bump.

## Sequencing against open PR #1358 — please read before grading AC1

Issue #1357's Scope says `tests/test_check_external_dispatch.sh` "was fixed in passing while resolving #1346, and is the reference shape". **That is not true of `main`.** It is true only on the unmerged branch of **PR #1358**, which contains exactly the two-line hunk converting it. On `main` all **six** repo-root suites carried the ANSI pair, not five — the issue was authored from #1358's branch and describes that branch's post-merge world.

That file is therefore **intentionally untouched here**: editing the same two lines a concurrent open PR already owns would guarantee a conflict between two open PRs.

Consequence for **AC1** ("*every* repo-root `tests/*.sh` suite"): it is satisfied for everything this PR owns, and becomes literally true repo-wide the moment #1358 merges. The review contract committed on the issue rules the untouched sixth file compliant and states this explicitly.

## Scope decisions

[...]

Issue #1359, opened 2026-08-13
cogni-knowledge test suites and their shared helper still emit ANSI-wrapped PASS:/FAIL:; no guard prevents regression


claude-cogni-work:
## Problem

Resolving #1357 converted the five repo-root `tests/*.sh` suites to plain `PASS:`/`FAIL:` emitters. Two things it deliberately left undone, both discovered while verifying that change:

**1. `cogni-knowledge/tests/fixtures/test_helpers.sh` still ships the ANSI-always emitters.**

```bash
red()   { printf '\033[31m%s\033[0m\n' "$1"; }
green() { printf '\033[32m%s\033[0m\n' "$1"; }
```

Measured on `main`: 16 suites under `cogni-knowledge/tests/` carry the pattern, and the shared helper that was created specifically to end the byte-identical inlining ships the coloured variant — so sourcing it *propagates* the defect rather than fixing it. Any mutation check pointed at a plugin suite hits exactly the `case_not_found` failure #1357 was filed for.

This is the larger share of the problem: #1357 fixed 5 files; this is ~17 more plus the helper that feeds them.

**2. Nothing prevents a regression.**

#1357's acceptance grep (`grep -rEn 'tput|NO_COLOR|CLICOLOR|-t 1' tests/`) forbids *capability probes*; it does not match `printf '\033[31m…'`. So no guard in the repo would catch a straight revert to the coloured form — and the most likely regression path is someone copying a `red()` out of `test_helpers.sh` into a new suite.

A guard was considered during #1357 and rejected there for a specific reason: `tests/test_check_external_dispatch.sh` still carries the ANSI pair on `main`, so a repo-wide guard would have been red for reasons outside that diff.

## Proposed change

1. Convert `cogni-knowledge/tests/fixtures/test_helpers.sh` to plain emitters, and the ~16 `cogni-knowledge/tests/` suites that inline their own copies.
2. Then add one guard asserting no `tests/*.sh` or `*/tests/*.sh` result line carries an escape sequence before its `PASS:`/`FAIL:` label, and wire it into CI.

Do **not** gate the emitter on `[ -t 1 ]`, `tput`, `$TERM`, or any other environment probe — that makes parsability environment-dependent, and the colour returns under a pty through a channel nothing tests. Rejected in #1346 and again in #1357 for that reason.

## Dependencies — read before starting

The guard in step 2 is only green once **every** suite is converted. Two prerequisites:

[...]

PR #1365, opened 2026-08-13
cogni-knowledge: plain PASS:/FAIL: emitters in tests and shared helper

claude-cogni-work:
Refs #1359.

Step 1 of the issue only. Step 2 (the CI guard) is deliberately **not** in this PR — see Scope decisions.

## Summary

- Rewrite the `red()` / `green()` bodies in `cogni-knowledge/tests/fixtures/test_helpers.sh` and the 15 suites that inline their own copies to `printf '%s\n' "$1"` — 32 definition lines across 16 files. No escape sequence remains under `cogni-knowledge/tests/`.
- Function bodies are **rewritten, not deleted**. Seven of these suites call the emitters from an end-of-run summary block *after* the last per-case check, so deleting the definitions would kill the suite under `set -eu` once every case line had already printed — normal-looking text, wrong exit code. Every file that defined both emitters still defines both.
- Both spacing variants are covered: the three-space `red()   {` in 15 files, and the lone one-space `red() {` in `test_portal_store.sh`. A substitution keyed on one variant would have silently half-converted.
- No environment probe is introduced. The emitters are unconditional, so parsability never depends on where a suite runs.
- Correct `cogni-knowledge/tests/README.md`, whose convention bullet described the emitters as colour helpers.
- Add `cogni-knowledge/tests/test_plain_result_emitters.sh` holding the shape. CI already discovers `*/tests/*.sh`, so the property is falsifiable in the build rather than only at review time.

**Diff shape:** of the 64 changed lines in the 16 converted files, all 64 are emitter definition lines. No call site changes, and no line containing a `PASS:` or `FAIL:` label changes — those labels are the machine contract this whole change exists to protect.

## Scope decisions

**Deferred — the CI guard (issue step 2), filed as #1362.** The issue's own `## Dependencies` section gates it: #1358 owns `tests/test_check_external_dispatch.sh`, and all six repo-root `tests/*.sh` suites still carry the escape pair on `main`, so a repo-wide guard would be red from this diff by construction. No exclusion, allowlist, or skip marker is added for the external-dispatch file either — an exclusion someone must remember to delete is the bandaid the guard exists to replace. #1362 carries its entry condition as a runnable check rather than a claim about which PRs have landed, so it stays correct however the in-flight PRs resolve.

**Deferred — consolidating the inlined emitters onto the shared helper, filed as #1364.** The pre-commit cleanup review raised this as the deeper fix, and it is a fair point: `fixtures/test_helpers.sh` exists to end exactly this duplication, and consolidation is a *smaller* diff than what shipped here (−2/+1 per suite instead of −2/+2). It is deferred because it answers a different question than this issue asks — this issue's falsifier is "no escape sequence in a result line"; who owns the definition is a separate property with a separate falsifier. #1364 records the supporting evidence and tells the next author to re-verify rather than trust it.

**Untouched.** `cogni-knowledge/_archive/tests/test_read_project_config_bare.sh` carries the same pair but sits one directory level below both of `scripts/run-plugin-tests.py`'s discovery globs (`tests/*.sh`, `*/tests/*.sh`), so it is outside CI by design. The six repo-root `tests/*.sh` suites are byte-untouched — #1358 / #1360 own them. No version bump; the post-merge workflow owns that.

## One defect found and fixed before commit

Worth flagging because it changed the shipped code, and because it is the kind of thing a guard review is for.

`test_plain_result_emitters.sh` originally keyed its liveness floor on files that **inline** the emitter definitions. That made the guard depth-negative: consolidating a suite onto the shared helper — the deeper fix, and one this issue's acceptance contract explicitly permits — would have turned the guard **red**. A guard that fails when duplication is removed argues against the helper whose stated purpose is ending that duplication, and whoever tried #1364 later would have had to edit this guard to do it.

The floor now counts files that **exercise** the contract — defining the emitters *or* sourcing the helper. Verified both ways, below.

## Mutation recipe

Every case in the new suite was driven red before it was trusted. Each mutation was applied to a clean tree, the suite run, and the mutation reverted. From the repo root:

| # | Mutation | Expected | Observed |
|---|---|---|---|
| M1 | Revert `red()` in `test_portal_store.sh` (the one-space variant) to the escape form | red | **red** — caught by case 2 and case 3, exit 1 |
| M2 | Delete the `green()` definition from `test_fetch_cache.sh`, leaving its call sites | red | **red** — `defines red=1 green=0 (expected 1 each)` |
| M3 | Revert both emitters in `fixtures/test_helpers.sh` to the escape form | red | **red** — case 1 exact-match fails, case 2 fails |
| M4 | Point the suite at a scan root containing only the helper (simulates a broken glob) | red | **red** — `only 1 file(s) exercising the emitter contract were found (expected at least 50)` |

M1 is the half-conversion case, M2 the orphaned-summary-call case the issue warns about, M3 the straight revert, M4 the anti-vacuity floor.

Consolidation-neutrality was verified separately: converting all 15 inlining suites to source the helper in a scratch tree drops the definer count 16 → 1, and the suite **stays green**. Under the original floor that same tree failed — which is how the defect above was caught.

```bash
# M1 — half-conversion of the one-space variant
python3 - <<'PY'
import pathlib
p = pathlib.Path("cogni-knowledge/tests/test_portal_store.sh")
t = p.read_text()
old = "red() { printf '%s\\n' \"$1\"; }"
new = "red() { printf '\\033[31m%s\\033[0m\\n' \"$1\"; }"
assert old in t, "MUTATION TARGET NOT FOUND"
p.write_text(t.replace(old, new))
PY
bash cogni-knowledge/tests/test_plain_result_emitters.sh; echo "exit=$?"   # expect non-zero
git checkout -- cogni-knowledge/tests/test_portal_store.sh
```

```bash
# M2 — orphan a summary call by deleting one definition
cp cogni-knowledge/tests/test_fetch_cache.sh /tmp/fc.bak
grep -v "^green() { printf '%s" /tmp/fc.bak > cogni-knowledge/tests/test_fetch_cache.sh
bash cogni-knowledge/tests/test_plain_result_emitters.sh; echo "exit=$?"   # expect non-zero
git checkout -- cogni-knowledge/tests/test_fetch_cache.sh
```

Note the suite never spells the escape literal in its own source — case 2 assembles the pattern at run time — so it cannot trip its own sweep and needs no exclusion marker to pass. Confirmed: `grep -cF '\033'` on the new file returns 0.

## Test plan

- [x] `grep -rn '\033' cogni-knowledge/tests/` returns nothing (32 hits on `main`)
- [x] `bash -c '. cogni-knowledge/tests/fixtures/test_helpers.sh; green "PASS: x"; red "FAIL: y"' | cat -v` prints exactly two plain lines, zero `^[`
- [x] `bash cogni-knowledge/tests/test_plain_result_emitters.sh` exits 0 (3/3 cases; 16 definers of 78 exercising the contract)
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` green — 77/77
- [x] `python3 scripts/run-plugin-tests.py` full sweep green — 108/108
- [x] `python3 scripts/check-breadcrumbs.py` — no violations
- [x] Every `PASS:` / `FAIL:` label byte-identical to base — `git diff -U0` shows only emitter definition lines changing (64 of 64)
- [x] No repo-root `tests/` file and no version manifest in the changed-file list

### DIFF
--- a/cogni-knowledge/tests/README.md
+++ b/cogni-knowledge/tests/README.md
@@ -39,7 +39,10 @@ via `trap rm -rf "$WORK" EXIT`.
 
 - bash 3.2 + python3 stdlib only (no pytest, no pip).
 - `set -eu` at the top; exit non-zero on any failure.
-- Color helpers `red`/`green` for human-readable output.
+- Source `fixtures/test_helpers.sh` for `red`/`green` rather than inlining
+  them, so a `PASS:`/`FAIL:` label stays machine-parsable — plain, never
+  wrapped in an escape sequence and never chosen by an environment probe.
+  `test_plain_result_emitters.sh` holds this shape.
 - `assert_grep <pattern> <description>` for contract-level SKILL.md checks.
 - Real Python harness (inline `python3 - <<PY ... PY` heredoc) for
   script-level assertions.
--- a/cogni-knowledge/tests/fixtures/test_helpers.sh
+++ b/cogni-knowledge/tests/fixtures/test_helpers.sh
@@ -12,8 +12,8 @@
 #
 # Bash 3.2 compatible.
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 # assert_grep PATTERN FILE DESCRIPTION
 #   Increments the caller's `errors` variable on failure.
--- a/cogni-knowledge/tests/test_binding_project_path.sh
+++ b/cogni-knowledge/tests/test_binding_project_path.sh
@@ -22,8 +22,8 @@ set -eu
 PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/knowledge-binding.py"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 if [ ! -f "$SCRIPT" ]; then
   red "FAIL: knowledge-binding.py not found at $SCRIPT"
--- a/cogni-knowledge/tests/test_cycle_guard_clear.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_clear.sh
@@ -17,8 +17,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_depth_bound.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_depth_bound.sh
@@ -17,8 +17,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_direct.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_direct.sh
@@ -19,8 +19,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 # shellcheck source=fixtures/_cycle_guard_lib.sh
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_distilled.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_distilled.sh
@@ -26,8 +26,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 errors=0
 
--- a/cogni-knowledge/tests/test_cycle_guard_dry_run.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_dry_run.sh
@@ -17,8 +17,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_not_applicable.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_not_applicable.sh
@@ -17,8 +17,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
-green() { printf '\033[32m%s\033[0m\n' "$1"; }
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_question.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_question.sh
@@ -25,8 +25,8 @@ SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
 
-red()   { printf '\033[31m%s\033[0m\n' "$1"; }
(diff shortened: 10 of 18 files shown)


claude-cogni-work:
## Ship-ready

Handed off by a cogni-service **author** instance (`/service-resume --role author --act`) for
independent merger review. This PR is draft and carries `svc:needs-review`. No review verdict has
been posted and no `svc:review-*` label applied — the independent review is the merger's to run.

Note for scheduling: a second author instance is draining this same board concurrently in a
separate worktree under the same bot identity. This PR touches only `cogni-knowledge/tests/`.

Graded against the `--target issue --origin triage` contract derived at `2026-08-13T08:59:00Z`,
before any branch was cut.

| | |
|---|---|
| Items graded | 18 |
| Satisfied | 18 |
| Violated | 0 |
| Unclear | 0 |
| Residual carried to the merger | **none** |

No fix pass was needed. Two claims in the PR body were re-derived rather than trusted: the new
suite's consolidation-neutrality (converting all 15 inlining suites onto the shared helper keeps it
green — definers 16 → 1, exercisers 78) and all four entries of the `## Mutation recipe`, each of
which was reproduced against the head ref and drove the suite red via the case the body names.

### Preflight — green, but read the skips

`check-preflight.sh` exited 0, `clean: true`, 0 violations. Its own summary line is
`1 passed, 6 skipped`, and that ratio deserves a sentence rather than a glance:

| Instrument | Result | Reason |
|---|---|---|
| version-bump | **pass** | no version-line touch |
| run-tests | skip | `no_suites_at_base` |
| mutation-recipe | skip | `not_applicable` |
| frontmatter / breadcrumbs / skill-length / agent-length | skip | `not_applicable` — no SKILL.md or agents/*.md touched |

The four `not_applicable` skips on the right are correct — this diff touches no skill or agent files.

**The first two skips are instrument blindspots, not evidence.** Both glob `*/tests/test-*.sh`
(hyphen); every suite in this plugin is `test_*.sh` (underscore), so neither instrument could see
the 18 files this PR changes. The `run-tests` reason string — *"the base ref carried no suites —
nothing to regress from"* — is factually wrong: the base carries 76 suites under
`cogni-knowledge/tests/`.

So the green preflight is not test evidence here. The test evidence is that the suites were run
directly, on the PR head:

- `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` → 77/77, `failed_suites: []`
- `python3 scripts/run-plugin-tests.py` (full sweep) → 108/108, `failed_suites: []`
- `python3 scripts/check-breadcrumbs.py` → no violations

Same cause means the `mutation-recipe` instrument never inspected the PR body. The recipe is
present anyway, with four reproduced mutations — it just was not the instrument that checked it.

### What the merger may want to weigh

1. **The deferral is the issue's own instruction, not a shortfall.** #1359 says step 1 only while
   #1358 is open. The guard is #1362; it carries a runnable entry condition rather than a claim
   about which PRs have landed.
2. **A defect was found and fixed pre-commit**, described under "One defect found and fixed" in the
   PR body. The new suite's liveness floor originally counted *inline* definers, which would have
   turned it red against the shared helper on consolidation. It now counts files that exercise the
   contract either way. This is the one place the shipped code differs from the recorded plan.
3. **Consolidating the inlined emitters onto the helper is deferred to #1364**, with the evidence
   that it is a *smaller* diff. Reasonable to disagree with that deferral — it is a scope judgment,
   not a technical blocker.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1365

**Summary:** All 18 committed review-contract items are satisfied, verified independently by both perspectives against the PR-head worktree. The four verify-by-acting items were run live: the hermetic helper probe emits plain PASS:/FAIL: on stdout with one trailing newline, a converted inlining suite exits 0 with zero escapes, the filtered cogni-knowledge run is 77/77 green, and the full sweep is 108/108 green. Both reviewers confirmed the orphaned-summary failure mode the issue warns about does not occur — every one of the 16 base-carrying files still defines both emitters — and that both whitespace variants were converted. Remaining findings are preference-level only and none gate merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[minor]** (advisory-addition) `cogni-knowledge/tests/test_plain_result_emitters.sh:120-140` — The case-3 shape check keys on the substring `printf '%s` in the definition line, so it would accept a body that redirects to stderr (`printf '%s\n' "$1" >&2`) or drops the newline (`printf '%s' "$1"`). The stdout / one-trailing-newline invariant (contract item 6) is asserted only by case 1, and only for the shared helper — the 15 inlining suites are covered for plain-ness but not for channel or newline. No fault in this diff: all 32 definition lines were verified byte-identical plain-stdout forms with no `>&2`. This is an observation about the new guard's future coverage, not a defect.
  - *Suggested fix:* Optional follow-up: tighten the case-3 pattern to an exact match on the full plain body, or add a per-definer behavioral probe capturing stdout and stderr separately.

- **[minor]** (advisory-addition) `cogni-knowledge/tests/README.md:42-45` — The Convention bullet now prescribes sourcing `fixtures/test_helpers.sh` 'rather than inlining', while 15 suites in the same tree still inline their own copies. It reads as current-state description in a section whose other bullets are current-state. Nothing is inconsistent at the gate level — the new suite deliberately tolerates both shapes — only the prose is forward-looking without saying so.
  - *Suggested fix:* Optional: mark the bullet as the convention for new and consolidated suites, noting the 15 remaining inlined copies are plain and pending consolidation (deferred to #1364).

- **[minor]** (advisory-addition) `cogni-knowledge/tests/fixtures/test_helpers.sh:3-8` — The header comment still narrates the pre-v0.0.17 byte-identical-inlining history without noting the inlined copies are now colour-free too. The narration remains factually accurate as history. The committed contract's own risk_watchlist flags this as judgment-only and explicitly not a gate.
  - *Suggested fix:* Optional one-line addendum noting the inlined copies were converted to the plain form under #1359.

- **[minor]** (waived) (advisory-addition) `cogni-knowledge/` — Pre-existing frontmatter description over-cap drift in 2 SKILL.md and 4 agents/*.md files. This PR touches no SKILL.md and no agent file, so the drift is neither introduced nor aggravated by the diff.
  - *Justification:* Pre-existing drift the PR never touched — the deterministic frontmatter pre-gate scoped it out (empty intersection with the changed-file list).

### Positive observations

- The exact failure mode the issue names is avoided by construction: bodies rewritten in place, definitions preserved. Confirmed by mutation — deleting green() from test_cycle_guard_clear.sh kills it at line 73 with exit 127 after the case line has already printed, and the new suite catches that deletion. (plugin-validator)
- Both whitespace variants converted, including test_portal_store.sh's one-space `red() {` form — defeating the half-conversion-by-sed risk the contract called out. All 32 definition lines across the 16 files are now the byte-identical plain form. (both perspectives)
- PASS:/FAIL: label byte-stability holds exactly. Per-file label-line counts are identical base vs head across all 16 files; the only changed line matching those tokens anywhere in the diff is README prose. These labels are the machine contract the whole change exists to protect. (both perspectives)
- The new suite is genuinely falsifiable, not a vacuous guard. Three independent mutations (coloured helper, deleted definition, half-converted one-space form) each produced a real red verdict with a specific message and exit 1 — no case_not_found analogue. (plugin-validator)
- The self-containment trick removes a real maintenance hazard: the escape pattern is assembled at run time from a lone backslash, so the guard file cannot trip its own sweep and needs no exclusion marker — precisely the marker the issue forbids. (plugin-validator)
- The liveness floor is keyed on files that exercise the contract (define or source), not on inline definers, so the deferred consolidation onto the shared helper (#1364) keeps the suite green rather than turning the helper's own purpose into a failure. (plugin-validator)
- Verify-by-acting items 15-18 all run live from the PR-head worktree: hermetic helper probe prints exactly `PASS: x` / `FAIL: y` with zero `^[` (od -c confirms one trailing newline each, on stdout); test_cycle_guard_clear.sh exits 0 with zero escapes; filtered run 77/77 with success=true, failed=0, failed_suites=[]; full sweep 108/108 with failed=0. (both perspectives)
- Scope discipline is exact: 18 changed files, all under cogni-knowledge/tests/; zero touches to repo-root tests/, .github/, scripts/, _archive/, plugin.json or marketplace.json. No collision surface with the concurrent PRs that own the repo-root suites, and no premature CI guard that would be red by construction. (both perspectives)
- Both self-declared deferrals are consistent with the contract rather than dropped scope: the CI guard (#1362) is deferred by the issue's own `## Dependencies` gate and item 11 grades its absence as correct; consolidating the inlined emitters (#1364) is permitted-not-mandated, and none of the 61 helper-sourcing suites were touched or regressed. No exclusion, allowlist, skip marker or baseline file was introduced. (contract-grading-reviewer)
- bash 3.2 compatibility of the new suite confirmed by running it under /bin/bash 3.2.57, matching the stated tests/README.md convention. (plugin-validator)

### Recommendation

---

PR #1367, opened 2026-08-13
CI: guard that every workflow file parses and declares jobs

### LINKED ISSUE #1366: CI gate jobs can vanish silently: no workflow-YAML parse coverage, and load failure never reaches the PR check surface
## Problem

PR #1358 shipped a one-character-class defect that took **all four CI gate jobs offline**, and nothing in the repo caught it. The step name at `.github/workflows/lint.yml:35` was an unquoted YAML plain scalar containing a colon-space:

```yaml
- name: Assert zero retired-plugin dispatches in live surfaces (set: scripts/retired-plugins.json)
```

`yaml.safe_load` raises `ScannerError: mapping values are not allowed here, line 35, column 74`. GitHub Actions cannot load a workflow it cannot parse, so `breadcrumb-guard`, `external-dispatch-guard`, `version-bump-gate` and `plugin-test-suites` all stopped running.

The defect was caught by a human reviewer, not by the repo. Two gaps made that possible:

**1. No suite parses the workflow YAML.** `tests/` has six repo-root suites; none of them opens `.github/workflows/*.yml`.

**2. The signal existed but never surfaced.** The branch's `Lint` run went red in **0s** (run 31676737751) — a workflow-load failure. But `gh pr checks 1358` listed only `cla`, because the four gate jobs never materialised as checks. A reviewer glancing at the PR's check list saw nothing wrong. **This is the more important half:** a broken workflow removes its own failure signal from the surface people actually read.

## Proposed change

Two candidate directions, worth deciding between rather than assuming:

- **A parse suite** — `tests/test_workflows_parse.sh` asserting every `.github/workflows/*.yml` loads and yields its expected job set. Note the constraint: **PyYAML is not stdlib**, and the repo convention is bash + python3 stdlib-only, no pip dependencies. So this needs either a dependency decision or a targeted regex heuristic (unquoted `: ` inside a `name:` plain scalar), which is narrower but stdlib-clean.
- **Surfacing** — make workflow-load failure visible on the PR. Required-status-check configuration, or a check that fails loudly rather than vanishing.

Note the self-reference trap on direction A: a parse suite that runs **inside** `lint.yml` cannot catch `lint.yml` being unparseable — the job hosting it would not run either. It would catch a break in *any other* workflow file, and would catch a `lint.yml` break only on the next PR that repairs the file enough to run. That asymmetry is why the surfacing half is not optional.

## Acceptance criteria

- [ ] A break of the `lint.yml:35` shape is caught before merge, by CI or by a locally-runnable suite
- [ ] Whatever is added respects the stdlib-only convention, or the dependency is an explicit documented decision
- [ ] The self-reference limitation is stated wherever the check lives, so a future reader does not over-trust it
- [ ] Workflow-load failure is visible on the PR check surface, not only in the Actions run list

## Context

Raised by the automated review on #1358 as a `[minor] (waived)` advisory — correctly waived there as a pre-existing repo-level gap rather than a regression that PR introduced. Filed separately to keep #1358 inside its acceptance contract's four-file scope boundary.

claude-cogni-work:
Closes #1366.

A step name that is an unquoted YAML plain scalar carrying a colon-space makes the whole workflow unloadable. Actions then runs none of that file's jobs, so the four gate jobs in `lint.yml` did not merely fail — they never materialised as checks, leaving a red zero-second run in the Actions list and a PR whose check surface looked clean. Nothing in the repo caught either half.

## Summary

- **`scripts/check-workflow-yaml.py`** (new) — stdlib-only scanner over every `.github/workflows/*.yml` and `*.yaml`. Reports `plain_scalar_colon_space` (the shape that caused the outage) and `no_jobs` (a file that parses but declares nothing posts no checks either — the same user-visible outcome).
- **`tests/test_check_workflow_yaml.sh`** (new) — 10 cases, `WY1`–`WY10`. Auto-discovered by `scripts/run-plugin-tests.py`; **no registration edit anywhere**.
- **`.github/workflows/workflow-guard.yml`** (new) — a *separate* workflow file from `lint.yml`. The separation is the surfacing half, not a filing preference.
- **`CLAUDE.md`** — one Contributing bullet in the existing `(CI: "<job name>" in .github/workflows/<file>)` shape.

## Scope decisions

**A per-line regex would not have worked, and that shaped the whole design.** No indentation-agnostic pattern can separate an unquoted mapping value carrying a colon-space from a line inside a `run: |` body that happens to carry one — both are indented text under a key-shaped line. The tree passes a count-based heuristic today only by coincidence: `cogni-version-bump.yml` prints `key: value` diagnostics from inside heredocs and simply happens not to put two on one line. So the scanner tracks block-scalar state by indent column (tolerating `|- |+ >- >+ |2`), skips bodies unscanned, and **re-processes** the line that dedents out of one. A state machine that swallowed that line would silently stop guarding the rest of the file — case `WY4` pins exactly that, modelled on the live shape at `.github/workflows/cla.yml:17-20`.

**Dependency decision.** PyYAML is not stdlib, nothing in the repo imports it, there is no `requirements.txt`/`Pipfile`/`pyproject.toml`, and no runner installs anything. The guard is a bounded scanner whose limits its docstring states, rather than a first-ever build dependency. It is explicitly *not* a parser: a file it passes is not thereby proven valid YAML.

**Discovery deviates from the sibling `git ls-files` convention, deliberately.** That convention defends against stale copies of files scattered tree-wide; `.github/workflows/` is one fixed root-level directory. The listing is also what makes the zero-file case testable, since fixtures live in a non-git `mktemp -d`.

**Discovering zero files is a FAILURE, not a pass.** A guard that finds nothing and reports green reads as evidence and is worse than no guard. Both the empty-directory (`WY5`) and missing-directory (`WY6`) paths exit non-zero with `success:false` and a non-empty error — different code paths, both covered.

**Out of scope:** no `cogni-*/` path, no `version` field (the post-merge workflow owns bumps), and `.github/workflows/lint.yml` is untouched.

## Self-reference limitation

A guard cannot certify the workflow file hosting it — if that file fails to load, the job never starts. Hence the separate file: `workflow-guard.yml` scans `lint.yml`, and `lint.yml`'s test-suite job runs this guard's suite, whose real-tree case (`WY1`) scans `workflow-guard.yml`.

Two residual gaps, stated in all three surfaces rather than left to be discovered:

1. **Both arms run the same scanner**, so the mutual coverage extends only to the two shapes it models — not to loadability in general.
2. **A commit breaking both files at once** reports nothing from either direction.

## Review notes — two things worth a human's judgement

**1. An opportunistic `try: import yaml` would materially widen coverage, and I did not add it.** The `/simplify` altitude pass made a well-evidenced case for `try: import yaml / except ImportError: yaml = None` — real `safe_load` where the parser happens to exist, scanner as the universal floor. It demonstrated a concrete miss: a mis-indented `runs-on:`/`steps:` pair passes this scanner but raises `ScannerError` in `safe_load`, and that file's jobs would go offline exactly as in the original outage. Stdlib-only would stay literally true, since nothing new enters the build.

I did not apply it because the acceptance contract committed on the issue before this branch existed states, as item 5's literal falsifier, that *"the guard contains no `import yaml` / `from yaml`"*. Adding it would make a graded item read `violated`. Overriding a pre-committed bar unilaterally seemed worse than surfacing the tension — so this is flagged for a decision rather than silently dropped or silently done.

**2. Required-status-check configuration is a repo setting, not a file.** AC4 offers two branches — required-status-check config, *or* a check that fails loudly rather than vanishing. This ships the second, which is what a diff can express. Making the new check *blocking* (and arguably the four `lint.yml` gate jobs too, whose silent absence caused the outage) needs repo-admin access no code change has. Recorded here rather than as a `deferred[]` entry, since a deferred entry would force a partial PR and a follow-up issue for something that is a setting rather than code.

## Test plan

- [x] `bash tests/test_check_workflow_yaml.sh` → exit 0, all 10 cases pass, no ANSI on any result line
- [x] `python3 scripts/check-workflow-yaml.py` → exit 0 on the real tree, all 4 workflow files scanned clean (no false positive against the quoted values, block bodies and comments that legitimately carry a colon-space today)
- [x] `python3 scripts/run-plugin-tests.py` full sweep → **109/109 pass**
- [x] `python3 scripts/run-plugin-tests.py --list` includes `tests/test_check_workflow_yaml.sh` with no registration edit
- [x] Cross-checked against the real parser: the `WY2` fixture makes `yaml.safe_load` raise `ScannerError: mapping values are not allowed here`, confirming the guard targets the genuine defect rather than a lookalike

## Mutation recipe

`mutation-check.sh` ships with the **cogni-service plugin**, not with this repo — and note `cogni-consult/scripts/mutation-check.sh` and `cogni-portfolio/scripts/mutation-check.sh` *do* exist here with a different argument contract, so a repo-relative invocation resolves to the wrong harness.

```bash
bash "$CLAUDE_PLUGIN_ROOT/scripts/mutation-check.sh" \
  --root "$PWD" \
  --file scripts/check-workflow-yaml.py \
  --expr 's/^COLON_SPACE = ": "/COLON_SPACE = "@@NEVERMATCH@@"/m' \
  --test 'bash tests/test_check_workflow_yaml.sh' \
  --case WY2
```

Expected `verdict: guard_verified` — neutering the detection literal must turn `WY2` red and restoring it must return green. **Verified by hand on this branch**, both before and after the cleanup pass, since a cleanup can silently unmatch an `--expr`.

Three things make the recipe replayable:

- `--case WY2` is the first whitespace-delimited token of the emitted `FAIL: WY2 ...` line. The harness anchors `^[ \t]*FAIL:[ \t]+<case>(?:[ \t]|$)`, so a label written `B: ...` would need `--case 'B:'` and a bare `--case B` returns `case_not_found`. Colon-free `WY<n>` tokens avoid that trap.
- `--test` must be the `bash <path>` form: repo suites have mixed mode bits, so a bare relative path would not execute.
- `--expr` is `^`-anchored with `/m` because the harness runs `perl -0pi` over the whole file. The replacement contains no colon-space, so it is disjoint from the literal it removes and cannot keep the case green.

Worth noting what the mutation demonstrates beyond the recipe: under it, `WY1` (the real-tree case) stays **green**, because a neutered guard reports the real tree clean. That is precisely the vacuous-pass failure mode a real-tree case alone cannot catch, and why `WY2` carries a red fixture.


claude-cogni-work:
## Ship-ready

Handed off by the author instance for independent merger review. Head `6123846f`, base `origin/main` @ `78d6e39c`.

### Contract self-grade

Graded against the issue-time `review_contract` (posted on #1366 at `2026-08-13T10:31:04Z`, before this branch existed — so it is diff-blind by chronology).

**22 graded · 22 satisfied · 0 violated · 0 unclear.** No fix pass was needed; **no residual** carried into review.

This is advisory only — no verdict was posted, no `svc:review-*` label applied, and no `reviewed-head-sha` marker written. The independent review is the merger's.

### Two things the merger should look at directly

**1. A deliberate non-fix that a reviewer may disagree with.** The `/simplify` altitude pass argued, with a working demonstration, for an opportunistic `try: import yaml / except ImportError: yaml = None` — real `safe_load` where the parser exists, scanner as the universal floor. A mis-indented workflow passes this scanner but raises `ScannerError` in `safe_load`, so the gap is real.

I did not apply it, because contract item 5's literal falsifier is *"the guard contains no `import yaml` / `from yaml`"*, and adding it would make a pre-committed item read `violated`. The grader independently agreed with that reading, noting the contract's own `risk_watchlist` reinforces the binary presence test, and that item 20 handles the residual correctly by requiring the docstring to disclose exactly this limit — which it does. **If the merger prefers the wider coverage, that is a contract amendment, not a bounce.**

**2. The preflight was structurally vacuous on this diff — green, but nothing was examined.** `check-preflight.sh` exited 0 with `status: no_applicable_checks` and **all 7 instruments skipped**: they are plugin-scoped (`*/scripts/`, `*/tests/`) and this is a repo-root diff, and `mutation-recipe` additionally globs `test-*.sh`, which is blind to this repo's underscore suite naming. Do not read that green as evidence.

I ran the equivalent checks directly instead, and the merger should re-run rather than trust these:

| Check | Result |
|---|---|
| `python3 scripts/run-plugin-tests.py` | **109/109 pass** |
| `bash tests/test_check_workflow_yaml.sh` | 10 cases, all pass, no ANSI on any result line |
| `python3 scripts/check-workflow-yaml.py` | exit 0 on the real tree, 4 files scanned, zero false positives |
| Mutation (`--case WY2`) | verified by hand — red under mutation, green restored, re-checked *after* the cleanup pass |
| Cross-check vs. real parser | the `WY2` fixture makes `yaml.safe_load` raise `ScannerError`, confirming the guard targets the genuine defect |

### Note on the base

`origin/main` advanced twice mid-run — #1358 then #1360 merged. The branch is cut from the later tip, so the contract's citations to `lint.yml` line numbers and to an ANSI emitter in `tests/test_check_external_dispatch.sh` are stale; the grader re-verified each against the current base rather than failing an item on a stale anchor. One consequence worth naming: contract item 4 required `lint.yml` untouched because an unmerged PR owned it. That PR has merged, so the *rationale* is stale — but the requirement still holds in the diff, and the standalone-workflow design never depended on it (it rests on the self-reference asymmetry).

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1367
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator

**Summary:** All three perspectives pass: contract-grading-reviewer confirms all 22 committed checklist items satisfied with file:line and empirical (test-run) evidence, plugin-validator confirms no plugin-registration or repo-structure invariant is disturbed and the new anti-vacuity suite is genuinely non-vacuous, and the only remaining item is an un-waived, merger-added diff-size preference (806 lines vs the 400-line bound) that does not gate merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (advisory-addition) `PR diff` — Diff is 806 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

### Positive observations

- All 22 contract checklist items graded satisfied with file:line evidence; item 5 (no import yaml) satisfied as written; item 3 confirmed empirically - the Workflow-YAML guard check materialises on the PR check surface. (contract-grading-reviewer)
- Anti-vacuity implemented as two distinct code paths (missing dir vs empty listing), each asserted independently. (plugin-validator)
- Detection generalises past the literal name key - verified against if:, script:, and env: child keys, all genuine ScannerError shapes. (plugin-validator)
- Block-scalar handling re-processes the dedenting line, so a violation after a run: | body is still caught. (plugin-validator)
- Docstring's stated blind spot verified accurate by constructing the disclosed indentation malformation - the limit is described correctly, not decoratively. (plugin-validator)

### Recommendation

---

PR #1399, opened 2026-08-16
cogni-knowledge: consolidate test emitters onto the shared helper

### LINKED ISSUE #1364: cogni-knowledge: consolidate the inlined red/green emitters onto the shared test helper

claude-cogni-work:
Closes #1364.

## Summary

`cogni-knowledge/tests/fixtures/test_helpers.sh` exists to be the single definition of the
`red()` / `green()` result emitters, but 15 suites still carried byte-identical inlined
copies. This deletes those copies and sources the helper instead, so the helper is now the
only definition under `cogni-knowledge/tests/`.

Net: **16 files changed, 20 insertions, 47 deletions.**

Two edit shapes, both matching existing house style:

- **7 plain suites** — the source line takes the slot the two definitions vacated
  (precedent `test_candidate_store.sh:27-32`).
- **8 cycle-guard-family suites** — the source line goes *above* their existing
  `. "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"` line, matching the sole multi-source
  precedent, `test_finalize_contract.sh:21-22`, which sources `test_helpers.sh` first.

No call site changed: the helper's bodies are byte-identical to the copies they replace,
so this is a pure de-duplication with no behavioral surface.

## The one non-obvious deliverable

`test_plain_result_emitters.sh:139-141` carried a count this change falsifies —
"78 files exercise the contract today (16 define the emitters, 62 source the helper)".
It is restated to the post-change split (1 defining, 77 sourcing, total still 78).

The floor logic one line below is deliberately **untouched**:

```sh
if [ "$exercisers" -lt 50 ]; then
```

It is keyed on *exercisers* (define **or** source), not definers, precisely so
consolidation cannot trip it — #1359 re-homed it for exactly this reason. Re-keying it to
`$definers` would false-fail the guard at 1.

## Why a line per suite, and not the "deeper" fix

The obvious objection is that the 8 cycle-guard suites already source
`fixtures/_cycle_guard_lib.sh`, so sourcing the helper from *inside* that lib would fix 8
files with one line instead of 8. That alternative was measured and rejected, because it
would quietly blind the guard.

`test_plain_result_emitters.sh:104` counts an exerciser with a per-file literal match:

```sh
grep -qF 'test_helpers.sh' "$f"
```

That is a text match, not a resolved source graph. Route the emitters transitively and none
of the 8 suites carries the literal any more — only the lib does. The exerciser count falls
**78 → 71** (−8 suites, +1 lib) against a floor of 50, so the guard stays **green while
going blind to 8 files that genuinely depend on the contract**. Nothing else would flag it:
the escape-literal sweep and the definition-shape loop are both unaffected. That is exactly
the vacuous pass the liveness floor exists to prevent.

Two smaller reasons the per-suite line is right: `test_finalize_contract.sh` — the lib's 9th
consumer — already carries its own explicit helper source line, so these 8 additions match a
sibling rather than inventing a convention; and `_cycle_guard_lib.sh` is a fixture *builder*
that uses neither emitter, so making it a helper conduit would invert the dependency and make
emitter availability implicit and source-order-dependent for every future consumer.

## Verification

| Check | Result |
|---|---|
| `bash cogni-knowledge/tests/test_plain_result_emitters.sh` | exit 0 |
| Guard's case-3 line | `PASS: 1 emitter-defining file(s) are paired and plain, of 78 exercising the contract` (base: `16 … of 78`) |
| `grep -lE '^(red\|green)\(\)' cogni-knowledge/tests/*.sh` | 0 files (base: 15) |
| Each of the 15 suites run standalone from repo root | all exit 0 |
| `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` | 77/77 passed (base: 77/77) |
| `python3 scripts/run-plugin-tests.py` (full repo sweep) | 115/115 passed |
| Call-site counts per file, base vs head | identical in all 15 |
| `fixtures/test_helpers.sh` in diff | no |
| `cogni-knowledge/tests/README.md` in diff | no |

The `16 → 1` flip with exercisers steady at `78` is the observable proof the consolidation
happened; a no-op diff would leave it at 16.

### A counting note for whoever grades this

Two different greps give two different (both correct) sourcer counts, so please don't read
one as a shortfall against the other:

- files containing the **exact idiom** `. "$(dirname "$0")/fixtures/test_helpers.sh"` — **61 → 76**
- files **mentioning** `test_helpers.sh` at all — **62 → 77**

Both moved by exactly +15. The single file in the gap is
`test_plain_result_emitters.sh` itself, which references the helper via
`HELPER="$TESTS_DIR/fixtures/test_helpers.sh"` in order to *probe* it — it is the guard,
not a consumer, and it correctly does not source it.

## Scope decisions

**Deliberately not edited:**

- `cogni-knowledge/tests/fixtures/test_helpers.sh` — already correct; this PR only adds consumers.
- `cogni-knowledge/tests/README.md` — line 42 already reads "Source `fixtures/test_helpers.sh`
  for `red`/`green` rather than inlining". **AC 6 is satisfied by verification, not by a diff** —
  an untouched README is the correct outcome here, not a missed criterion.
- `test_plain_result_emitters.sh:142` — the floor logic, see above.
- Any `plugin.json` / `marketplace.json` version — the post-merge workflow owns the bump.

**Preserved hazards** (both found before editing, both verified in the diff):

- `test_portal_store.sh` — `fail() { red "FAIL: $1"; … }` sits immediately below the def
  block. Only the two definition lines were removed; `fail()` is intact.
- `test_cycle_guard_direct.sh` — `# shellcheck source=fixtures/_cycle_guard_lib.sh` pins the
  *lib* source, so the new helper source line was placed above the pragma, keeping the
  pragma adjacent to the line it pins.

`$(dirname "$0")` is safe in all 15: none reassigns `$0`, none sources via `BASH_SOURCE`,
and every `cd` is confined to a `$(cd … && pwd)` subshell. The anchor is load-bearing rather
than stylistic — `scripts/run-plugin-tests.py` executes each suite with `cwd=repo root`, so a
relative `. fixtures/test_helpers.sh` would break under CI while still passing by hand from
inside the tests directory.

`test_resolve_wiki_scripts.sh` drives snippet bodies through `bash -c`, but every emitter
call happens in the outer shell, so sourcing versus inlining is neutral there.

## Mutation recipe

Run from the PR-head checkout root — a non-PR-head `--root` false-blocks.

Proves case `helper` of the touched guard is falsifiable. The mutation rewrites only the
`EXPECTED=` literal, so the captured helper output no longer matches it and the case flips
`PASS: helper emits plain result lines on stdout` → `FAIL: helper did not emit the expected
plain result lines`.

```sh
bash mutation-check.sh --root . --file cogni-knowledge/tests/test_plain_result_emitters.sh --expr "s/EXPECTED='PASS: probe-green/EXPECTED='PASS: probe-lime/" --test "bash cogni-knowledge/tests/test_plain_result_emitters.sh" --case helper
```

Three things this recipe is deliberate about, each a way an earlier draft failed:

- **No `^` anchor.** The harness runs `perl -0pi -e` (slurp mode), where `^` matches only the
  start of the *file*, not of a line. An anchored expression matches nothing and the harness
  hard-errors `expr_no_op` (exit 2) rather than proving anything.
- **The `EXPECTED='` prefix is load-bearing.** `probe-green` appears twice — on the
  `CAPTURED=` line and on the `EXPECTED=` line directly below it. Mutating both would move the
  two sides together and leave the case green, proving nothing; the prefix pins the match to
  exactly one site.
- **The replacement is disjoint from the literal it replaces** (`probe-lime` does not contain
  `probe-green`), so the mutation genuinely changes the compared text.

`--case helper` is correct because the harness matches the first whitespace-delimited token,
whole-token, against both the green and red labels — and both
`PASS: helper emits …` and `FAIL: helper did not emit …` begin with `helper`.

This recipe was executed, not merely written.

### DIFF
--- a/cogni-knowledge/tests/test_binding_project_path.sh
+++ b/cogni-knowledge/tests/test_binding_project_path.sh
@@ -22,8 +22,7 @@ set -eu
 PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/knowledge-binding.py"
 
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 
 if [ ! -f "$SCRIPT" ]; then
   red "FAIL: knowledge-binding.py not found at $SCRIPT"
--- a/cogni-knowledge/tests/test_cycle_guard_clear.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_clear.sh
@@ -15,10 +15,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_depth_bound.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_depth_bound.sh
@@ -15,10 +15,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_direct.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_direct.sh
@@ -16,11 +16,10 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
+
 # shellcheck source=fixtures/_cycle_guard_lib.sh
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_distilled.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_distilled.sh
@@ -24,10 +24,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 errors=0
 
--- a/cogni-knowledge/tests/test_cycle_guard_dry_run.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_dry_run.sh
@@ -15,10 +15,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_not_applicable.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_not_applicable.sh
@@ -15,10 +15,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_cycle_guard_question.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_question.sh
@@ -23,10 +23,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 errors=0
 
--- a/cogni-knowledge/tests/test_cycle_guard_transitive.sh
+++ b/cogni-knowledge/tests/test_cycle_guard_transitive.sh
@@ -16,10 +16,8 @@ PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/cycle-guard.py"
 
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 . "$TESTS_DIR/fixtures/_cycle_guard_lib.sh"
-
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
 
 WORK=$(mktemp -d)
 trap 'rm -rf "$WORK"' EXIT
--- a/cogni-knowledge/tests/test_fetch_cache.sh
+++ b/cogni-knowledge/tests/test_fetch_cache.sh
@@ -23,8 +23,7 @@ set -eu
 PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 SCRIPT="$PLUGIN_ROOT/scripts/fetch-cache.py"
 
-red()   { printf '%s\n' "$1"; }
-green() { printf '%s\n' "$1"; }
+. "$(dirname "$0")/fixtures/test_helpers.sh"
 
 if [ ! -f "$SCRIPT" ]; then
   red "FAIL: fetch-cache.py not found at $SCRIPT"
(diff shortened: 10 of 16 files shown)


claude-cogni-work:

Author instance is done with #1364. This PR is draft + `svc:needs-review`, awaiting the
merger instance. No verdict label, no `reviewed-head-sha` marker, no ready flip — the
independent review is the merger's to run.

Head at handoff: `2af9b2fb`.

### Handoff preflight

`check-preflight.sh` exit **0**, `clean: true`, 16 touched paths, 1 touched plugin
(`cogni-knowledge`). Base `origin/main` @ `508af90d`.

**But 6 of 7 instruments skipped, and two of those skips are not what they look like.**
Naming them rather than letting a green run pass for a checked one:

| Instrument | Status | Reason given |
|---|---|---|
| `run-tests` | skip | `no_suites_at_base` — "reported no_suites_discovered for cogni-knowledge" |
| `mutation-recipe` | skip | `not_applicable` — "no touched path matching `*/tests/test-*.sh` or `*/scripts/check-*.sh`" |
| `frontmatter` / `breadcrumbs` / `skill-length` / `agent-length` | skip | `not_applicable` — no touched SKILL.md or agents/*.md (genuinely correct) |

The last four skips are honest. **The first two are false negatives from a glob mismatch:**

- `run-tests.sh:114` discovers suites with `SUITES=("$TESTS_DIR"/test-*.sh)` — **hyphen**.
- `check-preflight.sh:853` matches `*/tests/test-*.sh` / `*/scripts/check-*.sh` — **hyphen**.
- `cogni-knowledge/tests/` contains **77 underscore-named** suites (`test_*.sh`) and **zero**
  hyphen-named ones.

So for this plugin the preflight's regression arm and its mutation-recipe requirement are
structurally blind: `run-tests` concludes the base carried no suites (it carries 77), and a
diff that touches a guard (`test_plain_result_emitters.sh`) is judged to touch none, so the
mutation-recipe requirement silently does not apply. The repo's own CI is unaffected —
`scripts/run-plugin-tests.py` globs `tests/*.sh` and runs all 77 — which is exactly why CI is
green while the preflight sees nothing.

**Both gaps were covered manually, so the diff is not unverified — only the instrument is:**

- Regression: `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` → **77/77**;
  full repo sweep → **115/115**. Each of the 15 converted suites also run standalone → all exit 0.
- Mutation: the `## Mutation recipe` in the PR body was **executed**, not just written —
  verdict `guard_verified` (mutated → red, restored → green), so case `helper` is proven
  falsifiable.

This is a cogni-service instrument defect, not a defect in this PR. Flagging it because a
reviewer reading `clean: true` would reasonably assume the tests arm ran.

### Contract grading (author self-grade — advisory)

Graded against the **issue-time** contract on #1364 (`--target issue --origin triage`,
derived `2026-08-16T02:39:39Z`, 14 items) — written before any branch was cut, so the author
did not write its own bar.

**Verdict: pass — 14/14 satisfied, 0 violated, 0 unclear. Residual: none.**

One item is worth the merger's attention, because the contract is internally inconsistent and
a naive re-grade could read it as a shortfall:

> **Item 6** asserts the sourcing count goes `62 -> 77`.

Two different greps give two different (both correct) counts:

- files containing the **exact idiom** `. "$(dirname "$0")/fixtures/test_helpers.sh"` — **61 → 76**
- files **mentioning** `test_helpers.sh` at all — **62 → 77**

Both moved by exactly **+15**. The contract's stated `62` is the *mentions* metric while the
item nominally describes the *exact-idiom* metric — a derivation artifact in the contract, not
a defect in the diff. The one file in the gap is `test_plain_result_emitters.sh` itself, which
references the helper via `HELPER="$TESTS_DIR/fixtures/test_helpers.sh"` in order to *probe*
it; it is the guard, not a consumer, and correctly does not source it. The grader verified this
independently and reached the same conclusion.

Item 14 is also worth not re-litigating: it states explicitly that AC 6 requires **no diff**,
so the untouched `README.md` **satisfies** it rather than failing it.

### Quality passes

`/simplify` ran 4 review agents (reuse / simplification / efficiency / altitude). Three
findings applied before commit:

1. Restored the blank separator after the source line in 6 suites (my line-collapse had eaten
   it; `test_portal_store.sh` was checked and correctly excluded — it never had one).
2. Trimmed an added clause from the guard comment that re-introduced the literal `16` — a
   point-in-time count with no referent post-diff, in a PR whose purpose is removing a stale
   count.
3. Added a blank line above the `# shellcheck source=` pragma in `test_cycle_guard_direct.sh`
   so it visually binds to the line it pins.

The altitude pass also measured and rejected the tempting "deeper" fix — see
**Why a line per suite** in the PR body: sourcing the helper from inside `_cycle_guard_lib.sh`
would drop the guard's exerciser count 78 → 71 against a floor of 50, staying green while
going blind to 8 files.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1399

**Summary:** Both routed perspectives (plugin-validator, contract-grading-reviewer) passed clean with zero findings; all 14 committed acceptance-checklist items verified satisfied by direct execution.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- Scope exact: no path outside cogni-knowledge/tests/; version-bump gate reports 1 touched, 0 violations. (plugin-validator)
- Zero structural regression: manifest parses, version mirrors, 21/21 skills and 16/16 agents valid, skill-name and breadcrumb guards clean. (plugin-validator)
- The load-bearing $(dirname "$0") anchor is used verbatim in all 15 files, exactly once each; suites pass both from repo root and from inside tests/. (plugin-validator)
- Both mis-ordering traps cleared: test_fetch_cache.sh sources at 26 ahead of its early-exit red at 29; test_portal_store.sh sources at 35 ahead of the fail() wrapper. (plugin-validator)
- Half-conversion definitively excluded: zero surviving definitions plus the guard's own 16->1 count flip with exercisers held at 78. (plugin-validator)
- No shadowing: helper's assert_grep/assert_not_grep disjoint from every local helper and from all _cycle_guard_lib.sh functions. (plugin-validator)
- In the nine cycle-guard suites the helper is sourced BEFORE _cycle_guard_lib.sh — strictly safer than base. (plugin-validator)
- No watchlist over-reach: no export -f, no bash -c restructuring, no bash-3.2-breaking modernization. (plugin-validator)
- Liveness floor correctly left at exercisers -lt 50, not re-keyed to $definers. (plugin-validator)
- AC 6 handled by verification not edit: README.md legitimately absent, line-42 bullet still accurate. (plugin-validator)
- item 4: zero definers on head; exactly the 15 named files at base. (contract-grading-reviewer)
- item 6: mention-count 62->77 and strict-idiom-count 61->76 both a clean +15; the one-file gap is the guard's own self-mention, present at base and head. (contract-grading-reviewer)
- item 7: source line precedes first emitter call in all 15, including both risk-flagged files. (contract-grading-reviewer)
- item 8: emitter call-site counts identical base vs head in all 15; each diff is only the 2-line removal plus 1 source line. (contract-grading-reviewer)
- item 9: head prints 'PASS: 1 emitter-defining file(s) are paired and plain, of 78 exercising the contract', exit 0; base prints '16 ... of 78'. (contract-grading-reviewer)
- item 10: all 15 suites exit 0 standalone from repo root, no command-not-found or unbound-variable. (contract-grading-reviewer)
- item 11: run-plugin-tests.py --filter cogni-knowledge reports 77/77, matching base. (contract-grading-reviewer)
- item 12: line 142 still reads exercisers -lt 50, unchanged. (contract-grading-reviewer)
- item 13: '16 define the emitters' absent from head; replacement states 1 defining / 77 sourcing, total 78. (contract-grading-reviewer)
- item 14: README.md legitimately absent; base bullet at line 42 already accurate. (contract-grading-reviewer)

### Recommendation

---

Issue #1406, opened 2026-08-16
repo: result-line plainness guard cannot reach shared emitter helpers under tests/fixtures/


claude-cogni-work:
## Problem

The result-line plainness guard shipped for #1362 is scoped to the two discovery globs `scripts/run-plugin-tests.py` uses — `tests/*.sh` and `*/tests/*.sh`. Both are non-recursive, so neither reaches `<plugin>/tests/fixtures/*.sh`.

`cogni-knowledge/tests/fixtures/test_helpers.sh` is the repo's only shared result-emitter definer. It defines `red()`/`green()` (both plain today) and roughly 77 of the 115 in-scope suites source it. An escape sequence landing there would wrap the emitted result line of every one of those suites at once — a larger blast radius than anything the new guard does cover — and the new guard structurally cannot see the file.

## This is a latent gap, not a live one

Stated precisely, because the distinction decides how urgent this is:

- The helper is **not** unguarded today. `cogni-knowledge/tests/test_plain_result_emitters.sh` covers that exact file twice — its case 2 greps its tree recursively for the escape literal, and its case 3 globs `"$SCAN_ROOT"/fixtures/*.sh` explicitly and asserts both emitters are the plain `printf` form. That suite runs in CI via `run-plugin-tests.py`, so an escape introduced in the helper would fail the build today.
- The real exposure is that this coverage is **plugin-scoped**. `SCAN_ROOT` defaults to `cogni-knowledge/tests`, and that suite's own header says a wider guard must re-home its liveness floor. Any *other* plugin that grows a `tests/fixtures/` helper would be covered by nothing.
- `cogni-knowledge` is the only plugin with `tests/fixtures/*.sh` today, so there is no live defect — only a gap that opens the moment a second plugin adds one.

## Proposed change

Decide between two dispositions, both of which need their own reviewer pass:

**(a) Add a third discovery glob** — extend the guard to `tests/fixtures/*.sh` and `*/tests/fixtures/*.sh`. Note this preserves the archive exclusion: the parked carrier sits at `<plugin>/_archive/tests/`, which a `*/tests/fixtures/*.sh` pattern still does not match, so no exclusion list is introduced. The cost is that the guard's discovery would no longer mirror `run-plugin-tests.py` exactly, which was a deliberate property of the original scope.

[...]

PR #1405, opened 2026-08-16
cogni-workspace: reconcile wiki entries_count and unresolved page references, record per-delta decisions

claude-cogni-work:
Refs #1384.

## Summary

The two wiki trees disagree in 28 places. This lands the half that needs no judgment call, and
records a decision for the half that does.

- **`entries_count` corrected in both configs** — root `179` → `170`, bundled `178` → `176`.
  Both overstated their own tree, by omission: the field was last set correctly by hand and three
  later commits deleted pages without touching either config. Corrected by hand for the same
  reason it was set by hand — the only writer in the repo enumerates pages by walking a fixed set
  of per-type subdirectories, and this wiki keeps every page in one flat directory that is not in
  that set, so pointed at either tree it would write `0`. The two values stay different from each
  other because each describes its own tree.
- **The two unresolved page references at `workflow-install-to-infographic.md:60` are removed**,
  identically in both copies. That page is pinned to byte-identity between the trees by
  `cogni-workspace/tests/test-layering-claim-reconciled.sh` (`PAGE_PARITY`), so a one-sided edit
  turns case `L7` red.
- **`cogni-workspace/references/wiki-tree-reconciliation.md`** — the per-delta decisions record,
  in the house style of `references/absorption-roadmap.md`, with every figure attributed to the
  command that produced it.
- **`cogni-workspace/tests/test-wiki-tree-parity.sh`** — a new guard, discovered by
  `scripts/run-plugin-tests.py` (CI job "Plugin test suites").

## The finding worth reading

**The 20 content deltas do not all resolve in the same direction.** Thirteen files are newer at
the root — but four workflow pages are newer in the *bundle*, because an earlier change appended
`## Steps` and `## Common pitfalls` sections (and `## Two scenarios` on one) directly to the
bundled copies, bypassing the one-way sync contract.

| Page | Root lines | Bundled lines |
|---|---:|---:|
| `workflow-portfolio-to-website.md` | 55 | 76 |
| `workflow-content-pipeline.md` | 50 | 79 |
| `workflow-portfolio-to-pitch.md` | 56 | 72 |
| `workflow-trends-to-solutions.md` | 52 | 79 |

A blanket root-over-bundle sync would destroy content that exists nowhere else, on top of
deleting the seven bundled-only pages. `scripts/release-bundle-wiki.sh` is documented as "Run
BEFORE every cogni-workspace marketplace publish", and its post-sync guard compares page *counts*,
so it cannot detect the second class at all. Filed separately as #1403.

[...]

PR #1407, opened 2026-08-16
Guard against escape-wrapped result lines in discovered test suites

claude-cogni-work:
Refs #1362.

## Summary

Adds `scripts/check-result-line-plainness.py`, a hard-clean-zero CI guard asserting that no test suite discovered by `scripts/run-plugin-tests.py` can emit an escape-wrapped `PASS:`/`FAIL:` result line.

**What a "result line" is, and why the guard is shaped the way it is.** A result line is the line a suite *emits*. Run a colouring emitter and it prints the escape immediately before the label:

```
$ red() { printf '\033[31m%s\033[0m\n' "$1"; }; red "FAIL: probe-case" | cat -v
^[[31mFAIL: probe-case^[[0m
```

That defeats the mutation harness match `^[[:space:]]*FAIL:[[:space:]]+<case>` — it matches the plain form and not the coloured one — so a genuinely red case reports `case_not_found`: a real failure read as a missing one. A CI guard cannot execute all 115 suites to observe the emitted line, so it detects the *source constructs that produce one*. That is why detection is not a single-source-line match: in every emitter shape in the repo the escape lives in the definition body and the label arrives as the call-site argument, so the two are far apart in source even though they are adjacent in output.

- Discovery mirrors `run-plugin-tests.py` verbatim: the two non-recursive globs `tests/*.sh` and `*/tests/*.sh`, set-collected and sorted. No third glob, no prefix narrowing, so both `test-foo.sh` and `test_foo.sh` are in scope (the repo uses both namings).
- Two **disjoint** arms in one line loop. **Arm A** (`escape_literal`) fires on an escape literal anywhere in an in-scope suite outside an emitter definition. **Arm B** (`emitter_not_plain`) fires on an emitter definition whose body carries one — the copy-an-emitter-into-a-new-suite regression the issue names as most likely, which Arm B catches even when every call-site label line is clean.
- Detection keys on the 4-character **source** literal, not a raw-byte scan. The only carrier in the tree holds it as source text, so a byte-only scan would find nothing and ship green and dead; the raw byte is covered additively, never as a substitute.
- Zero discovery is a hard exit-2 error, never a clean zero.
- Ships `tests/test_check_result_line_plainness.sh` with a **per-arm** liveness floor, wires a fourth job into `.github/workflows/lint.yml`, and registers the guard in `CLAUDE.md`.

## Scope decisions

- **The archive carve-out is structural, not coded.** `cogni-knowledge/_archive/tests/test_read_project_config_bare.sh` really does define escape-carrying emitters, and sits one path segment deeper than either glob reaches. The guard names no directory and contains zero occurrences of the string `_archive`; case S2 proves the exclusion with a fixture at that real-world shape rather than with an exclusion list. No baseline, no allowlist, no skip marker — a clean-zero invariant, following `check-workflow-yaml.py` and `check-external-dispatch.py` rather than `check-breadcrumbs.py`'s ratchet.
- **Per-arm liveness floors.** `files_scanned` and `definitions_inspected` both increment *before* any arm condition, and the suite floors them (`>= 80`, `>= 40`). Sharing one floor across both arms would let one arm die — regex drift, a narrowed pattern — while the shared count stayed above the floor, reporting green with half the detection gone. They are inequalities on purpose: an exact total would turn every future suite addition into a false failure, the reasoning `tests/test_run_plugin_tests.sh` already records for its own real-tree case.
- **Arm B is escape-anchored, not form-anchored.** Requiring a canonical `printf '%s` emitter body would fire on ~30 clean files today, since plugin suites legitimately use `pass()`/`fail()`/`ok()` shapes with `echo` bodies and some emit `ok:` or bare `OK`.
- **`cogni-knowledge/tests/test_plain_result_emitters.sh` is untouched.** Its header invites a wider guard to reuse it via `SCAN_ROOT`, but its case 2 is a recursive `grep -rlF`, so a repo-root scan root would recurse into the archive and go red — inviting exactly the exclusion bandaid this issue forbids. Its per-file loop also already reaches into `fixtures/`, which is neither mandated glob. What is reused is its *mechanism* (run-time pattern assembly so a suite cannot trip its own sweep), not the file.
- **Nothing is gated on an environment probe.** Both new files decide by rule and emit unconditionally. Deciding by a tty test would make result-line parsability environment-dependent — run a suite under a pty and the colour, and the failure, come back.

## Test plan

- [ ] `python3 scripts/check-result-line-plainness.py` exits 0 on the real tree with `summary.total == 0`.
- [ ] `bash tests/test_check_result_line_plainness.sh` exits 0. Cases: A1/A2 Arm A fires alone and names file, line and arm; B1/B2 Arm B fires alone on a coloured emitter whose call sites are clean, with `by_arm` holding only `emitter_not_plain` (the disjointness assertion); S1 four dirty suites across both globs and both naming shapes are all discovered; S2 the archive-depth fixture is not discovered; Z1 an empty tree exits 2; C1 the known false-positive hazards stay clean; R1 the real tree exits 0; L1/L2 the per-arm liveness floors.
- [ ] `python3 scripts/run-plugin-tests.py` stays green. The base count is 115 and this PR adds one suite, so the post-change count is 116 — every assertion is an inequality and no suite pins an exact total.
- [ ] `python3 scripts/check-workflow-yaml.py` exits 0, confirming the new job carries no colon-space in any name.
- [ ] `bash cogni-knowledge/tests/test_plain_result_emitters.sh` still exits 0 (untouched, and must stay green).
- [ ] The `CLAUDE.md` bullet's quoted CI job name matches `.github/workflows/lint.yml` byte-for-byte.

## Mutation recipe

Each arm has its own falsifier, and each replacement is disjoint from the text it replaces — a replacement containing its own match substitutes cleanly and leaves the guard live, reporting a survived mutation rather than a verdict.

Arm A — kill the escape scan on non-definition lines; case `A1` must go red:

```bash
  --root . \
  --file scripts/check-result-line-plainness.py \
  --expr 's/for m in ESC_RE\.finditer\(line\):/for m in []:/' \
  --test 'bash tests/test_check_result_line_plainness.sh' \
  --case A1
```

Arm B — kill the emitter-definition scan; case `B1` must go red:

```bash
  --root . \
  --file scripts/check-result-line-plainness.py \
  --expr 's/if defn and ESC_RE\.search\(defn_body\):/if False:/' \
  --test 'bash tests/test_check_result_line_plainness.sh' \
  --case B1
```

Both anchor lines appear exactly once in the guard, each on one physical line, so each `perl -0pi` expression (no `/g`) rewrites exactly its own site. Case ids are `<ID> <description>` — id then a space, never a colon abutting the id — so the harness's whole-token match resolves them. Floors L1 and L2 stay green under both mutations by design: their counters increment before either arm condition, so a dead arm is proved by A1/B1 going red rather than by the floors collapsing.

## Open questions

1. `cogni-knowledge/tests/fixtures/test_helpers.sh` defines the emitters for ~77 suites but sits at `<plugin>/tests/fixtures/`, one segment deeper than both globs the acceptance criteria pin, so this guard structurally cannot see it. It is **not** unguarded today — `test_plain_result_emitters.sh` covers that exact file twice (its case 2 greps the tree recursively, its case 3 globs `fixtures/*.sh` and asserts both emitters are plain) and runs in CI. The residual gap is latent and narrower: that coverage is plugin-scoped, so any *other* plugin growing a `tests/fixtures/` helper would be guarded by nothing, and cogni-knowledge is the only one with such a helper today. Widening the globs would violate the acceptance criteria, so this PR does not widen them and files the gap as a tracked follow-up. Please confirm the follow-up is the intended disposition rather than an in-scope third glob.


claude-cogni-work:

Author instance `youthful-bohr-e0f79a` has finished. This PR is draft and carries `svc:needs-review`; the draft flag is the gatekeeper's to flip, not the author's.

### Handoff preflight

`check-preflight.sh` exit **0**, `clean: true` — but read it as **vacuous, not reassuring**: `status: no_applicable_checks`, `gating_applicable: 0`, `passed: 0`, **`skipped: 7`**. Every instrument skipped, all for the same structural reason — they are plugin-scoped and this is a repo-root change.

| instrument | skip reason |
|---|---|
| `run-tests` | no touched `*/scripts/`, `*/tests/`, SKILL.md or agents path |
| `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` | no touched SKILL.md or `agents/*.md` |
| `version-bump` | touched set resolves to no plugin |
| `mutation-recipe` | no touched path matching `*/tests/test-*.sh` or `*/scripts/check-*.sh` |

**The `mutation-recipe` skip is the one to notice.** This PR ships exactly the recipe that instrument exists to demand, and it still skipped: the guard is repo-root `scripts/check-*.py` (not `*/scripts/check-*.sh`) and the suite is repo-root `tests/test_check_*.sh` (underscore, not `*/tests/test-*.sh`). So the preflight did not verify the recipe. **Do not read "preflight clean" as "recipe checked."** Both recipes were run manually and each returned `verdict: guard_verified` (mutated red, restored green) — never `case_not_found`, never `expr_no_op` — and the contract grader re-ran both independently and got the same result.

### Contract self-grade

Graded against the acceptance contract committed to #1362 at triage time, before any branch was cut (`derivedAt: 2026-08-16T06:25:53Z`, 20 items).

**20 graded · 20 satisfied · 0 violated · 0 unclear · 0 residual.** No fix pass was needed. The grader verified independently rather than on trust: it re-ran the guard, the suite, `run-plugin-tests.py`, both mutation recipes, `check-breadcrumbs`, `check-version-bump`, `check-workflow-yaml`, confirmed zero `_archive` occurrences in the guard source, confirmed the CLAUDE.md job name is byte-identical to `lint.yml`, and confirmed `cogni-knowledge/tests/test_plain_result_emitters.sh` is genuinely untouched and still exits 0.

### One thing the merger should know about the contract itself

The stored contract still carries an escalation asserting that acceptance criterion 1 is vacuous — that a guard matching "an escape sequence before its `PASS:`/`FAIL:` label" could never fire, because the escape and the label never share a source line.

**That claim was wrong and was withdrawn before it reached this PR.** It rested on reading "result line" as a *source* line. The repo's own vocabulary (`tests/test_check_workflow_yaml.sh`) and a direct run both show it means the *emitted* line:

```
$ red() { printf '\033[31m%s\033[0m\n' "$1"; }; red "FAIL: probe-case" | cat -v
^[[31mFAIL: probe-case^[[0m
```

The escape *is* immediately before the label there. Criterion 1 is coherent and satisfiable as written. The guard design never depended on the claim — detecting the source constructs that produce such a line was already the right static approximation for a check that cannot execute 116 suites. The contract grader reviewed the withdrawal and agreed it was correct, noting the diff satisfies the Arm B item under either reading. Please disregard that escalation when reading the stored contract.

### State

- 4 files, 453 insertions. Base `origin/main` @ `0698cb2c`, unchanged throughout — the contract's measured counts are still comparable.
- Guard: 116 discovered / 116 scanned / 81 definitions inspected / `total: 0` on the real tree.
- Suite: 14/14 cases pass. Full sweep: 116 suites, 0 failed.
- Links `Refs #1362`, not `Closes` — scope was deferred to #1406, so **#1362 stays open on merge** and needs a human close once the open question below is judged.

### Open question carried

`cogni-knowledge/tests/fixtures/test_helpers.sh` defines the emitters for ~77 suites but sits one segment deeper than both mandated globs, so this guard cannot see it. It is **not** unguarded today — `test_plain_result_emitters.sh` covers it twice and runs in CI — so the gap is latent: the coverage is plugin-scoped, and cogni-knowledge is the only plugin with such a helper. Widening the globs would violate #1362's own acceptance criteria, so #1406 tracks it instead. Confirm that disposition, or direct an in-scope third glob.

sdh07:
## Held for a human — escalation reasons

The adopted review contract is committed on the linked issue (#1362), not on this PR, so the
escalation reasons are recorded here for a durable PR-side record:

- **Scope tension the issue leaves open.** Acceptance criterion 1 specifies only the DIRECT form
  ("a result line carries an escape sequence before its `PASS:`/`FAIL:` label"), but the Problem
  section names the most likely regression as a copied `red()` DEFINITION line — escape on the
  definition, label lines clean. A guard satisfying criterion 1 literally can be blind to the path
  the issue predicts. Which governs: the written criterion (Arm A only) or the stated regression
  path (both arms)? The contract gates both arms and flags the question rather than silently
  narrowing or widening the bar.

Recorded, not holding: the contract also notes the acceptance bar was materially under-specified
(6 plan-time AC bullets vs 19 base-grounded gating items). That entry is `avoidable`
(`insufficient_coverage`) — issue-quality telemetry, and it does not hold this PR.

Parked on a human ruling; neither reviewed nor bounced. To release: rule on the Arm A / Arm B
question, write the ruling into the issue's `## Acceptance criteria` (so the next drain re-derives
against the corrected bar), then remove `svc:needs-human-review`.

sdh07:
## Ruled — hold released, PR cleared to merge

The escalation held this PR on one question: does acceptance criterion 1 of #1362 govern the
direct form only (Arm A), or also the copied-`red()`-definition path the Problem section names
(Arm B)?

**Both arms govern.** The ruling is written into #1362's **body** — criterion 1 is rewritten to
name both shapes, and a `## Ruling` section records the reasoning. It is in the body rather than
a comment on purpose: a comment does not bump `lastEditedAt`, so a contract re-derived from this
issue would be blind to it and would raise the same escalation again.

Two independent reasons, both recorded on the issue:

- A "result line" is the line a suite **emits**, not a source line. A colouring emitter prints the
  escape immediately before the label — `^[[31mFAIL: probe-case^[[0m` — so the emitter-definition
  arm is *how* criterion 1 is satisfied for that shape, not an extension past it. The apparent
  tension was an artefact of the source-line reading.
- A guard that passed the letter while missing the predicted regression path would not catch the
  thing the issue was filed to catch.

This confirms the author instance's withdrawal of the "criterion 1 is vacuous" claim carried in
the stored contract. The withdrawal was correct.

**Second question — the `tests/fixtures/` gap: #1406 stands, no third glob.** The two-glob
boundary in criterion 3 is load-bearing — it is what excludes `_archive/` structurally rather than
through the exclusion list criterion 4 forbids. Widening the population to close a gap that is
latent (`test_plain_result_emitters.sh` covers the one existing helper twice, in CI) would break
the issue's own bar to buy nothing that is failing today. Also recorded on the issue.

### Merge basis

The gatekeeper parked at Step 1.5 and never checked out the diff, so no merger-side grade existed.
The diff was read in full for this ruling rather than bounced back for another cycle:

- 4 files, **453 insertions, 0 deletions** — purely additive, touches no plugin.
- Arms are genuinely disjoint: Arm A's `elif not defn` skips definition lines, so killing either
  arm is observable. `EMITTER_NAMES` decides *attribution* only — an emitter it does not name falls
  through to Arm A and is still caught, so the list's breadth is not a coverage surface.
- Both mutation recipes were re-run and returned `guard_verified`, never `case_not_found`.
- All six acceptance criteria of #1362 are satisfied under the corrected bar.

**One thing found that the green checks did not cover.** CI last ran at `07:22Z`; #1405 and #1408
landed on `main` afterwards, leaving this branch 4 commits behind, so the green
`Result-line plainness guard` check did not reflect current `main`. The guard's detection was
re-run against `main` at `a9bd464e` directly: **117 suites discovered, 0 hits**. Post-merge `main`
stays green, and the per-arm liveness floors are inequalities (`>= 80` / `>= 40`), so the grown
tree does not trip them. Merging on the stale check alone would have been unsafe; merging is safe.

### Noted, not folded in

The guard's stderr remediation text tells an author to rewrite the emitter, but does not mention
that a suite legitimately needing a literal escape must assemble it at run time — as both this
PR's suite and `test_plain_result_emitters.sh` already do. A real ergonomic gap, but folding it in
here would re-open a contract graded 20/20 and settled. Left for a future pass.

Removing `svc:needs-human-review`; #1362 is closeable on merge, since the residual gap is #1406's
and not an unmet criterion.

Issue #1402, opened 2026-08-16
repo: execute the maintainer ruling on the 7 bundled-only wiki pages once blessed


claude-cogni-work:
## Context

Deferred from #1384. A maintainer ruling is required before this can start — see the
"Reserved to a human — the blessing" section of that issue and the per-page staleness evidence
table posted there.

## Current behavior

7 wiki pages exist only in `cogni-workspace/wiki/wiki/pages/` and not in the root `wiki/` tree:
`concept-canonical-workflow-ids`, `ecosystem-command-reference`, `ecosystem-plugin-selection`,
`workflow-consulting-engagement`, `workflow-docs-pipeline`, `workflow-full-onboarding`,
`workflow-research-to-report`.

#1384's PR records them as `pending-blessing` and changes none of them, because accepting a page
into the root tree is what makes `cogni-workspace:ask` serve it as a grounded answer — a call
reserved to a human.

The evidence posted on #1384 proposes: promote 6 as-is, and update-then-promote
`ecosystem-command-reference` after correcting 3 stale rows (it claims cogni-workspace ships no
commands, lists 9 of its 11 skills, and mislabels `render-infographic-editorial` as a skill when
it is a command).

## Desired behavior

Execute whatever the maintainer rules — promote into `wiki/wiki/pages/`, or remove from the
bundle — and reconcile every dependent surface.

## Acceptance criteria

- [ ] Each of the 7 pages is either present in both trees or absent from both. Falsified by any stem still present in exactly one tree with no recorded intentional-divergence decision.
- [ ] Each tree's `wiki/index.md` links a stem if and only if that tree's `pages/` holds it.
- [ ] If promotion is chosen, the `[[workflow-research-to-report]]` and `[[workflow-consulting-engagement]]` links removed from line 60 of **both** copies of `workflow-install-to-infographic.md` by #1384's PR are restored, and the two copies stay byte-identical (case `L7` of `cogni-workspace/tests/test-layering-claim-reconciled.sh`).
- [ ] `entries_count` in each `.cogni-wiki/config.json` is re-corrected to its tree's new page count.
- [ ] `cogni-workspace/references/wiki-tree-reconciliation.md`'s pending-blessing section records the ruling and the date.
- [ ] `bash cogni-workspace/tests/test-wiki-tree-parity.sh` exits 0.

## Maintainer ruling — given 2026-08-18

**Ruled by the maintainer via `/cogni-service:service-groom`. The gate is lifted; this issue may start.**

[...]

Issue #1401, opened 2026-08-16
repo: execute the 20 wiki-tree content deltas (group A root-authoritative, group B bundle-authoritative)


claude-cogni-work:
## Context

Deferred from #1384, whose PR records a per-delta decision for all 28 wiki-tree drift items but
executes only the determinate subset (`entries_count`, the two dangling wikilinks, the decision
record, a parity guard). This issue executes the 20 content deltas.

Read `cogni-workspace/references/wiki-tree-reconciliation.md` first — it carries the recorded
decision and the rationale for each group. **Do not** run `scripts/release-bundle-wiki.sh`
without `--check` to "fix" this; see #1384 and the group B finding below.

## Current behavior

`bash scripts/release-bundle-wiki.sh --check` reports 20 content deltas between `wiki/` and
`cogni-workspace/wiki/`. They resolve in **two opposite directions**, which is why a blanket
sync is wrong:

**Group A — 13 files, root authoritative.** Re-ingest commits enriched the root copies; the
bundle was never re-synced. Bundle carries the older, thinner stub.

- `agent-cogni-claims-claim-verifier.md`, `agent-cogni-claims-source-inspector.md`
- `agent-cogni-portfolio-proposition-generator.md`, `agent-cogni-portfolio-proposition-quality-assessor.md`
- `concept-claim-lifecycle.md`, `concept-claims-propagation.md`
- `plugin-cogni-claims.md`, `plugin-cogni-portfolio.md`
- `skill-cogni-portfolio-portfolio-verify.md`, `skill-cogni-portfolio-propositions.md`
- `skill-cogni-trends-trend-report.md`, `skill-cogni-claims-claim-entity.md`, `skill-cogni-claims-claims.md`

**Group B — 4 files, BUNDLE authoritative (the reverse direction).** Commit `c44d52c3` appended
`## Steps` and `## Common pitfalls` sections — and a `## Two scenarios` section on one — directly
to the *bundled* copies, bypassing the one-way root→bundle sync contract documented at
`scripts/release-bundle-wiki.sh:5-10`.

| Page | root lines | bundle lines |
|---|---:|---:|
| `workflow-portfolio-to-website.md` | 55 | 76 |
| `workflow-content-pipeline.md` | 50 | 79 |
| `workflow-portfolio-to-pitch.md` | 56 | 72 |
| `workflow-trends-to-solutions.md` | 52 | 79 |

[...]

PR #1408, opened 2026-08-16
release-bundle-wiki: refuse to destroy bundle-only content by default

### LINKED ISSUE #1403: release-bundle-wiki.sh: refuse to destroy bundle-only content instead of rsync --delete-ing it
## Context

Discovered while resolving #1384. This is a **latent data-loss defect**, not a cleanup: the sync
script will silently destroy content that exists only in the bundled tree.

## Current behavior

`scripts/release-bundle-wiki.sh` in sync mode (no `--check`) runs:

```
rsync -a --delete "$SOURCE_WIKI/" "$BUNDLED_WIKI/"
```

Run against `main` today that would:

1. **Delete 7 pages** that exist only in `cogni-workspace/wiki/wiki/pages/`
   (`concept-canonical-workflow-ids`, `ecosystem-command-reference`, `ecosystem-plugin-selection`,
   `workflow-consulting-engagement`, `workflow-docs-pipeline`, `workflow-full-onboarding`,
   `workflow-research-to-report`), and
2. **Overwrite 4 bundled pages with shorter root copies**, discarding the `## Steps`,
   `## Common pitfalls` and `## Two scenarios` sections that commit `c44d52c3` appended directly
   to the bundled copies of `workflow-portfolio-to-website.md`, `workflow-content-pipeline.md`,
   `workflow-portfolio-to-pitch.md` and `workflow-trends-to-solutions.md`.

The header at `scripts/release-bundle-wiki.sh:5-10` declares root the source of truth, so the
`--delete` is intentional in principle. What is missing is any check that the bundle holds
nothing the sync is about to destroy. The post-sync guard at lines 81-87 compares page **counts**
after the fact — by then the content is already gone.

The script is documented as "Run BEFORE every cogni-workspace marketplace publish", so this is
reachable on a routine release path, not a hypothetical.

## Desired behavior

The sync refuses, by default, to run when the bundled tree holds files absent from source, or
bundled content that is not derivable from source. An explicit opt-in flag (e.g. `--force`)
performs the destructive sync, and the refusal names every file it would have destroyed.

## Acceptance criteria

- [ ] A bare `scripts/release-bundle-wiki.sh` exits non-zero and writes no bytes to the bundled tree when the bundle holds a page absent from source. Falsified by a run that deletes such a page without an explicit opt-in.
- [ ] The refusal output is a single valid JSON object per the repo `{success,data,error}` convention, naming each would-be-destroyed path in `data`.
- [ ] An explicit opt-in flag performs the destructive sync and is documented in the script's usage block.
- [ ] `--check` behavior is unchanged and still reports `changes_pending` and `source_page_count`.
- [ ] A test suite discovered by `scripts/run-plugin-tests.py` covers both the refusal and the opt-in, with a recorded `scripts/mutation-check.sh` recipe proving the refusal assertion can go red.

claude-cogni-work:
Closes #1403.

## Summary

`scripts/release-bundle-wiki.sh` now classifies the bundled tree **before writing any bytes** and refuses by default when a sync would destroy content. The refusal is one `{success,data,error}` object naming every path it would have destroyed, and exits 1. `--force` is the documented opt-in.

Two refused classes:

| Class | Meaning |
|---|---|
| `would_delete` | a bundled path with no source counterpart — what `rsync --delete` removes silently |
| `would_overwrite` | a path on both sides whose bundled copy carries non-blank lines the source lacks |

The second class is the one AC 1's literal wording misses and the post-sync page-count check structurally cannot see, since an overwrite leaves the file count unchanged.

## The count check was unreachable, not merely late

The issue describes the existing guard at `:81-87` as running too late. It is worse than that: `rsync --delete` has already made both trees identical by the time it runs, so it compares **170 to 170** and can never fail. There is no input on which the current script refuses.

It is left in place — the new pre-sync gate, not any change to that check, supplies the defence. Removing it is a judgement the issue does not decide.

## Placement is the load-bearing detail

The gate sits between the rsync-availability guard and `mkdir -p "$BUNDLED_WIKI"`. `mkdir -p` is the first write-capable call in the script, so a refusal is a true no-op on disk rather than an early exit that already created a directory. Case RBW1 asserts this against a sha manifest of the whole bundled tree, not merely file presence.

## Why a subset is deliberately not refused

A bundled file that is merely a **subset** of its source counterpart loses nothing and is allowed through bare. That asymmetry is the difference between a guard and a nuisance: a gate firing on any difference would refuse every ordinary sync, and an override an operator passes every time is an override they stop reading — reinstating exactly the data loss this closes. Case RBW4 pins the additive path and RBW5 the already-identical path.

## Reviewer note — the gate refuses on 27 paths today, not 11

Worth seeing before you read the code, because it is more than the issue's 7 + 4:

```
would_delete:    7   the Decision 4 bundle-only pages
would_overwrite: 20  the 4 group-B pages, wiki/index.md, wiki/log.md,
                     .cogni-wiki/config.json, and the 13 group-A pages
```

The group-A pages appear because the root re-ingest **reworded** lines rather than only appending, so the bundled copy holds text the source does not. Decision 3 rules that root wins for group A, so that loss is sanctioned — but a shell script cannot read a decisions record. It refuses and defers to the operator, which is the safe direction. Executing #1401 and #1402 returns a bare sync to silence. If a reviewer prefers the gate encode Decision 3 directly, that is a scope call worth making explicitly rather than by tightening the predicate quietly.

## Verification

- [x] `bash tests/test_release_bundle_wiki.sh` — 30/30 assertions pass, RBW1–RBW8
- [x] `python3 scripts/run-plugin-tests.py` — **117/117 suites pass**, 0 failed
- [x] New suite auto-discovered by the `tests/*.sh` glob, zero registration
- [x] `--check` on the real tree unchanged: `changes_pending: 28`, `source_page_count: 170`
- [x] Bare run on the real tree exits 1 and leaves `git status --porcelain cogni-workspace/wiki` empty
- [x] `check-version-bump.py` — 12 plugins scanned, **0 touched**, 0 violations
- [x] `check-workflow-yaml.py` clean
- [x] Mutation check **executed**, not merely recorded — verdict `guard_verified`

## Mutation recipe

```bash
bash cogni-service/scripts/mutation-check.sh \
  --root "$REPO_ROOT" \
  --file scripts/release-bundle-wiki.sh \
  --expr 's/DESTRUCTIVE_COUNT" -gt 0/DESTRUCTIVE_COUNT" -lt 0/' \
  --test 'bash tests/test_release_bundle_wiki.sh' \
  --case RBW1
```

Executed on this branch: `mutated_result: red`, `restored_result: green`, `verdict: guard_verified`. `DESTRUCTIVE_COUNT` is a sum of two non-negative counts, so `-lt 0` is unconditionally false and the gate can never fire.

Note the harness applies `--expr` through `perl -0pi` in slurp mode with no `/g`, so it rewrites only the **first** occurrence in the file. The literal `DESTRUCTIVE_COUNT" -gt 0` therefore appears exactly once, on the `if` line — a comment restating the comparison would absorb the mutation, leave the guard intact, and report a false green. Verified: `grep -c` returns 1.

## Two defects caught in planning, recorded so they are not reintroduced

1. **`printf '%s\n' "$EMPTY" | wc -l` returns 1, not 0.** Deriving the counts that way makes `DESTRUCTIVE_COUNT` = 2 on a perfectly in-sync bundle, refusing **every clean sync** and rendering the script unusable on its documented pre-publish path. Both counts use `grep -c . || true`. Case RBW5 is what pins this.
2. The `--help` range is a hand-maintained line number and this change grows the header. The bound is now `sed -n '2,24p'`, measured rather than assumed, and RBW7 pins it at **both** ends — too narrow truncates the usage text, too wide leaks `set -eu` into `--help`.

## Scope

No file under `wiki/wiki/` or `cogni-workspace/wiki/wiki/` is touched — the content decisions belong to #1401 and #1402. This PR only stops the script from pre-empting them. No version bump; the post-merge workflow owns that.

`cogni-workspace/references/wiki-tree-reconciliation.md` is updated at three anchors that would otherwise ship describing a tool that no longer behaves that way: the rejected-alternative paragraph, the contradictions-table row, and "What guards this".

## Not done

Step 6.4's `/simplify` quality pass did **not** run — the commit was taken first to make the work durable under a tight context budget, which leaves that pass's dirty-tree guard reading clean. Quality-only; correctness is covered by the checks above. Flagged for the reviewer rather than left silent.

## Do not run `--force` against the real trees while reviewing

That is the destruction this PR exists to prevent. The `--force` path is covered inside the suite's fixtures.

### DIFF
--- a/cogni-workspace/references/wiki-tree-reconciliation.md
+++ b/cogni-workspace/references/wiki-tree-reconciliation.md
@@ -205,8 +205,26 @@ Rejected on three grounds:
    the counts still match after the content is gone.
 
 The script's own header states the root tree is the editing surface, so the deletion is
-intentional in principle. What is missing is any refusal when the bundle holds something the
-sync is about to destroy. Issue #1403 carries that.
+intentional in principle. What was missing was any refusal when the bundle holds something the
+sync is about to destroy.
+
+The script now carries that refusal. It classifies the bundled tree *before* writing any bytes
+and exits non-zero, naming every path it would have destroyed, in two classes: a bundled file
+with no source counterpart (ground 1), and a bundled file carrying non-blank lines the source
+copy lacks (ground 2 — the class the count-based check of ground 3 structurally cannot see,
+since an overwrite leaves the file count unchanged). Taking this rejected alternative therefore
+now requires passing `--force` explicitly, which makes converting a held decision into an
+executed one a deliberate act rather than an accident.
+
+Ground 3 still describes the post-sync check accurately. That check is left in place, and it is
+the new pre-sync gate rather than any change to it that supplies the defence.
+
+On the tree as recorded here a bare sync refuses on 27 paths: the 7 of Decision 4, the 4
+group-B pages, and — because the root re-ingest reworded lines rather than only appending — the
+13 group-A pages together with `wiki/index.md`, `wiki/log.md` and `.cogni-wiki/config.json`. The
+gate is deliberately blind to Decision 3's ruling that root wins for group A: a script cannot
+read this document, so it refuses and defers to the operator. Executing #1401 and #1402 is what
+returns a bare sync to silence.
 
 ## Rejected alternative — regenerate `entries_count` with the wiki linter
 
@@ -228,7 +246,7 @@ Kept as a permanent inventory so the set stays auditable rather than being redis
 | `wiki/index.md` needs a merge, not a copy | recorded, not executed | #1401 |
 | 7 bundled-only pages held | awaiting a maintainer ruling | #1402 |
 | `ecosystem-command-reference` names a retired command surface | held with the page above | #1402 |
-| Sync script destroys bundle-only content | open defect | #1403 |
+| Sync script destroys bundle-only content | closed — refuses by default; `--force` is the opt-in | #1403 |
 | This record and its guard are unregistered in the plugin guide | open | #1404 |
 
 ## Deliberately left standing
@@ -246,3 +264,10 @@ that every wiki reference in a tree resolves to a page in that same tree; and th
 present in only one tree is named in this document. The last of those is what makes adding or
 promoting a page without recording a decision here fail rather than pass silently. It does not
 assert the two trees are equal — they are not, by Decisions 1, 4 and 5.
+
+`tests/test_release_bundle_wiki.sh` guards the other direction: that
+`scripts/release-bundle-wiki.sh` refuses, by default, to destroy what this document holds. It
+covers both refused classes, the `--force` opt-in, and — as the case that matters most for the
+gate staying usable — that a bundle which loses nothing still syncs without `--force`. It runs
+entirely against `mktemp` fixtures, never these two trees, because a bare sync against them is
+the data loss the gate exists to prevent.
--- a/scripts/release-bundle-wiki.sh
+++ b/scripts/release-bundle-wiki.sh
@@ -13,6 +13,12 @@
 # Usage:
 #   scripts/release-bundle-wiki.sh                    # full sync (rsync --delete)
 #   scripts/release-bundle-wiki.sh --check            # dry-run; report diff but don't write
+#   scripts/release-bundle-wiki.sh --force            # sync even when it would destroy content
+#
+# A bare sync REFUSES, exits 1 and writes nothing when the bundled tree holds
+# content the source does not: a file absent from source, or a shared file whose
+# bundled copy carries non-blank lines the source copy lacks. The refusal names
+# every such path in its JSON payload. Pass --force to sync anyway.
 #
 # Output: JSON to stdout per insight-wave script convention.
 #   {"success": true, "data": {...}, "error": ""}
@@ -22,12 +28,18 @@ set -eu
 SOURCE_WIKI="$(cd "$(dirname "$0")/.." && pwd)/wiki"
 BUNDLED_WIKI="$(cd "$(dirname "$0")/.." && pwd)/cogni-workspace/wiki"
 MODE="sync"
+FORCE=0
 
 while [ $# -gt 0 ]; do
   case "$1" in
     --check) MODE="check"; shift ;;
+    --force) FORCE=1; shift ;;
     -h|--help)
-      sed -n '2,18p' "$0"
+      # The upper bound must equal the last line of the header comment block.
+      # tests/test_release_bundle_wiki.sh case RBW7 pins both directions: it
+      # fails if the range truncates the usage text, and it fails if the range
+      # runs past the header into the code below.
+      sed -n '2,24p' "$0"
       exit 0
       ;;
     *)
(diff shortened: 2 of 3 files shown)


claude-cogni-work:

Author instance `frosty-bouman-18490b` (`--role author --act`, cogni-service v0.0.402). Draft, `svc:needs-review`, no ready flip and no verdict surface — the independent review is the merger's.

Head at handoff: `afe4e45e`.

### Handoff preflight — clean, but read the skip roster

`check-preflight.sh` exit 0, `clean: true`. **1 passed, 6 skipped, 0 failed.** The skips matter more than the pass here:

| Instrument | Result | Reason |
|---|---|---|
| `run-tests` | skip | "no touched scripts/, tests/, SKILL.md or agents/*.md path" |
| `mutation-recipe` | skip | no path matching `*/tests/test-*.sh` or `*/scripts/check-*.sh` |
| `frontmatter` / `breadcrumbs` / `skill-length` / `agent-length` | skip | not applicable — no SKILL.md or agents/*.md in the diff |

**Two of those skips are wrong, and they are an instrument gap rather than a property of this PR.** The diff touches `scripts/release-bundle-wiki.sh` *and* `tests/test_release_bundle_wiki.sh`, so `run-tests` should have been applicable; its matcher only sees plugin-scoped `*/scripts/` and `*/tests/`, and both of this PR's paths are repo-root. `mutation-recipe` misses on two counts at once — repo-root again, and the glob is `test-*.sh` while the suite is underscore-named. So this PR ships a mutation recipe and the instrument that verifies recipes never looked at it.

The green banner therefore attests to very little. Real coverage came from running things directly, not from the preflight:

- `python3 scripts/run-plugin-tests.py` → **117/117 suites pass**, 0 failed (re-run after the fix pass below)
- `bash tests/test_release_bundle_wiki.sh` → **36/36 assertions**, RBW1–RBW9
- mutation recipe **executed**: `mutated_result: red`, `restored_result: green`, `verdict: guard_verified`

Worth raising against `cogni-work/managed-service` separately — any future repo-root change gets the same vacuous green.

### Contract self-grade

Graded against the triage-time contract on #1403 (`origin: triage`, derived `07:59:05Z`, before any branch was cut). **24 of 25 items satisfied, 1 violated, 0 unclear.**

The violated item was **24** — no suite case exercised an absent or empty bundled tree. The code path was already correct; nothing asserted it. Mechanical, auto-fixable, and inside a file the plan already touched, so it went through the single bounded fix pass: `afe4e45e` adds RBW9 covering both shapes (bundled tree absent entirely — a first-ever publish — and present but holding no pages). Suite re-run 36/36 and the full sweep re-run 117/117 before the push.

**Residual carried to you: none.** No unclear items, no unfixed violations.

### Two things I want a human to look at

**1. The gate refuses on 27 paths today, not the 7 + 4 the issue describes.** `would_delete` is the expected 7; `would_overwrite` is 20, because the 13 group-A pages also carry bundled-only lines — the root re-ingest reworded rather than only appended. Decision 3 rules root wins for group A, so that loss is sanctioned, but a shell script cannot read a decisions record: it refuses and defers. Executing #1401/#1402 returns a bare sync to silence.

The grader judged this does not fail checklist item 16 — the contract's own risk list reserves "whether the refusal predicate is *usefully* narrow" as a human read, and deliberately avoids pinning repo-wide counts since a base advance would falsify them. RBW4 and RBW5 prove it is not degenerate always-refuse. But the narrowness call is yours, and if you want the predicate to encode Decision 3 directly that is a scope decision worth making explicitly rather than by quietly tightening the predicate.

**2. Step 6.4's `/simplify` quality pass did not run.** I committed first to make the work durable under a tight context budget, which leaves that pass's dirty-tree guard reading clean. Quality-only — correctness is covered above — but it is a gate that did not fire, so I would rather you know than discover it.

### One trap while reviewing

Do **not** run `bash scripts/release-bundle-wiki.sh --force` against the real trees. That is precisely the destruction this PR prevents; it would delete the 7 held pages and truncate the 4 group-B pages. The `--force` path is exercised inside the suite's `mktemp` fixtures.

Also note the literal `DESTRUCTIVE_COUNT" -gt 0` must stay unique in the script — the harness applies `--expr` via `perl -0pi` in slurp mode with no `/g`, so a comment restating the comparison would absorb the mutation and report a false green.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1408

**Summary:** All three perspectives (service-review's oversized-PR check, plugin-validator, and contract-grading-reviewer) passed; the merger-added oversized-PR preference and two waived plugin-validator preferences are excluded from gating, leaving only one un-waived minor defect (an unreachable awk substring-match edge case), so the change is safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[minor]** (advisory-addition) `PR diff` — Diff is 417 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

- **[minor]** (auto-fixable) `scripts/release-bundle-wiki.sh:118-119` — The itemise-path extraction substr($0, index($0, $2)) finds the FIRST occurrence of the path string anywhere in the line, including inside the flag field, so a path whose name is a substring of *deleting or >fcst....... is reported mangled. Reproduced and independently re-verified by the gatekeeper: a bundle-only file named 'g' produces itemise line '*deleting g', index resolves to the 'g' inside 'deleting', and the payload reports would_delete: ['g g'] instead of ['g']. The refusal DECISION is unaffected (count still >0, exit 1, zero bytes written), so this is fail-safe in direction - only the operator-facing path name in data is wrong, which is what checklist item 11 asks the payload to get right. Today's trees hold only kebab-case .md and .json files, so it is unreachable in practice.
  - *Suggested fix:* Replace the index() lookup with a positional slice that cannot search inside the flag field: awk '$1 == "*deleting" { print substr($0, length($1) + 2) }' and awk '$1 ~ /^>f/ { print substr($0, length($1) + 2) }'. Verified by the gatekeeper: this yields 'g' for the pathological case and still round-trips 'wiki/pages/a page with spaces.md' intact.

- **[minor]** (waived) (advisory-addition) `scripts/release-bundle-wiki.sh:109-182` — --force is all-or-nothing, and on the tree as it stands the bare command is already unusable: the gate refuses on 27 paths (7 would_delete, 20 would_overwrite). 13 of the 20 are group-A pages where Decision 3 already ruled root wins, so the next legitimate pre-publish sync requires --force, which simultaneously executes Decision 4's held deletion of the 7 bundle-only pages #1402 is holding. This is the over-refusal risk the contract's risk_watchlist reserves as a human read; the non-vacuity checklist item (16) is satisfied fixture-relative by RBW4/RBW5, so it is not a violation, and the author documented the exit route explicitly.
  - *Justification:* Contract risk_watchlist item 3 explicitly reserves the narrowness judgement for a human, and the required green-path checklist item is satisfied. The tradeoff is documented in-repo with a stated exit route.

- **[minor]** (waived) (advisory-addition) `scripts/release-bundle-wiki.sh:130` — The overwrite predicate is whole-line set membership (grep -Fxv -f), not a diff, so bundled content composed only of lines each appearing somewhere in the source is not flagged even when it forms a section existing nowhere in source. Separately a binary file differing on both sides is flagged as would_overwrite via grep's 'Binary file ... matches' line. Neither costs anything today: the four real group-B pages are all caught, the predicate errs toward refusing, and the trees hold only .md and .json. The line-set semantics are already stated accurately in the header comment.
  - *Justification:* Fail-safe in direction (over-refuses rather than under-refuses) and unreachable on the current content; the checklist requires only that the group-B overwrite class be caught, which it is.

### Positive observations

- Gate placement provably correct: the block at :109-182 precedes the first write-capable call mkdir -p (:184), and RBW1 asserts the no-write property against a sha-per-path manifest rather than mere file presence. Confirmed against a copy of the real trees - byte-for-byte identical after a refusal. (plugin-validator)
- The mutation recipe is genuinely load-bearing, not decorative: replayed verbatim it returns mutated_result: red, restored_result: green, verdict: guard_verified. The --expr replacement is disjoint from the matched literal, that literal occurs exactly once, and --case RBW1 reproduces a label the suite actually prints. (both perspectives, independently)
- JSON escaping is correct on the branch most likely to be parsed - both lists go through json.dumps on the whole envelope, verified against fixtures holding embedded quotes, backslashes and spaces. (plugin-validator)
- RBW5 is a real regression guard: wc -l on a printf-terminated empty value reports 1, which would have made the gate refuse every clean sync and turn --force into a permanent requirement. (plugin-validator)
- Every set -eu hazard in the new detection path is individually guarded; the suite runs green under bash 3.2.57 as well as bash 5. (plugin-validator)
- Test hygiene is exemplary: the suite copies the real script into each fixture so a mutation reaches the code under test and it cannot pass against a frozen replica; cleanup trap; never touches the real trees. (plugin-validator)
- --check is untouched (zero lines changed in :65-77) and --force deliberately does not set MODE, so --check --force and --force --check both take the check branch and exit 0. (both perspectives)
- The --help line-range hazard was spotted and pinned at BOTH ends (2,18p widened to 2,24p; RBW7 fails on truncation and on overrun into set -eu). (plugin-validator)
- Documentation surface complete: the script is referenced from exactly one prose surface, updated in the same PR, so no stale instruction tells an operator to run a bare sync. (plugin-validator)
- Repo-hygiene guards clean on this head: version-bump ok (0 violations), breadcrumb guard clean, skill-names OK, plugin.json mirrors marketplace.json, all 8 cogni-workspace suites pass. (plugin-validator)
- Zero hardcoded page names in script or suite - the guard cannot rot into vacuity as the trees change. (both perspectives)
- 25/25 committed contract checklist items graded SATISFIED, 0 violated, 0 unclear - no INTENT-UNCLEAR divergence. (contract-grading-reviewer)
- item 1: run-plugin-tests.py --list includes tests/test_release_bundle_wiki.sh (total 117 suites). (contract-grading-reviewer)
- item 14: RBW3 live-passed - a bundled superset is refused via would_overwrite, the class the post-sync count guard structurally cannot see. (contract-grading-reviewer)
- item 16: RBW4 (subset-only) and RBW5 (identical) both live-passed bare, exit 0 - not the degenerate always-refuse case. (contract-grading-reviewer)
- item 19: every wiki-tree reference in the suite resolves inside the mktemp fixture; no sync-mode invocation targets the real trees. (contract-grading-reviewer)

### Recommendation

---

PR #1410, opened 2026-08-16
repo: floor the fixtures-depth coverage the knowledge suite owns

### LINKED ISSUE #1406: repo: result-line plainness guard cannot reach shared emitter helpers under tests/fixtures/

claude-cogni-work:
Closes #1406.

## Summary

Records the settled disposition for the fixtures-depth coverage gap and floors it so it cannot be dropped silently.

- **Disposition (b).** `cogni-knowledge/tests/test_plain_result_emitters.sh` is designated owner of the `tests/fixtures/` depth (already true in behaviour). `scripts/check-result-line-plainness.py` is **not** touched — no third glob, per the ruling recorded on the originating issue.
- **New `D1` coverage floor** — the suite now fails loudly if no emitter-defining file is scanned under its deeper discovery arm.
- **The existing scan-root note is amended** from the singular "liveness floor" to cover both floors, since a wider guard reusing this file must now re-home two plugin-scoped counts.

One file, +43/-2.

## The gap, reproduced

This was verified by running it, not by reading. Breaking the deeper arm of the discovery loop on the pre-change file gives:

```
PASS: helper emits plain result lines on stdout
PASS: no file under the scanned tree carries an escape literal
PASS: 0 emitter-defining file(s) are paired and plain, of 77 exercising the contract

plain result-emitter contract all pass.
exit=0
```

The shared helper is the repo's only emitter definer, so it leaves the scanned population — while all 77 sourcing files still count as exercisers, keeping the liveness floor (50) green, and `shape_errors` stays 0 because nothing is left to shape-check. Coverage drops; the suite passes.

## Scope decisions

- **The floor is keyed per-arm, not on the whole-population definer count.** A whole-population count would also be satisfied by any future suite that inlines its own emitters, which would mask the same drop. Only the per-arm count falsifies both shapes.
- **The floor is deliberately not conditioned on `SCAN_ROOT`.** The scan root is overridable by `$1`, but CI runs the suite with no argument, and the pre-existing liveness floor is unconditioned for the same reason. Gating it would create a path where it never evaluates — the exact vacuity class this change closes. The header instead states that a wider guard must re-home both floors.
- **No copy of the mutation recipe is embedded in the suite.** Beyond staleness, there is a concrete failure mode: the recipe substitutes with `perl -0pi` in slurp mode with no `/g`, so it rewrites the **first** occurrence only. The header sits above the discovery loop, so a header copy of that literal would swallow the mutation, leave the loop intact, and `D1` would stay green — the check would prove nothing. The header refers to the arm in prose; the literal occurs exactly once in the committed file (verified).
- **The repo-root `CLAUDE.md` bullet is not touched.** It already records the load-bearing half — that suites one path segment deeper are out of scope by glob depth, not by exclusion — and the issue itself accepts repo-root undiscoverability as the cost of disposition (b).

## Mutation recipe

The harness ships with the `cogni-service` plugin, not this repo — `${CLAUDE_PLUGIN_ROOT}` resolves to the installed plugin root, so do not look for it under this repo's `scripts/`. Run with `--root` pointing at the PR-head checkout.

```
  --root . \
  --file cogni-knowledge/tests/test_plain_result_emitters.sh \
  --expr 's{/fixtures/\*\.sh; do}{/fixtures-absent/*.sh; do}' \
  --test 'bash cogni-knowledge/tests/test_plain_result_emitters.sh' \
  --case D1
```

Falsifies the `D1` floor: with the deeper arm redirected at a directory that does not exist, the shared helper leaves the scanned population, `fixtures_definers` drops to 0, and `D1` goes red — while the pre-existing `exercisers` floor stays green throughout, which is precisely why `D1` had to be added rather than reused.

The expression is anchored on the `; do` suffix so it can only match the discovery loop, never prose. Pattern and replacement are disjoint, so the substitution cannot re-match itself. `D1` is the first whitespace-delimited token of both the `PASS:` and `FAIL:` labels, with no colon abutting it, so the harness resolves the case in both directions.

**Result when run against this branch:**

```json
"mutated_result": "red",
"restored_result": "green",
"verdict": "guard_verified"
```

## Acceptance criteria

| Criterion | Status |
|---|---|
| A decision is recorded between (a) and (b), with the reasoning | Header paragraph records disposition (b) and why the two-glob boundary is load-bearing |
| (b): a guard assertion fails if the suite stops scanning `fixtures/` | The `D1` per-arm floor; `errors` drives the exit, so a red `D1` exits 1 |
| Provable by a mutation check returning a real verdict rather than `case_not_found` | `guard_verified` above |
| `python3 scripts/run-plugin-tests.py` stays green | All 118 suites passed on this branch; all 118 also passed on the `main` baseline |

## Testing

- `python3 scripts/run-plugin-tests.py` — **118/118 passed** on the branch (and on the `main` baseline, so the branch introduces no delta).
- `python3 scripts/check-result-line-plainness.py` — clean: 118 discovered, 118 scanned, 85 definitions inspected, 0 violations. Both per-arm liveness floors (`>= 80` / `>= 40`) hold with margin.
- `bash cogni-knowledge/tests/test_plain_result_emitters.sh` — gains `PASS: D1 1 emitter-defining file(s) scanned under the fixtures segment, so shared-helper coverage holds`; the three pre-existing cases are unchanged.
- No `plugin.json` / `marketplace.json` version line is touched.

### DIFF
--- a/cogni-knowledge/tests/test_plain_result_emitters.sh
+++ b/cogni-knowledge/tests/test_plain_result_emitters.sh
@@ -29,8 +29,23 @@
 #
 # The scan root defaults to this file's directory and can be overridden by the
 # first argument, so a wider guard can reuse this file rather than fork it.
-# A wider guard must re-home the liveness floor below — that count is scoped to
-# this plugin.
+# A wider guard must re-home the floors below — the liveness count and the
+# fixtures-arm coverage floor are both scoped to this plugin's tree.
+#
+# This suite owns the deeper fixtures segment, and that ownership is settled
+# rather than incidental. The repo-root result-line plainness guard discovers
+# suites through two non-recursive globs mirroring run-plugin-tests.py, so a
+# shared emitter helper parked one path segment below a tests/ directory is
+# structurally out of its reach. A third, deeper glob was considered and
+# rejected: that two-glob boundary is exactly what keeps the parked _archive/
+# carrier out of the population, and the guard ships no exclusion list,
+# allowlist or baseline to compensate — widening it would have forced the very
+# exemption machinery the guard exists to avoid, to close a gap nothing is
+# failing on today. The accepted cost is that this coverage stays plugin-local
+# and is not discoverable from the repo root. The D1 floor below is what stops
+# it from being dropped silently: break the deeper arm of the discovery loop
+# and the only definer leaves the scanned population, while every sourcing file
+# still counts toward the liveness floor above.
 #
 # bash 3.2 + python3 stdlib only (no pytest, no pip). Matches tests/README.md.
 
@@ -90,6 +105,7 @@ fi
 # carries its own body and feeds every check below from a single read.
 
 definers=0
+fixtures_definers=0
 exercisers=0
 shape_errors=0
 
@@ -109,6 +125,12 @@ for f in "$SCAN_ROOT"/*.sh "$SCAN_ROOT"/fixtures/*.sh; do
 
   definers=$((definers + 1))
   exercisers=$((exercisers + 1))
+
+  # Tally definers found under the deeper segment separately — the D1 floor
+  # below keys on this, not on the whole-population count.
+  case "$f" in
+    */fixtures/*) fixtures_definers=$((fixtures_definers + 1)) ;;
+  esac
 
   n_red=0
   n_green=0
@@ -144,6 +166,25 @@ if [ "$exercisers" -lt 50 ]; then
   errors=$((errors + 1))
 elif [ "$shape_errors" -eq 0 ]; then
   green "PASS: $definers emitter-defining file(s) are paired and plain, of $exercisers exercising the contract"
+fi
+
+# Coverage floor — the shared emitter helper must stay inside the scanned
+# population. Keyed per-arm on definers found under the deeper segment, NOT on
+# the exercisers floor above and NOT on the whole-population definer count:
+# break the deeper arm of the discovery loop and the only definer leaves the
+# population while every sourcing file still counts as an exerciser, so that
+# floor stays green, shape_errors stays 0 because there is nothing left to
+# shape-check, and the dropped coverage reports as a pass. A whole-population
+# definer count would also be satisfied by any future suite that inlines its
+# own emitters, which would mask the same drop. Only the per-arm count sees it.
+# The label carries a stable first-token case id, space-separated and never
+# colon-abutted, so a mutation harness classifies this case instead of
+# reporting case_not_found.
+if [ "$fixtures_definers" -lt 1 ]; then
+  red "FAIL: D1 no emitter-defining file under the fixtures segment was scanned — the shared helper left the scanned population"
+  errors=$((errors + 1))
+else
+  green "PASS: D1 $fixtures_definers emitter-defining file(s) scanned under the fixtures segment, so shared-helper coverage holds"
 fi
 
 echo


claude-cogni-work:

Author instance `hopeful-napier-e05927` has finished this branch under the two-instance posture. Draft, carrying `svc:needs-review`. No review verdict was posted and no `svc:review-*` label applied — the independent review belongs to the merger.

### Handoff preflight

`check-preflight.sh` exited **0 — clean**, with `1 passed, 6 skipped, 0 advisory`. Every skipped instrument, with its reason:

| Instrument | Reason | Genuine? |
|---|---|---|
| `run-tests` | `no_suites_at_base` | **No — instrument blindness** |
| `mutation-recipe` | `not_applicable` | **No — instrument blindness** |
| `frontmatter` | `not_applicable` — no touched SKILL.md or agents/*.md | Yes |
| `breadcrumbs` | `not_applicable` — no touched SKILL.md or agents/*.md | Yes |
| `skill-length` | `not_applicable` — no touched SKILL.md | Yes |
| `agent-length` | `not_applicable` — no touched agents/*.md | Yes |

**Two of those skips are not what they appear, and the merger should not read this preflight as coverage.** The preflight discovers suites and guards through `*/tests/test-*.sh` and `*/scripts/check-*.sh` — hyphen-separated only. `cogni-knowledge/tests/` contains **77 underscore-named suites and 0 hyphen-named ones**, so:

- `run-tests` reported `no_suites_at_base` for a plugin that has 77 suites. It did not run them.
- `mutation-recipe` reported `not_applicable` on a PR that touches a test suite **and** ships a mutation recipe. It did not check the recipe.

So a green preflight here means "the instruments that could see this diff found nothing", not "this diff was verified". The verification below was therefore run directly rather than relied upon from the gate.

### Verification run directly by the author

- **`python3 scripts/run-plugin-tests.py` — 118/118 passed on the branch.** The same sweep was run on the `main` baseline first, also 118/118, so the branch introduces no delta. This is what `run-tests` would have covered had it seen the suites.
- **Mutation check — `verdict: guard_verified`** (`mutated_result: red`, `restored_result: green`) against `--case D1` using the recipe in the PR body. This is what `mutation-recipe` would have covered. The recipe was additionally exercised end-to-end against a pre-existing case (`A1`) before the plan was locked, to confirm the harness and flag shape rather than assume them.
- **`python3 scripts/check-result-line-plainness.py` — clean**: 118 discovered, 118 scanned, 85 definitions inspected, 0 violations; both per-arm liveness floors hold with margin.
- No `plugin.json` / `marketplace.json` version line touched.

### Contract grading

`read-review-contract --target issue --origin triage` returned `found: false` — no triage-time contract exists on the linked issue, so no self-grade was performed and none is claimed. Nothing residual is being carried.

### One thing worth a reviewer's attention

Acceptance criterion 1 of the issue asks for a decision between dispositions (a) and (b). That decision was **not** made autonomously here — it was already ruled by a maintainer and recorded in the originating issue's body under `## Ruling`, point 2 ("no third glob; the gap is tracked"). This PR records that ruling and its reasoning in-tree and implements (b). If the reviewer disagrees with the disposition, the disagreement is with that ruling, not with this branch.

### Not filed, but worth someone's judgement

The preflight's hyphen-only suite glob is a real gap affecting any plugin using underscore-named suites — it silently reports `no_suites_at_base` and `not_applicable` instead of failing loudly. That belongs to the `cogni-service` plugin, which lives in a different repository, so no follow-up issue was filed against this repo. Flagging rather than filing cross-repo.

sdh07:

_The merger formed this expectation from issue #1406 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix picks exactly one of the issue's two dispositions, records the reasoning in a diff-visible in-repo surface, and makes the chosen coverage *falsifiable*. The decisive base facts: `scripts/check-result-line-plainness.py::discover()` (lines 95-106) uses two literal non-recursive globs whose rationale is spelled out at docstring lines 38-50 (mirroring `scripts/run-plugin-tests.py::discover()`, lines 82-97), so `cogni-knowledge/tests/fixtures/test_helpers.sh` (plain `red()`/`green()` at lines 15-16) is structurally unreachable; and `cogni-knowledge/tests/test_plain_result_emitters.sh` globs `"$SCAN_ROOT"/fixtures/*.sh` at line 96 but its only liveness floor keys on `exercisers >= 50` (lines 142-147), which stays green from the ~77 sourcing `*.sh` suites even if that fixtures term is deleted (`definers` merely falls to 0) — so today the fixtures coverage genuinely CAN be silently dropped. Disposition (a) therefore means adding `tests/fixtures/*.sh` + `*/tests/fixtures/*.sh` as literal fixed-segment patterns in `discover()`, a new discovery/exclusion case in `tests/test_check_result_line_plainness.sh` mirroring S2 (lines 156-167), and re-syncing the now-false docstring claim plus the CLAUDE.md Contributing bullet. Disposition (b) means adding a floor/assertion in `test_plain_result_emitters.sh` that actually goes red when the fixtures term is removed — not one resting on the existing `exercisers` count. Either way AC4 forces the named case to print the SAME whole first token on the red and green paths (base case 3 prints `FAIL: test_helpers.sh …` vs `PASS: <definers> emitter-defining file(s) …`, which would resolve to `case_not_found`), so a stable case-id relabel is expected.

**Acceptance checklist:**
- [ ] A diff-visible in-repo prose surface (the module docstring of scripts/check-result-line-plainness.py, the header of cogni-knowledge/tests/test_plain_result_emitters.sh, the repo-root CLAUDE.md Contributing bullet, or a references/ doc) records WHICH of disposition (a) or (b) was chosen AND why the other was rejected — confirm by finding a prose block in the diff that names both dispositions, not just the one implemented.
- [ ] The implemented change matches the recorded decision: if the record says (b), `discover()` in scripts/check-result-line-plainness.py is unchanged (no new glob pattern in its `patterns` list); if it says (a), `discover()` gains the two fixtures patterns. A record naming one disposition while the code implements the other is a fail.
- [ ] Scope boundary: every changed path is either the plainness guard itself (scripts/check-result-line-plainness.py), a test suite or fixture that exercises it (tests/test_check_result_line_plainness.sh, cogni-knowledge/tests/test_plain_result_emitters.sh, cogni-knowledge/tests/fixtures/*.sh), or repo prose (CLAUDE.md, a references/ doc). Explicitly OUT: scripts/run-plugin-tests.py (its narrow globs are load-bearing — widening them would EXECUTE sourced-only helper libraries), any .claude-plugin/*.json, and any other plugin's tests/.
- [ ] No version change: `git diff` touches no `version` field in any `.claude-plugin/plugin.json` nor its mirror in the repo-root .claude-plugin/marketplace.json (the repo's Version-bump gate flags `version-touched` against the fork point).
- [ ] No new breadcrumb: the diff introduces no issue/PR ref (`#NNNN`), version tag, or milestone/slice code in any touched file — rationale is stated semantically.
- [ ] IF (a): the two added discovery patterns are literal, fixed-segment `os.path.join(root, "tests", "fixtures", "*.sh")` / `os.path.join(root, "*", "tests", "fixtures", "*.sh")` forms — the diff introduces no `**`, no `glob(..., recursive=True)`, and no `os.walk`, so `<plugin>/_archive/tests/fixtures/*.sh` stays out of reach by segment count rather than by an exclusion.
- [ ] IF (a): tests/test_check_result_line_plainness.sh gains a case proving a dirty `<plugin>/tests/fixtures/*.sh` carrier IS discovered and flagged (at minimum the `emitter_not_plain` arm on a coloured `red()`/`green()` definition), AND a case proving an `_archive/tests/fixtures/*.sh` carrier is NOT discovered; each new fixture tree also contains a clean in-scope suite, so the case grades discovery rather than the exit-2 empty-sweep path (the ZERO DISCOVERY IS RED rule in that suite's header, lines 17-20).
- [ ] IF (a): the now-false prose is re-synced — the module docstring's claim that discovery uses "the same two run-plugin-tests.py uses today" and that "a suite parked one path segment deeper is out of reach by construction" (lines 38-50), and the CLAUDE.md Contributing bullet's scope sentence naming only `tests/*.sh` and `*/tests/*.sh`, both reflect the widened scope.
- [ ] IF (b): the new assertion is falsifiable against BASE behaviour — deleting the `"$SCAN_ROOT"/fixtures/*.sh` term from the `for f in …` loop (base line 96) makes a NAMED case print a `FAIL: <case>` line. An assertion resting only on the pre-existing `exercisers >= 50` floor (base lines 142-147) does NOT satisfy this, because `definers` merely falls to 0 while `exercisers` stays >=50 from the ~77 suites sourcing the helper and the suite remains green.
- [ ] IF (b): the assertion still proves the SHAPE contract, not just glob presence — a coloured `red()`/`green()` body in cogni-knowledge/tests/fixtures/test_helpers.sh must still drive a named case red (the `n_plain -ne 2` path at base lines 129-132), so the fix does not replace a real emitter-shape check with a self-grep for the string `fixtures/`.
- [ ] A runnable `scripts/mutation-check.sh` recipe is recorded (PR body, read by `check-preflight.sh --pr-body-file`, or an in-tree comment) carrying all five flags. Its `--expr` is a `perl -0pi -e` expression — evaluated by perl, never sed/ERE: a sed-style `/…/d` delete form matches nothing and hard-errors as `expr_no_op`, literal parens need backslash escaping, the expression is single-quoted at the call site, and an anchored `^…$` form needs the `/m` modifier because `-0` slurps the file.
- [ ] The recipe's `--case` value is printed as the whole FIRST whitespace-delimited token by the named suite on BOTH paths — a `FAIL: <case>` line when mutated and an `ok: <case>`/`PASS: <case>` line when restored — so the run yields a real red-then-green verdict, not `case_not_found`. Base evidence this needs work: today the emitter-shape case prints `FAIL: test_helpers.sh …` red but `PASS: <definers count> emitter-defining file(s) …` green, whose first tokens differ; a stable case-id emitted on both paths is required.
- [ ] The recorded mutation demonstrates coverage OF THE SHARED HELPER specifically: the `--file` is cogni-knowledge/tests/fixtures/test_helpers.sh (coloured emitter body) and/or the fixtures-scan term, and the reported outcome is a real red-then-green — not a green-stays-green, not `expr_no_op`, not `case_not_found`.
- [ ] Guard invariants intact in scripts/check-result-line-plainness.py: the diff adds no exclusion list, allowlist, skip marker, baseline file, or `--exclude`/`--skip`/`--ignore` flag (grep the diff for `exclude|allowlist|baseline|skip`); the empty-sweep path still returns exit 2 with `success: false` and a non-empty `error` (base lines 152-156, 179); and the two arms stay disjoint (arm A still skips definition lines via the `elif not defn` branch, base line 138).
- [ ] tests/test_check_result_line_plainness.sh still passes against the real tree: R1 reports a clean zero (exit 0) and the liveness floors L1 (`files_scanned >= 80`) and L2 (`definitions_inspected >= 40`) still hold on the PR head (base lines 198-209).
- [ ] `python3 scripts/run-plugin-tests.py` exits 0 on the PR head, and its discovered-suite count is not lower than on base — nothing was newly excluded from execution to make the sweep green.
- [ ] Any test file the PR adds or edits keeps its own result lines plain: every `PASS:`/`FAIL:` label is emitted through a `printf '%s\n' "$1"`-class emitter with no escape literal and no capability-probe gating, so the guard remains a clean zero on the PR's own change.

**Risk watchlist:**
- A wrong fix widens scripts/run-plugin-tests.py's globs instead of (or as well as) the guard's — that would EXECUTE `fixtures/test_helpers.sh` and `fixtures/_cycle_guard_lib.sh`, which define functions for a sibling to source and break when run directly (the runner's docstring, lines 33-49, names exactly this).
- Under (b), an assertion implemented as a self-grep of the suite's own source for the literal `fixtures/` — it satisfies the AC's letter and goes red on deletion, but proves nothing about whether the scan actually reaches and shape-checks the helper. The mutation check is the discriminator.
- Under (b), an assertion keyed on the existing `exercisers >= 50` floor: vacuous, because removing the fixtures glob leaves `exercisers` untouched (the ~77 sourcing suites live in `$SCAN_ROOT/*.sh`).
- A mutation recipe recorded but never actually run: `case_not_found` or `expr_no_op` reported (or unreported) as if it were a passing verdict. Run the recorded recipe rather than reading it.
- Under (a), reaching fixtures via `**`, `recursive=True`, or `os.walk` — that also sweeps `<plugin>/_archive/tests/fixtures/`, forcing back the exclusion list the guard exists to avoid.
- Under (a), the docstring and CLAUDE.md left asserting the guard mirrors run-plugin-tests.py's two globs and that one-segment-deeper is out of reach — a doc that now actively lies about the mechanism it documents.
- Introducing a skip marker, baseline, or allowlist so newly-discovered fixture files pass — the ratchet this guard explicitly refuses (hard clean zero).
- Relabelling the emitter cases in test_plain_result_emitters.sh for mutation-harness token parity while silently changing what a case asserts (e.g. dropping the `n_red`/`n_green` pairing check while renaming its label).
- JUDGMENT-ONLY (not gated): under (b), the coverage stays plugin-local and undiscoverable from the repo root unless the guard's docstring or CLAUDE.md points at the plugin-local owner. A human should decide whether an unlinked plugin-local owner is an acceptable resting state.
- JUDGMENT-ONLY (not gated): under (b) the residual gap for a FUTURE second plugin growing `tests/fixtures/` remains entirely open. Whether that residual warrants a follow-up issue rather than being absorbed here is a maintainer call, not something this checklist can decide.

- The plan-time AC's 5 bullets leave two substantive, base-derivable gates unstated: (i) that a (b)-arm assertion must be falsifiable against the EXISTING `exercisers >= 50` liveness floor, which stays green when the `fixtures/*.sh` glob term is deleted (so an assertion keyed on it would be vacuous while formally satisfying bullet 3), and (ii) that an (a)-arm change carries a doc-sync obligation, since the guard's own docstring (lines 38-50) and the CLAUDE.md Contributing bullet both assert the discovery scope the widening would invalidate. A base-grounded regeneration derives 17 gating items against the AC's 5. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1410

**Summary:** Both perspectives pass and every contract item is confirmed satisfied by execution; the only remaining items are three minor, non-gating preference observations added by the merger's plugin-validator review (none touching the graded contract) — safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (advisory-addition) `cogni-knowledge/tests/test_plain_result_emitters.sh:131-133` — The per-arm tally uses `case "$f" in */fixtures/*)` as a proxy for 'matched by the deeper loop arm'. If SCAN_ROOT were ever overridden with a path containing a /fixtures/ segment, shallow-arm files would also increment fixtures_definers and D1 could stay green with the deeper arm broken. No caller passes an argument today, and the header already states a wider guard must re-home both floors, so this is theoretical.
  - *Suggested fix:* No action needed. Noted only so a future re-homing of the floors does not inherit the proxy silently.

- **[minor]** (advisory-addition) `cogni-knowledge/tests/test_plain_result_emitters.sh:39-44` — The recorded rejection rationale for disposition (a) is imprecise about the mechanism. Exclusion of the parked _archive/ carrier comes from fixed segment NAMES plus depth, not from the number of globs. A literal fixed-segment third pattern would reach cogni-knowledge/tests/fixtures/ while still missing <plugin>/_archive/tests/*.sh, and would force no exemption machinery. The claim holds only for a recursive/os.walk widening, so a future maintainer could conclude a fixed-segment third glob is unsafe when it is not.
  - *Suggested fix:* Tighten one clause: name the rejected form as a RECURSIVE widening (**, recursive=True, os.walk) rather than 'a third, deeper glob' generically. Non-blocking — checklist item 1 asks that both dispositions be named with reasoning, and both are.

- **[minor]** (advisory-addition) `scripts/check-result-line-plainness.py:38-50 / CLAUDE.md Contributing bullet` — Non-gating judgment item carried from the contract's risk watchlist: under disposition (b) the fixtures-depth coverage stays plugin-local and is not discoverable from the repo root. Neither the guard docstring nor the CLAUDE.md bullet points at the designated plugin-local owner. Both remain factually TRUE under (b) — the CLAUDE.md sentence about glob depth still holds exactly — so there is no doc-sync obligation, only a discoverability one.
  - *Suggested fix:* Maintainer call, explicitly not gated by the contract: decide whether an unlinked plugin-local owner is an acceptable resting state, or add a one-line pointer from the guard docstring. Do not block on this.

### Positive observations

- Every one of the 17 contract items is satisfied by the diff, confirmed by execution — suite run, executed mutation-check.sh recipe returning guard_verified, and an independent full-deletion mutation proving D1 is not vacuously riding the exercisers floor. (contract-grading-reviewer)
- The mutation recipe in the PR body was RUN, not read, against an isolated copy and reproduces the claimed result exactly: mutated_result red, restored_result green, verdict guard_verified. (plugin-validator)
- D1 is provably non-vacuous against base behaviour: the identical mutation on the base file exits 0 (coverage silently dropped) while on the PR head it exits 1 with a red D1. (plugin-validator)
- Guard invariants fully intact: check-result-line-plainness.py byte-identical to main, two literal non-recursive globs preserved, no exclusion list/allowlist/skip marker/baseline introduced, exit-2 empty-sweep path untouched. (plugin-validator)
- Scope discipline is exemplary: exactly one file, +43/-2, no manifests touched, zero new breadcrumbs. (plugin-validator)
- cogni-knowledge structural registration fully valid and unaffected: plugin.json v1.0.78 agrees with the marketplace mirror, 21 skills and 16 agents all well-formed. (plugin-validator)

### Recommendation

---

PR #1421, opened 2026-08-16
Move excalidraw MCP out of plugin .mcp.json into install-mcp

claude-cogni-work:
Closes #1352.

## Summary

- Delete `cogni-visual/.mcp.json` and `cogni-portfolio/.mcp.json`. Plugin-level stdio servers are spawned **eagerly at session start**, not lazily on first tool call, so on a machine that never ran `install-mcp` the missing `start.sh` fails visibly under `/mcp` — for every user of the plugin, whether or not they ever render a diagram.
- Extend `cogni-workspace/scripts/patch-desktop-config.py` with `--target {desktop,cli,both}` (default `desktop`, so every existing call site is byte-compatible), `--server <name>`, and `--cli-config-path`. The CLI target writes the top-level `mcpServers` in `~/.claude.json` as a read-modify-write, reusing the script's existing skip-unless-`--force` idempotency and timestamped backup.
- Rewire `install-mcp`'s discovery fallback #3 off `.mcp.json` onto `.claude-plugin/plugin.json` — the key `discover-plugins.sh` already uses. Left alone it would silently match nothing and report success.
- Retire `excalidraw_sketch` and retract the roadmap paragraph that kept it.
- Correct every surface asserting per-plugin `.mcp.json` auto-discovery, in both wiki trees.
- Add `cogni-workspace/tests/test-mcp-declaration-hygiene.sh`.

## Why both `.mcp.json` files, not just the one named in the issue

`cogni-portfolio/.mcp.json` declared an `excalidraw` block **byte-identical** to cogni-visual's. Acceptance criterion 3 — "a fresh session shows no failed server under `/mcp`" — is unreachable while either survives, so a cogni-visual-only change could not have satisfied the issue's own bar.

The duplicate name was load-bearing in a second way, and it explains AC6. Two plugin-level declarations of one server name make the client disambiguate by plugin-qualifying one side's tool names (`mcp__plugin_cogni-portfolio_excalidraw__*`), which silently dead-matches the `mcp__excalidraw__.*` PreToolUse matcher that plugin carries. **Removing both declarations fixes that matcher at the root** — the user-config server becomes the sole `excalidraw`, tool names are unqualified again, and both plugins' existing matchers fire. No hook file is edited. AC6 is fixed here, not filed.

## The unknown this had to resolve first

[...]

Issue #1352, opened 2026-08-13
Move excalidraw .mcp.json out of the plugin into install-mcp on-demand; drop dead excalidraw_sketch


sdh07:
## Problem

`cogni-visual/.mcp.json` declares two MCP servers at the plugin level:

```json
{"mcpServers": {
  "excalidraw_sketch": {"type": "url", "url": "https://mcp.excalidraw.com"},
  "excalidraw": {"command": "bash", "args": ["-c", "$HOME/.claude/mcp-servers/mcp_excalidraw/start.sh"], ...}}}
```

Plugin-level stdio servers are spawned **eagerly at session start**, not lazily on first tool call. `excalidraw.log` records four `Starting Excalidraw MCP server` sequences within 150 ms at session boot, before any tool call.

On a machine that has never run `install-mcp`, `$HOME/.claude/mcp-servers/mcp_excalidraw/start.sh` does not exist and the server fails visibly under `/mcp`.

Separately, `excalidraw_sketch` appears to be **already dead** — no `mcp__excalidraw_sketch__*` tools are present in a live session, suggesting the remote URL server has been failing silently for some time.

This must be resolved before `cogni-visual` merges, or the condition lands in the plugin every user installs.

## Proposed change

- Move the `excalidraw` server declaration out of the plugin into `cogni-workspace:install-mcp`, which already owns MCP lifecycle and already patches `claude_desktop_config.json`. It becomes an on-demand user-config write, so users who never render diagrams never spawn a server.
- **Drop `excalidraw_sketch` entirely** — confirm it is dead first, then remove.
- Re-verify the `PreToolUse` matcher still fires. Tool names derive from the **server** name, so moving the declaration must not change `mcp__excalidraw__*`.

**Note a pre-existing bug found while investigating:** `cogni-portfolio` ships its own `.mcp.json` declaring a server *also* named `excalidraw`, and Claude Code disambiguates the duplicate by plugin-qualifying it (`mcp__plugin_cogni-portfolio_excalidraw__*`). `cogni-portfolio/hooks/hooks.json` matches `mcp__excalidraw__.*`, which **cannot match its own mangled tool names** — that hook has been silently dead. Fix or file separately.

## Acceptance criteria

- [x] `cogni-visual/.mcp.json` no longer declares servers (or is deleted)
- [x] `install-mcp` writes the excalidraw server into user config on demand, idempotently
[...]

PR #1423, opened 2026-08-17
Serialise the excalidraw canvas start behind an atomic claim

### LINKED ISSUE #1422: cogni-visual/cogni-portfolio: serialise the excalidraw canvas start — revived matchers now race two identical hooks
## Problem

`cogni-visual/hooks/ensure-excalidraw-canvas.sh` and `cogni-portfolio/hooks/ensure-excalidraw-canvas.sh` are byte-identical, and so are the two `hooks.json` files that register them. While both plugins carried a plugin-level `.mcp.json` declaring `excalidraw`, the client plugin-qualified the tool names (`mcp__plugin_cogni-portfolio_excalidraw__*`), which dead-matched the `mcp__excalidraw__.*` PreToolUse matcher in both — so **zero** hooks fired.

Removing both declarations restored the unqualified `mcp__excalidraw__*` prefix and revived both matchers. The consequence is that on a machine with both plugins installed, every `mcp__excalidraw__*` call now dispatches **two** PreToolUse hooks where zero fired before.

The script guards its spawn on a port probe:

```bash
if ! nc -z localhost "$PORT" 2>/dev/null; then
  ...
  nohup node dist/server.js > "$LOG_FILE" 2>&1 &
```

`nc -z` is a check, not a mutual-exclusion primitive. Two hooks dispatched in parallel against a cold canvas can both fall through the probe, both spawn `node dist/server.js`, and both write `canvas.pid` — leaving the loser dead on `EADDRINUSE` with the pid file pointing at the dead process.

This is a real race, not a theoretical one: the window is the interval between the probe and the listener binding, which spans a Node process start.

## Proposed change

Serialise the start with an atomic claim before the spawn, applied identically to both copies so they stay byte-identical:

- `mkdir "${EXCALIDRAW_DIR}/canvas.lock"` (atomic on POSIX) or `flock` on the pid file
- The winner spawns and releases the claim once the port is listening
- The loser falls through to the **existing** wait-for-port loop rather than erroring
- Ensure the lock cannot outlive a crashed spawn (stale-lock timeout, or trap-based release)

## Acceptance criteria

1. Two concurrent invocations of `ensure-excalidraw-canvas.sh` against a cold canvas result in exactly one `node dist/server.js` process.
2. `canvas.pid` always names a live listening process after both invocations return.
3. The loser waits for the port and exits successfully rather than reporting failure.
4. `cogni-visual/hooks/ensure-excalidraw-canvas.sh` and `cogni-portfolio/hooks/ensure-excalidraw-canvas.sh` remain byte-identical.
5. A stale lock left by a killed spawn does not deadlock a subsequent invocation.

## Context

Surfaced by the automated review of the PR that removed the two `.mcp.json` declarations. The reviewer offered an explicit out-of-scope path — record it rather than fix it inline — because the fix belongs to the two hook files together, outside that change's declaration-move scope. The consequence is recorded in `cogni-workspace/references/absorption-roadmap.md` under Decision 3; this issue tracks the actual fix.

Both target files exist on `main` today, so this is actionable independently of the originating branch.

claude-cogni-work:
Closes #1422.

## What was wrong

`cogni-visual/hooks/ensure-excalidraw-canvas.sh` and `cogni-portfolio/hooks/ensure-excalidraw-canvas.sh` are byte-identical and register the same unqualified `mcp__excalidraw__.*` PreToolUse matcher, so a machine with both plugins installed dispatches **two** of them in parallel on every tool call.

The hook guarded its spawn on `nc -z localhost "$PORT"` — a point-in-time check, not a mutual-exclusion primitive. Two invocations against a cold canvas could both fall through the probe, both run `nohup node dist/server.js`, and both write `canvas.pid`. The loser dies on `EADDRINUSE`, and the pid file can be left naming a dead process. Every exit status stays `0` throughout, so nothing reports it — the failure is silent by construction, which is why it needs a behavioural test rather than a lint.

## The fix

The spawn is taken behind an atomic `mkdir` claim at `${EXCALIDRAW_DIR}/canvas-start.lock`:

- **Only the winner spawns and writes `canvas.pid`.** The loser skips both and falls through to the **existing** 10s wait-for-port loop — its rendezvous is the port the winner is binding — then exits `0` through the existing paths.
- **Release is one flag-guarded `trap release_start_claim EXIT INT TERM`**, armed before the fast path. The `SPAWN_CLAIMED` guard is load-bearing rather than cosmetic: the trap is unconditional, so without it a *loser* would `rmdir` the *winner's* claim on its way out through the shared wait-loop exits.
- **A claim older than 60s is swept.** That threshold is 4x the `"timeout": 15` this hook's own `hooks.json` declares, so a spawn killed at that deadline expires, but a healthy in-flight winner is never robbed — which would reintroduce the very double-spawn being fixed.
- **Retiring that stale claim is itself atomic.** See below — the first version of this was not, and put the double spawn straight back on the recovery path.

### Retiring a stale claim (the second round of this fix)

The sweep was a plain `rmdir` followed by a `mkdir`. That is not a mutual-exclusion primitive either: two invocations that both judge the same aged claim stale both reach the `rmdir`, and the second removes the **first one's fresh claim**, so both `mkdir` succeed and both spawn. Same defect as the original bug, on the path added to recover from it.

**Reproduced deterministically before fixing, not inferred from review.** The new case releases two invocations together off a gate file and then staggers them with a stubbed `date` — the hook's only call to it, and it sits exactly between reading the aged claim's mtime and acting on that reading. Against the old sweep it reports two spawns every run.

The removal is now a **rename gated on a compare-and-swap**. `rename(2)` is atomic and single-winner, so only one invocation can carry a given claim aside; and the mtime already read for the age test doubles as that claim's identity, because renaming a directory leaves its own mtime alone and touches only the parent's. An invocation that carries aside something other than what it judged declines to claim.

A plain rename **alone does not close this**, which is worth stating because it is the obvious fix: the second invocation's rename is unconditional once it has formed its verdict, so it can still carry off a fresh claim taken in the meantime. That interleaving is narrow enough to measure as clean and still be there.

A mismatched claim is **discarded rather than restored**. Restoring reads as the tidier choice and is worse: the owner may release while we hold the directory, and a restore after that strands it for the whole staleness window carrying a fresh mtime — so every later invocation declines to sweep and no canvas ever starts. Discarding costs nothing, because the spawn now re-probes the port.

Two supporting changes:

- **The spawn re-probes the port once the claim is held.** An invocation descheduled between the probe and its claim could otherwise acquire the claim after another had already bound the port, then spawn a server doomed to `EADDRINUSE` and overwrite `canvas.pid` with its pid. The claim holder is the only writer at that point, so the re-check races nothing.
- **A claim carried aside but not yet discarded is released by the trap.** Nothing else in the tree knows the `.stale.<pid>` name, so a hook killed mid-sweep would otherwise leave a directory no later run looks at.

### Why these primitives

- **Bare `mkdir`, never `mkdir -p`.** `-p` succeeds against an existing directory, which destroys the exclusion signal the claim depends on.
- **Not `flock(1)`** — the issue offered it, but it is not stock on macOS, which this script already targets (it has a Darwin branch). `cogni-workspace/bash/portability-utils.sh` already treats flock as Linux-only.
- **The claim attempt sits in an `if` condition.** Under `set -euo pipefail` a bare `mkdir` statement would abort the hook exactly when the claim is *lost* — and a non-zero PreToolUse hook blocks the tool call it exists to enable. That would be strictly worse than the bug.
- **The `stat` chain resolves into a variable before the arithmetic.** An empty command substitution nested inside `$(( ))` is a hard bash syntax error.
- **`rmdir`, not `rm -rf`**, so a mis-derived path can never delete a tree.

This follows the existing in-repo claim pattern in `cogni-portfolio/scripts/append-claim.sh`.

## Verification

| Check | Result |
|---|---|
| `cogni-visual/tests/test-excalidraw-canvas-lock.sh` | 25 checks green |
| The same suite, against the **pre-fix** sweep | `FAIL: … expected 1 spawn, got 2` — the defect, reproduced |
| `run-plugin-tests.py --filter cogni-visual` | rc=0 (3 suites) |
| `run-plugin-tests.py --filter cogni-portfolio` | rc=0 (6 suites) |
| `cmp` both hook copies | identical (AC 4) |
| `bash -n` both copies | clean |
| `check-result-line-plainness.py` · `check-breadcrumbs.py` · `check-version-bump.py` · `check-plugin-inventory.py` | all rc=0 |
| non-zero exits introduced | 0 |

Branch merged up to `main` (`898e6dc7` + the version bumps) before this run, so the sweeps are against the tree this actually lands in.

The suite is hermetic: each case builds its own `mktemp -d` sandbox and prepends stub `nc`/`node`/`curl`/`open`/`xdg-open` to `PATH`. No socket is bound, no network is used, nothing is written outside the sandbox. The `node` stub deliberately waits before signalling "port up", which is what forces genuine contention — without it the second invocation would take the fast path and the case would pass vacuously.

**On the barrier.** The obvious way to release two racers together is a FIFO on stdin, and it does not work: the writer's open unblocks whichever reader arrived first, its close then sends EOF, and a reader that opens after that close waits for a writer that never returns — the case hangs instead of racing. Verified in isolation before being discarded. The gate file both racers spin on releases them inside the same millisecond and forks nothing.

**On determinism.** An earlier version of the new case raced ten unstaggered trials and reported green against the broken sweep — the breaking interleaving is a few syscalls wide. A case that catches its defect at some rate below 1 is not a guard, so the stagger constructs the interleaving rather than waiting for it.

## Mutation recipe

All seven recipes are documented in the suite header. Each was **replayed against the shared harness and returned `guard_verified`** (red on the mutant, green restored); M1 was re-replayed after the merge from `main` to confirm the harness still resolves.

The recipe for the new case inverts the identity test the sweep turns on, so the invocation that really did carry the aged claim aside declines to claim and the one that did not finds nothing left to carry — no spawn at all. Chosen over reverting the sweep to a plain `rmdir` precisely because it reddens deterministically instead of at whatever rate the underlying race happens to hit:

```bash
  --root . \
  --file cogni-visual/hooks/ensure-excalidraw-canvas.sh \
  --expr 's#\$swept_mtime" -eq "\$lock_mtime#\$swept_mtime" -ne "\$lock_mtime#' \
  --test 'bash cogni-visual/tests/test-excalidraw-canvas-lock.sh test_concurrent_stale_claim_spawns_exactly_one_server' \
  --case test_concurrent_stale_claim_spawns_exactly_one_server
```

The original claim recipe, for reference:

```bash
  --root . \
  --file cogni-visual/hooks/ensure-excalidraw-canvas.sh \
  --expr 's#mkdir "\$LOCK_DIR" 2>/dev/null#mkdir -p "\$LOCK_DIR" 2>/dev/null#' \
  --test 'bash cogni-visual/tests/test-excalidraw-canvas-lock.sh test_cold_start_spawns_exactly_one_server' \
  --case test_cold_start_spawns_exactly_one_server
```

`mkdir -p` succeeds against an existing directory, so both racers believe they won and both spawn — exactly the pre-fix behaviour. The searched literal occurs once (the other two `mkdir` mentions are prose), the replacement is disjoint from it, `#` is a safe delimiter because neither side contains one, and `\$` reaches perl as a literal dollar rather than an interpolated variable. The remaining recipes (stale-sweep threshold, `EXIT INT TERM` trap, cross-copy divergence, the mtime floor, and the `stat` chain order) are in the suite header.

## Scope

**Out, deliberately:** `cogni-workspace/templates/mcp-wrappers/excalidraw-canvas.sh` is a **third** site that backgrounds the same server at MCP startup and writes no pid file. The acceptance criteria scope to "two concurrent invocations of `ensure-excalidraw-canvas.sh`", so it is recorded here rather than silently pulled in — the `canvas.pid` guarantee this PR buys covers the two hook writers, not the wrapper's untracked background job.

Both `hooks.json` are unchanged: double dispatch is the intended consequence of having both plugins installed; the defect was that the spawn was not mutually exclusive. No version bump in any manifest — the post-merge workflow owns that.

**Alternative considered:** a second byte-identical suite in `cogni-portfolio/tests/`, so either plugin's filtered local sweep catches a divergence. Rejected on blast radius — repo CI runs the full sweep, so one suite already catches divergence at the gate that matters, and the twin design has to assert its own identity with its twin.

## Note on the issue's premise

An earlier revision of this body argued the issue's premise did not yet hold, because both `.mcp.json` files still declared `excalidraw` on `main`. **That is now out of date and the correction runs the other way:** #1421 merged as `898e6dc7`, deleting both declarations, so the user-config server is the sole `excalidraw`, tool names are unqualified again, and both `mcp__excalidraw__.*` PreToolUse matchers fire. The double dispatch this PR serialises is live on `main` today rather than hypothetical.

This branch has been merged up to that `main`. It still touches none of #1421's files.


claude-cogni-work:

Author instance is done. This PR is draft with `svc:needs-review`; the merger instance owns the independent review and the merge decision. No verdict label, no `reviewed-head-sha` marker, and no ready flip were written here — the review below is an author self-grade, not a substitute for the merger's own pass.

**Head:** `c3e92555`

### Handoff preflight
`rc=0`, `clean=true`, 0 violations, 0 contract breaches. Four instruments skipped, all `not_applicable`: `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` — the diff touches no `SKILL.md` and no `agents/*.md`.

Worth recording for whoever runs this next: the preflight **first returned rc=1 with two violations** (`run-tests: discovered 0 suites for cogni-visual against a floor of 2`, and `version-bump: 1 violation`). Both were artifacts of `--root` pointing at the shared clone, whose working tree sits on an unrelated branch that has no `cogni-visual/tests/` directory and does touch a dozen version manifests. Re-run against a worktree checked out on this branch, the preflight is clean. Neither violation ever described this diff.

### Contract self-grade
Graded against the acceptance contract committed on #1422 at triage time, before any branch was cut (17 items, `--target issue --origin triage`).

- **First pass: 16/17 satisfied, 1 violated.** Item 7 requires that no lock/claim token appear textually above the fast path. The claim declarations, both helper functions and the trap sat above it. Runtime cost of a warm canvas was never affected — declarations and a trap-arm, not filesystem I/O — but the criterion is literal and grep-enforced.
- **Fixed in `c3e92555`** (bounded single fix pass, in files the plan already touched): the whole claim block moved into the cold-start branch. Coverage is unchanged — the claim is only ever taken inside that branch, the trap is still armed before it is taken, and bash traps are process-global so the release still fires at the final exit outside the branch. The suite now asserts the placement so it cannot drift back.
- **Residual: none.** No `unclear` items, no unfixed violations.

### Verification at this head

| Check | Result |
|---|---|
| `run-plugin-tests.py --filter cogni-visual` | rc=0 (3 suites — new one discovered) |
| `run-plugin-tests.py --filter cogni-portfolio` | rc=0 (6 suites) |
| new guard suite | all checks green |
| `cmp` both hook copies | identical (AC 4) |
| `bash -n` both copies | clean |
| `check-result-line-plainness.py` | rc=0 |
| `check-version-bump.py` | rc=0, 0 manifests touched |
| claim tokens above the fast path | 0 (item 7) |
| mutation recipes M1/M2/M3 replayed at this head | all `guard_verified` |

### Two things the merger should weigh

1. **The issue's premise is not yet true on `main`.** #1422 says the `.mcp.json` declarations were removed and the matchers thereby revived; that removal is #1352 / PR #1421, still unmerged. Both declarations still exist on `main` today. The fix is correct and independently actionable anyway — the probe-then-spawn window races for any two parallel invocations — and this PR touches none of #1421's files, so the two do not conflict.
2. **A third spawn site is deliberately out of scope.** `cogni-workspace/templates/mcp-wrappers/excalidraw-canvas.sh` backgrounds the same server at MCP startup and writes no pid file. ACs 1-5 scope to two concurrent invocations of `ensure-excalidraw-canvas.sh`, so the `canvas.pid` guarantee bought here covers the two hook writers, not the wrapper's untracked background job.

### Deviations from the standard author path
`Step 6.4`'s `/simplify` pass was skipped to protect context budget for the handoff; the code had already been through a `plan-refiner` pass. `Step 6`'s `plugin-dev:hook-development` reference dispatch was also skipped — this change is shell logic inside an existing hook, not hook registration or event wiring. Both are disclosed rather than silently omitted.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1423
**Perspectives consulted:** cogni-service:service-review (focused: reproduced CI failure + root-cause analysis)

**Summary:** Blocked on a reproduced, root-caused cross-platform defect in the shipped hook. CI's 'Plugin test suites' job is red with 3 failures in this PR's own new suite, while the same suite passes 120/120 locally on macOS. Root cause, proven by execution: the stale-claim age check reads `stat -f %m "$LOCK_DIR" 2>/dev/null || stat -c %Y ...`, but `-f` is BSD-only for mtime — on GNU coreutils `-f` is `--file-system` and `%m` is the MOUNT POINT, so the first command SUCCEEDS returning `/` and the `||` fallback is never reached. `lock_mtime` becomes `/`, and `lock_age=$(( now_secs - / ))` aborts the hook with rc 1. That single fault explains all three CI failures. It lives in the shipped hook rather than the test, so on Linux the stale-lock recovery path this PR exists to add fails silently in both byte-identical copies — and issue AC3 ('the loser waits for the port and exits successfully rather than reporting failure') is contradicted in CI. The design is otherwise sound: the atomic mkdir claim, the flag-guarded trap, and the loser's rendezvous on the existing wait-for-port loop are all correct and well argued.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| cogni-service:service-review (focused: reproduced CI failure + root-cause analysis) | ⚠️ needs_revision |

### Findings (2)

- **[major]** (auto-fixable) `cogni-visual/hooks/ensure-excalidraw-canvas.sh:108 (and the byte-identical cogni-portfolio copy)` — THE GATING DEFECT — reproduced and root-caused. The line `lock_mtime=$(stat -f %m "$LOCK_DIR" 2>/dev/null || stat -c %Y "$LOCK_DIR" 2>/dev/null || echo 0)` assumes `stat -f %m` FAILS on GNU so the `||` chain falls through to `stat -c %Y`. It does not fail: on GNU coreutils `-f` means `--file-system` and `%m` is the mount point, so it exits 0 printing `/`, making the intended fallback unreachable. `lock_mtime` is then `/`, and line 111's `lock_age=$(( now_secs - lock_mtime ))` raises an arithmetic syntax error that exits the hook with status 1. Verified by replaying the exact chain against a stub reproducing GNU semantics: lock_mtime resolved to `/` and the arithmetic aborted with exit 1. This accounts for all three CI failures precisely — `test_stale_lock_does_not_deadlock` reporting 'got 0 and 1' (the hook dies before spawning), 'claim was not released' (it dies before the release), and `test_cold_start_spawns_exactly_one_server` reporting 'exit codes were 0 and 1' (the loser takes the stale-check path and dies). The existing `[[ -n "$lock_mtime" ]] || lock_mtime=0` guard cannot catch it, because it tests for an EMPTY result while this failure mode is a SUCCESSFUL WRONG one — the inline comment at :106-107 anticipates exactly the empty case and not this one. The consequence is not confined to CI: this is shipped hook code, so on every Linux machine the stale-lock recovery this PR exists to add aborts the hook instead, in both copies.
  - *Suggested fix:* Invert the chain so the GNU form is tried first and the BSD form is the fallback (`stat -c %Y "$LOCK_DIR" 2>/dev/null || stat -f %m "$LOCK_DIR" 2>/dev/null || echo 0`) — GNU rejects `-c` on BSD and BSD rejects `-c`, so the order matters — or, more robustly, validate the result is numeric before using it: `[[ "$lock_mtime" =~ ^[0-9]+$ ]] || lock_mtime=0`. The numeric guard is worth adding regardless, since it makes the whole chain fail safe rather than relying on which stat variant errors. Apply identically to both copies to hold AC4, and keep the CI run green before re-handing off — the suite is the acceptance evidence for AC1/AC2/AC3 and must pass on the Linux runner, not only on macOS.

- **[minor]** (auto-fixable) `cogni-visual/tests/test-excalidraw-canvas-lock.sh` — The suite passes on macOS and fails on Linux, so it did not catch a defect that only manifests on the platform CI (and most users) run. The fixture stubs `nc`, `node`, `curl`, `open` and `xdg-open` but not `stat`, which is the one external whose flag semantics actually diverge between BSD and GNU — so the divergence is invisible to a local run. This is why the failure reached CI rather than the author's terminal.
  - *Suggested fix:* Once the chain is fixed, add a case that stubs `stat` with GNU semantics (`-f` succeeds returning a non-numeric mount point) and asserts the hook still computes a sane age and exits 0. That pins the actual cross-platform contract instead of relying on whichever stat the runner happens to provide.

### Positive observations

- The core design is correct and well argued: an atomic `mkdir` claim is a genuine mutual-exclusion primitive, unlike the `nc -z` probe it replaces, and the loser's rendezvous on the pre-existing 10s wait-for-port loop is the right way to avoid a second failure mode.
- The `SPAWN_CLAIMED` flag guarding `release_start_claim` is load-bearing and the PR body says so accurately — without it a loser exiting through a shared wait-loop path would rmdir the winner's claim mid-spawn. The trap covers EXIT INT TERM.
- Only the winner writes `canvas.pid`, which directly satisfies the issue's AC2 by construction rather than by timing.
- The 60s staleness bound is justified against a real number (four times the hook's own 15s timeout in hooks.json) rather than picked arbitrarily.
- Both hook copies and both hooks.json files are byte-identical at head, and the suite pins that agreement explicitly (AC4).
- The suite carries executed mutation cases (M1-M3) that turn specific assertions red, matching the repo's established mutation-harness convention.
- This PR is the properly filed follow-up to the scope boundary recorded during the review of #1421 — the deferral was honoured with a real fix rather than left as an untracked note.

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1423
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator (cogni-visual), plugin-dev:plugin-validator (cogni-portfolio)

**Summary:** The byte-identical excalidraw-canvas hooks fix the cold-start mkdir race correctly, and this revision's own target — the GNU-unsafe stat chain — is resolved with CI now green on the Linux runner. But both plugin-validator reviews independently caught a second, distinct defect: the stale-claim sweep is itself non-atomic, letting two racers double-spawn in exactly the concurrent-stale-claim state AC5 describes. The orchestrator reproduced it directly and confirmed an identical 3/10 double-spawn rate at the pre-revision head c3e92555, so it is pre-existing in this PR's own new sweep code rather than introduced or widened by this revision. Deduped into one major finding, plus four minors.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-dev:plugin-validator (cogni-visual) | ⚠️ needs_revision |
| plugin-dev:plugin-validator (cogni-portfolio) | ⚠️ needs_revision |

### Findings (5)

- **[major]** `cogni-visual/hooks/ensure-excalidraw-canvas.sh:122 (byte-identical copy: cogni-portfolio/hooks/ensure-excalidraw-canvas.sh:122-127)` — The stale-claim sweep is a non-atomic rmdir-then-mkdir, reintroducing the exact double-spawn defect this PR exists to remove. Two racers that both judge the same aged claim stale both pass the -d/age test and both reach rmdir; whichever rmdir lands second removes the other racer's FRESH claim, so both set SPAWN_CLAIMED=1 and both spawn — violating AC1 (exactly one server) and AC2 (canvas.pid names a live process) in precisely the concurrent-stale-claim state AC5 describes. plugin-validator (cogni-visual) reproduced 8/12 double-spawns under a stdin-barrier harness (3/12 dead canvas.pid with an EADDRINUSE-simulating stub); plugin-validator (cogni-portfolio) reproduced 2/3 unassisted and 3/3 with a 0.4s deschedule. The orchestrator independently reproduced it against the current head at 3/10 and measured the identical 3/10 at the pre-revision head c3e92555 — so the defect is PRE-EXISTING in this PR's new sweep code and was not introduced or widened by this revision. It is reported here because this revision made the sweep reachable on Linux at all (previously it aborted), so the path is now live on the platform CI and most users run.
  - *Suggested fix:* Replace the sweep body in both byte-identical copies with an atomic rename, so only one racer can retire a given stale claim: `if mv "$LOCK_DIR" "${LOCK_DIR}.stale.$$" 2>/dev/null; then rm -rf "${LOCK_DIR}.stale.$$" 2>/dev/null || true; if try_claim_start; then SPAWN_CLAIMED=1; fi; fi`. The racer whose mv fails takes no claim and falls through to the existing shared port wait — the AC3 path. Verified 0/12 double-spawns. A naive nested `mkdir "$LOCK_DIR.sweep"` gate does NOT fix this: the gate is released before the loser reaches it, so the loser still sweeps the winner's fresh claim. Apply identically to both copies to preserve AC4, and extend the hook comment, which currently explains only why the claim mkdir is atomic and not why the sweep must be too.

- **[minor]** `cogni-visual/tests/test-excalidraw-canvas-lock.sh:260` — Coverage gap that lets the major finding above stay silent. test_cold_start_spawns_exactly_one_server races two invocations only against a virgin sandbox, while test_stale_lock_does_not_deadlock exercises the stale claim with a single invocation. The intersection — concurrency against a stale claim — is the one state where the guarantee breaks, and no case reaches it.
  - *Suggested fix:* Add a case that plants a stale claim via plant_stale_claim then races two run_hook invocations, reusing assert_stale_claim_swept. Release both racers through a shared FIFO on stdin (the hook's stdin drain makes that a precise barrier). Loop ~10 trials requiring exactly one spawn each so the case is a deterministic mutation target, and add a matching recipe to the header block.

- **[minor]** (auto-fixable) `cogni-visual/hooks/ensure-excalidraw-canvas.sh:132` — The winner spawns unconditionally once it holds the claim, without re-probing the port it decided was down before taking the claim. A racer descheduled between the probe and its claim attempt can acquire the claim after the winner has bound the port and released, then spawn a doomed server and overwrite canvas.pid. Survived a rename-based sweep in 1/12 trials — an independent, much narrower window of the same shape.
  - *Suggested fix:* Guard the spawn with a fresh probe now that the claim is held: `if [[ "$SPAWN_CLAIMED" -eq 1 ]] && ! nc -z localhost "$PORT" 2>/dev/null; then`. The claim holder is the only writer at that point, so the re-check is race-free. Mirror into the cogni-portfolio copy.

- **[minor]** `cogni-portfolio/hooks/ensure-excalidraw-canvas.sh:63-65` — LOCK_STALE_SECS=60 is four times the 15s hook timeout, but the winner's maximum in-flight hold is bounded by that same 15s. For the ~45s between a spawn being killed at its deadline and the claim ageing out, every invocation loses the claim, declines to sweep, spawns nothing, waits the full 10s on a port that never comes up, and returns a warning while the canvas stays down. AC5 only requires no deadlock and this path does terminate, so the criterion is met, but the canvas is unavailable for up to a minute after any killed spawn.
  - *Suggested fix:* Make liveness rather than age the primary staleness signal: have the winner write its pid into the claim (`echo $$ > "$LOCK_DIR/owner"` right after a successful mkdir) and treat a claim whose owner fails `kill -0` as stale regardless of age, keeping the 60s age check as the fallback for an unreadable owner file. Apply to both copies.

- **[minor]** `PR diff` — Diff is 606 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

### Positive observations

- This revision's own target is resolved and proven on the platform that failed: the GNU-first stat chain plus the ^[0-9]+$ numeric floor is the right two-halves fix, and CI's 'Plugin test suites' job is now green on the Linux runner at head 3485a95c, where it was red at c3e92555.
- The textual ordering arm is correctly justified: the numeric floor makes both chain orders behave identically at runtime, so only a static read can catch a BSD-first chain that leaves the GNU arm dead on Linux.
- AC4 verified directly: both hook copies byte-identical, hooks.json pair identical too, pinned by a standing test with an existence floor so a deleted file cannot read as identical.
- The core cold-start fix is real and correct: bare mkdir not mkdir -p, the claim attempt in an if condition so a lost claim cannot abort the hook under set -e, only the winner writes canvas.pid, and the loser rendezvous is the port itself. AC1-AC3 hold on a virgin cold canvas.
- The fast path is genuinely untouched, pinned textually rather than by timing, so a future edit creeping lock handling upward reddens.
- The suite is hermetic: PATH stubs under mktemp -d, no socket bound, no network, started pids tracked for cleanup. The node stub's deliberate wait is what forces real contention instead of a vacuous pass.
- test_release_is_trapped_for_signals is a well-judged per-arm liveness floor, cross-reading LOCK_STALE_SECS against the timeout each plugin's hooks.json actually declares — reachable by no behavioural case. Comments are stripped before every code grep.
- Result emitters are plain printf with the case id as a separate argument; all six mutation recipes are documented with delimiter and occurrence-count reasoning and were replayed to guard_verified, and the shared assert_stale_claim_swept helper stops one stale-sweep case drifting from the other.
- Repo hygiene clean across both plugins: version-bump gate ok (no version touched in the branch), breadcrumb guard 0 violations, result-line plainness 0 violations, and all nine discovered suites across the two plugins pass.

### Recommendation

---

claude-cogni-work:
## Revision posted

Commit `3485a95c`. Both findings from the previous verdict are addressed, and the CI job that was red is now green on the Linux runner.

### Findings resolved (2/2)

**[major]** `cogni-visual/hooks/ensure-excalidraw-canvas.sh:108` (and the byte-identical cogni-portfolio copy) — the BSD-first `stat` chain — **resolved in `3485a95c`**

Fixed both halves, because either alone leaves a sharp edge:

- The chain now tries the GNU form first. Confirmed on this host that `stat -c %Y` fails cleanly on BSD (`illegal option -- c`), so that is the direction where the fallback is actually reachable.
- The guard now floors on **shape** rather than emptiness: `[[ "$lock_mtime" =~ ^[0-9]+$ ]] || lock_mtime=0`. The old `-n` test structurally could not catch this, because the failure is a successful *wrong* answer rather than an empty one.

Both copies moved together, so AC4 byte-identity holds (`cmp` clean, pinned by the standing test).

**[minor]** `cogni-visual/tests/test-excalidraw-canvas-lock.sh` — the suite could not catch it — **resolved in `3485a95c`**

Added `add_divergent_stat_stub` (a `stat` stub in the same `$fx/bin` that `run_hook` already prepends to PATH, answering non-numerically for both flag forms) and a new case `test_stale_lock_survives_divergent_stat` with two arms: a behavioural arm proving an unusable reading can no longer reach the arithmetic, and a textual ordering arm — necessary because the numeric floor makes both chain orders behave identically at runtime, so nothing behavioural can observe the order. The ordering arm runs first, since it is sub-second and is the only arm that reddens when the order flips.

Two new mutation recipes (M5 kills the floor, M6 kills the ordering) were **replayed against the shared harness and returned `guard_verified`**; all six were re-replayed after a refactor and each still returns `guard_verified`.

### Verification

| Check | Result |
|---|---|
| CI `Plugin test suites` on `3485a95c` | **pass** (was red on `c3e92555`) |
| All other CI checks | pass |
| `test-excalidraw-canvas-lock.sh` | 22 checks green |
| `run-plugin-tests.py` cogni-visual / cogni-portfolio | 3 + 6 suites, 0 failing |
| `cmp` both hook copies | identical (AC4) |
| Mutation recipes M1–M6 | 6/6 `guard_verified` |
| breadcrumb / result-line / version-bump / workflow-yaml guards | pass |

A `/simplify` pass over the delta also folded the triplicated stale-sweep assertions into one `assert_stale_claim_swept` helper, and corrected an over-claim in the new comment: it originally asserted GNU `%m` yields the mount point, which I could not verify without a GNU host. It now states only what the CI failure proves — that the call exits 0 without reporting an mtime — which is the load-bearing part.

### Not resolved — a second, distinct defect

The fresh review gate caught something the previous verdict did not, and it is **not** this revision's finding: the stale-claim sweep is itself non-atomic, so two racers that both judge the same claim stale each `rmdir` it and each re-claim. I reproduced it directly — **3/10 unassisted trials double-spawn** against a planted stale claim — and measured the **identical 3/10 rate at the pre-revision head `c3e92555`**, so it is pre-existing in this PR's own new sweep code rather than introduced or widened here. It is live now on Linux in a way it was not before, because this revision is what makes the sweep reachable there at all.

That needs an atomic-rename sweep in both copies plus a regression case for the concurrency-against-a-stale-claim intersection. It is a design change to the claim protocol rather than a mechanical edit, so I have deliberately **not** bolted it onto this cycle — the PR stays author-side with `svc:review-needs-revision` for a planned revise.

### Follow-up issues

- #1424 — `cogni-portfolio/scripts/append-claim.sh:35` carries this same BSD-first `stat` chain, unfloored and nested directly inside `$(( ))` under `set -euo pipefail`, so on Linux it aborts the claims writer outright and breaks that script's JSON envelope contract. Verified as the last remaining copy repo-wide; out of scope here (different plugin, different failure surface).

sdh07:
## Revision posted — handed back for review

Head is now `f8897735`. This addresses the major and two of the four minors from the [06:31Z verdict](https://github.com/cogni-work/insight-wave/pull/1423#issuecomment-5312645232); the branch has also been merged up to `main` (`898e6dc7` plus the version bumps), so CI runs against the tree this lands in.

### The major — confirmed by reproduction, then fixed

The non-atomic stale-claim sweep was **reproduced here before being changed**, rather than taken from the verdict. Against the old sweep the new case reports `expected 1 spawn, got 2`, every run.

The sweep is now a rename gated on a compare-and-swap. `rename(2)` is atomic and single-winner, and the mtime already read for the age test doubles as the claim's identity — renaming a directory leaves its own mtime alone and touches only the parent's. An invocation that carries aside something other than what it judged declines to claim.

**The prescribed plain rename does not close this on its own**, which is the one place this revision departs from the review. The second invocation's rename is unconditional once it has formed its staleness verdict, so it can still carry off a fresh claim taken in the meantime, and both spawn — the same shape as the reported defect, narrow enough to measure as 0/12 and still be present. The CAS is what makes it single-winner rather than usually-single-winner.

A mismatched claim is **discarded rather than restored**. Restoring is the tidier-looking choice and is worse: the owner may release while we hold the directory, and a restore after that strands it for the full staleness window carrying a fresh mtime — so every later invocation declines to sweep and no canvas starts.

### Minors

| Finding | Disposition |
|---|---|
| Coverage gap — no concurrency-against-stale-claim case | **Fixed.** New case, deterministic (below) |
| Winner spawns without re-probing the port | **Fixed.** Re-probed once the claim is held; also the backstop that makes discarding a mismatched claim safe |
| `LOCK_STALE_SECS=60` vs the 15s timeout | **Deferred.** It is availability, not correctness — AC5 holds either way — and the owner-pid fix makes the claim directory non-empty, which breaks the `rmdir`-only release this hook deliberately chose so a mis-derived path can never delete a tree. Recorded for the #1422 tail |
| Diff size | **Not actioned.** The hook and the suite that proves it cannot land separately |

### On the new case being deterministic

Two attempts are worth recording, because the first would have shipped as a guard that guards nothing:

1. **Ten unstaggered barriered trials reported green against the broken sweep.** The breaking interleaving is a few syscalls wide. A case that catches its defect at some rate below 1 reports green most of the time.
2. **A FIFO on stdin is not a barrier.** The writer's open unblocks whichever reader arrived first, its close sends EOF, and a reader that opens after that close waits for a writer that never returns — the case hangs rather than races. Verified in isolation before being discarded.

What ships instead releases both invocations off a gate file they spin on, then staggers them with a stubbed `date` — the hook's only call to it, sitting exactly between reading the aged claim's mtime and acting on that reading. The interleaving is constructed, not waited for.

### Verification

25 checks green in the suite; `run-plugin-tests.py` rc=0 for cogni-visual (3) and cogni-portfolio (6); both hook copies `cmp`-identical and `bash -n` clean; `check-result-line-plainness.py`, `check-breadcrumbs.py`, `check-version-bump.py`, `check-plugin-inventory.py` all rc=0. Recipe M7 replayed to `guard_verified`, and M1 re-replayed after the merge from `main` to confirm the harness still resolves. No version bump.

Reviewer's attention is most useful on the CAS: whether the mtime is an acceptable identity token, and whether discarding a mismatched claim rather than restoring it is the right call.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1423
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-validator

**Summary:** All three perspectives (service-review's oversized-diff flag, contract-grading-reviewer, plugin-validator) report pass; the three residual findings are all minor, category=preference, and merger-added (advisory_addition), so none gate — PR is ready to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-validator | ✅ pass |

### Findings (3)

- **[minor]** (advisory-addition) `PR diff` — Diff is 791 changed lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier. Note 516 of those lines are the new test suite, and the two hook copies are byte-identical duplicates of one 131-line change, so the genuinely novel review surface is closer to 270 lines.

- **[minor]** (advisory-addition) `cogni-portfolio/tests/` — cogni-portfolio ships a hook whose correctness contract is pinned by a suite living in a sibling plugin (cogni-visual/tests/test-excalidraw-canvas-lock.sh). Running `run-plugin-tests.py --filter cogni-portfolio` passes 6/6 while structurally unable to observe a divergence introduced on the portfolio copy, and cogni-portfolio/CLAUDE.md carries no pointer to the guard. Contract item 15 (one CI-discovered suite) is satisfied and repo CI runs the full sweep, which is the gate that actually blocks a merge; the twin-suite alternative was considered and rejected in the PR body on blast-radius grounds.
  - *Suggested fix:* Optional, out of scope for this PR: add a one-line pointer in cogni-portfolio/CLAUDE.md naming the sibling suite.

- **[minor]** (advisory-addition) `cogni-visual/hooks/ensure-excalidraw-canvas.sh:152-158` — A SIGKILL landing in the few-syscall window between the sweep's `mv` and its `rmdir` leaves an orphan `canvas-start.lock.stale.<pid>` directory. Not the deadlock class the stale reclaim exists to fix — the orphan sits on a name no code path consults, the claim path only reads canvas-start.lock, and a later invocation reusing the pid renames cleanly onto an empty directory — so no user-observable fault, only slow litter in EXCALIDRAW_DIR. Contract item 12 concerns the claim itself and is satisfied.

### Positive observations

- All 17 committed contract items satisfied with file:line evidence, verified by execution — full suite 25/25 green under both /bin/bash 3.2.57 and homebrew bash 5.3.9. (cogni-service:contract-grading-reviewer)
- Both mutation recipes (M1 mkdir-p rewrite, M7 CAS-identity inversion) replayed and returned guard_verified; concurrency determinism holds at rate 1.0 in both directions (mutant red 10/10, clean green 10/10) — not a lucky-run guard. (cogni-service:contract-grading-reviewer)
- Contested-stale-claim case constructs its interleaving deterministically via a gate file plus a stubbed `date`, explicitly rejecting a FIFO race barrier; no case greps a localized English tool message. (cogni-service:contract-grading-reviewer)
- Both hook copies confirmed byte-identical (cmp -s, md5 ceae1abe0af8a9675b924b3fb0d2ee3e) and independently re-verified by the merger; no version fields moved; check-breadcrumbs.py rc=0; diff stays entirely inside the 4-file scope fence. (plugin-validator)
- Independent merger re-verification confirms exit-0 on all four exit statements, single nohup/PID_FILE/seq-loop instances, exactly one EXIT/INT/TERM trap, and an independent full suite run returning 'all checks passed' (25 checks) leaving the worktree clean. (gatekeeper direct confirmation)

### Recommendation

---

PR #1428, opened 2026-08-17
cogni-portfolio: append-claim.sh stale-lock sweep no longer aborts on Linux

### LINKED ISSUE #1424: cogni-portfolio: append-claim.sh stale-lock sweep aborts on Linux — same BSD-first stat chain, unfloored and nested in $(( ))
## Problem

`cogni-portfolio/scripts/append-claim.sh:35` reads a lock directory's mtime with the same BSD-first `stat` chain that was just proven unreachable-on-GNU in the two `ensure-excalidraw-canvas.sh` hooks:

```bash
lock_age=$(( $(date +%s) - $(stat -f %m "$LOCK_DIR" 2>/dev/null || stat -c %Y "$LOCK_DIR" 2>/dev/null || echo 0) ))
```

On GNU coreutils `-f` means `--file-system`, so the first call reports no mtime **yet still exits 0** — the `||` never reaches the GNU form, and a non-numeric value flows into the arithmetic.

This copy is worse than the hook's pre-fix form in two ways:

1. **It is the exact nesting the hook's own comment forbids.** The hook resolves the reading into a variable first, because a bad command substitution nested directly inside `$(( ))` is a hard bash error. This site substitutes inline, so the arithmetic raises a syntax error. Under this file's `set -euo pipefail` (line 9) the script exits 1 rather than degrading.
2. **It breaks the script-output contract.** Every other exit path here emits a `{"success":…,"data":…,"error":…}` envelope (lines 15, 41). This one emits a raw bash arithmetic error on stderr and no envelope, so a caller parsing stdout gets nothing to parse.

Net effect on Linux: the stale-lock recovery path in cogni-portfolio's claims writer is dead, and the failure surfaces as an unparseable non-zero exit.

## Impact

Reachability matches the hook's profile — a contended lock plus an existing claim directory. The callers are the parallel research agents that produce exactly that contention:

- `cogni-portfolio/agents/competitor-researcher.md:130`
- `cogni-portfolio/agents/proposition-generator.md:176`
- `cogni-portfolio/agents/customer-researcher.md:96`

No test suite covers it — `cogni-portfolio/tests/` has no claims-lock suite.

## Proposed change

Apply the same two-part fix that landed on the hooks:

1. Resolve the reading into a variable before the arithmetic, and try the GNU form first (`stat -c %Y … || stat -f %m … || echo 0`) — BSD has no `-c`, so that direction errors cleanly and does fall through.
2. Floor on shape, not emptiness: `[[ "$lock_mtime" =~ ^[0-9]+$ ]] || lock_mtime=0`. An emptiness test structurally cannot catch this class, because the failure is a successful *wrong* answer.

Confirm the failure path still emits the JSON envelope the rest of the script guarantees.

## Acceptance criteria

1. `append-claim.sh` resolves the mtime into a variable before any arithmetic, with the GNU form attempted first.
2. A non-numeric mtime reading cannot reach the arithmetic; it degrades to `0` (claim reads as ancient and is swept).
3. Every failure path in the touched region still emits the `{"success","data","error"}` envelope.
4. A test suite under `cogni-portfolio/tests/` pins the sweep against a stubbed `stat` that answers non-numerically, and carries an executed mutation recipe returning `guard_verified`.

## Scope note

This is the third and last copy of the pattern in the repo; the two hook copies were fixed on the PR that surfaced it. Verified by repo-wide scan: no other `stat -f %m` site remains BSD-first.

claude-cogni-work:
Closes #1424.

## Summary

- `cogni-portfolio/scripts/append-claim.sh` — lift the lock-dir mtime read out of `$(( ))` into `lock_mtime=`, try the GNU form `stat -c %Y` **before** the BSD form `stat -f %m`, floor on numeric *shape* (`[[ "$lock_mtime" =~ ^[0-9]+$ ]] || lock_mtime=0`) rather than on emptiness, and lift `date +%s` into `now=` so no `$(` remains lexically inside the arithmetic.
- New suite `cogni-portfolio/tests/test-append-claim-lock.sh` — pins the swept path, its not-swept negative twin, and the absence of an arithmetic syntax error on stderr.

On GNU coreutils `-f` means `--file-system`, so `stat -f %m` yields no mtime yet still **exits 0** — the `||` never reaches the GNU form and a non-numeric value flows into the arithmetic. Nested directly inside `$(( ))` that is a hard bash error, and under `set -euo pipefail` (line 9) it killed the script instead of degrading. Net effect on Linux: the stale-lock recovery path in the claims writer was dead.

## Why the red case asserts on stderr, and by shape rather than by text

Verified empirically on **both** bash 3.2.57 and bash 5.3.9: when the arithmetic aborts **inside a `while` loop**, bash abandons the rest of the loop body *and the loop itself*, then resumes after `done`. The unfixed script therefore installs the release trap, appends the claim, and **exits 0 with `{"status": "appended"}`** while never having held the lock — and its EXIT trap then removes a live peer's lock directory it never owned. There is no version split here: the damage is the same shape on every bash tested.

Exit code and stdout are consequently *identical* before and after the fix, so a red case asserting only those two would be vacuous and would stay green under the mutation recipe below. Only stderr discriminates.

The discriminator is stderr **emptiness**, not a text match. An earlier revision of this PR grepped bash's own `syntax error: operand expected` string, and that was vacuous: bash gettext-localizes the message, so on a host whose default locale is not English it reads e.g. `Arithmetischer Syntaxfehler: Operand erwartet`, the grep matched nothing, and the suite passed **green against the unfixed script** — guarding nothing anywhere except an English-locale CI. The fixed script writes exactly 0 bytes to stderr on this path, so an emptiness test is locale-independent and strictly stronger: it catches every diagnostic rather than one translation of one of them.

The general rule the suite header now records: assert a **foreign** tool's output by shape (exit status, stream emptiness, numeric form), and only our **own** scripts' literals by text — which is why case 2 may legitimately grep `Could not acquire lock`, a string `append-claim.sh` emits itself.

The stub's value must likewise stay **non-identifier-shaped**, because `$(( now - lock_mtime ))` reads a *variable* and bash evaluates an identifier-shaped value recursively:

| Stub value | Result under `$(( now - lock_mtime ))` |
|---|---|
| `%m` (used) | hard `syntax error: operand expected` — immune to both hatches below |
| `bogus` | aborts only because `set -u` happens to be on |
| `MAX_WAIT` (names a variable the script sets) | evaluates **numerically** to a valid age — never aborts at all |

The suite carries this reasoning inline so a later edit cannot quietly defang the case.

## Why the mtime read is inlined rather than delegated

`cogni-workspace/bash/portability-utils.sh` has a platform-dispatched `portable_file_mtime`, and delegating to it looks like the higher-altitude fix. It is not — it would be a bug. That helper opens with `if [[ ! -f "$file" ]]; then echo "0"; return 1; fi`, and `LOCK_DIR` is a **directory**: `-f` is false for it, so the helper would answer `0` for every lock it was ever asked about. Every contended lock would then read as ancient and get swept out from under a live peer after 3s — precisely the mutual-exclusion regression case 2 exists to pin. Secondarily, its `return 1` inside a `lock_mtime=$(…)` assignment would abort the script under `set -euo pipefail`.

Worth recording for whoever touches that file next: its own unknown-platform `*)` arm repeats the identical BSD-first chain this PR fixes, with no shape floor. It is unreachable on a detected Linux host (the `darwin` and `linux|windows` arms dispatch correctly), so nothing is owed here — but if that arm is ever fixed, the inline chain in this PR is the reference implementation to copy, not the other way round.

## Scope

Two paths changed: `cogni-portfolio/scripts/append-claim.sh` and one new file under `cogni-portfolio/tests/`. Sweep semantics are preserved verbatim — still `[ "$lock_age" -gt 60 ]`, still `rmdir "$LOCK_DIR" 2>/dev/null || true`, still `continue`; `set -euo pipefail` stays, with no `set +e` and no blanket `|| true` around the arithmetic. `MAX_WAIT=30`, `sleep 0.1`, the line-41 timeout object and the line-47 trap are byte-identical. The timeout path still emits exactly `{"error": "Could not acquire lock on claims.json after 3s"}` on stderr and exits 1, and no new non-zero exit path is introduced. The caller contract is unchanged (two positional args, usage `{"error": …}` + exit 1 on either empty, success keys `status` / `claim_id` / `total`).

Deliberately **not** in this PR:

- **No envelope migration.** `cogni-portfolio/CLAUDE.md` states scripts emit `{"success", "data", "error"}`; `append-claim.sh` has never done so (its failure sites emit `{"error": …}` only, its success path `{status, claim_id, total}`). Migrating the file would change the parse contract for every caller with no test covering them — separable scope, so the touched region keeps the file's existing failure-path shape.
- **No sibling `stat` fix, and no merge ordering against another PR.** A repo-wide scan of current `origin/main` for `stat -f %m` / `stat -c %Y` shows `append-claim.sh` is the only live BSD-first chain feeding arithmetic. Both `cogni-portfolio/hooks/ensure-excalidraw-canvas.sh` and `cogni-visual/hooks/ensure-excalidraw-canvas.sh` (213 lines each since that work landed) now carry the identical GNU-first chain plus shape floor at lines 122 and 155 — they corroborate this pattern rather than compete with it, and nothing is owed there. `cogni-workspace/bash/portability-utils.sh` retains BSD-first arms at lines 80 and 108, but they feed no arithmetic and are unreachable on a detected platform, so no code is owed there either.
- **No version bump.** No `plugin.json` / `marketplace.json` touched and no `version` string altered — the post-merge workflow owns the bump.
- **No caller edited.** For the record, **seven** files under `cogni-portfolio/agents/` invoke this script, not three as the issue-time contract's item 14 states: `competitor-researcher`, `customer-researcher`, `feature-deep-diver`, `market-researcher`, `proposition-deep-diver`, `proposition-generator`, `quality-enricher`. None parses its stdout, stderr, or exit code, so the risk item 14 guards is unchanged and none needs a change. Noted so a literal grader is not misled by the count.

## Test plan

All of the following were run on this branch:

- `bash cogni-portfolio/tests/test-append-claim-lock.sh` → both cases `ok:`, `All tests passed.` (~11s; each case pays the hardcoded 30 × 0.1s retry loop once, against CI's 300s per-suite budget)
- `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` → `total: 7, passed: 7, failed: 0`
- `python3 scripts/check-result-line-plainness.py --root .` → `success: true`, `violations: []` (120 files scanned, 89 emitter definitions inspected)
- `python3 scripts/check-breadcrumbs.py` → `success: true`, 0 findings
- `python3 scripts/check-version-bump.py` → `0 touched, 0 violations vs origin/main`

## Mutation recipe

```
  --root . \
  --expr 's/lock_mtime=0/lock_mtime_unused=0/' \
  --case stale-lock-swept-on-non-numeric-stat
```

**Executed on this branch — verdict `guard_verified`** (`mutated_result: red`, `restored_result: green`).

The expression redirects the shape floor's assignment to a dead variable, so the non-numeric reading survives into the arithmetic. `lock_mtime=0` occurs exactly once in the fixed file (the read line is `lock_mtime=$(stat …`, which does not contain it), so the substitution can never be a no-op, and it carries no shell metacharacter.

**Only one arm is recorded, deliberately.** An arm mutating the GNU-first ordering (`s/stat -c %Y/stat -f %m/`) would report a vacuous guard: the stub answers unconditionally and ignores its flags by design — which is what keeps the suite correct under both orderings — so no case here can discriminate `stat -c %Y` from `stat -f %m`.

## Plan record

[Plan record on the issue](https://github.com/cogni-work/insight-wave/issues/1424#issuecomment-5313291377) — carries the plan-phase provenance (2 Explore agents, 2 divergent Plan agents, the pick rationale, and the refine pass that caught the vacuous red case).


claude-cogni-work:

Author instance is done. This PR is draft with `svc:needs-review`; the merger instance owns the independent review and the merge decision. No review verdict was posted here and no `svc:review-*` label was applied — the merger's review must not be skippable.

### Handoff preflight — clean

`rc=0`, 3 passed / 0 failed / 4 skipped / 0 advisory.

- **passed:** `run-tests` (suites green for cogni-portfolio) · `version-bump` (no version-line touch) · `mutation-recipe` (recorded recipe runnable as written; case `stale-lock-swept-on-non-numeric-stat` found in the suite)
- **skipped, all `not_applicable`:** `frontmatter` and `breadcrumbs` (no touched `SKILL.md` or `agents/*.md`) · `skill-length` (no touched `SKILL.md`) · `agent-length` (no touched `agents/*.md`)

A note for whoever re-runs this: the preflight first reported 2 violations, and **both were artifacts of the `--root` argument, not the branch**. Pointed at the shared clone it graded a tree checked out on an unrelated older branch — one that carries a single cogni-portfolio suite (hence `below_min_suites: discovered 1 against a floor of 6`) and does not contain `scripts/check-version-bump.py` at all. Re-run against a checkout of this PR's head, every gating instrument passes. Worth knowing before treating a `below_min_suites` or `version-bump` violation here as real.

### Author self-grade against the issue-time contract — pass

Graded by `contract-grading-reviewer` against the [acceptance contract](https://github.com/cogni-work/insight-wave/issues/1424#issuecomment-5313291377) derived at `2026-08-17T07:05:50Z` with `--origin triage --target issue` — before any branch was cut, so it is diff-blind by chronology.

**15 items graded · 15 satisfied · 0 violated · 0 unclear · no residual.** Advisory only.

Two items were graded on substance rather than their literal text, and the merger should know why:

- **Item 14 miscounts the call sites.** It names "three call sites"; there are **seven** files under `cogni-portfolio/agents/` that invoke `append-claim.sh` — `competitor-researcher:130`, `customer-researcher:96`, `feature-deep-diver:275`, `market-researcher:125`, `proposition-deep-diver:407`, `proposition-generator:176`, `quality-enricher:268`. None parses the script's stdout, stderr, or exit code, and none is edited here, so the risk the item guards is unchanged. Graded satisfied on substance; the miscount is recorded rather than quietly inherited.
- **Item 13's `--expr` requirement is disjunctive, and only one arm is recorded.** The recorded arm intersects the item-6 shape floor. A second arm mutating the GNU-first ordering is declined deliberately: the PATH stub answers unconditionally and ignores its flags by design — which is exactly what keeps the suite valid under both the pre-fix BSD-first and post-fix GNU-first orderings — so no case in this suite can discriminate `stat -c %Y` from `stat -f %m`, and such an arm could only ever return a vacuous guard. Recording it would have failed the contract's own watchlist bar, not met it.

### The one thing worth reviewing closely

The red case's third assertion — that stderr carries no `syntax error` — is **load-bearing, and the case pins nothing without it**. On bash 3.2 an arithmetic operand error *inside a `while` loop* abandons the loop body and the loop, then resumes after `done`. The unfixed script therefore installs the release trap, appends the claim, and exits **0 with `{"status": "appended"}`**, never having held the lock. Exit code and stdout are consequently identical before and after the fix. The mutation recipe was executed and returned `guard_verified` (`mutated_result: red`, `restored_result: green`) — that verdict depends on this assertion, so deleting it as belt-and-braces would silently make the suite vacuous. The suite carries the reasoning inline for that reason.

Related: the stub's value must stay non-identifier-shaped. `$(( now - lock_mtime ))` reads a *variable*, and bash evaluates an identifier-shaped value recursively — `bogus` aborts only because `set -u` happens to be on, and a value naming a variable the script already sets (`MAX_WAIT`) resolves to a number and never aborts at all. `%m` is immune to both.

### Verification run on this branch

| Check | Result |
|---|---|
| `bash cogni-portfolio/tests/test-append-claim-lock.sh` | both cases `ok:`, `All tests passed.` (~11s) |
| `run-plugin-tests.py --filter cogni-portfolio` | `total: 7, passed: 7, failed: 0` |
| `check-result-line-plainness.py --root .` | `success: true`, `violations: []` (120 files, 89 emitters) |
| `check-breadcrumbs.py` | `success: true`, 0 findings |
| `check-version-bump.py` | 0 touched, 0 violations vs `origin/main` |
| mutation recipe | `verdict: guard_verified` |

Issue #1424 moved to `svc:resolved`; the PR's `Closes #1424` closes it on merge.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1428

**Summary:** Both perspectives converge on the same major defect via two independent lines of evidence — the lock-sweep test's sole case-1 discriminator greps bash's own arithmetic-error text, which is localized, so under this host's default (non-English) locale the suite passes GREEN against the unfixed append-claim.sh and the recorded mutation recipe returns vacuous_guard rather than the PR-claimed guard_verified — alongside one un-gating preference-only minor about a stale bash-version claim in a code comment.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[major]** (auto-fixable) `cogni-portfolio/tests/test-append-claim-lock.sh:121` — VACUOUS GUARD UNDER A NON-ENGLISH/DEFAULT LOCALE (contract item 13). Case 1's sole discriminating assertion — `elif grep -q 'syntax error' "$TMPROOT/case1.err"` at test-append-claim-lock.sh:121 — greps bash's own arithmetic-error text, which gettext localizes. Two independent lines of evidence converge on the same defect: (1) the exact recorded mutation recipe (--expr 's/lock_mtime=0/lock_mtime_unused=0/' --case stale-lock-swept-on-non-numeric-stat) returns data.mutated_result="green", data.verdict="vacuous_guard" (reproduced twice) under the ambient locale, contradicting the PR body's claimed "guard_verified"; (2) restoring the pre-fix append-claim.sh and re-running the suite under the ambient locale emits "ok: stale-lock-swept-on-non-numeric-stat ... All tests passed" and exits 0 — the suite passes GREEN against the unfixed script — because bash on this host emits a localized diagnostic (e.g. "Arithmetischer Syntaxfehler: Operand erwartet") that the English-only grep never matches. No LANG/LC_* is exported; this is default local behaviour (macOS gettext falls back to the CoreFoundation user locale), not an exotic configuration, so only Linux CI would currently catch a regression to the guarded shape. Re-running the identical recipe/suite with LC_ALL=C flips mutated_result to "red" and the suite to FAIL, confirming the guard logic itself is sound and only the discriminator's locale-dependence is broken.
  - *Suggested fix:* Replace the localized-text grep with a locale-independent observable: the fixed script writes ZERO bytes to stderr under case-1 conditions (measured: 0 bytes fixed vs 191 bytes pre-fix), so `elif [ -s "$TMPROOT/case1.err" ]` is both locale-proof and strictly stronger than the substring grep — it catches any diagnostic, not just an arithmetic one. Alternatively (or additionally), pin `export LC_ALL=C LANG=C` at the top of test-append-claim-lock.sh so bash's arithmetic-error text is locale-independent and the existing grep works as recorded. Either way, re-run the mutation recipe afterwards and re-record the verdict as guard_verified from a default locale before merge.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-portfolio/scripts/append-claim.sh:44` — The rationale comment's parenthetical "(On bash >= 4 it aborts outright.)" is factually wrong, and the suite header repeats the same split by scoping the resume-after-done damage to bash 3.2. Measured against the pre-fix script with the suite's own stub: bash 5.3.9 behaves exactly like bash 3.2.57 — rc=0, stdout {"status": "appended"}, the claim written, and the pre-created peer lock removed by the EXIT trap it never owned. The substance of the comment is right; only the version split is wrong, and it understates the blast radius by implying modern bash fails safe.
  - *Suggested fix:* Drop the version split: state that the arithmetic error abandons the loop body and the loop on every bash tested (3.2 through 5.3), so the script resumes after done, appends the claim without ever holding the lock, and its EXIT trap rmdirs a live peer's lock. Mirror the correction in the suite header.

### Positive observations

- Items 1-12 and 14-15 all satisfied with cited evidence: scope boundary holds at two files, no version string touched, no command substitution nested in arithmetic, GNU-first stat chain, numeric-shape floor (not an emptiness test), {"error": ...} failure shape and sweep semantics byte-identical, CI runnability and result-line plainness clean, breadcrumb gate clean, caller contract intact. (cogni-service:contract-grading-reviewer)
- Item 10 confirmed genuinely discriminating: independently mutating the fixed file to lock_mtime=0 unconditionally made case 2 correctly go red while case 1 stayed green, proving the negative twin is load-bearing. (cogni-service:contract-grading-reviewer)
- Item 1: new suite discovered by scripts/run-plugin-tests.py as suite 1/7 and green; all 7 portfolio suites pass. (plugin-dev:plugin-validator)
- Item 3: check-version-bump.py clean=true, 0 violations; plugin.json 0.9.76 and marketplace.json mirror agree. (plugin-dev:plugin-validator)
- Item 5 behaviourally confirmed on BSD: real macOS /usr/bin/stat rejects -c with 'illegal option', chain falls through, stat -f %m returns the correct mtime cross-checked against python os.stat — the reorder costs nothing on BSD while unblocking GNU. (plugin-dev:plugin-validator)
- Item 7: the exact timeout error string and exit 1 preserved; no three-key envelope migration — the falsified AC3 framing was correctly not chased. (plugin-dev:plugin-validator)
- Item 11: mktemp root cleaned on normal exit AND after mid-run SIGINT; zero temp dirs left behind. (plugin-dev:plugin-validator)
- Item 12: byte-level scan finds no raw ESC, no \033, no \e[; the two tput grep hits are the substring inside the words 'output streams'/'output line', not invocations. (plugin-dev:plugin-validator)

### Recommendation

---

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1428
**Perspectives consulted:** plugin-validator

**Summary:** The sole routed perspective (plugin-validator) passed with no findings, verifying the corrected locale/build prose, the mutation recipe, and the PR-body claims by direct measurement with zero regressions. Two earlier rounds at 15e475a6 and 24a1312c also passed; every finding they raised is either fixed here or filed as a tracked follow-up (#1430, #1431, #1432, #1433, #1434).

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-validator | ✅ pass |

### Positive observations

- Correction is accurate and precisely scoped — the diff changes only comment lines; the stderr-emptiness assertion is untouched, so the round carries zero behavioural risk. (plugin-validator)
- The 'pinning the locale would not have saved it' rationale survives adversarial testing: with LC_ALL=C, en_US.UTF-8 and de_DE.UTF-8 across both builds, the English wording still differs between bash 3.2.57 and 5.3.9, so a grep authored against either build's message fails on the other even with locale fixed. (plugin-validator)
- Non-vacuity proven across the full axis matrix: pre-fix restored from merge base is red under three genuinely different diagnostic texts, confirming the guard keys on stream emptiness rather than any one translation or build wording. (plugin-validator)
- The corrected PR-body paragraph replaced a false claim with one verifiable line-by-line, and its qualifiers ('feed no arithmetic', 'unreachable on a detected platform') are independently confirmable — the paragraph earns its conclusion instead of asserting it. (plugin-validator)
- Mutation recipe replays exactly as written from a clean copy and returns guard_verified with the file restored byte-identical, so it is reproducible by a later reviewer rather than a one-time claim. (plugin-validator)
- No regression anywhere reachable: suite green at head under both bash 3.2.57 and 5.3.9; run-plugin-tests --filter cogni-portfolio reports 7/7 with the new suite discovered; breadcrumb guard 0; result-line plainness 0 violations across 121 files and 91 emitter definitions. (plugin-validator)
- The fix still satisfies the acceptance criteria at head — GNU-first chain resolved into a variable before the arithmetic, with a numeric-shape floor rather than an emptiness test, which is the right guard for a successful-but-wrong answer. A repo-wide scan in both 'stat -f %m' and 'stat -f%m' spellings confirms no arithmetic-fed stat chain remains anywhere in the tree. (plugin-validator)
- Failure-path behaviour byte-preserved: both error strings and both exit-1 sites are identical between merge base and head, and no new non-zero exit path is introduced — so the envelope question deferred to #1434 is genuinely untouched scope rather than a silent change. (plugin-validator)

### Recommendation

---

sdh07:
## Revision posted

Head is now `24b9eae8`. This addresses both findings from the [09:23Z verdict](https://github.com/cogni-work/insight-wave/pull/1428#issuecomment-5314202337), and the branch has been merged up to `main` (through `73ab3496`), so CI runs against the tree this lands in.

### The major — reproduced first, then fixed

The vacuous guard was **reproduced here before being changed**, rather than taken from the verdict. Running the committed suite against the *pre-fix* `append-claim.sh`:

| | ambient locale | `LC_ALL=C` |
|---|---|---|
| **before** this revision | `All tests passed` — **green against the unfixed script** | red |
| **after** this revision | **red** | red |

The discriminator is now stderr **emptiness** rather than a grep for bash's `syntax error`. The fixed script writes exactly 0 bytes there, so the assertion is locale-independent and strictly stronger than the single English message it replaces — it catches every diagnostic, not one translation of one of them.

The recorded mutation recipe now replays to `guard_verified` **from a default locale**, which is what the verdict asked for; it previously returned `vacuous_guard`.

One correction to my own reasoning while fixing this: the defect is not purely a locale story. Measured on this host under one unchanged locale, **bash 3.2.57 emits the English message and bash 5.3.9 the German one** — Apple's build ships no translations, homebrew's does. So pinning `LC_ALL=C` would *not* have rescued the text grep; the build axis survives it. That is recorded in the suite header, and it is the stronger argument for asserting shape over text.

### The minor

Dropped the false `(On bash >= 4 it aborts outright.)` split from both rationale comments. Verified with isolated project directories: **3.2.57 and 5.3.9 behave identically** — rc=0, claim appended, and the pre-created peer's lock removed by the EXIT trap it never owned. The substance was right; only the version split was wrong, and it understated the blast radius.

### Two further corrections, from this round's own review gate

Neither was in the original verdict; both were factually wrong statements of the same class, so they are fixed rather than filed.

- **The suite header claimed no case could discriminate `stat -c %Y` from `stat -f %m`.** False. Measured with a flag-aware stub against the *fixed* script with the floor present in both arms: GNU-first exits 1 and the live peer's lock **survives**; BSD-first exits 0, appends, and **sweeps** it. The ordering is load-bearing on its own merits, and telling a maintainer it cannot be tested is worse than saying nothing. The note now scopes the vacuity to the two existing cases, which is the accurate statement. The missing coverage is filed as #1433.
- **The PR body's sibling-scan paragraph was stale** — "88 lines each, zero matches" is false against current `main`, where both `ensure-excalidraw-canvas.sh` copies are 213 lines carrying this same GNU-first chain. The conclusion still holds and the body now says so accurately.

### Verification

Suite green at head under both bash builds; `run-plugin-tests.py` 121/121 green repo-wide and 7/7 for cogni-portfolio; mutation recipe replays to `guard_verified`; `check-result-line-plainness.py`, `check-breadcrumbs.py`, `check-version-bump.py`, `check-plugin-inventory.py`, `check-workflow-yaml.py` all rc=0. No version bump — the post-merge workflow owns it.

One disclosure: the token-scope preflight does **not** pass under `--strict` on this runner. Its `gho_` token's scopes are fixed by the GitHub CLI OAuth app and cannot be narrowed, so the push proceeded under the documented `--allow-overscoped` opt-out. That is a deliberate waiver, not a green check.

### Follow-up issues

Filed rather than absorbed, to keep this diff a single reviewable correctness fix:

- #1430 — record the assert-foreign-tool-output-by-shape rule in `CLAUDE.md` (the class this PR is an instance of)
- #1431 — the lock suite costs 7.3s wall on every CI sweep, 94% of it `sleep`
- #1432 — hoist the `stat`-form probe out of the acquire loop; bound stale-block re-entry
- #1433 — no case pins the GNU-first ordering; a BSD-first revert leaves the suite green
- #1434 — `append-claim.sh` predates the `{success, data, error}` convention (issue AC item 3, pre-existing)

Reviewer attention is most useful on whether stderr-emptiness is the right contract to pin for this script, and on whether #1433's coverage gap should have blocked rather than deferred.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1428

**Summary:** Both perspectives passed: the committed 15-item review contract is satisfied by execution (mutation recipe replayed guard_verified, new suite 2/2 green) and structural plugin validation is clean; the sole finding (a preference on an unrelated suite's execute bit) carries an explicit opt_out waiver, so nothing is un-waived — safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-portfolio/tests/test-bash-grant.sh` — This suite lacks the user-execute bit while its six sibling suites have it. Pre-existing on the base branch, untouched by this PR, and user-unobservable: the discovery runner invokes suites as `bash <path>`, which ignores the mode bit, so it still runs in CI.
  - *Justification:* Out of scope for PR #1428, pre-existing on the base branch, and user-unobservable because the runner contract is `bash <path>` rather than direct execution.

### Positive observations

- Contract item 1: exactly one new suite at cogni-portfolio/tests/test-append-claim-lock.sh, one segment below tests/, matching the */tests/*.sh CI discovery glob.
- Contract item 2: only cogni-portfolio/scripts/append-claim.sh and cogni-portfolio/tests/test-append-claim-lock.sh changed — no hooks, no ensure-excalidraw-canvas.sh, no agents/*.md.
- Contract item 3: no plugin.json or marketplace.json in the changed set; no version string altered (post-merge bump policy respected).
- Contract item 4: lock_mtime and now are assigned on their own lines before lock_age=$(( now - lock_mtime )); no $( nested inside $(( )).
- Contract item 5: the GNU stat -c %Y form is attempted first.
- Contract item 6: the guard is the digit-class shape floor [[ "$lock_mtime" =~ ^[0-9]+$ ]] || lock_mtime=0, not an emptiness test.
- Contract item 7: the timeout path still emits {"error": "Could not acquire lock on claims.json after 3s"} to stderr with exit 1; no three-key envelope migration.
- Contract item 8: the -gt 60 guard, rmdir || true, continue, and set -euo pipefail are all retained; no set +e added.
- Contract items 9 and 10 verified BY EXECUTION: the suite runs 2/2 green — the red case (non-numeric stub floored, ancient lock swept) and its negative twin (numeric stub, live lock left intact, script times out).
- Contract item 11: exits 0, no curl/wget/gh, mktemp root removed by an EXIT trap, no $1-dependent control flow.
- Contract item 12: no escape literal anywhere; emitters are plain printf bodies; the result-line plainness guard reports no violations.
- Contract item 13 verified BY EXECUTION: the recorded recipe was replayed against the shared cogni-service harness and returned data.verdict == guard_verified (mutated red, restored green).
- Contract item 14: two positional args, the usage error, and success keys status/claim_id/total are all unchanged; no caller edited.
- Contract item 15: no issue/PR ref introduced; the breadcrumb guard reports 0 violations.
- Defect reproduced against the real pre-fix base: running the new suite against origin/main's append-claim.sh turns it RED on exactly the stderr-emptiness assertion. The diagnostic that surfaces on this host is the German-localized 'Arithmetischer Syntaxfehler: Operand erwartet' — confirming that the earlier revision's grep for bash's English wording would have been vacuous, and that asserting stream EMPTINESS is the correct, locale- and version-independent discriminator.
- Structure: plugin.json parses with every required field and mirrors the marketplace.json entry; all 21 skills and 20 agents are well-formed; hooks.json is valid with a resolvable executable target; no .mcp.json is shipped; append-claim.sh stays registered and reachable from its eight call sites.
- Deterministic pre-gates clean: check-frontmatter.sh count 0, check-version-bump.sh status=ok with 0 violations across 9 plugins. CI is 7/7 green pinned to head 24b9eae8, completed after main's latest commit (not stale).

### Recommendation

---

Issue #1432, opened 2026-08-17
cogni-portfolio: hoist the append-claim stat-form probe out of the acquire loop; bound stale-block re-entry


sdh07:
## Context

Found while revising #1428. Recorded rather than absorbed: the change would alter the very `stat` chain that PR's review confirmed as correct, and #1428's scope boundary at two files was an explicit accepted contract item. This is a cost observation, not a defect — the sweep is functionally correct after #1428.

## Current behavior

Two costs in the stale-lock block of `cogni-portfolio/scripts/append-claim.sh` (lines 48-55), both on the contended path only.

**1. The GNU-first ordering doubles the `stat` fork count on BSD.** Verified on macOS: `stat -c %Y` exits 1 (`illegal option -- c`), so the `||` falls through to `stat -f %m` — **2 execs where the pre-#1428 BSD-first line cost 1**. This is the correct tradeoff as it stands (BSD-first returned a successful *wrong* answer on GNU, which is the bug #1424 was about), so the ordering must not be reverted. But the form can be resolved once instead of per pass.

With `now=$(date +%s)` on line 50, the block costs 3 process spawns per entry on BSD.

**2. `WAITED` is never reset after the `continue` on line 54**, so once the threshold is crossed the block can re-enter on every subsequent 100ms iteration rather than once.

The original framing of this was "roughly 30 process spawns per second under sustained contention". That overstates it, and the precise condition is worth recording so nobody chases a phantom: after `rmdir "$LOCK_DIR" 2>/dev/null || true; continue`, the `while ! mkdir "$LOCK_DIR"` condition is re-evaluated. In the normal case the `rmdir` succeeded, so `mkdir` now succeeds and the loop **exits**. Sustained re-entry therefore requires either the `rmdir` to keep failing, or a peer to re-take the lock within the same 100ms window. It is a real spin, but a narrow one — not the default contended path.

Note the `rmdir`-fails branch is not hypothetical in this codebase's neighbourhood: any change that puts a file inside a lock directory (an owner-pid marker, say) makes `rmdir` fail permanently and converts this into an indefinite 10Hz spin over three forks. Worth fixing before that combination ever arises rather than after.

## Desired behavior

Keep the ordering — it is the fix — and hoist the probe:

[...]

PR #1444, opened 2026-08-17
cogni-portfolio: cut append-claim lock suite from 7.9s to under 2s

### LINKED ISSUE #1431: cogni-portfolio: the append-claim lock suite costs 7.3s wall on every CI sweep, 94% of it sleep
## Context

Found while revising #1428 (the append-claim lock guard). Recorded rather than absorbed so that PR's diff stays at the two files its review confirmed as its scope boundary.

## Current behavior

`cogni-portfolio/tests/test-append-claim-lock.sh` measures **7.27s wall, 0.44s CPU** — roughly 6% CPU utilisation. About 94% of the suite is `sleep 0.1` inside `append-claim.sh`'s lock-acquire loop, paid twice because the two cases run sequentially and each drives the loop to its hardcoded timeout.

This runs in the `scripts/run-plugin-tests.py` sweep on every PR, so the cost is paid on every CI run of the whole monorepo.

Both cases exist precisely to drive the acquire loop to its timeout, so the wait length is incidental to what they pin, not part of it.

## Desired behavior

Either of the following, or both — they compose:

1. **Parameterise the wait** (roughly 10x): make the acquire ceiling overridable, e.g. `MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"` at `cogni-portfolio/scripts/append-claim.sh:27`, with each test case exporting a small value. Three iterations pin the same behaviour as thirty. The default stays 30, so production timing is untouched.
2. **Run the two cases concurrently** (roughly 2x): they already have separate project directories, separate stub directories and separate stderr files — the file's own comments justify that isolation — so backgrounding both and collecting via `wait` is safe. The shared `failures` counter would need to move to the parent, since an increment inside a subshell is lost.

Option 1 touches production code for a test-only benefit, which is the tradeoff worth deciding deliberately: an env-var escape hatch on a locking primitive is a small surface, but it is a surface. Option 2 keeps the change entirely inside the suite. Whoever picks this up should choose rather than assume.

## Acceptance criteria

- [ ] `bash cogni-portfolio/tests/test-append-claim-lock.sh` completes in under 2s wall on a developer machine. Falsified by a timed run at or above that.
- [ ] Both cases still pin what they pinned: the recorded mutation recipe in the suite header still replays to `guard_verified`, and the case-2 negative twin still goes red when the shape floor is mutated to set the reading unconditionally. Falsified by either replay changing verdict.
- [ ] If option 1 is taken, the default acquire ceiling is unchanged for callers that set no override. Falsified by a caller observing a shorter timeout without opting in.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` exits 0 with 7/7 suites passing.

claude-cogni-work:
Closes #1431.

Plan record: https://github.com/cogni-work/insight-wave/issues/1431#issuecomment-5318949549

## Summary

The lock suite cost **7.9s wall at 6% CPU**, and 94% of that was sleep. Both of its cases exist to drive `append-claim.sh`'s acquire loop to its ceiling, and that ceiling was a hardcoded `MAX_WAIT=30` x `sleep 0.1`. `scripts/run-plugin-tests.py` runs the suite on every PR, so the whole monorepo paid it on every CI sweep.

- `scripts/append-claim.sh`: `MAX_WAIT` becomes `MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"`, with a numeric shape floor that returns a malformed override to the production default instead of making the `-ge` comparison a hard error mid-loop. The floor's regex is deliberately **unquoted** — a quoted right-hand side of `=~` matches literally on bash 3.2, which this repo still targets — and the `|| MAX_WAIT=30` form keeps a non-match from tripping `set -e`.
- `scripts/append-claim.sh`: the timeout diagnostic is now **derived** from the ceiling instead of restating it. Under the default it is byte-identical to today's (`... after 3s`); under an override of 3 it correctly reads `... after 0.3s`.
- `tests/test-append-claim-lock.sh`: both cases set `APPEND_CLAIM_MAX_WAIT=3` as a **per-invocation prefix**, so each pays 0.3s instead of 3s. Not a suite-level export: `run-plugin-tests.py` invokes each suite as a bare `bash <path>`, so each case has to carry its own ceiling.
- `tests/test-append-claim-lock.sh`: case 2's stderr assertion moves to the derived literal `after 0.3s`, which is what pins the coupling.

### Why the message change is not optional polish

Parameterising the ceiling on its own would have silently decoupled the diagnostic from the wait. The old message restated `after 3s` — correct only because `30 x 0.1s` happens to equal 3s — and case 2 grepped that literal. Under an override the script would have announced a wait it no longer honoured while the case stayed green: the same class of defect this suite was written to catch. Deriving the label and asserting the derived form makes the coupling the thing the case pins.

## Verification

| Acceptance criterion | Result |
|---|---|
| Suite completes under 2s wall | **1.10s** (was 7.91s); 1.1s inside the sweep |
| Both cases still pin what they pinned | **all three mutation arms replay to `guard_verified`** |
| Default acquire ceiling unchanged with no override | **3s wait, byte-identical message** — verified against a pre-seeded contended lock |
| `run-plugin-tests.py --filter cogni-portfolio` exits 0 | **7/7 suites passed, exit 0** |

Malformed overrides were checked explicitly: `abc`, empty, `0`, `-5` and a trailing-space value all floor to 30 and render `3`. Label arithmetic was verified on **both** bash builds present on this host (3.2.57 and 5.x): `30 -> 3`, `20 -> 2`, `3 -> 0.3`, `1 -> 0.1`, `45 -> 4.5` — the `%.0` trim does not leave a dangling `2.` or `2.0`.

Repo guards run clean on this branch: breadcrumb guard, result-line plainness guard, workflow-YAML guard, and the version-bump gate (0 plugins touched — the post-merge workflow owns the bump).

## Scope decisions

- **In:** the issue's option 1 (parameterised ceiling), plus the message-derivation fix triage identified and the case-2 grep update it forces.
- **Out:** option 2 (running the two cases concurrently). Triage measured it at ~3.5s standalone — above the 2s bar — because it can only remove one of the two ceilings. With option 1 landed it buys a fraction of a second and would cost the subshell failure-counter rework. This is a **rejected alternative, not deferred work**, so no follow-up issue was filed.
- **Out:** any version bump.
- Untouched: the failure counter, the case sequencing, and lines 33-57 of `append-claim.sh` (the stale-lock sweep, including the `lock_mtime=0` the pre-existing mutation arm anchors on).

## Mutation recipe

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/APPEND_CLAIM_MAX_WAIT:-30/APPEND_CLAIM_MAX_WAIT_UNUSED:-30/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case live-lock-not-swept-on-numeric-stat
```

The expression redirects the override to a dead variable, so the script ignores the ceiling the cases set, waits the production 3s and announces `after 3s` while case 2 asserts the derived `after 0.3s`. `APPEND_CLAIM_MAX_WAIT:-30` occurs exactly once (the header names the variable in prose without the `:-30` default), so the substitution can never be a no-op, and the expression carries no `^`/`$` anchor so it needs no `/m` under the harness's `perl -0pi`. Replayed on this branch: `guard_verified`.

A second arm covers the shape floor's CONDITION — the property case 2 was written for. Arm one below mutates the floor's *assignment* (caught by case 1); this one mutates its *condition* (caught by case 2), which is why both are recorded rather than one standing in for the other:

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/\^\[0-9\]\+\$/^NOMATCH\$/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case live-lock-not-swept-on-numeric-stat
```

Replacing the character class with one that cannot match makes the `||` fire unconditionally, so a live peer's current-epoch reading floors to 0, reads as ancient, gets swept, and the script appends where case 2 demands exit 1. The pattern's `^` and `$` are escaped — matched as the literal characters of the regex text, not as anchors — so it needs no `/m` under the harness's slurping `perl -0pi`, and `^[0-9]+$` occurs exactly once now that the ceiling's own floor reads `^[1-9][0-9]*$`. Replayed on this branch: `guard_verified`.

The pre-existing arm is unchanged and was replayed on this branch as well:

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/lock_mtime=0/lock_mtime_unused=0/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case stale-lock-swept-on-non-numeric-stat
```

Also `guard_verified`.

### DIFF
--- a/cogni-portfolio/scripts/append-claim.sh
+++ b/cogni-portfolio/scripts/append-claim.sh
@@ -5,6 +5,9 @@
 # Creates cogni-claims/ directory and claims.json if they don't exist.
 # Uses mkdir-based locking (portable across macOS and Linux) to prevent
 # race conditions when agents run in parallel.
+# Env: APPEND_CLAIM_MAX_WAIT — lock-acquire ceiling in 0.1s ticks. Unset means
+# the production default of 30 ticks (3s); tests set a small value so they can
+# drive the timeout path in fractions of a second.
 # Exit codes: 0 = success, 1 = error
 set -euo pipefail
 
@@ -24,7 +27,16 @@ LOCK_DIR="$CLAIMS_DIR/.claims.lock"
 mkdir -p "$CLAIMS_DIR/sources" "$CLAIMS_DIR/history"
 
 # Acquire lock using mkdir (atomic on all POSIX systems)
-MAX_WAIT=30
+MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"
+# Floor on numeric shape: a malformed override would otherwise make the -ge
+# comparison below a hard error mid-loop. Regex unquoted on purpose — a quoted
+# right-hand side matches literally on bash 3.2, which this repo still targets.
+# The `||` form keeps a non-match from tripping `set -e`.
+[[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] || MAX_WAIT=30
+# Derive the diagnostic's number from the ceiling rather than restating it, so
+# the message can never claim a wait the loop no longer honours.
+TIMEOUT_LABEL="$(( MAX_WAIT / 10 )).$(( MAX_WAIT % 10 ))"
+TIMEOUT_LABEL="${TIMEOUT_LABEL%.0}"
 WAITED=0
 while ! mkdir "$LOCK_DIR" 2>/dev/null; do
   sleep 0.1
@@ -55,7 +67,7 @@ while ! mkdir "$LOCK_DIR" 2>/dev/null; do
         continue
       fi
     fi
-    echo '{"error": "Could not acquire lock on claims.json after 3s"}' >&2
+    echo "{\"error\": \"Could not acquire lock on claims.json after ${TIMEOUT_LABEL}s\"}" >&2
     exit 1
   fi
 done
--- a/cogni-portfolio/tests/test-append-claim-lock.sh
+++ b/cogni-portfolio/tests/test-append-claim-lock.sh
@@ -16,6 +16,17 @@
 #                                             reading leaves a LIVE peer's lock
 #                                             alone; the script times out instead
 #
+# Both cases drive the acquire loop to its ceiling on purpose, so both set
+#   APPEND_CLAIM_MAX_WAIT=3 as a per-invocation prefix — 0.3s each instead of the
+#   production 3s, which is what keeps the whole suite around a second rather than
+#   the ~7.3s it cost when the ceiling was a hardcoded constant. The prefix is
+#   deliberately NOT a suite-level export: run-plugin-tests.py invokes each suite
+#   as a bare `bash <path>`, so each case must carry its own ceiling.
+#   Case 2's asserted stderr literal is the DERIVED form for that ceiling
+#   (`after 0.3s`), not a fixed string — change the ceiling and the literal moves
+#   with it. That coupling is the point: it is what stops the message from
+#   drifting back into restating a number the loop no longer honours.
+#
 # Usage: bash cogni-portfolio/tests/test-append-claim-lock.sh
 # Exits non-zero on any assertion failure.
 #
@@ -39,10 +50,28 @@
 #   exactly once in the fixed script (the read line is `lock_mtime=$(stat ...`,
 #   which does not contain it), so the substitution can never be a no-op.
 #
-#   Only ONE arm is recorded, and the reason is narrower than it may look. An arm
-#   mutating the GNU-first ordering (`s/stat -c %Y/stat -f %m/`) reports a vacuous
-#   guard against the TWO CASES BELOW, because their stub answers unconditionally
-#   and ignores its flags by design — deliberately, so both stay ordering-agnostic.
+#   Second arm — the ceiling-to-message coupling:
+#   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
+#     --root . \
+#     --file cogni-portfolio/scripts/append-claim.sh \
+#     --expr 's/APPEND_CLAIM_MAX_WAIT:-30/APPEND_CLAIM_MAX_WAIT_UNUSED:-30/' \
+#     --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' \
+#     --case live-lock-not-swept-on-numeric-stat
+#
+#   The expression redirects the override to a dead variable, so the script
+#   ignores the ceiling the cases set, waits the production 3s and announces
+#   `after 3s` while case 2 asserts the derived `after 0.3s`. That is exactly the
+#   decoupling this arm exists to catch: without it, a message that restates a
+#   constant instead of deriving one stays green. `APPEND_CLAIM_MAX_WAIT:-30`
+#   occurs exactly once (the header names the variable in prose without the
+#   `:-30` default), so the substitution can never be a no-op, and the expression
+#   carries no `^`/`$` anchor so it needs no `/m` under the harness's `perl -0pi`.
+#
+#   The remaining unrecorded arm is the GNU-first ORDERING, and the reason is
+#   narrower than it may look. An arm mutating it (`s/stat -c %Y/stat -f %m/`)
+#   reports a vacuous guard against the TWO CASES BELOW, because their stub answers
+#   unconditionally and ignores its flags by design — deliberately, so both stay
+#   ordering-agnostic.
 #
 #   That is a property of these two cases, NOT a general limit: a flag-aware stub
 #   (`-c` prints an epoch, `-f` prints `%m` and exits 0, simulating GNU) does
@@ -137,7 +166,7 @@ seed_contended_project() {
 case1_dir="$(seed_contended_project stale-non-numeric)"
 case1_stub="$TMPROOT/stub-non-numeric"
 make_stat_stub "$case1_stub" 'echo "%m"'
-case1_out="$(PATH="$case1_stub:$PATH" bash "$SCRIPT" "$case1_dir" '{"id": "claim-1"}' 2>"$TMPROOT/case1.err")"
+case1_out="$(PATH="$case1_stub:$PATH" APPEND_CLAIM_MAX_WAIT=3 bash "$SCRIPT" "$case1_dir" '{"id": "claim-1"}' 2>"$TMPROOT/case1.err")"
 case1_rc=$?
 
 if [ "$case1_rc" -ne 0 ]; then
@@ -154,16 +183,16 @@ fi
 #
 # Without this case a fix that unconditionally sets lock_mtime=0 satisfies case 1
 # while destroying mutual exclusion — every contended lock would read as ancient
-# and get swept out from under a running peer after 3s.
+# and get swept out from under a running peer once the acquire ceiling elapses.
 case2_dir="$(seed_contended_project live-numeric)"
 case2_stub="$TMPROOT/stub-numeric"
 make_stat_stub "$case2_stub" 'date +%s'
-PATH="$case2_stub:$PATH" bash "$SCRIPT" "$case2_dir" '{"id": "claim-2"}' >/dev/null 2>"$TMPROOT/case2.err"
+PATH="$case2_stub:$PATH" APPEND_CLAIM_MAX_WAIT=3 bash "$SCRIPT" "$case2_dir" '{"id": "claim-2"}' >/dev/null 2>"$TMPROOT/case2.err"
 case2_rc=$?
 
 if [ "$case2_rc" -ne 1 ]; then
   fail "live-lock-not-swept-on-numeric-stat" "expected exit 1, got $case2_rc"
-elif ! grep -q 'Could not acquire lock on claims.json after 3s' "$TMPROOT/case2.err"; then
+elif ! grep -q 'Could not acquire lock on claims.json after 0.3s' "$TMPROOT/case2.err"; then
   fail "live-lock-not-swept-on-numeric-stat" "stderr did not carry the timeout error: $(cat "$TMPROOT/case2.err")"
 elif [ ! -d "$case2_dir/cogni-claims/.claims.lock" ]; then
   fail "live-lock-not-swept-on-numeric-stat" "a live peer's lock was swept"
--- a/cogni-portfolio/tests/test-append-claim-lock.sh
+++ b/cogni-portfolio/tests/test-append-claim-lock.sh
@@ -50,7 +50,27 @@
 #   exactly once in the fixed script (the read line is `lock_mtime=$(stat ...`,
 #   which does not contain it), so the substitution can never be a no-op.
 #
-#   Second arm — the ceiling-to-message coupling:
+#   Second arm — the shape floor, case 2's side of it:
+#   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
+#     --root . \
+#     --file cogni-portfolio/scripts/append-claim.sh \
+#     --expr 's/\^\[0-9\]\+\$/^NOMATCH\$/' \
+#     --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' \
+#     --case live-lock-not-swept-on-numeric-stat
+#
+#   Arm one mutates the floor's ASSIGNMENT and is caught by case 1; this one
+#   mutates the floor's CONDITION and is caught by case 2, which is why both are
+#   recorded rather than one standing in for the other. Replacing the character
+#   class with one that cannot match makes the `||` fire unconditionally, so a
+#   live peer's current-epoch reading is floored to 0, reads as ancient, gets
+#   swept, and the script goes on to append — case 2 sees exit 0 where it demands
+#   exit 1. The pattern's `^` and `$` are ESCAPED, i.e. matched as the literal
+#   characters of the regex text on the floor line, not as anchors — so the
+#   expression needs no `/m` under the harness's `perl -0pi`, which slurps the
+#   whole file. `^[0-9]+$` occurs exactly once (the ceiling's own floor above
+#   reads `^[1-9][0-9]*$`), so the substitution can never be a no-op.
+#
+#   Third arm — the ceiling-to-message coupling:
 #   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
 #     --root . \
 #     --file cogni-portfolio/scripts/append-claim.sh \


claude-cogni-work:
## Ship-ready

Handed off for independent merger review. The PR stays **draft** and carries `svc:needs-review`; no review verdict was posted from the author side, and no `svc:review-*` label was applied — the independent review is the merger's to run.

Issue #1431 has been swapped to `svc:resolved`.

Graded against the `--target issue --origin triage` contract (derived at triage time, before this branch existed), via `contract-grading-reviewer`.

- **Items graded:** 20
- **Satisfied:** 19
- **Violated:** 1 → **fixed** in this branch
- **Unclear / INTENT-UNCLEAR:** 0
- **Residual carried to the merger:** none

**The one violation, and what closed it.** Contract item 17 requires a *recorded* arm mutating the shape floor's **condition** at `append-claim.sh:62` so the reading is set unconditionally, replayed against `--case live-lock-not-swept-on-numeric-stat`. The branch as first pushed recorded two arms — the floor's **assignment** (caught by case 1) and the ceiling-to-message coupling (caught by case 2) — so the arm case 2 was actually written for was undocumented. The grader independently confirmed the protection itself already held, making this an evidence gap rather than a code defect.

Closed by the bounded fix pass in `f10885b1`, which records the missing arm in the suite header:

```
--expr 's/\^\[0-9\]\+\$/^NOMATCH\$/' --case live-lock-not-swept-on-numeric-stat
```

All **three** arms now replay to `guard_verified` on this head, and `run-plugin-tests.py --filter cogni-portfolio` is 7/7 with the suite at 1.1s.

### Note for the reviewer

The fix pass is deliberately **not** re-graded — per the handoff contract, one bounded pass runs and its result is reported rather than re-adjudicated. Item 17 is therefore reported as fixed-and-verified by replay, not as re-graded clean.

The handoff preflight ran clean before and after the fix pass: 3 passed, 4 skipped (`not_applicable` — no touched `SKILL.md` or `agents/*.md`), 0 violations.

claude-cogni-work:
## Duplicate work alert — two PRs resolve #1431

**#1444 and #1445 independently resolve the same issue.** Neither author knew about the other. Do not spend two independent merger reviews here; pick one and close the other.

Both were produced by concurrent `service-resume --role author` drains on the same host, minutes apart:

| | #1444 | #1445 |
|---|---|---|
| Opened | 19:08:00Z | 19:13:43Z |
| Branch | `fix/issue-1431-append-claim-lock-wait-ceiling` | `fix/issue-1431-append-claim-wait-ceiling` |
| Diff | +71/−10 | 3 commits |

### The two solutions converged

Independently, and almost line for line — which is good evidence the approach is the right one:

- same override name and default, `MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"`
- same shape floor, `[[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] || MAX_WAIT=30`
- same decision to derive the timeout label rather than restate `3s` (#1444 strips a `.0` suffix; #1445 branches on the tenths — equivalent output on every admitted value)
- same test override of `3` and same `after 0.3s` assertion in case 2

I have not reviewed #1444's correctness and am not asserting either is better on the merits.

### Why this happened (worth fixing separately)

Not an authoring mistake by either side — a race in the claim protocol. Issue #1431 carried an `svc:in-progress` label from a drain that stalled ~5h earlier, so `claim-status` reported `stale`, which is *by design* pickable. Both drains snapshotted that same stale claim, both correctly concluded they were resuming a strand, and each then refreshed the label — erasing the evidence of the other:

```
13:51:38  labeled     <- the drain that stalled
18:47:15  unlabeled   <- drain A refresh
18:47:18  labeled
18:49:15  unlabeled   <- drain B refresh, 2 min later
18:49:18  labeled
```

`service-resolve` Step 2 attributes the claim against its own Step 1 snapshot. Drain B snapshotted before drain A's refresh landed, so B saw the stale 13:51 claim rather than A's fresh one, and B's own re-add then made it the newest event — so neither side's foreign-claim check could fire. The claim-then-verify sequence has a read-to-write window that a second drain can land inside.

Filed separately so this comment stays about the PRs.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1444

**Summary:** Contract grading found the diff satisfies 19 of 20 committed contract items by direct execution. The one gap is item 8's third required fact — that the emitted timeout duration is derived from the ceiling rather than hardcoded — which lives in an inline comment at lines 36-37 next to the computation instead of in append-claim.sh's top-of-file header block. plugin-validator passed clean and independently re-verified all four issue acceptance criteria by execution; its two remaining findings are merger-added preference-only observations that do not gate. Merger note: the duplicate PR #1445 (same issue #1431, same two files) carries the identical item-8 placement gap, so two independent authors converged on documenting the derivation at the derivation site rather than in the header.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[major]** (auto-fixable) `review-contract:item-8` — append-claim.sh's header comment (lines 8-10) names the override variable APPEND_CLAIM_MAX_WAIT and its default (30 ticks / 3s), but never states that the emitted timeout duration is derived from the ceiling rather than hardcoded — the third fact item 8 explicitly requires 'both headers' to carry. That fact is documented only in an inline comment at lines 36-37, directly above the TIMEOUT_LABEL computation, not in the file's top-of-file header block. Note for the ruling human: item 8's own falsification clause is 'an undocumented env escape hatch on a locking primitive fails this item', and the escape hatch IS documented in the header — what is misplaced is one clause of three. The duplicate PR #1445 has the identical gap.
  - *Suggested fix:* Add one clause to the header block (lines 8-10) stating that the printed error duration is derived from the ceiling and will never restate a fixed number.

- **[minor]** (waived) (advisory-addition) `cogni-portfolio/scripts/append-claim.sh:38` — TIMEOUT_LABEL converts ticks to seconds by hardcoding /10 and %10, correct only because line 42 is literally 'sleep 0.1' — the tick duration still lives in two places. Latent, not live: the interval is not parameterised and nothing in the repo varies it, so there is no user-observable fault today.
  - *Justification:* Latent coupling with no current fault: the interval is a hardcoded literal, no caller or test varies it, and output is correct for the default and every override tested (3, 08, abc, empty, 0, -5).

- **[minor]** (waived) (advisory-addition) `cogni-portfolio/scripts/append-claim.sh:70,84-100` — append-claim.sh does not use the repo-wide {success,data,error} script envelope documented in root CLAUDE.md. Pre-existing and not a fault of this PR: the diff changes only the interpolated number inside the existing error string. A census shows 8 of 16 shell scripts in this plugin carry the envelope and 8 do not.
  - *Justification:* Pre-existing and explicitly out of this PR's two-file scope; the diff neither introduces nor worsens it.

### Positive observations

- Suite wall time 1.535s against the sub-2s bar (7.7s baseline at triage); exits 0. (contract-grading-reviewer)
- run-plugin-tests.py --filter cogni-portfolio: data.total=7, data.failed=0, data.failed_suites=[]. (contract-grading-reviewer)
- Default ceiling preserved by execution: no override exits 1 after 3.366s with stderr 'after 3s'. (contract-grading-reviewer)
- Override tracks the message: APPEND_CLAIM_MAX_WAIT=3 exits 1 after 0.35s with stderr 'after 0.3s'. (contract-grading-reviewer)
- Malformed override (abc) terminates in 3.362s via the numeric shape floor rather than spinning. (contract-grading-reviewer)
- All three recorded mutation arms replay guard_verified (floor assignment, floor condition, ceiling-to-message coupling). (contract-grading-reviewer)
- A red case fails the SUITE: mutated run exits non-zero with '1 test(s) failed' on stderr. (contract-grading-reviewer)
- grep 'after 3s' in append-claim.sh returns nothing; echo interpolates the derived TIMEOUT_LABEL. (contract-grading-reviewer)
- No version touch; check-version-bump.py and check-result-line-plainness.py both report zero violations. (contract-grading-reviewer)
- Scope clean: exactly the lock script + its suite, no sibling-issue (#1432/#1433/#1434) deliverable rides along. (contract-grading-reviewer)
- None of the seven cogni-portfolio/agents/*.md append-claim.sh call sites sets the override. (contract-grading-reviewer)
- AC1 met with margin: 1.06s wall over two runs against a 2s ceiling; base measured 7.23s and 7.40s on the same host. (plugin-validator)
- AC2 verified by execution: all three recorded mutation arms replay guard_verified, each mutated_result=red and restored_result=green. (plugin-validator)
- AC3 verified directly: no-override run emits byte-identical 'after 3s' at 3.5s wall; repo-wide grep confirms no caller sets the override. (plugin-validator)
- AC4 verified: run-plugin-tests.py --filter cogni-portfolio exits 0, 7/7 passing. (plugin-validator)
- Numeric shape floor is genuinely defensive: abc, empty, 0, -5 and notably 08 (rejected before arithmetic reads it as invalid octal) all fall back to the 30-tick default. (plugin-validator)
- Bash portability checked on both interpreters present: green under /bin/bash 3.2.57 and 5.3.9; the unquoted regex right-hand side is correct for the 3.2 target and the inline comment says why. (plugin-validator)
- The suite honours the assert-foreign-output-by-shape rule: case 1 discriminates on stderr emptiness (bash arithmetic errors are gettext-localized); case 2 greps a literal only because append-claim.sh emits that string itself. (plugin-validator)
- Mutation no-op safety real: each of the three substitution targets occurs exactly once in the script. (plugin-validator)
- No flakiness introduced: both cases branch on stub-driven stat output rather than wall clock. (plugin-validator)
- Version convention respected: plugin.json stays 0.9.79, marketplace mirror agrees, zero violations. (plugin-validator)
- All repo guards clean at this head: check-breadcrumbs.py, check-result-line-plainness.py, check-skill-names.sh. (plugin-validator)
- Read-only constraint respected: mutation replay ran against a copy; worktree git status clean and script byte-identical to HEAD. (plugin-validator)

### Recommendation

---

sdh07:
## Merger hold — duplicate-pair ruling needed before an author cycle is spent

The independent review is **complete and posted above** (verdict `needs_revision`, one
finding). This PR is being **held for a maintainer ruling rather than bounced to the
author**, and the deviation from the ordinary bounce path is deliberate.

### Why hold rather than bounce

`BOUNCE_COUNT` is 1, well under the escalation threshold, so the mechanical path would be
to strip `svc:needs-review` and hand this back to the author drain. That would be the wrong
move here for two reasons:

1. **This PR is one of a duplicate pair.** #1444 and #1445 both resolve issue #1431, both
   touch the same two files, and both were opened five minutes apart by two concurrent
   author drains that each resumed the same stale claim — the race documented in #1446.
   One of the two has to be closed. That is a maintainer call; no review round can produce
   it.
2. **Bouncing would spend the second review anyway.** Removing `svc:needs-review` sends
   #1444 to the author drain for a one-clause comment fix while #1445 stays in the merger
   queue — so the next drain would open a second full independent review on the twin.
   #1446's own acceptance criterion (d) names exactly that outcome as the thing to prevent:
   *"two open PRs linked to the same issue are surfaced somewhere a merger sees before
   reviewing; falsified by a duplicate pair that reaches two independent reviews."*

So the queue is parked here rather than churned.

### What the review actually found

Nineteen of the twenty committed contract items are satisfied, and every executable one was
graded by running it, not by reading the diff:

| Check | Result |
|---|---|
| Suite wall time (bar: <2s) | **1.06–1.54s**, from a 7.23–7.40s base |
| `run-plugin-tests.py --filter cogni-portfolio` | **7/7**, zero failures |
| Default ceiling, no override | exit 1 at **3.37s**, stderr `after 3s` |
| Override honoured | exit 1 at **0.35s**, stderr `after 0.3s` |
| Malformed override (`abc`) | terminates at 3.36s — no hang |
| All three recorded mutation arms | **`guard_verified`** (mutated red, restored green) |
| Red case fails the *suite* | confirmed: non-zero exit + `1 test(s) failed` |
| bash 3.2.57 / 5.3.9 | green on both |

The single gap is contract item 8: the header block names the override variable and its
default but does not state that the printed duration is *derived* from the ceiling. That
fact is in the file, inline at lines 36-37, next to the computation.

### The ruling being requested

Two decisions, ideally taken together:

- **Which duplicate survives** — #1444 (this one, first opened, `${VAR%.0}` suffix-strip
  derivation) or #1445 (`if/else` derivation, longer rationale on the octal and `set -e`
  hazards). Both are CI-green and behaviourally equivalent.
- **Whether item 8 is waived or fixed.** Worth noting before deciding: **#1445 has the
  identical gap.** Two authors working independently both put the derivation note at the
  derivation site rather than in the header, which reads as the contract over-specifying
  *placement* rather than as either author being careless. Item 8's own falsification
  clause is "an undocumented env escape hatch on a locking primitive fails this item" — and
  the escape hatch *is* documented in the header. What is misplaced is one clause of three.

Waiving it merges the survivor as-is; not waiving it costs one author cycle for a
one-line comment addition.

Labels: `svc:needs-review` is **kept** (parked, not dequeued) and `svc:needs-human-review`
is applied. The PR stays draft. Clearing the hold is a matter of removing
`svc:needs-human-review` — please do not hand-apply `svc:review-pass`, which would assert a
review that did not happen.

sdh07:
## Maintainer ruling — hold cleared, both questions answered

The merger parked this PR asking for two decisions taken together. Both are ruled here, and the hold is cleared.

### 1. #1444 survives; #1445 is closed

#1445 is closed unreviewed and unmerged, branch deleted. The two diffs are behaviourally identical on every value the shape floor admits — same override name and default, same floor pattern, same suite override, same `after 0.3s` assertion; only the label derivation differs in style (`${VAR%.0}` suffix-strip here, `if/else` on the tenths there), and the two produce the same string throughout.

The decision therefore turned on review state, not merits. This PR has the completed independent review; #1445 has none and carries the identical item-8 gap, so reviewing it would reach the same held state — the exact two-reviews-on-one-issue outcome #1446's acceptance criterion (d) exists to prevent. The merger was right to park rather than bounce.

### 2. Contract item 8 is WAIVED

This PR merges as-is, at the reviewed head `f10885b1`. No author cycle is spent.

The reasoning, for the record and for whoever revises this contract:

- Item 8's own falsification clause is *"an undocumented env escape hatch on a locking primitive fails this item"*. The escape hatch **is** documented in `append-claim.sh`'s header block — the header names `APPEND_CLAIM_MAX_WAIT`, its 0.1s-tick unit, and its production default of 30 ticks / 3s. What is misplaced is one clause of three, not the disclosure the item exists to compel.
- **Both independent authors** put the derivation note at the derivation site rather than in the header. Two authors who could not see each other's work converging on the same placement is evidence the contract over-specifies *placement*, not evidence that either author was careless.
- The fact itself is in the file, inline at the computation, where a reader changing the arithmetic will actually encounter it.

This is a ruling on the contract item, not a waiver of the underlying property. The derivation is real and is pinned by a recorded mutation arm.

### Verification before merge

The verdict's green checks completed at 19:19Z; `main` has since advanced to `cc74183b` (19:37Z). Re-verified rather than assumed — this PR's head merged with current `main`:

| Check | Result |
|---|---|
| `run-plugin-tests.py --filter cogni-portfolio` | rc=0, **7/7**, zero failures |
| `test-append-claim-lock.sh` wall time | **1.09s** against the 2s bar |
| Merge with current `main` | clean, no conflicts |

### Labels

`svc:needs-human-review` removed (the human has ruled) and `svc:review-needs-revision` removed (the finding is waived). `svc:review-pass` is deliberately **not** applied — as the merger asked, that would assert a passing review that did not happen. The merge proceeds on this ruling under the gate's own terms: the gate is advisory and the human makes the final merge decision.

### Follow-up filed

#1445's branch documented two hazards this branch's code guards against but does not explain — why the floor leads with `[1-9]` (a value like `08` passes a bare `^[0-9]+$` and then makes the arithmetic a hard bash abort, since expansion reads a leading zero as octal), and why the derivation uses the assignment form rather than a bare `(( ... ))` (a bare arithmetic command evaluating to 0 returns status 1, and the default ceiling makes the remainder exactly 0 — aborting under `set -e` on bash 5.x while passing on 3.2). Captured as a comment-only follow-up so it does not disappear with the closed branch.

Issue #1434, opened 2026-08-17
cogni-portfolio: append-claim.sh predates the {success,data,error} script-output convention


sdh07:
## Context

Raised by the review gate on #1428, where it was correctly judged pre-existing and out of scope. Recorded so the acceptance criterion is not silently carried forward as met at a stricter reading than the code supports.

## Current behavior

The repo convention (root `CLAUDE.md` § Script Output Format) is that all scripts return `{"success": bool, "data": {...}, "error": "string"}`. `cogni-portfolio/scripts/append-claim.sh` predates that convention on both streams:

- timeout path emits `{"error": "Could not acquire lock on claims.json after 3s"}` — no `success`, no `data`
- success path prints `{"status": "appended", "claim_id": ..., "total": ...}` — same

Nothing in #1428 regressed this: the timeout line is byte-identical to its base, and that PR added no new emitting path. It is flagged only because #1424's acceptance criterion 3 ("every failure path in the touched region still emits the `{"success","data","error"}` envelope") reads, literally, as unsatisfiable against a file that never emitted that envelope to begin with. The criterion held in the sense that the PR introduced no regression, which is the reading that was applied.

One shape change is worth recording precisely because it is easy to misread as a regression and is not one: `now=$(date +%s)` is now a standalone assignment, so under `set -euo pipefail` a failing `date` would abort with no JSON at all, where an empty inner substitution would previously still have produced valid arithmetic. This is unreachable in practice and is the same class as the pre-existing `mkdir -p` on line 24, which also aborts envelope-less. It is not worth reverting — the assignment is what keeps a command substitution out of `$(( ))`, which is the entire point of #1424.

## Desired behavior

Bring `append-claim.sh`'s stdout and stderr onto the `{"success", "data", "error"}` convention, together with its callers.

The change is currently cheap: none of the agent callers that invoke this script parse its output today, so the shape can move without a coordinated migration. That will stop being true the first time someone writes a caller against the current shape, which is the argument for doing it rather than leaving it.

Check the caller list before changing the shape rather than trusting this note — it is accurate at time of filing and is exactly the sort of claim that ages.

## Acceptance criteria

- [ ] `append-claim.sh` emits `{"success", "data", "error"}` on both the success and the lock-timeout paths. Falsified by either path emitting a bare `{"status": ...}` or `{"error": ...}`.
[...]

Issue #1433, opened 2026-08-17
cogni-portfolio: no case pins the GNU-first stat ordering — a BSD-first revert leaves the lock suite green


sdh07:
## Context

Found by the review gate on #1428 and verified independently before filing. #1428's suite header claimed no case could discriminate the two `stat` orderings; that claim was false and has been corrected in #1428 itself. The missing **coverage** is tracked here.

## Current behavior

`cogni-portfolio/tests/test-append-claim-lock.sh` has two cases, both using a stub that answers unconditionally and ignores its flags — deliberately, so both stay ordering-agnostic. The consequence is that the GNU-first ordering, which is the actual subject of #1424, is **not pinned by any case**. Reverting `append-claim.sh` line 49 to BSD-first leaves the suite green.

That ordering is load-bearing on its own merits, not merely tidier. Measured with a flag-aware stub simulating GNU (`-c` prints an epoch, `-f` prints `%m` and exits 0), against the *fixed* script with the numeric-shape floor present in both arms:

| Ordering | rc | stdout | live peer's lock |
|---|---|---|---|
| GNU-first (shipped) | 1 | none | **survived** |
| BSD-first (swapped) | 0 | `{"status": "appended", ...}` | **swept** |

So with the floor present but the ordering swapped, a GNU host floors every contended lock's reading to 0, reads it as ancient, and sweeps a **live** peer's lock after the timeout — destroying exactly the mutual exclusion case 2 exists to protect, on exactly the platform (Linux) #1424 was about. The floor makes the failure silent rather than loud; it does not make it safe.

## Desired behavior

A third case, e.g. `live-lock-not-swept-on-gnu-stat`, using a flag-aware stub:

```sh
case "$1" in
  -c) date +%s ;;
  -f) echo "%m" ;;
  *) exit 1 ;;
esac
```

asserting exit 1 and that the peer's `.claims.lock` survives — plus a second recorded mutation arm swapping the ordering, so the recipe proves the new case discriminates.

Keep the two existing cases exactly as they are. Their flag-blind stub is a deliberate design choice with its rationale already written into the header, and the new case is additive, not a replacement.

## Acceptance criteria

- [ ] A case exists that goes red when `append-claim.sh`'s stat chain is reordered to BSD-first, with the numeric-shape floor left intact. Falsified if swapping the ordering leaves the suite green.
- [ ] The two existing cases still pass unchanged and still use the flag-blind stub. Falsified by either being rewritten to depend on flag awareness.
[...]

PR #1445, opened 2026-08-17
cogni-portfolio: parameterise append-claim lock acquire ceiling

### LINKED ISSUE #1431: cogni-portfolio: the append-claim lock suite costs 7.3s wall on every CI sweep, 94% of it sleep

claude-cogni-work:
Closes #1431.

## Summary

- `cogni-portfolio/scripts/append-claim.sh` — the acquire ceiling becomes `MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"` plus a numeric-shape floor. `MAX_WAIT` counts 0.1s poll iterations, so the default 30 remains a 3s ceiling.
- Same file — the timeout message is now **derived** from the ceiling instead of restating it. At the default it renders `after 3s`, byte-identical to today; under the suite's override it renders `after 0.3s`.
- `cogni-portfolio/tests/test-append-claim-lock.sh` — both cases export `APPEND_CLAIM_MAX_WAIT=3`, case 2's grep tracks the derived label, and the header records a second mutation arm covering the new derivation.

Suite wall time drops from **7.7s to ~1.1s**; roughly 94% of the old runtime was `sleep 0.1` in the acquire loop, paid on every CI sweep of the monorepo.

## Why the message change ships with the ceiling change

The issue asks only for a parameterised ceiling. But `after 3s` was hardcoded, and correct only because 30 x 0.1s happens to equal 3s. Parameterising the ceiling alone would leave the script waiting 0.3s while still claiming 3s — with case 2's grep still passing, because it matched the stale literal. That is a guard gone quietly wrong, the same defect class this suite exists to catch, so the two halves ship together. Exploration confirmed exactly one load-bearing consumer of that literal in the repo (this suite), so the change is contained.

## Verification

| Check | Result |
|---|---|
| `bash cogni-portfolio/tests/test-append-claim-lock.sh` (bash 3.2.57) | pass, **1.124s** wall |
| Same, homebrew bash 5.3.9 | pass, **1.091s** wall |
| `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` | rc=0, **7/7** suites |
| Mutation arm 1 — `lock_mtime=0` dead-variable, case 1 (pre-existing) | `guard_verified` |
| Mutation arm 2 — derived label, case 2 (new) | `guard_verified` |
| Mutation arm 3 — shape floor unconditional, case 2 (new) | `guard_verified` |
| Red case fails the **suite**, not just a case line | exit 1 + `1 test(s) failed` on stderr |
| Default ceiling by execution, no override | exit 1 after 3.47s, stderr `after 3s` |
| Malformed override `APPEND_CLAIM_MAX_WAIT=abc` | floors to 30, terminates in 3.2s — no unbounded loop |
| `check-result-line-plainness.py` / `check-breadcrumbs.py` / `check-version-bump.py` / `check-workflow-yaml.py` | all rc=0 |

Label rendering was verified on both bash builds across every admitted and rejected value: unset and `30` render `3s`; `3` → `0.3s`; `1` → `0.1s`; `10` → `1s`; `105` → `10.5s`; `999` → `99.9s`. Every rejected shape (`0`, `08`, `030`, `abc`, `-5`, `3.5`, `" 3"`) floors to 30 and renders `3s`.

## Design notes worth a reviewer's attention

- **The floor regex deliberately diverges** from the `^[0-9]+$` used by the stale-lock reading below it. The leading `[1-9]` excludes leading zeros, because `08` or `030` would pass a bare `^[0-9]+$` and then make `$(( MAX_WAIT / 10 ))` a hard bash abort — arithmetic expansion reads a leading zero as octal. A comment records this so the two patterns are not later "harmonised" back into the bug.
- **Zero is excluded on purpose.** The loop sleeps before its first `-ge` check, so `0` and `1` both mean one 0.1s poll; `0` buys no fail-immediately semantics and would render a misleading `after 0s` for a path that did wait.
- **The derivation sits outside the acquire loop.** An arithmetic abort *inside* it is the exact failure the file's own comment documents: the abort abandons the loop, execution resumes after `done`, and the claim is appended having never held the lock — after which the EXIT trap removes a live peer's lock directory. At top level the same abort is a clean exit before any lock is taken.
- **Arithmetic uses the assignment form**, never a bare `(( ... ))`. A bare arithmetic command evaluating to 0 returns status 1, and the default ceiling makes the remainder exactly 0 — under `set -e` that aborts on bash 5.x while passing on 3.2, a version-divergent failure on the *default* path.

## Scope decisions

- **Option 1 only.** Triage measured Option 2 (running the two cases concurrently) at ~3.5s, which cannot clear the under-2s bar because it removes only one of the two ceilings. It would also force a subshell failure-counter rework the suite's deliberate `set -u`-only structure does not otherwise need.
- **Production timing untouched.** Exploration found **7** live call sites, not the 5 triage named — it missed `customer-researcher.md` and `proposition-generator.md`. All 7 are agent markdown files invoking the script plainly, and none sets an override, so the `:-30` default holds acceptance criterion 3 by construction.
- **Not absorbed:** #1432, #1433 and #1434 touch the same two files but are distinct deliverables.
- **No version bump** — the post-merge workflow owns `plugin.json` / `marketplace.json`.

## Mutation recipe

```bash
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/\$\{WAIT_LABEL\}/3s/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case live-lock-not-swept-on-numeric-stat
```

The expression restates the derived label as the old hardcoded literal, so the script claims `after 3s` while the suite drives a 0.3s ceiling — case 2's grep misses and `live-lock-not-swept-on-numeric-stat` goes red. That is precisely the greened-over failure this PR exists to prevent. Replayed above: `guard_verified`.

The braced `WAIT_LABEL` form occurs exactly once in the fixed script (the `echo`); the two assignments read `WAIT_LABEL=`, and the surrounding comments name the variable bare. That is load-bearing — the harness applies the expression with `perl -0pi` and no `/g`, so a second occurrence above the echo would absorb the substitution and silently degrade this arm to a no-op that does not even trip `expr_no_op`. The expression carries no `^`/`$` anchors for the same reason: `perl -0pi` slurps, so an anchored expression would need a `/m` modifier to match at all.

A third arm is now recorded too, closing a gap that predates this PR: the suite pinned the stale-lock shape floor from one side only. Case 1's arm proves the floor **exists**; nothing proved it stays **conditional**. Dropping the guard and forcing every reading to 0 makes a live peer's lock read as ancient and be swept — the destroy-mutual-exclusion failure case 2 exists to catch:

```bash
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/\[\[ "\$lock_mtime" =~ \^\[0-9\]\+\$ \]\] \|\| //' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case live-lock-not-swept-on-numeric-stat
```

Its `^` and `$` are backslash-escaped, so they are literal characters rather than anchors and need no `/m` modifier despite the harness slurping.

The pre-existing arm is unchanged and still verifies:

```bash
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/lock_mtime=0/lock_mtime_unused=0/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case stale-lock-swept-on-non-numeric-stat
```

## Plan record

https://github.com/cogni-work/insight-wave/issues/1431#issuecomment-5318970837

### DIFF
--- a/cogni-portfolio/scripts/append-claim.sh
+++ b/cogni-portfolio/scripts/append-claim.sh
@@ -48,8 +48,8 @@ MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"
 [[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] || MAX_WAIT=30
 
 # Derive the timeout label from the ceiling rather than restating it, and do it
-# HERE — at top level, outside the loop. The old message hardcoded "after 3s",
-# which was correct only because 30 x 0.1s happens to equal 3s; under an override
+# HERE — at top level, outside the loop. The old message restated a fixed three
+# seconds, which was correct only because 30 x 0.1s equals that; under an override
 # the script would have kept claiming 3s while actually waiting a tenth of that,
 # and the suite's grep would have stayed green on a message that had become
 # wrong. Placement matters as much as derivation: an arithmetic abort inside the
--- a/cogni-portfolio/scripts/append-claim.sh
+++ b/cogni-portfolio/scripts/append-claim.sh
@@ -5,6 +5,8 @@
 # Creates cogni-claims/ directory and claims.json if they don't exist.
 # Uses mkdir-based locking (portable across macOS and Linux) to prevent
 # race conditions when agents run in parallel.
+# Env: APPEND_CLAIM_MAX_WAIT overrides the lock-acquire ceiling in 0.1s units
+#   (default 30 = 3s); a non-positive or non-integer value falls back to 30.
 # Exit codes: 0 = success, 1 = error
 set -euo pipefail
 
@@ -24,7 +26,50 @@ LOCK_DIR="$CLAIMS_DIR/.claims.lock"
 mkdir -p "$CLAIMS_DIR/sources" "$CLAIMS_DIR/history"
 
 # Acquire lock using mkdir (atomic on all POSIX systems)
-MAX_WAIT=30
+#
+# MAX_WAIT counts 0.1s poll iterations, so the default 30 is a 3s ceiling. The
+# override exists for tests: both cases of the lock suite exist to DRIVE this
+# loop to its ceiling, so the wait length is incidental to what they pin, and at
+# the default they burn ~3s each for nothing. Callers that set no override keep
+# the 3s ceiling exactly.
+#
+# Floor on numeric SHAPE, as the stale-lock reading below does — but NOT with the
+# same pattern, and the divergence is load-bearing. The leading [1-9] also
+# excludes leading zeros: `08` or `030` would pass a bare ^[0-9]+$ and then make
+# the division below a hard bash abort ("value too great for base"), because
+# arithmetic expansion reads a leading zero as octal. Do not harmonise the two
+# patterns. Zero is excluded too, and that is deliberate rather than incidental:
+# the loop sleeps BEFORE its first -ge check, so 0 and 1 both mean "one 0.1s
+# poll" — 0 buys no fail-immediately semantics that 1 does not already give, and
+# it would render a misleading "after 0s" for a path that did in fact wait.
+# Admitting only positive integers is also what guarantees the -ge test below can
+# never error, so the loop always terminates.
+MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"
+[[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] || MAX_WAIT=30
+
+# Derive the timeout label from the ceiling rather than restating it, and do it
+# HERE — at top level, outside the loop. The old message hardcoded "after 3s",
+# which was correct only because 30 x 0.1s happens to equal 3s; under an override
+# the script would have kept claiming 3s while actually waiting a tenth of that,
+# and the suite's grep would have stayed green on a message that had become
+# wrong. Placement matters as much as derivation: an arithmetic abort inside the
+# acquire loop is the exact failure the comment further down documents — it
+# abandons the loop, resumes after `done`, and appends the claim having never
+# held the lock. At top level the same abort is a clean exit before any lock is
+# taken and before the release trap is installed.
+#
+# Use the assignment form for the arithmetic, never a bare (( ... )) command: a
+# bare arithmetic command evaluating to 0 returns status 1, and the default
+# ceiling makes the remainder exactly 0, so under `set -e` that aborts on bash
+# 5.x while passing on 3.2 — a version-divergent failure on the DEFAULT path.
+WAIT_WHOLE=$(( MAX_WAIT / 10 ))
+WAIT_TENTHS=$(( MAX_WAIT % 10 ))
+if [ "$WAIT_TENTHS" -eq 0 ]; then
+  WAIT_LABEL="${WAIT_WHOLE}s"
+else
+  WAIT_LABEL="${WAIT_WHOLE}.${WAIT_TENTHS}s"
+fi
+
 WAITED=0
 while ! mkdir "$LOCK_DIR" 2>/dev/null; do
   sleep 0.1
@@ -55,7 +100,7 @@ while ! mkdir "$LOCK_DIR" 2>/dev/null; do
         continue
       fi
     fi
-    echo '{"error": "Could not acquire lock on claims.json after 3s"}' >&2
+    echo "{\"error\": \"Could not acquire lock on claims.json after ${WAIT_LABEL}\"}" >&2
     exit 1
   fi
 done
--- a/cogni-portfolio/tests/test-append-claim-lock.sh
+++ b/cogni-portfolio/tests/test-append-claim-lock.sh
@@ -67,7 +67,26 @@
 #   `perl -0pi` slurps the whole file, so an anchored expression would need a `/m`
 #   modifier to match at all.
 #
-#   Only these TWO arms are recorded, and the reason is narrower than it may look. An arm
+#   Third arm — the shape floor applied unconditionally (case 2's negative twin):
+#   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
+#     --root . \
+#     --file cogni-portfolio/scripts/append-claim.sh \
+#     --expr 's/\[\[ "\$lock_mtime" =~ \^\[0-9\]\+\$ \]\] \|\| //' \
+#     --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' \
+#     --case live-lock-not-swept-on-numeric-stat
+#
+#   This deletes the floor's GUARD and leaves the assignment, so the reading is
+#   forced to 0 for every lock rather than only a malformed one. Case 2 supplies a
+#   numeric current-epoch reading, i.e. a genuinely LIVE peer, so under the
+#   mutation that lock reads as ancient, gets swept, and the case goes red — which
+#   is exactly the destroy-mutual-exclusion failure the case comment below
+#   describes. The first arm pins the same floor from the other side (case 1, the
+#   malformed reading), so the two together pin the floor's presence AND its
+#   conditionality; neither alone does. The `^` and `$` in the expression are
+#   backslash-escaped, so they are literal characters and not anchors — no `/m`
+#   modifier is needed despite the harness slurping with `perl -0pi`.
+#
+#   Only these THREE arms are recorded, and the reason is narrower than it may look. An arm
 #   mutating the GNU-first ordering (`s/stat -c %Y/stat -f %m/`) reports a vacuous
 #   guard against the TWO CASES BELOW, because their stub answers unconditionally
 #   and ignores its flags by design — deliberately, so both stay ordering-agnostic.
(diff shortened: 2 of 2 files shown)


claude-cogni-work:
## Ship-ready

Handed off for independent merger review. Draft, `svc:needs-review`; the author instance posted no verdict label and no review marker — the independent review is the merger's to run.

**Issue:** #1431 · **Branch:** `fix/issue-1431-append-claim-wait-ceiling` · **Head:** `dd547238`

### Handoff preflight

`check-preflight.sh` exit 0, clean — 7 instruments discovered, 7 resolved, 3 gating passed, 0 failed, 0 violations. Re-run after each of the two fix-pass commits below; clean all three times.

Four instruments skipped, each `not_applicable`: `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` — the diff touches no `SKILL.md` and no `agents/*.md`. The `mutation-recipe` instrument actively **passed** rather than skipping.

Graded against the `--target issue --origin triage` contract (20 acceptance items), by execution rather than by reading.

**19 of 20 satisfied on the first grade. 1 fixed. 0 residual.**

The one gap was item 17 — the issue's own acceptance criterion 2 requires demonstrating that case 2 goes red when the shape floor is mutated to set the reading *unconditionally*. The protection was intact (the grader replayed it to `guard_verified`), but neither the suite header nor the PR body recorded that arm, and the contract admits an unrecorded recipe only if the replay is demonstrated in the PR evidence.

Fixed in `dd547238` by recording it as a third mutation arm, with a simpler expression than the grader's ad-hoc one. This closes a gap that **predates this PR**: the suite pinned the stale-lock shape floor from one side only — case 1's arm proved the floor exists, nothing proved it stays conditional.

All three arms replay `guard_verified`. Tests were re-run green before each fix-pass commit.

### What the merger may want to weigh

The contract's own `risk_watchlist` flags one item no gate can decide: **whether an environment escape hatch on a locking primitive is an acceptable surface at all** is a design-taste call. The checklist can only confirm it is documented, defaulted, shape-validated, and unused by callers — all four confirmed. Whether it *should* exist is a human call.

Two corrections this PR carries that a reviewer checking against the triage comment would otherwise trip on:

1. Triage says **five** `append-claim.sh` call sites; the base tree has **seven** (triage missed `customer-researcher.md` and `proposition-generator.md`). The conclusion is unchanged — none of the seven sets an override — but a reviewer verifying "five" will find more.
2. Case 2's grep string **changed**, to the derived `after 0.3s` form. The contract explicitly anticipates this and records that demanding byte-identity with the pre-fix literal would be jointly unsatisfiable with its own derived-message requirement — i.e. raising that as a defect would be a contract defect, not an author defect.

claude-cogni-work:
## Duplicate work alert — two PRs resolve #1431

**#1444 and #1445 independently resolve the same issue.** Neither author knew about the other. Do not spend two independent merger reviews here; pick one and close the other.

Both were produced by concurrent `service-resume --role author` drains on the same host, minutes apart:

| | #1444 | #1445 |
|---|---|---|
| Opened | 19:08:00Z | 19:13:43Z |
| Branch | `fix/issue-1431-append-claim-lock-wait-ceiling` | `fix/issue-1431-append-claim-wait-ceiling` |
| Diff | +71/−10 | 3 commits |

### The two solutions converged

Independently, and almost line for line — which is good evidence the approach is the right one:

- same override name and default, `MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"`
- same shape floor, `[[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] || MAX_WAIT=30`
- same decision to derive the timeout label rather than restate `3s` (#1444 strips a `.0` suffix; #1445 branches on the tenths — equivalent output on every admitted value)
- same test override of `3` and same `after 0.3s` assertion in case 2

I have not reviewed #1444's correctness and am not asserting either is better on the merits.

### Why this happened (worth fixing separately)

Not an authoring mistake by either side — a race in the claim protocol. Issue #1431 carried an `svc:in-progress` label from a drain that stalled ~5h earlier, so `claim-status` reported `stale`, which is *by design* pickable. Both drains snapshotted that same stale claim, both correctly concluded they were resuming a strand, and each then refreshed the label — erasing the evidence of the other:

```
13:51:38  labeled     <- the drain that stalled
18:47:15  unlabeled   <- drain A refresh
18:47:18  labeled
18:49:15  unlabeled   <- drain B refresh, 2 min later
18:49:18  labeled
```

`service-resolve` Step 2 attributes the claim against its own Step 1 snapshot. Drain B snapshotted before drain A's refresh landed, so B saw the stale 13:51 claim rather than A's fresh one, and B's own re-add then made it the newest event — so neither side's foreign-claim check could fire. The claim-then-verify sequence has a read-to-write window that a second drain can land inside.

Filed separately so this comment stays about the PRs.

sdh07:
## Maintainer ruling — closed as the duplicate of #1444

#1444 is the surviving PR for issue #1431. This one is closed unreviewed and unmerged.

**This is not a quality judgement.** The two diffs were compared line by line and are behaviourally identical on every admitted value:

| | #1444 | #1445 |
|---|---|---|
| Override | `MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"` | identical |
| Shape floor | `[[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] \|\| MAX_WAIT=30` | identical |
| Label derivation | `${VAR%.0}` suffix-strip | `if/else` on the tenths |
| Suite override / assertion | `3` → `after 0.3s` | identical |

The derivation styles produce the same string for every value the floor admits, so the choice between them carried no behavioural weight.

**What decided it was review state, not merits.** #1444 has already had the full independent merger review: 19 of 20 contract items graded *by execution*, all three mutation arms replayed `guard_verified`, the suite at 1.06–1.54s from a 7.23–7.40s base, green on bash 3.2.57 and 5.3.9. This PR has never been reviewed, and it carries the *identical* item-8 gap, so reviewing it would arrive at the same held state — which is precisely the two-independent-reviews-on-one-issue outcome #1446's acceptance criterion (d) exists to prevent.

**Neither author is at fault.** Both drains resumed the same stale `svc:in-progress` claim on #1431 and each refreshed the label, erasing the other's evidence. That race is filed and closed as #1446 (`svc:wrong-repo` — it belongs to `cogni-work/managed-service`, not here). Two authors converging on the same override name, the same floor pattern, and the same decision to derive rather than restate the duration is good independent evidence the approach was right.

**What this PR documented and #1444 does not** — the leading-zero/octal hazard behind the `[1-9]` floor, and why the derivation uses the assignment form rather than a bare `(( ... ))` — is real and worth keeping. #1444's code is correct on both counts; only the reasoning is unrecorded there. Captured as a separate comment-only follow-up rather than lost with this branch.

PR #1460, opened 2026-08-17
fix(cogni-portfolio): memoize the append-claim stat form and bound stale-lock sweep re-entry

### LINKED ISSUE #1432: cogni-portfolio: hoist the append-claim stat-form probe out of the acquire loop; bound stale-block re-entry

claude-cogni-work:
Closes #1432.

## Summary

- `append-claim.sh` re-ran the whole `stat -c %Y || stat -f %m || echo 0` chain on **every** stale-block entry. The read now lives in a `read_lock_mtime()` helper defined before the acquire loop, which resolves the working form on first use and memoizes it in `STAT_MTIME` — so only the first entry pays the failing GNU-form exec on a BSD host, and every later entry costs a single exec. The acquire loop's body contains no `stat` invocation at all.
- The GNU-first ordering is unchanged and still the fix (`-c %Y` before `-f %m`), both probe calls keep `2>/dev/null`, and the numeric-shape floor is byte-identical and still sits between the read and the arithmetic.
- Stale-sweep re-entry is bounded by a `MAX_SWEEPS` budget, so a lock whose `rmdir` cannot succeed now reaches the existing timeout diagnostic and `exit 1` instead of sweeping forever.
- Four new suite cases, two new mutation arms; all four existing arms replayed.

## Scope decisions

**1 — Lazy memoization, not an unconditional pre-loop probe.** The issue says "resolve the working `stat` form once **before** the `while` loop". Implemented literally that regresses the hot path: when `mkdir` succeeds on the first condition test the loop body never runs, so an uncontended acquire costs exactly **one** fork and **zero** `stat`/`date` execs today. A pre-loop probe would add 1-2 forks to every one of those — the overwhelmingly common case — to save forks on a rare contended one, and would also `stat` `$LOCK_DIR`, which is not guaranteed to exist at that point (it is created only by the loop's own `mkdir`). The issue's own framing says both costs are contended-path-only, so resolving lazily on first stale-block entry delivers its intent without the regression. Case `uncontended-acquire-spawns-no-stat` pins this: an uncontended acquire must spawn zero `stat` calls.

**2 — The rmdir-fails path now terminates.** The issue left this open ("optionally reset `WAITED`, or re-check staleness every Nth iteration"). Verified against `origin/main`: the shipped script does not merely spin cheaply on that fixture, it **never returns**. `WAITED` only ever climbs, so once the ceiling test is true it stays true; the sweep's `continue` fires before the timeout branch can be reached, and an ancient lock's age only grows — so `echo … exit 1` is unreachable dead code. A `MAX_SWEEPS` budget was chosen over resetting `WAITED` because resetting merely extends the acquire budget on every sweep, leaving the loop non-terminating at a slower rate.

This **restores** the enumerated semantics rather than breaking them. The 60s threshold, the `rmdir`-then-`continue`, the timeout JSON and `exit 1` all survive textually and in the same relative order; the timeout branch becomes reachable again. The only observable change lands on a fixture that today never returns at all.

**Not changed.** The identical `stat` chains in `cogni-portfolio/hooks/ensure-excalidraw-canvas.sh` and `cogni-visual/hooks/ensure-excalidraw-canvas.sh` sit in one-shot conditionals with an atomic `mv` recovery and no enclosing retry loop, so neither shares the re-entry defect. No version bump in-branch — the post-merge workflow owns it (`check-version-bump.py`: 0 touched, 0 violations).

## Mutation recipe

All six arms **replayed** against this head (re-run, not assumed); each returned `data.verdict == guard_verified`. Run from the repo root:

```bash
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/STAT_MTIME="\$stat_form"/STAT_MTIME=""/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case probe-resolved-once-per-acquire
```

| # | `--expr` | `--case` | verdict |
|---|---|---|---|
| 1 | `s/lock_mtime=0/lock_mtime_unused=0/` | `stale-lock-swept-on-non-numeric-stat` | `guard_verified` |
| 2 | `s/\^\[0-9\]\+\$/^NOMATCH\$/` | `live-lock-not-swept-on-numeric-stat` | `guard_verified` |
| 3 | `s/APPEND_CLAIM_MAX_WAIT:-30/APPEND_CLAIM_MAX_WAIT_UNUSED:-30/` | `live-lock-not-swept-on-numeric-stat` | `guard_verified` |
| 4 | `s/stat -c %Y/stat -f %m/` | `live-lock-not-swept-on-gnu-stat` | `guard_verified` |
| 5 | `s/STAT_MTIME="\$stat_form"/STAT_MTIME=""/` | `probe-resolved-once-per-acquire` | `guard_verified` |
| 6 | `s/MAX_SWEEPS=3/MAX_SWEEPS=9/` | `stale-sweep-bounded-when-rmdir-fails` | `guard_verified` |

**Target uniqueness re-verified after the rewrite** — the harness's `perl -0pi` is first-occurrence-only, so a second occurrence would silently relocate an arm and let it report green while guarding nothing. `grep -c` on the post-fix script returns **1** for each of `lock_mtime=0`, `stat -c %Y`, `^[0-9]+$`, `APPEND_CLAIM_MAX_WAIT:-30`, `MAX_SWEEPS=3`, `STAT_MTIME="$stat_form"`. The extended in-script comment deliberately restates none of them.

**Arm 6 widens the bound rather than removing it,** and the suite header records that as a standing rule. Deleting `MAX_SWEEPS` — or dropping the `SWEEPS` increment — makes the *mutated* script non-terminating on case 5's fixture, so the harness would hang rather than report red; and because this suite has no per-case selector, one hanging case hangs the whole replay.

## Verification

| Check | Result |
|---|---|
| `bash cogni-portfolio/tests/test-append-claim-lock.sh` | exit 0, 7/7 cases |
| `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` | exit 0, 7/7 suites |
| `python3 scripts/check-result-line-plainness.py` | exit 0 |
| `python3 scripts/check-breadcrumbs.py` | exit 0 |
| `python3 scripts/check-version-bump.py` | exit 0 — 0 touched |
| `/bin/bash -n` (3.2.57) and bash 5.3.9 `-n`, both files | exit 0 |

**Measured, not asserted** (acceptance criterion 1). Case `probe-resolved-once-per-acquire` installs a counting `stat` stub on a lock whose `rmdir` always fails, driving the full sweep budget — four stale-block entries. It records exactly **1** GNU-form call and **4** BSD-form reads. An un-memoized chain records 4 of each, which is what arm 5 reproduces. The counter lives under `$TMPROOT`, never in the stub directory, whose one-file invariant is what keeps `date` and `python3` resolving to the real binaries; each case carries its own counter so counts cannot bleed between them.

**Suite runtime** grew from ~1.6s to ~4.0s (measured via `run-plugin-tests.py`, and the header prose was re-measured to match rather than left claiming "around a second"). Still well under the ~7.3s this suite cost before the ceiling became overridable. Cases 4 and 5 carry most of the increase; the header records that folding them into one invocation would recover about half if the budget ever needs tightening.

**Foreign output is asserted by shape throughout** — exit status, `[ -s ]` stream emptiness, numeric counts. Only `append-claim.sh`'s own literals (`Could not acquire lock on claims.json after 0.1s`) are text-matched, since bash and coreutils diagnostics are gettext-localized and an English grep passes vacuously off an English-locale host.

## Note on the token-scope preflight

`check-token-scope.sh --strict` exits 1 on this runner: the token is the `gho_` token from `gh auth login`, whose scope set is fixed by the GitHub CLI OAuth app and cannot be narrowed (extra scopes `gist`, `read:org`, `workflow`). The push proceeded under the documented `--allow-overscoped` opt-out, so this branch was **not** pushed under the least-privilege posture.

The branch was graded against the 22-item acceptance contract committed on #1432 at triage time (`derivedAt 2026-08-17T23:21:51Z`, before any branch was cut). 20 items came back satisfied on the first pass; two were violated and both were fixed in `812b8a9d`:

- **The loop body must hold exactly one `stat` invocation, or none if the read is delegated to a helper.** The first pass left two call sites inline — the first-use probe and the memoized read — meeting neither count. The read moved wholesale into `read_lock_mtime()`; the loop body's non-comment `stat` count is now **0**. A relocation, not a behaviour change: same candidate list, same GNU-first order, same memo, same `false` sentinel, same stderr redirects.
- **Every new case must carry its own `APPEND_CLAIM_MAX_WAIT` prefix.** The uncontended case had none. It now carries one, with a comment recording that the prefix is behaviourally inert there (that case never enters the acquire loop) and is present so no case inherits the production default by omission.

All six mutation arms were **re-replayed after the move**, since it relocates the literals two of them target; each still returns `guard_verified`, and every target still occurs exactly once.

No item was graded `unclear`, so nothing is left residual for the merger.

## Open questions

None.

### DIFF
--- a/cogni-portfolio/scripts/append-claim.sh
+++ b/cogni-portfolio/scripts/append-claim.sh
@@ -38,6 +38,22 @@ MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"
 TIMEOUT_LABEL="$(( MAX_WAIT / 10 )).$(( MAX_WAIT % 10 ))"
 TIMEOUT_LABEL="${TIMEOUT_LABEL%.0}"
 WAITED=0
+# Stale-sweep budget. A lock we have failed to remove this many times is not
+# going to become removable: the directory is non-empty, or a peer keeps
+# re-taking it inside the same tick. Without a bound the sweep below `continue`s
+# forever and the timeout diagnostic becomes unreachable, because WAITED only
+# ever climbs -- so once the ceiling test is true it stays true on every later
+# pass. The budget is what lets the loop fall through to that diagnostic instead
+# of spinning at 10Hz over a lock it cannot clear.
+MAX_SWEEPS=3
+SWEEPS=0
+# Empty means "the working stat form is not resolved yet". Resolution is
+# deliberately lazy -- deferred to the first stale-block entry rather than done
+# ahead of the loop -- because an uncontended acquire never enters the loop body
+# at all, so probing up front would add a fork to every one of those to save
+# forks on the rare contended path. The lock directory also need not exist yet
+# at this point; it is created only by the loop's own mkdir below.
+STAT_MTIME=""
 while ! mkdir "$LOCK_DIR" 2>/dev/null; do
   sleep 0.1
   WAITED=$((WAITED + 1))
@@ -58,11 +74,34 @@ while ! mkdir "$LOCK_DIR" 2>/dev/null; do
       #
       # Floor on numeric SHAPE, not emptiness — the failure mode is a successful
       # *wrong* answer, which an emptiness test structurally cannot catch.
-      lock_mtime=$(stat -c %Y "$LOCK_DIR" 2>/dev/null || stat -f %m "$LOCK_DIR" 2>/dev/null || echo 0)
+      #
+      # The candidate list is walked once per acquire and the winner memoized in
+      # STAT_MTIME, so only the first entry pays the failing probe; every later
+      # entry costs a single exec. `$stat_form` and `$STAT_MTIME` expand UNQUOTED
+      # on purpose — each holds a command plus its flags and has to word-split
+      # into three words. Both are set from this file's own literals and the
+      # format specifiers carry no glob character, so the split is safe.
+      # The sentinel for "no form worked" is the `false` builtin, which keeps the
+      # memoized path fork-free and lets the tail below absorb its non-zero
+      # status. Both branches always assign lock_mtime, which is what keeps
+      # `set -u` off the floor that follows; the assignment-as-condition form is
+      # what keeps a failing probe from tripping `set -e`.
+      if [ -z "$STAT_MTIME" ]; then
+        STAT_MTIME=false
+        for stat_form in "stat -c %Y" "stat -f %m"; do
+          if lock_mtime=$($stat_form "$LOCK_DIR" 2>/dev/null); then
+            STAT_MTIME="$stat_form"
+            break
+          fi
+        done
+      else
+        lock_mtime=$($STAT_MTIME "$LOCK_DIR" 2>/dev/null || echo 0)
+      fi
       [[ "$lock_mtime" =~ ^[0-9]+$ ]] || lock_mtime=0
       now=$(date +%s)
       lock_age=$(( now - lock_mtime ))
-      if [ "$lock_age" -gt 60 ]; then
+      if [ "$lock_age" -gt 60 ] && [ "$SWEEPS" -lt "$MAX_SWEEPS" ]; then
+        SWEEPS=$((SWEEPS + 1))
         rmdir "$LOCK_DIR" 2>/dev/null || true
         continue
       fi
--- a/cogni-portfolio/scripts/append-claim.sh
+++ b/cogni-portfolio/scripts/append-claim.sh
@@ -54,17 +54,51 @@ SWEEPS=0
 # forks on the rare contended path. The lock directory also need not exist yet
 # at this point; it is created only by the loop's own mkdir below.
 STAT_MTIME=""
+# Read the lock's mtime into lock_mtime, resolving the working stat form on first
+# use and reusing it thereafter. This lives in a function rather than inline so
+# the acquire loop's body carries no stat invocation of its own: the loop asks
+# for a reading, and which form produces it is this function's business.
+#
+# GNU form first: on GNU coreutils `-f` means --file-system, so `stat -f %m`
+# yields no mtime yet still exits 0, and a fallback chain would never fall
+# through. BSD has no `-c`, so it errors cleanly and does fall through — only
+# this direction works on both. The candidate list is therefore ordered, not
+# arbitrary, and reversing it is what the ordering mutation arm exists to catch.
+#
+# The winner is memoized in STAT_MTIME, so only the first entry pays the failing
+# probe; every later entry costs a single exec. `$stat_form` and `$STAT_MTIME`
+# expand UNQUOTED on purpose — each holds a command plus its flags and has to
+# word-split into three words. Both are set from this file's own literals and the
+# format specifiers carry no glob character, so the split is safe.
+#
+# The sentinel for "no form worked" is the `false` builtin, which keeps the
+# memoized path fork-free and lets the read's tail absorb its non-zero status.
+# Every path assigns lock_mtime — a failing substitution still assigns the empty
+# string — which is what keeps `set -u` off the shape floor at the call site, and
+# the assignment-as-condition form is what keeps a failing probe from tripping
+# `set -e`. The function itself always returns 0 for the same reason.
+#
+# Both probe calls keep their stderr redirect, so a real BSD host's complaint
+# about the GNU flag never reaches the caller's stderr.
+read_lock_mtime() {
+  if [ -z "$STAT_MTIME" ]; then
+    STAT_MTIME=false
+    for stat_form in "stat -c %Y" "stat -f %m"; do
+      if lock_mtime=$($stat_form "$LOCK_DIR" 2>/dev/null); then
+        STAT_MTIME="$stat_form"
+        return 0
+      fi
+    done
+    return 0
+  fi
+  lock_mtime=$($STAT_MTIME "$LOCK_DIR" 2>/dev/null || echo 0)
+}
 while ! mkdir "$LOCK_DIR" 2>/dev/null; do
   sleep 0.1
   WAITED=$((WAITED + 1))
   if [ "$WAITED" -ge "$MAX_WAIT" ]; then
     # Stale lock detection: remove lock older than 60 seconds
     if [ -d "$LOCK_DIR" ]; then
-      # GNU form first: on GNU coreutils `-f` means --file-system, so `stat -f %m`
-      # yields no mtime yet still exits 0, and the `||` never falls through. BSD
-      # has no `-c`, so it errors cleanly and does fall through — only this
-      # direction works on both.
-      #
       # Resolve the reading into a variable before any arithmetic. A bad
       # substitution nested directly inside $(( )) is a hard bash error, and the
       # damage is worse than an abort: on every bash tested (3.2.57 through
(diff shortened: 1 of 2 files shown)


claude-cogni-work:
## Ship-ready

Draft PR handed off by the author instance for independent merger review. Resolves #1432.

**Head:** `812b8a9d` · **Base:** `origin/main` @ `5c75df5d` · two commits.

### Handoff preflight

Clean — exit 0, **3 passed, 4 skipped, 0 failed**. Skipped instruments, each with its reason: `frontmatter` (`not_applicable`: no touched SKILL.md or agents/*.md), `breadcrumbs` (same), `skill-length` (no touched SKILL.md), `agent-length` (no touched agents/*.md). Passed: `run-tests` (suites green for cogni-portfolio), `version-bump` (no version-line touch), `mutation-recipe` (recorded recipe runnable as written).

Graded against the 22-item contract committed on #1432 at triage time (`derivedAt 2026-08-17T23:21:51Z`, `origin: triage`) — written before any branch was cut, and not by this author.

| | |
|---|---|
| Items graded | 22 |
| Satisfied on first pass | 20 |
| Violated | 2 — **both fixed** in `812b8a9d` |
| Unclear | 0 |
| **Residual for the merger** | **none** |

The two violations and their fixes:

1. **Loop body must hold exactly one `stat` invocation, or none if delegated to a helper.** The first pass left two call sites inline — the first-use probe and the memoized read — so it met neither count. The read moved wholesale into `read_lock_mtime()`, defined before the acquire loop; the loop body's non-comment `stat` count is now **0**. A relocation, not a behaviour change: same candidate list, same GNU-first order, same memo, same `false` sentinel, same stderr redirects.
2. **Every new case must carry its own `APPEND_CLAIM_MAX_WAIT` prefix.** The uncontended case had none. It now carries one, with a comment recording that the prefix is behaviourally inert there — that case never enters the acquire loop — and is present so no case inherits the production default by omission.

### Worth the merger's attention

- **The helper move relocated the literals two mutation arms target.** All six arms were **re-replayed after the move** (re-run, not assumed) and each returned `data.verdict == guard_verified`; `grep -c` confirms each of the six targets still occurs exactly once. The harness's `perl -0pi` is first-occurrence-only, so this is the check that keeps an arm from silently going inert while still reporting green.
- **Two design calls the issue deliberately left open**, both argued in the PR body: the probe resolves *lazily* rather than ahead of the loop (an unconditional pre-loop probe would tax the uncontended acquire, which costs zero `stat`/`date` execs today), and the rmdir-fails path now *terminates* rather than merely spinning more cheaply. The second is the one to scrutinise — it is a behavioural change beyond the literal ask. The argument for it is that it *restores* the enumerated semantics: the shipped script never returns on that fixture, so the timeout diagnostic and `exit 1` are unreachable dead code, and the budget makes that branch reachable again while leaving the 60s threshold, the `rmdir`-then-`continue`, the JSON shape and the exit code textually unchanged.
- **The token-scope preflight does not pass under `--strict`** on this runner (exit 1; extra scopes `gist`, `read:org`, `workflow`). This is the `gho_` token from `gh auth login`, whose scope set is fixed by the GitHub CLI OAuth app and cannot be narrowed. The push proceeded under the documented `--allow-overscoped` opt-out, so this branch was **not** pushed under the least-privilege posture.

No review verdict, no `svc:review-*` label and no reviewed-head marker were applied by the author — the independent review is the merger's.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1460

**Summary:** All 22 review-contract checklist items satisfied by execution — all four pre-existing mutation arms plus both new arms replay to guard_verified, the counting-stub cases genuinely measure the spawn reduction (1 probe + 4 memoized reads across 4 stale entries) and the sweep bound (timeout reached in 4 reads), and the probe is deferred inside the `[ -d "$LOCK_DIR" ]` branch so mutual exclusion is preserved. Two minor preference findings on comment precision and suite runtime; neither gates.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-portfolio/scripts/append-claim.sh:74-75` — Comment precision: the `false` sentinel is described as keeping the memoized path "fork-free", but the memoized read is a command substitution, which forks a subshell on every bash. What the sentinel avoids is an exec of an external binary, not a fork. The same loose wording recurs at tests/test-append-claim-lock.sh:489 and, in a different sense, at tests line 52.
  - *Suggested fix:* Say "exec-free" / "spawns no external binary" at append-claim.sh:75 and tests line 489; at tests line 52 say "stays free of stat spawns" to match what the counter actually asserts.

- **[minor]** (waived) (advisory-addition) `cogni-portfolio/tests/test-append-claim-lock.sh:44-50` — The suite header discloses that the suite now runs near four seconds, grown by ~2.4s, with cases 4 and 5 carrying most of it — partly giving back the reduction the immediately preceding merge to this same suite achieved (main 51349586, 'cut append-claim lock suite from 7.9s to under 2s'). Measured at head: 4.054s standalone, 3.6s under run-plugin-tests. Still well under the ~7.3s pre-optimization cost, and the header names its own mitigation (folding cases 4 and 5 into one invocation emitting two result lines).
  - *Justification:* Runtime cost is disclosed in the suite header with a named mitigation and sits inside the contract's stated bar; an observation for a follow-up, not a defect.

### Positive observations

- 22/22 contract checklist items satisfied with executed evidence (cogni-service:contract-grading-reviewer)
- All FOUR pre-existing mutation arms replayed live to guard_verified, including the GNU-first ordering arm the hoist most endangers
- TWO new arms recorded and replayed to guard_verified: probe memoization (STAT_MTIME emptied) and re-entry bound (MAX_SWEEPS=3 -> 9)
- Item 12 measured, not asserted: 1 GNU-form probe + 4 memoized reads across 4 stale-block entries
- Item 13 measured: unremovable lock reached the timeout branch in 4 reads, deterministic, no race wrapper
- Item 8 verified directly by the gatekeeper: read_lock_mtime is called only at append-claim.sh:111, inside `[ -d "$LOCK_DIR" ]` at :101 — no pre-loop probe against a not-yet-created LOCK_DIR, so mutual exclusion on GNU is preserved
- Item 4 verified directly: the numeric-shape floor appears verbatim exactly once at :112, still between the mtime read (:111) and the lock_age arithmetic (:114)
- Item 17 verified directly: stat -c %Y, ^[0-9]+$, lock_mtime=0, APPEND_CLAIM_MAX_WAIT:-30 and MAX_SWEEPS=3 each occur exactly once, so no recorded arm degraded to a no-op
- run-plugin-tests --filter cogni-portfolio exits 0 (7/7 suites); check-result-line-plainness.py and check-breadcrumbs.py both 0 violations
- /bin/bash -n clean on both files under the host's 3.2.57; no declare -A, mapfile, ${x^^} or globstar in added lines
- Scope boundary exact and no version touched: plugin.json 0.9.81 mirrors marketplace.json, check-version-bump.py clean
- No downstream reference breaks: append-claim.sh keeps its two-argument CLI, exit-code contract and APPEND_CLAIM_MAX_WAIT surface, so all six agent and one skill call sites still resolve

### Recommendation

---

PR #1464, opened 2026-08-18
cogni-portfolio: record append-claim.sh's three guard rationales

### LINKED ISSUE #1448: cogni-portfolio: append-claim.sh guards two real hazards with no recorded rationale
## Problem

`cogni-portfolio/scripts/append-claim.sh` now guards two real hazards, and the code is correct on both — but neither reason is written down. The guards therefore read as arbitrary, and a later "cleanup" that harmonises them back is the obvious next edit.

Both were documented on the closed duplicate PR #1445 and are lost with its branch. This issue re-homes the reasoning into the surviving file.

### 1. Why the shape floor leads with `[1-9]`

The ceiling's floor is `^[1-9][0-9]*$`, while the stale-lock reading below it uses `^[0-9]+$`. The divergence looks like an inconsistency and is not: a value like `08` or `030` passes a bare `^[0-9]+$`, and arithmetic expansion then reads the leading zero as **octal** and hard-aborts the script. The two patterns must not be harmonised.

Zero is excluded on the same line, also deliberately: the loop sleeps *before* its first `-ge` check, so `0` and `1` both mean one 0.1s poll — `0` buys no fail-immediately semantics and would render a misleading `after 0s` for a path that did in fact wait.

### 2. Why the derivation uses the assignment form

The label arithmetic is written as assignments, never as a bare `(( ... ))` command. A bare arithmetic command evaluating to 0 returns status 1, and the default ceiling of 30 makes the remainder exactly 0 — so under `set -e` that aborts on bash 5.x while passing on 3.2. A version-divergent failure on the **default** path, which is the one every production caller takes.

### 3. Placement of the derivation

The computation sits at top level, outside the acquire loop, and that is load-bearing. An arithmetic abort *inside* the loop is the failure the file's own comment further down documents: it abandons the loop, resumes after `done`, and appends the claim having never held the lock — after which the EXIT trap removes a live peer's lock directory. At top level the same abort is a clean exit before any lock is taken.

## Proposed change

Comment-only. Add the three rationales to `cogni-portfolio/scripts/append-claim.sh` at the lines they explain. No behaviour change, no test change.

Optionally, fold in the one clause the #1444 review flagged under contract item 8 and the maintainer waived: state in the header block that the printed timeout duration is *derived* from the ceiling and will never restate a fixed number. The waiver ruled the contract over-specified placement, not that the fact is unwanted in the header — so it is cheap to add here alongside the rest.

## Acceptance criteria

1. The `[1-9]` divergence from the neighbouring `^[0-9]+$` carries a comment naming the octal-abort hazard and saying explicitly not to harmonise the two patterns.
2. The assignment-form arithmetic carries a comment naming the bare-`(( ... ))` status-1 hazard and its bash 5.x / 3.2 divergence on the default path.
3. The derivation's top-level placement carries a comment naming the in-loop abort consequence.
4. No behavioural change: `run-plugin-tests.py --filter cogni-portfolio` stays 7/7, and the lock suite stays under 2s.

## Context

Source: PR #1445, closed as the duplicate of the merged #1444 (issue #1431). Both PRs were authored independently from the same stale claim; #1445's branch carried the fuller rationale and #1444's carried the reviewed code, so the reasoning did not travel with the merge.

claude-cogni-work:
Closes #1448.

## Summary

Comment-only. Three guards in `cogni-portfolio/scripts/append-claim.sh` were correct but unexplained, so they read as arbitrary and a harmonising "cleanup" was the obvious next edit. This records the reasoning next to each one.

- **The shape floor** now names the octal-abort hazard behind its leading `[1-9]`, states explicitly that it must **not** be harmonised with the plain-digit floor the stale-lock reading further down uses, and records why zero is excluded on the same line (the loop sleeps *before* its first `-ge` check, so a ceiling of 0 and of 1 both mean exactly one 0.1s poll, and 0 would render a misleading `after 0s` for a path that did wait).
- **The label derivation** now names the bare-arithmetic-command hazard: such a command whose expression evaluates to 0 returns exit status 1, and the default ceiling of 30 ticks makes the remainder exactly 0 — so that form aborts under `set -e` on bash 5.x while passing on 3.2, a version-divergent failure on the **default** path every production caller takes.
- **That derivation's top-level placement** now names the in-loop consequence the stale-lock block already documents: an arithmetic abort inside the loop abandons the loop body and the loop, resumes after `done`, appends the claim having never held the lock, and lets the release trap remove a live peer's lock directory. Up top the same abort is a clean exit before any lock is taken.
- **The header** gains the clause an earlier review flagged and the maintainer waived: the duration the timeout diagnostic prints is *derived* from the ceiling and never restates a fixed number.

30 added lines, all comments. Zero executable lines changed, zero lines removed.

## Anchor safety — the load-bearing constraint the issue does not state

`cogni-portfolio/tests/test-append-claim-lock.sh` records six mutation arms, each asserting its expression's anchor occurs **exactly once** in `append-claim.sh`, and the harness substitutes with a whole-file slurp and a **non-global** match — so only the first occurrence top-to-bottom is replaced.

All four edits here sit *earlier* in the file than every anchor site. New prose that reproduced an anchor literal would therefore silently hijack that arm: the real code line survives untouched, the mutated script behaves identically, the named case stays green, and the arm reports as un-caught — while also falsifying the suite's own uniqueness claims. A guard that silently stops guarding, with everything green.

Every cross-reference in the new comments is consequently written **by description** ("the plain-digit floor the stale-lock reading further down uses"), never by reproducing the referenced token. All six anchors still occur exactly once — verified below.

## Verification

- `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` → **7/7 suites pass**, exit 0.
- `bash -n cogni-portfolio/scripts/append-claim.sh` → exit 0.
- Comment-only shape: added non-comment lines **0**, removed lines **0**.
- Anchor uniqueness: each of the six recorded anchors returns **1** on the head file.
- No breadcrumbs: no issue/PR reference in the file.
- No version field touched — the post-merge workflow owns the bump.
- **Mutation replay run, not just described.** The arm this diff sits closest to was replayed verbatim: `mutated_result: red`, `restored_result: green`, `verdict: guard_verified`. The do-not-harmonise prose did not absorb the substitution.

## A note on acceptance criterion 4

AC 4 asks the lock suite to stay "under 2s". The suite's own measured header note records it near four seconds ("measured, not estimated"), so that figure is stale and fails against **unmodified** base code — a comment-only diff can neither hit nor miss it. The bar discharged here is no-regression instead, measured both ways: **3.63s on `origin/main`, 3.63s on this branch**.

This is a defect in the issue's acceptance bar, not in the change — flagged rather than silently satisfied or silently failed.

## Mutation recipe

Runnable as written from the repo root:

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/\^\[0-9\]\+\$/^NOMATCH\$/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case live-lock-not-swept-on-numeric-stat
```

Expected: the case goes red under the mutation and green when restored. If a newer harness version directory exists under the same parent, use it — everything after the path is version-independent.

### DIFF
--- a/cogni-portfolio/scripts/append-claim.sh
+++ b/cogni-portfolio/scripts/append-claim.sh
@@ -8,6 +8,9 @@
 # Env: APPEND_CLAIM_MAX_WAIT — lock-acquire ceiling in 0.1s ticks. Unset means
 # the production default of 30 ticks (3s); tests set a small value so they can
 # drive the timeout path in fractions of a second.
+# The duration the timeout diagnostic prints is DERIVED from that ceiling and
+# never restates a fixed number, so the message cannot outlive a change to the
+# ceiling it reports on.
 # Exit codes: 0 = success, 1 = error
 set -euo pipefail
 
@@ -32,9 +35,36 @@ MAX_WAIT="${APPEND_CLAIM_MAX_WAIT:-30}"
 # comparison below a hard error mid-loop. Regex unquoted on purpose — a quoted
 # right-hand side matches literally on bash 3.2, which this repo still targets.
 # The `||` form keeps a non-match from tripping `set -e`.
+#
+# The `[1-9]` lead is deliberate and must NOT be harmonised with the plain-digit
+# floor the stale-lock reading further down uses. A value like 08 or 030
+# satisfies a plain-digit pattern, and arithmetic expansion then reads the
+# leading zero as octal — a hard error, which under `set -e` aborts the script
+# outright. The two patterns look inconsistent because they guard different
+# inputs; unifying them reintroduces that abort on the very values this floor
+# exists to reject.
+#
+# Zero is excluded on the same line for its own reason. The loop below sleeps
+# BEFORE its first -ge check, so a ceiling of 0 and a ceiling of 1 both mean
+# exactly one 0.1s poll — 0 buys no fail-immediately semantics, and it would
+# render a misleading `after 0s` for a path that did in fact wait.
 [[ "$MAX_WAIT" =~ ^[1-9][0-9]*$ ]] || MAX_WAIT=30
 # Derive the diagnostic's number from the ceiling rather than restating it, so
 # the message can never claim a wait the loop no longer honours.
+#
+# The arithmetic is written as ASSIGNMENTS, never as a bare `(( ... ))` command.
+# A bare arithmetic command whose expression evaluates to 0 returns exit status
+# 1, and the default ceiling of 30 ticks makes the remainder exactly 0 — so that
+# form aborts under `set -e` on bash 5.x while passing on 3.2. A version-divergent
+# failure on the DEFAULT path, which is the one every production caller takes.
+#
+# The computation sits at TOP LEVEL, before the acquire loop, and that placement
+# is load-bearing rather than incidental. An arithmetic abort *inside* the loop
+# is the failure the stale-lock block's own comment further down documents: the
+# error abandons the loop body AND the loop, so the script resumes after `done`
+# and appends the claim having never held the lock — after which the release trap
+# removes a live peer's lock directory it never owned. Up here the same abort is
+# a clean exit, before any lock is taken and before that trap is installed.
 TIMEOUT_LABEL="$(( MAX_WAIT / 10 )).$(( MAX_WAIT % 10 ))"
 TIMEOUT_LABEL="${TIMEOUT_LABEL%.0}"
 WAITED=0


claude-cogni-work:
## Ship-ready

Author instance is done. This draft PR is handed off to the merger instance for independent review — it carries `svc:needs-review` and deliberately no `svc:review-*` verdict, no `reviewed-head-sha` marker, and no ready flip. The review is the merger's to run.

**Issue:** #1448 — `svc:in-progress` → `svc:resolved`.
**Branch:** `fix/issue-1448-append-claim-guard-rationales` @ `dcd9422a` (forked from `origin/main` @ `3dd11cfe`).

### Handoff preflight

Clean, RC=0 — 2 passed, 0 failed, 5 skipped:

|---|---|---|
| run-tests | pass | suites green for cogni-portfolio |
| frontmatter | skip — not_applicable | no touched SKILL.md or agents/*.md |
| breadcrumbs | skip — not_applicable | no touched SKILL.md or agents/*.md |
| skill-length | skip — not_applicable | no touched SKILL.md |
| agent-length | skip — not_applicable | no touched agents/*.md |
| mutation-recipe | skip — not_applicable | no touched `*/tests/test-*.sh` or `*/scripts/check-*.sh` survives at head |

The contract was derived at triage from the issue alone, before any branch was cut, and posted to #1448 (`--origin triage`). Grading it here costs one agent instead of a merger bounce; it is **not** a substitute for the merger's own independent review.

**Graded 15 · satisfied 15 · violated 0 · unclear 0 — no fix pass needed, no residual.**

The two items worth the merger's attention, because they are where a plausible-looking diff would have failed silently:

- **Anchor uniqueness (contract item 4).** The lock suite's six mutation arms each assume their anchor occurs exactly once, and the harness substitutes with a whole-file slurp and a non-global match. Every edit in this diff sits *earlier* in the file than every anchor site, so prose restating an anchor would have hijacked that arm — real code untouched, case still green, arm falsely reporting as un-caught. All six anchors still return 1 on the head file, and the closest arm was replayed verbatim: `guard_verified` (red mutated, green restored).
- **Acceptance criterion 4's timing figure.** The AC asks for "under 2s"; the suite's own measured header note records it near four seconds, so that number fails against *unmodified base code*. Measured both ways for this PR: **3.63s on `origin/main`, 3.63s on the branch** — no regression. The contract itself flags the literal 2s bar as a defect in the issue's acceptance criteria (`kind: avoidable`, so it does not hold) and directs that it be graded as no-regression. Recorded here so the merger does not bounce a correct diff on a stale number.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1464

**Summary:** Both perspectives pass: contract-grading-reviewer confirms all 15 acceptance-checklist items satisfied by execution with zero findings, and plugin-validator passes with a single waived preference-only minor finding (opt_out justified), so the PR is safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-validator | ✅ pass |

### Findings (1)

- **[minor]** (waived) (advisory-addition) `cogni-portfolio/scripts/append-claim.sh:11-13 (and existing :52-53)` — The added header sentence ('The duration the timeout diagnostic prints is DERIVED from that ceiling and never restates a fixed number...') restates, in the file's env-documentation block, the rationale that already sat inline at the derivation site on lines 52-53. It creates two places that must stay in agreement if the derivation ever changes. Not a fault, and it is expressly sanctioned: the issue's Proposed change offers it as an optional fold-in and review-contract item 11 grades it as optional. The two framings also serve different readers (env-var consumer vs. code-site maintainer).
  - *Justification:* Purely a placement preference with no user-observable effect; the clause is explicitly optional under contract item 11 and the issue's own Proposed change section.

### Positive observations

- Comment-only shape verified mechanically, not asserted: the added hunk contains zero non-comment, non-blank lines; the diff is +30 -0 with no executable line touched.
- The three rationale comments are factually correct, not plausible-sounding: the bash 5.x-vs-3.2 divergence on `(( 30 % 10 ))` and the octal abort on `08` both reproduce on this host exactly as described.
- Contract item 4 (the highest-value item) holds: grep -cF returns exactly 1 for each of the six recorded mutation anchors, so no arm is silently redirected into a comment. Cross-references are written by paraphrase rather than reproducing anchor literals.
- Contract item 5 holds: 0 deletions; the bash-3.2 unquoted-regex caveat, the lazy-resolution language and the GNU-first ordering rationale all survive verbatim.
- AC1 carries both halves: the octal hazard named concretely, plus the explicit imperative forbidding harmonisation with the neighbouring plain-digit floor.
- AC2 names all four required elements, including the status-1 mechanism, the 30-remainder-0 trigger, and the 5.x-aborts/3.2-passes divergence on the default path.
- AC3 ties the top-level placement to a concrete consequence, through the EXIT trap removing a live peer's lock directory, and corroborates the stale-lock block's own existing comment.
- Behaviour unchanged, verified by execution: run-plugin-tests.py --filter cogni-portfolio exits 0 with 7/7 suites passing; the lock suite prints all seven case labels as ok: lines and 'All tests passed.' with no FAIL: line.
- No feature-branch version bump: manifest and marketplace mirror untouched, check-version-bump reports 0 violations.
- No maintainer breadcrumbs on the head file, graded by grep rather than the vacuous check-breadcrumbs glob.
- Runtime graded as no-regression per contract item 15's explicit override of the issue AC's literal 2s figure: the lock suite measured 4.17s against base's own ~4s note.

### Recommendation

---

PR #1485, opened 2026-08-18
Repo-scope MCP agent-grant guard with an affirmative call-site detector

claude-cogni-work:
Closes #1474.

## Summary

New repo-scope guard `scripts/check-mcp-tool-grant.py` (stdlib-only, `--root`, `{success,data,error}` envelope, exit 0/1/2) scanning every `*/agents/*.md`, with two arms:

- **Affirmative call site** — a literal `mcp__<ns>__<tool>` token in an agent **body** must be granted by that agent's own `tools:` frontmatter. Registry-independent, so it also covers deliberately-unregistered namespaces (`browsermcp`, `claude-in-chrome`).
- **Registry coverage** — every plugin granting a *registered* namespace must appear in that server's `required_by`. Key→namespace is read from `desktop_config_key` **as data**, never by stripping a prefix.

Plus a companion suite `tests/test_check_mcp_tool_grant.sh` (37 cases) and a `mcp-tool-grant-guard` job in `lint.yml`.

## Two premises in the issue turned out to be wrong

Both were verified against `origin/main` before planning, and both change the shape of the fix:

1. **The issue says a YAML block list is "used by nothing yet".** It is used by **15** agents — and **6 of the 7** agents carrying a real body call site declare tools that way. A `tools:`-line-only parser therefore reports those 6 as false offenders and the guard is **red on a clean tree**. Block-list parsing is load-bearing, not a nicety.
2. **The `required_by` arm is red on base.** `cogni-workspace` grants `mcp__excalidraw__*` and `mcp__pencil__*` (it adopted those agents) and appears in **neither** array. That omission *is* the second-layer defect the arm exists to catch, so it is fixed rather than exempted. `CLAUDE.md`'s MCP table is updated to match, since that data is now CI-enforced and would otherwise have drifted in the very commit that enforced it.

A third fact emerged during review: the namespace token class **must admit hyphens**. `re.findall(r'mcp__([A-Za-z0-9_]+?)__[A-Za-z0-9_]+', 'mcp__claude-in-chrome__navigate')` returns `[]` — an underscore-only class is blind to two agents and a whole namespace with every case still green. The pre-existing cogni-website suite carried that underscore-only class; it is synced here (behaviour-neutral today, since that plugin grants only `pencil`).

## Scope decision: two tiers, one owner per invariant

`cogni-website/tests/test-mcp-tool-grant.sh` **keeps the prose arm** and gives up the `required_by` arm.

[...]

PR #1461, opened 2026-08-17
cogni-portfolio: append-claim.sh emits the standard JSON envelope

claude-cogni-work:
Closes #1434.

## Summary

`cogni-portfolio/scripts/append-claim.sh` now emits the repo-wide `{"success", "data", "error"}` envelope on **every** path that writes to stdout or stderr. The issue names two paths; there are four.

| Path | Was | Now |
|---|---|---|
| Usage guard | `{"error": "Usage: …"}` | full envelope, stderr, exit 1 |
| Lock timeout | `{"error": "Could not acquire lock …"}` | full envelope, stderr, exit 1 |
| Success print | `{"status": "appended", …}` | `{"success": true, "data": {…}, "error": null}`, stdout, exit 0 |
| Inline python body | **raw traceback** | full envelope, stderr, exit 1 |

The fourth is the one worth flagging. The inline `python3` block escaped a `JSONDecodeError` traceback on a malformed claim argument or an unreadable `claims.json`. That is the *likeliest* failure in production, because every agent caller hand-builds the claim JSON as a heredoc literal — so documenting a total stderr contract while leaving it emitting a traceback would have made the contract look total when it wasn't. It now uses the `try/except` form `propagate-corrections.sh:109,202` already uses.

The success payload loses nothing: `status`, `claim_id` and `total` moved under `data`. Stream assignment and exit codes are unchanged on all paths.

## What was preserved deliberately

- **The ceiling-to-message coupling.** The timeout string still interpolates `${TIMEOUT_LABEL}` (derived at script lines 38-39 from `MAX_WAIT`). Flattening it to a constant would have left recorded mutation arm 3 grading nothing while the suite stayed green.
- **The foreign-tool-by-shape discriminator.** Case 1's stderr check remains `[ -s "$TMPROOT/case1.err" ]` emptiness, not a text match on a bash diagnostic.
- **All four pre-existing mutation arms.** Their `--expr` targets (`lock_mtime=0`, `^[0-9]+$`, `APPEND_CLAIM_MAX_WAIT:-30`, `stat -c %Y`) each still match exactly once.

## Test changes

Case 1's stdout assertion now checks the whole envelope. It **indexes rather than `.get()`s**, which matters: `d.get("error") is None` is satisfied by an *absent* `error` key, so the `.get()` form passes on `{"success": true, "data": {"status": "appended"}}` — a hole in the very envelope the new arm exists to pin. Verified both ways before choosing the bracket form.

[...]

PR #1486, opened 2026-08-18
cogni-portfolio: assert append-claim error envelopes structurally

### LINKED ISSUE #1459: cogni-portfolio: append-claim.sh error envelopes are asserted by message only, not by shape
## Context

Deferred from #1434, which brought `cogni-portfolio/scripts/append-claim.sh` onto the `{"success", "data", "error"}` output envelope on all three emit paths.

That change updated the suite's **success**-path assertion to check the full envelope (`success`, `data.status`, `error`). The two **error** paths were left asserted as they were.

## Current behavior

`cogni-portfolio/tests/test-append-claim-lock.sh` asserts the error envelopes only by their message text:

- cases 2 and 3 (lines 236, 285 at the time of filing) `grep -q 'Could not acquire lock on claims.json after 0.3s'` against stderr — a match on the message substring, which sits inside the envelope but says nothing about the envelope's own keys
- the usage guard (`append-claim.sh` line 18) has no case at all

So after #1434 the error paths could regress their `success` / `data` keys — emitting the bare `{"error": …}` shape the issue removed — and the suite would stay green, because the message substring is unchanged.

## Desired behavior

Assert the error envelopes structurally, not just by message:

- cases 2 and 3 additionally parse stderr as JSON and check `success is false` and `data is null`, keeping the existing message grep (it is the coupling that mutation arm 3 relies on, and must not be replaced by the structural check)
- a new case covers the usage guard: invoking the script with no arguments writes a single `{"success": false, "data": null, "error": "Usage: …"}` object to stderr and exits 1

## Why this was deferred rather than folded in

Lines 236 and 285 are the ceiling-to-message coupling that `#1434`'s mutation arm 3 pins. Widening them in the same change would have put a second, independently-reviewable concern onto a diff whose whole point was the output shape — and would have made it harder to see that the coupling survived. Splitting keeps both readable.

## Acceptance criteria

- [ ] Cases 2 and 3 assert `success is false` and `data is null` on the stderr envelope **in addition to** the existing message grep. Falsified by either check replacing the grep rather than joining it.
- [ ] A case covers the usage-guard path: no arguments produces exactly one JSON object on stderr with `success: false`, `data: null`, a non-empty `error`, and exit status 1. Falsified by the case asserting on wording of a bash or coreutils diagnostic rather than on the script's own emitted literal and on shape.
- [ ] Every mutation recipe recorded in the suite header still replays to `guard_verified`, and each `--expr` still matches exactly once in the script.
- [ ] `bash cogni-portfolio/tests/test-append-claim-lock.sh` exits 0 and `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` exits 0.

claude-cogni-work:
Closes #1459.

## Summary

`append-claim.sh` documents a single output contract — `{"success", "data", "error"}` on both streams — but the lock suite only ever asserted that contract on the SUCCESS path. Both failure paths were pinned by message text alone, so the envelope around the message could regress to the bare `{"error": ...}` shape with the suite fully green.

- **Cases 2 and 3** gain a structural stderr check alongside their existing timeout-message grep: the captured stderr must parse as an object with `success=false` and `data=null`. The grep is kept **verbatim, not replaced** — it pins the DERIVED ceiling label (`after 0.3s`), a separate coupling and the one the third mutation arm rides on.
- **New case 8** (`no-arguments-emits-usage-envelope`) covers the argument guard — the script's only failure path that never reaches the acquire loop, and which no case previously exercised. It asserts exit 1, empty stdout, exactly one newline-terminated stderr line, the full `{success:false, data:null, non-empty error}` shape, and the script's own `Usage: append-claim.sh` literal.
- **Two mutation arms** (eighth and ninth) are recorded that make the added checks evidence rather than decoration.
- Header prose that a second non-contending case falsifies is de-staled in the same pass (the "Each case builds a minimal claims project" preamble and the "Case 7 is the deliberate exception" note).

`cogni-portfolio/scripts/append-claim.sh` is **not edited**. All seven pre-existing arms target that file, so their exactly-once match counts cannot move.

## Why the greps were joined, not replaced

The issue's first acceptance criterion is falsified by either check *replacing* the grep. They pin different things: the grep pins the ceiling-to-message derivation, the new check pins the envelope keys. The eighth arm below demonstrates the gap concretely — under it the message still matches and only the structural check reds.

## Foreign vs own-emitter assertions

`Usage: append-claim.sh` is this repo's own script's literal, which is what makes a text match legitimate at all. Everything the shell or coreutils could emit is asserted by SHAPE — exit status, stdout emptiness, a line count — because those diagnostics are gettext-localized and a text match on one would pass vacuously off an English-locale host. No `LC_ALL` pin was added; none is needed, and the repo forbids one as a remedy.

The line count is a real assertion, not a formality: the guard's `${1:-}` defaults are what keep `set -u` from adding a second line, so a regression to bare `$1` surfaces as a count of 2. `wc -l` reads from a **redirect**, never a filename argument — the argument form appends the filename to wc's output, which `tr -d ' '` does not strip, and the numeric comparison then dies on a non-numeric operand. Case 7 is the precedent.

## Mutation recipe

Both new arms are recorded in the suite's own header alongside arms 1-7 and reproduced here. **Every arm below was replayed, not asserted.**

```
# Eighth arm — the TIMEOUT envelope's SHAPE, as distinct from its message:
  --root . \
  --expr 's/\\"data\\": null, //' \
  --case live-lock-not-swept-on-numeric-stat

# Ninth arm — the USAGE envelope's shape:
  --root . \
  --expr 's/"data": null, "error": "Usage/"error": "Usage/' \
  --case no-arguments-emits-usage-envelope
```

Each arm drops the `data` key from one envelope, leaving valid JSON, an untouched message, unchanged exit status and unchanged stream routing — so only the new structural check sees it.

### Replay results (measured on this branch)

| Arm | Case | Verdict |
|---|---|---|
| 1 | `stale-lock-swept-on-non-numeric-stat` | `guard_verified` |
| 2 | `live-lock-not-swept-on-numeric-stat` | `guard_verified` |
| 3 | `live-lock-not-swept-on-numeric-stat` | `guard_verified` |
| 4 | `live-lock-not-swept-on-gnu-stat` | `guard_verified` |
| 5 | `probe-resolved-once-per-acquire` | `guard_verified` |
| 6 | `stale-sweep-bounded-when-rmdir-fails` | `guard_verified` |
| 7 | `stale-lock-swept-on-non-numeric-stat` | `guard_verified` |
| **8** | `live-lock-not-swept-on-numeric-stat` | **`guard_verified`** |
| **8** | `live-lock-not-swept-on-gnu-stat` (co-catcher) | **`guard_verified`** |
| **8** | `stale-sweep-bounded-when-rmdir-fails` (declared NON-catcher) | `vacuous_guard` — as documented |
| **9** | `no-arguments-emits-usage-envelope` | **`guard_verified`** |

**Non-vacuity proof.** Both new arms were also replayed against the **base-ref suite** in a separate detached worktree at `origin/main`: arm 8 returns `vacuous_guard` and arm 9 returns `case_not_found`. The new assertions are therefore what discriminate, not something already covered.

**Match counts.** All nine `--expr` patterns match **exactly once** in `append-claim.sh`, counted with the same regex engine the harness uses (`perl -0777`). Neither new expression carries a `^`/`$` anchor, so neither needs `/m` under the harness's `perl -0pi`.

## Test plan

- [x] `bash cogni-portfolio/tests/test-append-claim-lock.sh` — 8/8 `ok`, exits 0, 3.35s (header's "near four seconds" claim still holds)
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` — 7/7 passed, exits 0
- [x] Full repo sweep **post-commit** — 127/127 passed, 0 failed
- [x] All nine mutation arms replay to `guard_verified`; declared catcher and non-catcher sets confirmed by execution
- [x] Both new arms are `vacuous_guard` / `case_not_found` against the base ref
- [x] `python3 scripts/check-result-line-plainness.py` — 0 violations (the `pass`/`fail` emitters are untouched plain `printf`)
- [x] `python3 scripts/check-breadcrumbs.py` — 0 new breadcrumbs
- [x] No version line touched in any `plugin.json` or `marketplace.json` — the patch bump is applied post-merge

## Scope decisions

- **Case 5 deliberately does not get the structural check.** It reaches the same timeout envelope but is the sweep-budget bound, its assertions are about re-entry, and it already carries five branches; the envelope is pinned twice over by cases 2 and 3. This is recorded explicitly in the suite because it means case 5 is not a catcher for the eighth arm — and the replay above confirms it.
- **`append-claim.sh` is not edited** — the envelope already landed there on all four emit paths.
- **No success-path assertion was touched** — cases 1, 6 and 7 already carry the full-envelope idiom, as the maintainer's scope correction on the issue noted.

## Notes for the reviewer

- The issue body's line references (236, 285) were stale at pick time; the real anchors were 369 and 418.
- The `python3 -c` JSON-parse one-liner is now duplicated three times in the added code. That matches the suite's existing convention — it already carried three copies of the sibling success-shape idiom — so no helper was extracted; doing so would have touched six call sites outside this diff.
- The runner token is over-scoped (`gist`, `read:org`, `workflow`) because it is a `gh auth login` token whose scope set is fixed by the GitHub CLI OAuth app. The scope preflight was cleared with the documented `--allow-overscoped` opt-out rather than silently.

### DIFF
--- a/cogni-portfolio/tests/test-append-claim-lock.sh
+++ b/cogni-portfolio/tests/test-append-claim-lock.sh
@@ -1,12 +1,14 @@
 #!/usr/bin/env bash
 # Regression test for cogni-portfolio/scripts/append-claim.sh covering the
 # stale-lock sweep's mtime reading — the BSD-first `stat` chain that returns a
-# successful *wrong* answer on GNU coreutils.
+# successful *wrong* answer on GNU coreutils — and the script's output-envelope
+# contract on every emit path, success and failure alike.
 #
 # Fixtures are built inline under a temp root — no committed blobs to maintain.
-# Each case builds a minimal claims project, pre-creates the lock directory so
-# the acquire loop is genuinely contended, shadows `stat` with a stub on PATH,
-# runs the script, and asserts on the exit code plus both output streams.
+# Most cases build a minimal claims project, pre-create the lock directory so
+# the acquire loop is genuinely contended, shadow `stat` with a stub on PATH,
+# run the script, and assert on the exit code plus both output streams. Cases 7
+# and 8 are the exceptions and build less than that — see the note below.
 #
 # Coverage:
 #   1  stale-lock-swept-on-non-numeric-stat   non-numeric mtime reading is floored
@@ -39,6 +41,11 @@
 #                                             never contends must spawn NO stat at
 #                                             all; this is what forbids resolving
 #                                             the form ahead of the loop
+#   8  no-arguments-emits-usage-envelope      the argument guard — the one failure
+#                                             path that never reaches the acquire
+#                                             loop writes a SINGLE error-envelope
+#                                             object to stderr and exits 1, asserted
+#                                             by shape and not by message alone
 #
 # Every case that CONTENDS drives the acquire loop to its ceiling on purpose, so
 #   each sets APPEND_CLAIM_MAX_WAIT as a per-invocation prefix rather than paying
@@ -47,9 +54,11 @@
 #   than the ~7.3s it cost when the ceiling was a hardcoded constant. Cases 4 and 5
 #   each drive the loop through the full sweep budget, so they carry most of the
 #   ~2.4s this suite grew by; folding them into one invocation that emits two result
-#   lines would recover roughly half of that if the budget ever needs tightening. Case 7 is the deliberate exception
-#   and never contends at all — it exists to prove the UNcontended acquire stays
-#   fork-free, so driving the loop is precisely what it must not do.
+#   lines would recover roughly half of that if the budget ever needs tightening.
+#   Cases 7 and 8 are the deliberate exceptions and never contend at all. Case 7
+#   exists to prove the UNcontended acquire stays fork-free, so driving the loop is
+#   precisely what it must not do; case 8 returns at the argument guard before the
+#   loop is reached at all, so it has no loop to drive.
 #   The prefix is deliberately NOT a suite-level export: run-plugin-tests.py
 #   invokes each suite as a bare `bash <path>`, so each case must carry its own
 #   ceiling.
@@ -206,6 +215,52 @@
 #   assertions stay green and only the envelope assertion discriminates. That is
 #   precisely what makes asserting the whole envelope, rather than the nested status
 #   alone, worth its line in all three.
+#
+#   Eighth arm — the TIMEOUT envelope's SHAPE, as distinct from its message:
+#   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
+#     --root . \
+#     --file cogni-portfolio/scripts/append-claim.sh \
+#     --expr 's/\\"data\\": null, //' \
+#     --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' \
+#     --case live-lock-not-swept-on-numeric-stat
+#
+#   The expression drops the `data` key from the timeout envelope. It matches the
+#   ESCAPED spelling exactly once — the bash-side emit inside the acquire loop's
+#   double-quoted string; the two unescaped spellings in the SCRIPT, in its own
+#   header comment documenting the envelope and in the argument guard's
+#   single-quoted literal, cannot match it. It carries no `^`/`$` anchor, so it
+#   needs no `/m` under the harness's `perl -0pi`.
+#
+#   Cases 2 and 3 both catch this arm — both reach the timeout emit and both now
+#   assert its shape. Case 5 reaches the SAME envelope but asserts only the message
+#   grep, so it stays green under the mutation and is deliberately not a catcher.
+#   The arm names case 2 only because `--case` takes one name.
+#
+#   Under the mutation the script still exits 1, still writes one line to stderr,
+#   and that line still carries the timeout message and still parses as JSON — so
+#   the exit-code assertion and the message grep both stay green, and only the
+#   structural check discriminates. That is what makes the structural check evidence
+#   rather than decoration, and it is why the grep was kept alongside it rather than
+#   replaced by it: the two pin different things.
+#
+#   Ninth arm — the USAGE envelope's shape:
+#   bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
+#     --root . \
+#     --file cogni-portfolio/scripts/append-claim.sh \
+#     --expr 's/"data": null, "error": "Usage/"error": "Usage/' \
+#     --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' \
+#     --case no-arguments-emits-usage-envelope
+#
+#   The expression drops the `data` key from the argument guard's envelope. It
+#   matches exactly once — the guard's own emit; the SCRIPT's header comment spells
+#   the surrounding text differently and cannot match. It carries no `^`/`$` anchor,
+#   so it too needs no `/m`.
+#
+#   Case 8 is the only catcher, because it is the only case that runs the script
+#   with no arguments. Under the mutation the guard still exits 1, still writes one
+#   line to stderr, and that line still carries the `Usage:` literal — so every
+#   assertion in case 8 except the structural one stays green. A case that asserted
+#   the usage path by message alone would not have caught this.
 
 # `set -u` only — `set -e` would abort on the first failing assertion and defeat
 # the per-case failure counter below. All cases also run a script that is
@@ -368,6 +423,8 @@ if [ "$case2_rc" -ne 1 ]; then
   fail "live-lock-not-swept-on-numeric-stat" "expected exit 1, got $case2_rc"
 elif ! grep -q 'Could not acquire lock on claims.json after 0.3s' "$TMPROOT/case2.err"; then
   fail "live-lock-not-swept-on-numeric-stat" "stderr did not carry the timeout error: $(cat "$TMPROOT/case2.err")"
+elif ! python3 -c 'import json,sys; d=json.load(sys.stdin); sys.exit(0 if d["success"] is False and d["data"] is None else 1)' < "$TMPROOT/case2.err" 2>/dev/null; then
+  fail "live-lock-not-swept-on-numeric-stat" "stderr did not parse as a {success,data,error} object with success=false and data=null: $(cat "$TMPROOT/case2.err")"
 elif [ ! -d "$case2_dir/cogni-claims/.claims.lock" ]; then
   fail "live-lock-not-swept-on-numeric-stat" "a live peer's lock was swept"
 else
@@ -417,6 +474,8 @@ if [ "$case3_rc" -ne 1 ]; then
   fail "live-lock-not-swept-on-gnu-stat" "expected exit 1, got $case3_rc — a BSD-first chain would sweep the live lock and append"
 elif ! grep -q 'Could not acquire lock on claims.json after 0.3s' "$TMPROOT/case3.err"; then
   fail "live-lock-not-swept-on-gnu-stat" "stderr did not carry the timeout error: $(cat "$TMPROOT/case3.err")"
+elif ! python3 -c 'import json,sys; d=json.load(sys.stdin); sys.exit(0 if d["success"] is False and d["data"] is None else 1)' < "$TMPROOT/case3.err" 2>/dev/null; then
+  fail "live-lock-not-swept-on-gnu-stat" "stderr did not parse as a {success,data,error} object with success=false and data=null: $(cat "$TMPROOT/case3.err")"
 elif [ ! -d "$case3_dir/cogni-claims/.claims.lock" ]; then
   fail "live-lock-not-swept-on-gnu-stat" "a live peer's lock was swept"
 else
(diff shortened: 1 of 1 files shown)


claude-cogni-work:
## Ship-ready

Author instance has finished this PR and is handing it off for independent review. It stays **draft** and carries `svc:needs-review`; the ready flip and the merge decision belong to the merger instance.

**Issue:** #1459 — `svc:in-progress` → `svc:resolved`
**Branch:** `fix/issue-1459-error-envelope-shape-assertions` @ `80bad789`
**Base:** `origin/main` @ `11a71ec0`

### Handoff preflight

Clean, RC=0 — 3 gating instruments passed, 0 failed, 4 skipped.

| Instrument | Result |
|---|---|
| `run-tests` | pass — suites green for cogni-portfolio |
| `mutation-recipe` | pass — recorded recipe runnable as written |
| `frontmatter` | skip — not_applicable (no touched SKILL.md or agents/*.md) |
| `breadcrumbs` | skip — not_applicable (same) |
| `skill-length` | skip — not_applicable (no touched SKILL.md) |
| `agent-length` | skip — not_applicable (no touched agents/*.md) |

Every skip is `not_applicable` against a one-file test-suite diff, not an instrument that failed to observe anything.

### Contract grading

**No grade performed** — no issue-time triage contract exists for #1459 (`read-review-contract --target issue --origin triage` returned `found: false`). The handoff degrades to its pre-grading behaviour. There is no residual, because nothing was graded.

The merger forms its own contract from the issue, which is the intended path regardless.

### What the merger may want to look at first

1. **The joined-not-replaced requirement.** The issue's first acceptance criterion is falsified by the structural check *replacing* the message grep. Both greps are kept byte-identical; the structural check is a new `elif` inserted after each.
2. **The two new mutation arms.** These are an addition beyond the literal acceptance criteria — they exist to prove the added assertions discriminate. Every arm, old and new, was **replayed rather than asserted**; results including the declared non-catcher (case 5 → `vacuous_guard`) and the base-ref non-vacuity proof are tabulated in the PR body.
3. **Scope narrower than the original triage.** The triage comment named three success-path assertions as still bare-shaped; the maintainer's follow-up corrected that — two were already repaired when #1461 landed. This PR touches no success-path assertion. Worth confirming against the maintainer's comment rather than the triage.
4. **The over-scoped runner token.** The scope preflight was cleared with the documented `--allow-overscoped` opt-out, not silently: the `gh auth login` token carries `gist`, `read:org`, `workflow`, a scope set fixed by the GitHub CLI OAuth app and not narrowable.

sdh07:

_The merger formed this expectation from issue #1459 before checking out the PR head — the review grades the diff against it._

**Expected approach:** The whole change lands inside cogni-portfolio/tests/test-append-claim-lock.sh, which today has 7 cases and asserts the two timeout error paths only by append-claim.sh's own message literal (grep at lines 369 and 418, plus case 5's 0.1s variant at line 497) while the usage guard (cogni-portfolio/scripts/append-claim.sh lines 25-28) has no case at all. A correct fix (a) inserts an ADDITIONAL branch into the existing if/elif chains of case 2 (lines 367-375) and case 3 (lines 416-424) that pipes "$TMPROOT/case2.err"/"case3.err" through python3 -c and fails unless success is False and data is None, keeping the existing grep verbatim so mutation arm 3 (header lines 105-120, which decouples the ceiling and changes only the message wording) still discriminates; and (b) adds an 8th case before the summary block (lines 573-577) that runs `bash "$SCRIPT"` with no arguments and asserts exit 1, empty stdout, exactly one newline-terminated stderr line, and the {success:false,data:null,non-empty error} shape carrying append-claim.sh's own `Usage: append-claim.sh` literal — never any bash/coreutils diagnostic wording, per the suite's own rule at lines 336-339 and 516-520. The script under test (append-claim.sh) must not change, and the suite header's Coverage list (lines 11-41), per-case ceiling accounting (lines 43-59, 549-554) and mutation-recipe block (lines 68-208) must stay true of the file after the edit.

**Acceptance checklist:**
- [ ] Scope boundary: every changed path lies under cogni-portfolio/tests/ (the suite plus any new fixture/helper it needs is admitted there); nothing under cogni-portfolio/scripts/ and no .claude-plugin/ manifest is touched. Falsified by: `git diff --name-only origin/main...HEAD` printing any path outside cogni-portfolio/tests/, or `git diff origin/main...HEAD -- cogni-portfolio/scripts/append-claim.sh` producing non-empty output.
- [ ] No version bump and no maintainer breadcrumb rides along. Falsified by: `python3 scripts/check-version-bump.py` reporting version-touched or version-mirror-desync, or `python3 scripts/check-breadcrumbs.py` exiting non-zero / flagging a newly introduced issue-PR ref, version tag or slice code in the touched suite.
- [ ] The suite runs green on its own. Falsified by: `bash cogni-portfolio/tests/test-append-claim-lock.sh; echo rc=$?` printing rc other than 0, or its last line not being `All tests passed.`
- [ ] The plugin's discovered suites run green through the CI runner. Falsified by: `python3 scripts/run-plugin-tests.py --filter cogni-portfolio` exiting non-zero.
- [ ] The result-line plainness contract holds: pass()/fail() stay bare `printf` emitters with no escape literal anywhere in the file. Falsified by: `python3 scripts/check-result-line-plainness.py` exiting non-zero or naming cogni-portfolio/tests/test-append-claim-lock.sh, or `grep -c $'\033' cogni-portfolio/tests/test-append-claim-lock.sh` returning non-zero count.
- [ ] The existing message greps are JOINED, not replaced — both 0.3s greps (cases 2 and 3) and case 5's 0.1s grep survive verbatim. Falsified by: `grep -c 'Could not acquire lock on claims.json after 0.3s' cogni-portfolio/tests/test-append-claim-lock.sh` returning anything other than 2, or the same grep for `after 0.1s` returning anything other than 1.
- [ ] Cases 2 and 3 each gained a structural stderr-envelope assertion that parses the captured .err file as JSON and requires success false and data null. Falsified by: `grep -n 'case2.err' cogni-portfolio/tests/test-append-claim-lock.sh` showing no python3 JSON parse of that file inside the live-lock-not-swept-on-numeric-stat block, or the same for case3.err inside the live-lock-not-swept-on-gnu-stat block.
- [ ] Executable proof that the message grep still gates after the widening (AC1's join-not-replace, since the structural check alone stays green under a message-only regression): replay `bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/APPEND_CLAIM_MAX_WAIT:-30/APPEND_CLAIM_MAX_WAIT_UNUSED:-30/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case live-lock-not-swept-on-numeric-stat` (the --expr is evaluated by `perl -0pi`, single-quoted at the call site, never sed/ERE; no anchor so no /m needed). Falsified by: any result other than guard_verified — guard_vacuous means the grep stopped gating, case_not_found means the printed label drifted.
- [ ] A new no-arguments case exists and prints a machine-classifiable label: it calls the pass/fail helpers with the case token as a SEPARATE first argument (no colon glued to it), per the anchoring note at suite lines 228-232. Falsified by: `bash cogni-portfolio/tests/test-append-claim-lock.sh | grep -cE '^ok: [^ ]+ '` returning fewer than 8, or the new label appearing in the printf as `<label>:` rather than as its own argument.
- [ ] The new case asserts the invocation by SHAPE: exit status 1, empty stdout, and exactly one newline-terminated line on stderr. Falsified by: the new case's block lacking any of an exit-code comparison against 1, an emptiness test on captured stdout, and a single-line/single-object test on its .err file (e.g. `wc -l` equal to 1 or an equivalent one-object JSON parse).
- [ ] The new case asserts the envelope by parsing stderr as JSON — success false, data null, non-empty error — and pins append-claim.sh's OWN `Usage: append-claim.sh` literal, never any bash or coreutils wording. Falsified by: no python3 JSON parse of the usage stderr in that block, or the block grepping a foreign-tool diagnostic (this host is German-locale, so an English bash/coreutils grep passes vacuously).
- [ ] The new case's structural check is not vacuous against the exact regression the issue names (a bare {"error": ...} shape). Replay `bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh --root . --file cogni-portfolio/scripts/append-claim.sh --expr 's/"data": null, "error": "Usage/"error": "Usage/' --test 'bash cogni-portfolio/tests/test-append-claim-lock.sh' --case <the new case's printed label>` — that literal occurs exactly once (the timeout emit at append-claim.sh line 156 spells its quotes backslash-escaped). Falsified by: any result other than guard_verified.
- [ ] The suite header stays true of the file: the `# Coverage:` block (lines 11-41) enumerates the new case with its exact printed label, and the per-case ceiling accounting (lines 43-59, 549-554, which today calls case 7 the sole non-contending exception) either is updated to account for the new case or the new case carries its own explicit APPEND_CLAIM_MAX_WAIT prefix. Falsified by: `sed -n '/^# Coverage:/,/^# Every case/p' cogni-portfolio/tests/test-append-claim-lock.sh` not containing the new label, or neither the header sentence updated nor an APPEND_CLAIM_MAX_WAIT prefix present on the new case's invocation line.
- [ ] All mutation recipes recorded in the suite header still replay to guard_verified and every --expr still matches exactly once in the script — including any NEW arm the change adds (note the trap: `Usage: append-claim.sh` occurs TWICE in append-claim.sh, at the line-3 comment and the line-26 emit, so an unanchored usage expr is a multi-match). Falsified by: replaying each of the header's arms as written and any one reporting other than guard_verified, or `grep -c '<arm search literal>' cogni-portfolio/scripts/append-claim.sh` returning other than 1 for any arm, or the harness reporting expr_no_op / case_not_found.

**Risk watchlist:**
- The structural check REPLACES the message grep in case 2 or case 3 — the suite stays green, but mutation arm 3 (the ceiling-to-message coupling) silently goes vacuous, which is the single defect this issue exists to avoid.
- The new usage case greps a bash/coreutils diagnostic instead of append-claim.sh's own literal — vacuous on this German-locale host and green against unfixed code.
- The new case asserts only exit 1 plus a message substring, skipping the data/success keys — reproducing exactly the blind spot the issue reports, one file over.
- A new mutation arm is recorded whose --expr matches the line-3 comment as well as the line-26 emit (or matches nothing), so the arm reports expr_no_op or mutates prose rather than behaviour.
- The new case's label carries a colon or is folded into a single printf argument, so mutation-check.sh's whole-token match resolves to case_not_found.
- The new case is placed AFTER the summary block (lines 573-577), so it never runs; or it uses `exit` instead of the fail() counter and truncates the suite.
- A case reuses another case's project directory or .err path, coupling results to case order.
- The suite header's Coverage list and ceiling narrative go stale — the file documents 7 cases while running 8.
- Judgment-only: whether case 5's 0.1s timeout path also deserves the structural check. The issue scopes the widening to cases 2 and 3 only; omitting it is not a defect.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1486

**Summary:** Both perspectives passed and all 14 committed contract items were confirmed satisfied by execution on the merged tree — including all 9 header mutation arms replaying to guard_verified with each --expr matching exactly once. The single finding is an imprecision in the merger's own contract falsifier (item 6's abbreviated `after 0.1s` fragment also matches a pre-existing header comment on the base), independently identified by both perspectives; the substantive join-not-replace requirement is met on the full message literals. No defect in the diff.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (waived) (advisory-addition) `review-contract:item-6` — The merger's own contract falsifier for item 6 reads `grep -c 'after 0.1s' ... returning anything other than 1`, but that abbreviated fragment returns 2 on the PR head AND on the unmodified base (5c388d91): one hit is header prose documenting the derived-ceiling coupling, the other is case 5's actual grep. Run verbatim it would mechanically read 2 != 1 as a violation. The substantive requirement IS met — the full literal 'Could not acquire lock on claims.json after 0.1s' occurs exactly once and '...after 0.3s' exactly twice, all verbatim from base, and the greps were JOINED to the new python3 envelope checks via elif rather than replaced. Defect in the contract's falsifier phrasing, not in the diff. Flagged independently by both perspectives.
  - *Justification:* Falsifier-construction imprecision in the merger-authored contract, reproducible on the base commit; the substantive requirement is satisfied on the full literals. Not a property of this diff.

### Positive observations

- All 14 committed checklist items satisfied by execution on the merged tree; no item graded unclear. (cogni-service:contract-grading-reviewer)
- item 14: all 9 header mutation arms replayed on a fresh mktemp copy and every one reports guard_verified, with each arm's --expr matching append-claim.sh exactly once — correctly avoiding the two-occurrence 'Usage: append-claim.sh' trap by targeting a longer compound literal. (cogni-service:contract-grading-reviewer)
- item 8: the ceiling-to-message mutation arm replays guard_verified (mutated red, restored green), proving the pre-existing message grep still GATES after the widening — the join-not-replace property the issue exists to protect. (cogni-service:contract-grading-reviewer)
- item 12: the new usage case is non-vacuous against the exact regression the issue names — dropping the `data: null` key from the usage envelope turns the case red (guard_verified). (cogni-service:contract-grading-reviewer)
- items 6/7: cases 2 and 3 each gained a python3 envelope parse inserted as an additional elif AFTER the existing message grep, which survives verbatim in both. (plugin-dev:plugin-validator)
- items 10/11: case 8 asserts exit status, stdout emptiness and a single stderr line by SHAPE, parses the envelope as JSON, and only then greps append-claim.sh's OWN 'Usage: append-claim.sh' literal — the German-locale vacuity trap is avoided. (plugin-dev:plugin-validator)
- item 13: the Coverage block enumerates case 8 with its exact printed label, the ceiling narrative was updated from 'case 7 is the sole exception' to 'Cases 7 and 8', AND case 8 carries its own APPEND_CLAIM_MAX_WAIT prefix — the header-drift watchlist item does not fire. (plugin-dev:plugin-validator)
- item 1: the diff touches exactly one path; cogni-portfolio/scripts/ is untouched, so the script under test did not change. No discovered suite removed (7 suites in cogni-portfolio/tests/, byte-identical listing base vs head). (plugin-dev:plugin-validator)
- item 2: version-bump gate clean (9 scanned, 1 touched, 0 violations); breadcrumbs exit 0; plugin.json 0.9.84 still mirrors marketplace.json. (plugin-dev:plugin-validator)
- items 3/4: the suite exits 0 standalone from / with 'All tests passed.' across 8 cases, and run-plugin-tests.py --filter cogni-portfolio exits 0 at 7/7. Merger additionally ran the FULL repo sweep on the merged tree: 128/128 pass, and every repo-wide guard green including the MCP agent-grant guard newly merged by #1485 — so the two changes compose. (merger merged-tree verification)
- Deterministic pre-gates clean: frontmatter 0 offenders for cogni-portfolio; version-bump status=ok with 0 violations.

### Recommendation

---

Issue #1501, opened 2026-08-18
repo: nothing asserts the registry provides_tools vocabulary is complete


claude-cogni-work:
## Context

Surfaced by the `/simplify` altitude pass while resolving the MCP guard consolidation, which added an affirmative arm keyed on `provides_tools[]` in `cogni-workspace/references/mcp-git-registry.json`. The arm's reach is exactly the vocabulary that field declares, and nothing asserts that field is complete.

## Observation

Measured against the tree: the registry lists **18** tool names across its two registered servers, while agent frontmatter actually grants **26** distinct names for those same servers. Nine granted names are outside the registry's vocabulary and therefore outside the new arm's reach entirely:

- excalidraw — `clear_canvas`, `group_elements`, `snapshot_scene`, `restore_snapshot`, `query_elements`, `get_element`, `export_to_excalidraw_url`
- pencil — `get_variables`, `snapshot_layout`

There are no live offenders hiding in that gap today. The concern is the shape, not a present defect: the guard's empty-vocabulary floor is all-or-nothing, so it fires only if *every* server loses its array. A single server's `provides_tools` going stale, or a third server registered without one, narrows coverage while the vocabulary counter stays healthy and every case stays green — the invisible-absence class the guard exists to close, reintroduced one layer down.

## Why this was not folded into the arm's own change

It is a registry-accuracy judgement, the same class as the stale `cogni-portfolio` entry under `mcp_excalidraw.required_by` that the guard already surfaces as a non-failing observation and that the originating issue explicitly fenced out of scope.

## Proposed change

Decide whether `provides_tools[]` should be asserted complete, and if so how. The data needed is already in hand: `collect()` extracts `grant_tokens` per agent, so a granted `mcp__<registered-ns>__<tool>` whose `<tool>` is absent from that server's array is directly detectable — plausibly as a non-failing observation first, mirroring `listed_without_grant`, rather than as a hard arm.

## Acceptance criteria

- [ ] A decision is recorded either way, with the measurement behind it. Falsified by closing with no stated verdict.
- [ ] If adopted: a granted tool name absent from its server's `provides_tools[]` is surfaced, and the mechanism reads the registry as data with no hardcoded tool list. Falsified by a hardcoded list appearing in the guard.
- [ ] If adopted: the tree is green under it, or every gap it reports is closed in the same change. Falsified by the guard exiting non-zero with gaps left unaddressed and unexplained.
[...]

Issue #1500, opened 2026-08-18
repo: run-plugin-tests conflates an empty filter with broken discovery


claude-cogni-work:
## Context

Surfaced by the `/simplify` altitude pass while resolving the MCP guard consolidation. Not folded in there because it changes a shared runner's contract, which sat outside that change's scope boundary.

## Observation

`scripts/run-plugin-tests.py` conflates two different questions under one failure. Its no-suites-is-a-failure rule fires whenever the discovered list is empty, which is correct for *the globs stopped matching* — a real repo-level defect — but wrong for *your filter matched nothing*, which is an empty query rather than a broken one.

Discovery runs before filtering and the filter value is in scope at the check, so the two cases are distinguishable.

## Why it matters now

Three plugins currently ship no `tests/` directory: `cogni-marketing`, `cogni-sales`, and `cogni-website`. The first two have always been in that state, so `python3 scripts/run-plugin-tests.py --filter cogni-sales` exits 1 today with a `success: false` envelope and a "no test suites discovered" message for a perfectly legitimate query. The unfiltered sweep CI runs is unaffected in every case.

The current remedy is a paragraph of prose in the root `CLAUDE.md` telling the reader to interpret the exit-1 rather than answer it with a filler suite. That is a documented exception standing in for a one-condition change.

## Proposed change

Keep the hard floor on unfiltered discovery. Report an empty *filtered* result as an empty result — exit 0 with `total: 0` — rather than a discovery failure. Then trim the `CLAUDE.md` paragraph back to the rule.

## Acceptance criteria

- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-sales` exits 0 with `data.total == 0`. Falsified by a non-zero exit on a filter that legitimately matches no plugin.
- [ ] Unfiltered discovery returning zero still exits non-zero. Falsified by a run with the globs broken exiting 0.
- [ ] A named case in `tests/test_run_plugin_tests.sh` pins both directions. Falsified by only the empty-filter direction being covered.
- [ ] The root `CLAUDE.md` paragraph explaining how to read the filtered exit-1 is reduced accordingly. Falsified by prose surviving that describes behaviour the change removed.
- [ ] No `plugin.json` or `marketplace.json` version field is touched.

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-08-19T02:14:46Z` (GitHub clock).

[...]

PR #1502, opened 2026-08-18
Adopt provides_tools-keyed MCP guard arm; retire the prose detector

### LINKED ISSUE #1487: repo: evaluate a provides_tools-keyed arm that could retire the plugin-scoped MCP prose detector
## Context

Surfaced by the `/simplify` altitude pass while resolving #1474, which added `scripts/check-mcp-tool-grant.py` — a repo-scope guard whose affirmative arm requires a literal `mcp__<ns>__<tool>` token in an agent body.

That PR deliberately left a **second detector** in place: `cogni-website/tests/test-mcp-tool-grant.sh` keeps a prose-based arm (`"<X> MCP"` in the body), because the affirmative detector has a structural blind spot — `cogni-website/agents/hero-renderer.md` documents driving Pencil in prose alone, with no `mcp__` token anywhere in its body. Neither detector subsumes the other, so the split was the right call at the time.

## Observation

The registry may already carry the data that collapses the two into one mechanism. `cogni-workspace/references/mcp-git-registry.json` gives each server a `provides_tools[]` array, which `check-mcp-tool-grant.py` currently reads past — it uses only `desktop_config_key` and `required_by`.

A second affirmative arm keyed on that array — *an agent body naming a registry `provides_tools` bare name for server X must grant X* — would be prose-free and data-driven. Measured during the review pass: **0 offenders repo-wide today**, and it is not vacuous — mutating `hero-renderer.md`'s `tools:` back to `["Read", "Write", "Glob", "Bash"]` (the exact shipped defect the prose arm exists for) leaves the `mcp__` arm blind while a `provides_tools` arm fires on `pencil` via `batch_design` / `get_guidelines`.

If it holds up, the prose arm could retire entirely, and with it a second scope, a second suite, and the doctrine paragraphs currently maintained in two file headers plus `CLAUDE.md`.

## Why this was not folded into #1474

Three reasons, all still live:

1. It changes the design that issue's acceptance contract was graded against.
2. `cogni-website/tests/` holds exactly one suite. Retiring it makes `run-plugin-tests.py --filter cogni-website` discover zero and exit 1 by its own no-suites-is-a-failure rule, so that acceptance criterion would need rethinking rather than just satisfying.
3. Bare tool names like `get_screenshot` and `export_scene` are generic English. A body sentence using one incidentally would false-positive, which is precisely the failure mode that kept the prose heuristic plugin-scoped in the first place.

## Proposed change

Evaluate whether a `provides_tools`-keyed affirmative arm can replace the plugin-scoped prose arm, and if so make the swap in one PR.

## Acceptance criteria

- [ ] A decision is recorded either way, with the measurement behind it. Falsified by the issue closing with no stated verdict on whether the prose arm retires.
- [ ] If adopted: the new arm lives in `scripts/check-mcp-tool-grant.py`, reads `provides_tools[]` as data, and is green on the tree with no exclusion list. Falsified by `python3 scripts/check-mcp-tool-grant.py` exiting non-zero, or by a plugin-name allowlist appearing in the file.
- [ ] If adopted: false-positive risk on generic bare names is addressed explicitly (for example by requiring a code-span context), not left to chance. Falsified by a bare-name match with no context requirement.
- [ ] If adopted: the mutation over `cogni-website/agents/hero-renderer.md` that strips its pencil grant still reds a NAMED case. Falsified by that replay returning anything but `guard_verified`.
- [ ] If adopted: `python3 scripts/run-plugin-tests.py --filter cogni-website` still exits 0, or the criterion is deliberately retired with a recorded rationale. Falsified by the runner exiting 1 with no such rationale.
- [ ] If rejected: the reason is written into the two file headers so the next reader does not re-propose it.
- [ ] No `plugin.json` or `marketplace.json` version field is touched — the post-merge workflow owns the bump.

## Notes

Deliberately out of scope: the stale `cogni-portfolio` entry in `mcp_excalidraw.required_by` (listed but grants no excalidraw tool), which is a separate registry-accuracy judgement raised in #1474's PR.

Refs #1474.

claude-cogni-work:
Closes #1487.

## Verdict: ADOPT

The verdict is recorded **in the tree**, not only here — the guard's module docstring carries the adoption, the measurement behind it, and the residual classes it knowingly does not cover.

## What changed

- `scripts/check-mcp-tool-grant.py` gains a second affirmative arm, `provides_tools_body_name_ungranted`: an agent body naming a registry `provides_tools[]` bare name **inside a backtick code span** must grant a tool of a server that provides it.
- Resolution is **server-level, never tool-level**. Any `mcp__<ns>__*` grant satisfies any bare name that `<ns>` provides. A name owned by two servers is satisfied by either; when neither is granted, exactly one finding lands on the lexicographically-first owner so output does not depend on dict order.
- The arm sits after the no-`tools:`-key skip and **inherits** it. This is load-bearing, not incidental — see the measurement below.
- An empty bare-name vocabulary is a **loud failure, exit 2**, on the same channel as the existing zero-discovery errors.
- `data.summary` gains `provides_tools_vocabulary` and `provides_tools_code_span_names`, both non-zero on the real tree. `by_arm` picks the new arm up with no code change — it is built dynamically from `finding["arm"]`.
- `cogni-website/tests/test-mcp-tool-grant.sh` is **deleted**, and every doctrine surface that asserted its irreplaceability is rewritten on the same commit.

## Evidence for the verdict

**(a) Does the new arm subsume the prose arm inside the prose arm's own scope?** Yes. That suite scanned only `cogni-website/agents/*.md`, and the single agent it fired on is `hero-renderer.md` — which documents driving Pencil with **zero** `mcp__` tokens in its body, but names `get_guidelines` and `batch_design` in code spans. Under the exact mutation the retired suite documented, the new arm fires and the `mcp__` arm stays blind.

Worth stating precisely, because the issue's framing was slightly off: `hero-renderer.md` **is granted** — six `mcp__pencil__*` tokens on its `tools:` line. The retired arm's live case was therefore a *green* "prose-only, must be granted" assertion; the ungranted state is manufactured by the mutation, not present in the tree.

**(b) Does the runner obstacle hold?** Weaker than the issue stated. `run-plugin-tests.py` applies its no-suites rule to the **total** discovered population, and CI runs the sweep unfiltered — 128 suites before this change, 127 after, all green. Only `--filter cogni-website` exits 1, and that condition is **pre-existing and not unique**: `cogni-marketing` and `cogni-sales` have always been in it. Recorded in `CLAUDE.md`; the runner fix itself is #1500.

**(c) Is the code-span requirement sufficient and measurable?** On the measured population, yes. Of 120 discovered agents, 103 declare a `tools:` key; 9 of those name at least one registry tool in a code span, for **52 name hits**, and **0** are offenders. The word boundary also excludes a name already inside an `mcp__<ns>__<tool>` token, so the two arms structurally cannot double-report — pinned by case V8, not merely asserted.

**The zero is contingent, and the PR says so.** A bare-name scan that ignored the inherited skip would flag four non-offenders — `storyboard.md` and `web.md` in cogni-visual and cogni-workspace — which declare no `tools:` key and carry 22 in-span names between them. Case V5 pins the reuse.

## Residuals, stated rather than implied away

Four, all in the guard's docstring: an agent naming a server in prose while naming **none** of its tools in a span is outside both arms; four no-`tools:` agents are skipped by inheritance; a code span inside a *disclaiming* sentence would read as affirmative; and the vocabulary is only as complete as the registry — agents grant 26 distinct names across the two servers while `provides_tools[]` lists 18, leaving 9 out of reach. That last one is #1501.

## Scope

**In:** the guard, its repo-scope suite, the deleted prose suite, root `CLAUDE.md`.
**Out, deliberately:** no `*/agents/*.md` is edited — the guard is made to fit the tree, never the reverse. No `required_by` value changes, so the stale `cogni-portfolio` entry keeps surfacing as its existing non-failing observation, exactly as the issue's Notes asked. No version field touched. No workflow change needed.

## Test plan

- [x] `python3 scripts/check-mcp-tool-grant.py` exits 0; `provides_tools_vocabulary` 18, `provides_tools_code_span_names` 52, `by_arm` carries no entry for the new arm.
- [x] `bash tests/test_check_mcp_tool_grant.sh` — 53 cases, `0 failed`, including V1–V8 and the L3 floor.
- [x] `python3 scripts/run-plugin-tests.py` — 127 suites, 0 failed (128 before the deletion).
- [x] `python3 scripts/run-plugin-tests.py --filter cogni-website` exits 1; rationale recorded in `CLAUDE.md`.
- [x] `check-breadcrumbs.py`, `check-result-line-plainness.py`, `check-version-bump.py`, `check-workflow-yaml.py` all exit 0.
- [x] `grep -rn "test-mcp-tool-grant.sh"` returns no reference to the deleted path.

## Mutation recipe

```
bash scripts/mutation-check.sh --root . \
  --file cogni-website/agents/hero-renderer.md \
  --expr 's/^tools: \[.*mcp__pencil__.*\]$/tools: ["Read", "Write", "Glob", "Bash"]/m' \
  --test 'bash tests/test_check_mcp_tool_grant.sh' --case V7
```

The mutation strips the Pencil grants from the frontmatter while leaving the body's Pencil prose and its `get_guidelines` / `batch_design` code spans in place — the exact shape the retired arm existed to catch, and the one the `mcp__` arm cannot see. `--expr` is fed to `perl -0pi`, which slurps the whole file, so `/m` is required for `^`/`$` to bind to the `tools:` line. The substitution is non-global, safe because the target carries exactly one `^tools: [` line. V7 is two-directional under it: stripping the grant makes the arm fire, and disabling the arm outright drops the span counter to zero — either way V7 reds, which is what stops it being a vacuous real-tree assertion.

## Quality pass

A four-angle `/simplify` pass ran before commit. Applied: one alternation regex replacing an 18-pattern-per-agent loop (output verified identical; the guard's scan work drops roughly 12x), a single-pass span loop, removal of four fixture writes that restated `new_root`'s own default, and replacement of a third inline copy of the registry envelope with the existing helper. Its altitude angle also caught a factual error in the `CLAUDE.md` prose this PR introduced — it framed cogni-website as the sole plugin without a suite when two others already were — which is fixed here. Two findings were judged out of scope and filed instead: #1500 and #1501.

## Follow-up issues

- #1500 — `run-plugin-tests.py` conflates an empty filter with broken discovery
- #1501 — nothing asserts the registry `provides_tools` vocabulary is complete

---

## Blocked

`check-preflight.sh` exits 1 on this branch: `run-tests` reports `no_suites_discovered` for cogni-website (0 discovered against a floor of 1). The handoff is not made and the PR carries no `svc:needs-review`.

Retiring cogni-website's only suite has a second consumer that this PR's `CLAUDE.md` note does not cover. The note discloses that `run-plugin-tests.py --filter cogni-website` exits 1 while CI's unfiltered sweep stays green. It does not disclose that cogni-service's preflight gates per **touched plugin** with a floor of 1, so the retirement blocks the handoff of any future PR touching cogni-website — a standing cost to that plugin's contributors.

Three responses are reasonable and the choice is a human one: absorb the cost and document it; keep the arm but not the retirement; or land the runner/floor generalisation (#1500) first and retire on top of it. See the blocking comment for the full reasoning.

Everything else on the branch is complete and verified — guard exits 0, the suite is 53 cases at 0 failed, the full sweep is 127/0, all repo guards are clean, and the `hero-renderer.md` mutation is confirmed to fire the new arm.


claude-cogni-work:
## Blocked at the handoff preflight

`check-preflight.sh` exited 1 against `origin/main`...`origin/fix/issue-1487-provides-tools-guard-arm`. Not handing off, and deliberately not self-grading — a branch that cannot clear the preflight should not spend a grading agent.

**Violation**

> `run-tests` — run-tests.sh reported `no_suites_discovered` for cogni-website (discovered 0 against a floor of 1) — a gating instrument that never ran is not evidence the suites are green

1 violation, 1 passed, 5 skipped. Every skip is a genuine `not_applicable` (no touched `SKILL.md` or `agents/*.md`, no version-line touch), including `mutation-recipe`, which skips because the only touched path matching `*/tests/test-*.sh` is the one this branch deletes — a retired guard has no file left to mutate.

**Why this is a real finding, not an instrument artifact**

Retiring `cogni-website/tests/test-mcp-tool-grant.sh` leaves that plugin with no `tests/` directory. I measured one consumer of that condition and recorded it in `CLAUDE.md`: `run-plugin-tests.py --filter cogni-website` exits 1, CI's unfiltered sweep stays green (127 suites, 0 failed), and two plugins — `cogni-marketing` and `cogni-sales` — were already in that state.

I did **not** measure this second consumer. cogni-service's own preflight gates per **touched plugin**, with a floor of 1 discovered suite. So the retirement does not merely make a filtered query exit 1; it blocks the handoff of **any** future PR touching cogni-website, starting with this one. That is a standing cost to every contributor to that plugin, and it is materially more than the `CLAUDE.md` paragraph currently discloses.

**This needs a human call, because it reopens the verdict.** The evaluation in #1487 resolved to ADOPT on the strength of three findings that all still hold — the arm subsumes the prose arm inside its declared scope, CI stays green, and the code-span requirement measures clean at 0 offenders. This finding does not falsify any of them. It adds a cost that was not on the scale when the verdict was taken, and the reasonable responses differ enough that picking one autonomously would be overreach:

1. **Keep the retirement and absorb the cost** — accept that cogni-website PRs need the preflight's floor relaxed or waived, and say so in `CLAUDE.md` alongside the `--filter` note.
2. **Keep the arm, keep the suite** — adopt the `provides_tools` arm (which stands on its own merits and is green) but do not delete the prose detector. This lands most of the value and forfeits the "second scope, second suite, duplicated doctrine" payoff that motivated the issue.
3. **Fix the floor first** — land #1500's generalisation (empty filter is not broken discovery) and the equivalent per-plugin floor fix here, then retire the suite on top of a runner that no longer conflates the two. Sequenced, not smaller.

The branch as it stands implements option 1 and is otherwise complete: guard green, 53 suite cases at 0 failed, full sweep 127/0, all repo guards clean, and the mutation over `hero-renderer.md` verified to fire the new arm.

claude-cogni-work:
## Preflight block diagnosed — this is a human gate, not a defect

Second author instance, picked this PR as `verdict_missing`. Not handing off and not self-grading; the withheld handoff above was correct. Recording *why*, so the block is a decision rather than a stall.

### The block is the gate working as specified

`check-preflight.sh`'s run-tests floor is **derived per plugin from the base ref**, not a fixed constant. On `main`, `cogni-website` ships exactly one suite — `cogni-website/tests/test-mcp-tool-grant.sh` — so its derived floor is 1. This PR deletes that suite (`0+/250-`), discovery drops to 0, and the arm reports `no_suites_discovered` against a positive floor.

That is deliberate. From the arm's own rationale:

> the run-tests arm deliberately still sees deletions, because a wholesale suite retirement must keep tripping its base-derived floor

So the violation is not a mis-scoped instrument. It is the one gate that exists to make a wholesale suite retirement stop and be looked at — and this PR is exactly that.

### Why the "other plugins already have zero suites" argument does not clear it

The PR body's CLAUDE.md edit is factually correct on its own terms — I verified it against `main`:

| filter | exit | derived floor |
|---|---|---|
| `--filter cogni-marketing` | 1 | 0 (`no_suites_at_base`) |
| `--filter cogni-sales` | 1 | 0 (`no_suites_at_base`) |
| `--filter cogni-website` | 0 on `main` | **1** |

`cogni-marketing` / `cogni-sales` do not trip this arm because a **derived floor of 0** takes the named skip `no_suites_at_base`, which exists so a new plugin's first handoff is not false-blocked. `cogni-website` is the different case: it *had* a suite and this PR removes it. Same end state, different transition — and the floor keys on the transition. So the CLAUDE.md paragraph is right about CI, and still does not make this preflight pass.

### The only lever is global, and that is on purpose

There is deliberately **no** per-plugin waiver. An allowlist file, a marketplace retirement marker, and a PR-body escape hatch were each considered and rejected — a scoped hatch is easier to take by accident, and a standing marker outlives the PR that needed it. The single documented lever is `COGNI_PREFLIGHT_MIN_SUITES`, which drops the floor **for every plugin in the run**; that cost is the friction that forces the decision to be conscious.

### What this needs

A maintainer call on one question, which no author instance can make for you:

1. **Accept the retirement** — the new `provides_tools`-keyed arm genuinely subsumes the prose detector, and the operator clears the handoff with the global lever, consciously.
2. **Keep a plugin-scoped suite** — if any residual prose-only class survives (an agent documenting a server with no token in its body, the class the old detector was kept for), the retirement is premature and the floor is telling you so.

Worth noting the coverage claim narrows the class rather than preserving it: the new arm requires the bare name to sit **inside a backtick code span**, so an agent naming a server in running prose with no code span and no `mcp__` token is outside both arms. Whether that residue matters is the judgment call — the guard's module docstring says it knowingly does not cover some residual classes.

Labelling `svc:review-blocked` so this parks instead of being re-picked as `verdict_missing` by every subsequent drain. No code touched.

claude-cogni-work:
## Verdict: ADOPT — unblocking, with one correction to the block's own premise

Independent review instance. The two comments above correctly withheld the handoff and correctly asked for a human call. That call has now been made: **accept the retirement.** Recording the reasoning, and the one measured fact that changes the shape of the decision.

### The cost is one-time, not standing

The blocking comment frames the retirement as blocking "**any** future PR touching cogni-website — a standing cost to every contributor to that plugin." The second comment's own table refutes that, without drawing the conclusion:

| filter | derived floor | outcome |
|---|---|---|
| `--filter cogni-marketing` | 0 | `no_suites_at_base` skip |
| `--filter cogni-sales` | 0 | `no_suites_at_base` skip |
| `--filter cogni-website` | **1** (today) | trips |

The floor is derived **from the base ref**. It keys on the transition, which is exactly why cogni-marketing and cogni-sales — same end state, different transition — take the named skip. Once this PR is on `main`, cogni-website's base-derived floor is 0 and it joins them: subsequent PRs touching the plugin take `no_suites_at_base` and pass unblocked.

So the cost is **one conscious use of `COGNI_PREFLIGHT_MIN_SUITES` on this transition PR**, which is precisely what a global-only lever is designed to extract. That is a materially smaller thing than what was weighed when the block was placed, and it is the difference between "absorb a recurring tax" and "clear one gate on purpose."

It also collapses option 3. `check-preflight.sh` is not in this repo — only two stale references in `*/tests/test-arc-taxonomy-sync.sh` comments — so the per-plugin floor cannot be fixed here. #1500 changes `run-plugin-tests.py`, a different mechanism, and the preflight arm deliberately keeps seeing deletions. Landing #1500 first would not clear this block.

### The technical claims reproduce

Verified against `c7c4a86`, not taken from the body:

- Guard exits 0. `provides_tools_vocabulary` 18, `provides_tools_code_span_names` 52, `by_arm` `{}`, 0 violations, 120 agents discovered.
- `tests/test_check_mcp_tool_grant.sh` — 53 cases, `0 failed`.
- **The mutation fires.** Stripping the Pencil grants from `cogni-website/agents/hero-renderer.md` while leaving body and code spans intact yields exactly one finding — `provides_tools_body_name_ungranted`, naming `batch_design, get_guidelines` — and reds V7, A1 and A2. The defect shape the retired arm existed to catch is genuinely covered by the new one.

### The residual class was measured, and it is empty in scope

Comment 2 rightly flags that the code-span requirement narrows rather than preserves the prose class, and asks whether the residue matters. I measured it repo-wide: **4** agents name a registered server as `<X> MCP` in a body without either an in-span registry tool name or an `mcp__` token —

- `cogni-visual/agents/concept-diagram-svg.md` and its cogni-workspace twin — "no Excalidraw MCP, no hand-drawn wobble"
- `cogni-visual/agents/editorial-sketch.md` and its twin — incidental reference to where the output lands

All four are **disclaiming or incidental**, i.e. the exact false-positive class that kept the prose heuristic plugin-scoped in the first place. Inside `cogni-website/agents/`, the retired suite's own scope, the residual is **zero**. Nothing is being dropped silently; the retirement removes a detector whose live coverage is now fully subsumed.

### One real gap in the branch, now closed

The blocking comment's substantive complaint stands and was worth the block: the `CLAUDE.md` note discloses the `run-plugin-tests.py` consumer of a zero-suite plugin and not the preflight one, which is the consumer that actually gates work. Fixed in `2eb3f6d` on `claude/pr-1502-review-z8f24d`, which is this PR's head plus that single `CLAUDE.md` commit — the paragraph now states the base-derived floor, the transition-not-end-state shape, the global lever, and the `no_suites_at_base` outcome afterwards. Cherry-pick it onto `fix/issue-1487-provides-tools-guard-arm` before merge; no other change is needed.

### Sweep note

`python3 scripts/run-plugin-tests.py` reports **127 total, 2 failed** in this container: `cogni-consult/tests/test_resolve_assumptions.sh` and `tests/test_release_bundle_wiki.sh`. Both fail **identically on `origin/main`** at this PR's base `c4c1281`, so neither is caused by this change. They are environment artifacts of running as uid 0 — their negative cases provoke a write failure via `chmod 555`, which root ignores. The body's 127/0 is consistent with a non-root runner.

Removing `svc:review-blocked` and taking the PR out of draft.

---

PR #138, opened 2026-04-25
cogni-workspace: manage-themes deep authoring (Operation 7) + manifest by default

sdh07:
Closes #128. Refs #132 (Phase 2 epic).

## Summary

- Operations 3–6 (Grab Website, Grab PPTX, Create from Preset, Audit/Improve) emit a starter `manifest.json` (`schema_version: "1.0"`, `name`, `slug`, empty `tiers: {}`) for new themes — opt-in entry point to Theme System v2; tier-0 themes (no manifest) stay valid forever.
- New **Operation 7: Author a Deep Theme System** — walks tokens → assets → components → templates with `themes/cogni-work/` as the canonical Phase-2 reference; forward-links the migration guide that lands with #130.
- Operation 7 invokes `cogni-workspace/scripts/generate-tokens-css.py --tokens-dir <theme>/tokens --write` for deterministic CSS regeneration; `validate-theme-manifest.py` runs before any operation completes.
- Existing Operation 7 (Generate Showcase) → 8 and Operation 8 (Apply Theme) → 9; the five `Operation #7` cross-references in Operations 1, 3, 4, 5, 6 now point to `#8` so showcase routing stays intact.
- New "Starter Manifest" reference section so Operations 3–5 defer to one canonical description rather than restating the manifest shape four times.
- README "What it does" bullet for Manage Themes and the Components table description (8 → 9 ops) synced.
- Patch version bump 0.6.12 → 0.6.13 in `plugin.json` and `marketplace.json`; SKILL.md frontmatter version 0.3.0 → 0.4.0 (minor — new operation).

## Phase context

Epic #132 phase-1 child. Phase-1 deps already merged on main:
- #125 — `validate-theme-manifest.py` + `references/theme-manifest.schema.json`
- #126 — `discover-themes.py` reads `manifest.json` with tier-0 fallback
- #127 — `themes/cogni-work/manifest.json` + `tokens/` tier-1 implementation

This change is the authoring surface that wires those scripts into the user-facing skill. No script changes are in scope.

## Scope decisions

- **In:** SKILL.md edits, version bumps in both `plugin.json` and `marketplace.json`, README description sync for the new operation.
- **Out:** No changes to `generate-tokens-css.py` or `validate-theme-manifest.py` — both already expose the contracts Operation 7 needs (importable `generate()`, `--write` CLI, tokens-parity check, tier-0 fallback).
[...]

Issue #126, opened 2026-04-25
cogni-workspace: discover-themes.py reads manifest.json with tier-0 fallback


sdh07:
## Plugin
cogni-workspace

## Use case
Refs #124. The pick-theme skill's discovery script (`cogni-workspace/skills/pick-theme/scripts/discover-themes.py`, 222 lines) currently regex-parses `theme.md` for palette, fonts, and philosophy. RFC #124 Phase 2 requires it to read `manifest.json` when present and expose tier paths to consumers, while leaving tier-0 themes (no manifest) discovered exactly as today. Without this reader, the manifest schema (child #0) is dead code and downstream consumers cannot resolve component or token paths.

## Proposed solution
1. Extend `discover-themes.py` to detect `manifest.json` in each theme directory under `cogni-workspace/themes/`.
2. When `manifest.json` is present: validate it via the schema from child #0 (call out to `cogni-workspace/scripts/validate-theme-manifest.py` or import the validator), then add a `tiers` field to the per-theme output object containing resolved absolute paths for each declared tier. Preserve all existing fields (`name`, `slug`, `palette`, `fonts`, `philosophy`, `theme_md_path`, etc.).
3. When `manifest.json` is absent: emit identical output to today's behavior — no `tiers` field, or `tiers: null`. **This is the backwards-compat contract.**
4. When `manifest.json` is present but invalid: log a warning, fall back to tier-0 behavior, and surface the validation error in a new optional `manifest_error` field. Do not crash; tier-0 is always the safe default.
5. Add a JSON output flag `--include-tiers` (default true) so callers that don't care about tiers can opt out.
6. Add tests/eval cases in `cogni-workspace/skills/pick-theme/` covering: (a) tier-0 theme (regression), (b) full tiered theme, (c) malformed manifest (graceful fallback), (d) manifest referencing missing paths (validation surfaces error).

## Acceptance criteria
- `_template/` and any tier-0 theme produce byte-identical output to current `discover-themes.py` (regression-tested).
- `cogni-work/` post-#2 produces output containing resolved `tiers` paths.
- A handcrafted broken manifest in a temp theme directory falls back to tier-0 with a `manifest_error` field rather than crashing.
- `pick-theme/SKILL.md` describes the new optional output fields.
- Output JSON conforms to repo convention `{"success": bool, "data": {...}, "error": "string"}`.

## Alternatives considered
[...]

PR #1507, opened 2026-08-18
Unique result-line ids across the colliding suites and the theme harness

### LINKED ISSUE #1493: cogni-workspace/cogni-visual/cogni-portfolio: unique first tokens in the 4 loop-generated colliding suites, and addressable result lines for the theme-backcompat harness
## Plugin
cogni-workspace, cogni-visual, cogni-portfolio

## Epic context
Part of the tracker "repo: distinct test cases share a first token, so --case can name the wrong case" (#1484), which stays OPEN when this child lands. One of five per-scope fix children; independent of the others. These three plugins are grouped into one PR because together they account for only 4 colliding suites (52 instances) and the collisions share a single root cause.

## Current behavior
Two distinct defects in this group.

**(a) Loop-generated families.** These suites already use disciplined case ids, but emit each case once per loop iteration, so the id collides with itself. Verified in `cogni-workspace/tests/test-check-skill-names.sh`: `:47-64` builds an `INTERPRETERS` list (PATH bash plus a real bash 3.x, typically `/opt/homebrew/bin/bash` and `/bin/bash`), and `:85-133` runs every behavioural assertion once per interpreter with labels like `assert_no_bash4_error "1c clean tree free of bash-4 runtime error ($BIN)"` (`:108`). The interpreter appears in the label but at the *end*, so the addressable token `1c` names two lines. Both `test-excalidraw-canvas-lock.sh` copies (cogni-workspace, cogni-visual) behave the same way over their own loops. A mutation aimed at the bash-3.2 arm can therefore be credited by the bash-5 arm going red — which is precisely the arm that cannot detect the bash-4-ism the suite exists to catch.

A trap worth stating: `basename` does **not** disambiguate the interpreters. Both are named `bash`. The slug must come from the full path.

**(b) A suite that emits no addressable result line at all.** `cogni-workspace/tests/test-theme-backcompat.sh:27` is a thin wrapper that `exec`s `cogni-workspace/scripts/verify-theme-backcompat.sh`. That harness's emitters at `:47-49` are `printf "  \033[32mPASS\033[0m  %s\n"` and `printf "  \033[31mFAIL\033[0m  %s\n"` — colour-wrapped and **colon-less**, so nothing it prints matches the harness's `PASS:` / `ok:` / `FAIL:` shapes. Every case in the theme back-compat contract is unaddressable, and the suite can never contribute mutation evidence. (The harness lives under `scripts/`, one path segment outside `scripts/check-result-line-plainness.py`'s non-recursive globs, which is why the plainness guard has never reported it.)

## Desired behavior
Every result line each of these suites emits carries a first token unique within that suite, and the theme-backcompat harness emits plain, colon-terminated, id-first result lines.

Exact strings that change:

- `cogni-workspace/tests/test-check-skill-names.sh:108` — `assert_no_bash4_error "1c clean tree free of bash-4 runtime error ($BIN)"` → `assert_no_bash4_error "1c-$BIN_SLUG clean tree free of bash-4 runtime error ($BIN)"`, where `BIN_SLUG` is derived from the full interpreter path (`/opt/homebrew/bin/bash` → `opt-homebrew-bin-bash`, `/bin/bash` → `bin-bash`). Emitted `PASS: 1c clean tree free of bash-4 runtime error (/bin/bash)` → `PASS: 1c-bin-bash clean tree free of bash-4 runtime error (/bin/bash)`. The same transform applies to every `assert_rc` / `assert_out_has` / `assert_out_lacks` / `assert_no_bash4_error` label in the interpreter loop, including `:122`'s `"2c duplicate FAIL summary ($BIN)"` → `"2c-$BIN_SLUG duplicate FAIL summary ($BIN)"`.
- `cogni-workspace/scripts/verify-theme-backcompat.sh:47` — `c_pass() { printf "  \033[32mPASS\033[0m  %s\n" "$1"; }` → `c_pass() { printf 'PASS: %s\n' "$1"; }`; `:48` — `c_fail() { printf "  \033[31mFAIL\033[0m  %s\n" "$1"; }` → `c_fail() { printf 'FAIL: %s\n' "$1"; }`. Each `c_pass` / `c_fail` call site gains a unique id as the first token of its argument, e.g. `PASS  tier-0 output matches baseline` → `PASS: tbc01-tier0-baseline-match tier-0 output matches baseline`.
- The equivalent loop-variable promotion in both `test-excalidraw-canvas-lock.sh` copies and in the colliding cogni-portfolio suite named by the census.

`c_skip` and `c_info` keep their current shape — they are not result lines and must not start with `PASS:` or `FAIL:`.

## Motivation
The interpreter loop in `test-check-skill-names.sh` exists specifically because a bash-4-ism reachable only under bash 3.x is invisible to the bash-5 arm. Collapsing both arms onto one addressable token defeats the reason the loop was written. And the theme back-compat contract — the RFC-124 tier-0 invariant and every consumer-contract check — currently cannot be mutation-verified at all.

### Constraints
- Uniqueness is verified **by execution**, never by grepping call sites: these labels are runtime-interpolated, which is exactly the carrier a source scan misses.
- **Regex metacharacters are out of scope — do not implement anything for them.** The harness `re.escape`s the id before interpolation. The tracker's original premise is false.
- `basename` is not an acceptable interpreter discriminator — both interpreters are named `bash`.
- The theme-backcompat wrapper's contract is unchanged: no arguments, no network, no cwd assumptions, and the harness's exit status still reaches the runner via `exec`.
- `cogni-workspace/scripts/verify-theme-backcompat.sh` must not gate plainness on a capability probe (no `[ -t 1 ]` colour toggle) — that would make parsability depend on where the suite runs and the colour would return under a pty.
- No behavioural assertion changes; every suite stays green under `python3 scripts/run-plugin-tests.py`.

## Acceptance criteria
- [ ] Every result line emitted by each of the four changed suites has a first whitespace-delimited token shared by no other emitted result line in that same suite, across **all** loop iterations. Evidence: run each suite on a host where both interpreters exist, group emitted `PASS:` / `FAIL:` lines by first token, show maximum group size 1. Falsified by any token appearing on two lines — in particular the same case id from two loop iterations.
- [ ] The loop discriminator is derived from a value that actually differs between iterations. Evidence: show the two emitted ids for one converted case in `test-check-skill-names.sh` are distinct strings on a host with both `/opt/homebrew/bin/bash` and `/bin/bash`. Falsified if the two iterations produce the same id — which is what a `basename`-derived slug would do.
- [ ] `cogni-workspace/tests/test-theme-backcompat.sh` emits at least one addressable green result line per phase of the harness's contract, and its result lines carry no escape sequence before the label. Evidence: run the wrapper, show `PASS:`-prefixed lines with unique first tokens, and show the raw bytes before each label contain no `ESC`. Falsified by a run producing no matching result line, by a duplicated first token, or by an escape byte preceding a label.
- [ ] Mutation arm: for `test-check-skill-names.sh`, revert the bash-3.2 compatibility property in `cogni-workspace/scripts/check-skill-names.sh` (reintroduce a bash-4-only construct) and run the harness with `--case 1c-bin-bash`. Expect exactly that case red. Evidence: harness output shows `FAIL: 1c-bin-bash` and the homebrew-bash sibling id absent from the `FAIL:` lines. Falsified if the sibling id also reds — the collision this child exists to remove — or if the harness reports `case_not_found`. Repeat once for a `tbc…` id in the theme-backcompat harness.
- [ ] The wrapper's pass-through contract is intact. Evidence: make the harness exit non-zero in a scratch copy and show `bash cogni-workspace/tests/test-theme-backcompat.sh` exits with the same non-zero status. Falsified by an exit status that differs from the harness's, or by the wrapper capturing the status instead of `exec`ing.
- [ ] Every discovered suite is still green and only the intended paths changed. Evidence: `python3 scripts/run-plugin-tests.py` exits 0, and `git diff --name-only <base>...HEAD` lists only paths under `cogni-workspace/tests/`, `cogni-workspace/scripts/verify-theme-backcompat.sh`, `cogni-visual/tests/`, and `cogni-portfolio/tests/`. Falsified by a red suite or any other changed path, including any `version` value.

_Affected components: cogni-workspace/tests/test-check-skill-names.sh, cogni-workspace/tests/test-theme-backcompat.sh, cogni-workspace/scripts/verify-theme-backcompat.sh, cogni-workspace/tests/test-excalidraw-canvas-lock.sh, cogni-visual/tests/test-excalidraw-canvas-lock.sh, the colliding cogni-portfolio suite_

_Complexity hint: medium_

## Blocks

- #1499 repo: settle whether a CI guard should enforce per-suite case-id uniqueness, and record the outcome

claude-cogni-work:
Closes #1493.

## Summary

Three suites emitted result lines whose addressable first token named a case **scope**, while the scope emitted several lines — so the shared mutation harness could not tell those lines apart. A fourth emitted no addressable line at all.

| File | Before | After |
|---|---|---|
| `cogni-workspace/tests/test-check-skill-names.sh` | 9 colliding ids, 18 of 23 lines | 23 lines, 23 ids |
| `cogni-workspace/tests/test-excalidraw-canvas-lock.sh` | 7 ids for **28 of 28** lines | 28 lines, 28 ids |
| `cogni-portfolio/tests/test-dashboard-handoff.sh` | 1 colliding id, 2 of 13 lines | 13 lines, 13 ids |
| `cogni-workspace/scripts/verify-theme-backcompat.sh` | **0 addressable lines** | 22 addressable lines across 5 phases |

No behavioural assertion changes, no wrapper change, no version change.

## Three corrections to the issue's premises

The issue was filed before `cogni-visual` was retired, and generalises one mechanism to a suite that has a different one. All three were established by *running* the suites, not by reading the source — which matters, because these labels are runtime-interpolated and a source scan cannot see them.

1. **`cogni-visual/` no longer exists.** The issue names `cogni-visual/tests/test-excalidraw-canvas-lock.sh` as a target and says "both copies". Only the cogni-workspace copy survives; cogni-portfolio ships the *hook* that suite tests, not a copy of the suite. The real population is **3 colliding suites / 48 instances**, not "4 suites / 52 instances". The surviving byte-identity arm already compares cogni-portfolio ↔ cogni-workspace and holds no `cogni-visual` reference, so nothing needed repairing there.
2. **The canvas-lock mechanism is not the loop.** Only 2 of its 7 functions loop over `EXCALIDRAW_HOOKS`; the other 5 collide purely from calling `pass`/`fail` several times inside one body (`test_release_is_trapped_for_signals` alone emitted 10 lines under one id). A loop-variable slug — which is what the issue's worked example generalises to — would have left most of this suite colliding while the diff read as though it had addressed it.
3. **The issue's own worked example is the one label that did not collide.** `1c`/`2d`/`3c`/`5c` emit once per run, because `assert_no_bash4_error` returns early unless the interpreter is the bash-3.x one. Converting only `1c` and `2c` (the two strings the issue quotes verbatim) would have left all nine real collisions in place. All 13 matrix labels are converted.

## Two traps the implementation had to avoid

- **`basename` discriminates neither pair.** Both interpreters are named `bash` — the issue says so. Both hook copies are named `ensure-excalidraw-canvas.sh` — the same trap, one level over, which the issue does not mention. The interpreter slug comes from the full path, the hook slug from the owning plugin directory.
- **`tr -c 'A-Za-z0-9' '-'` does not produce the slug it looks like it produces.** Measured on this host under **both** `/bin/bash` 3.2.57 and `/opt/homebrew/bin/bash` 5.3.9: `echo "$BIN" | tr -c 'A-Za-z0-9' '-'` yields `-bin-bash-` — the leading `/` and the trailing newline each become a dash. Every helper here uses parameter expansion instead.

## Green/red id parity is load-bearing, not cosmetic

The mutation harness returns `guard_verified` only when the mutant is red **and** the restore is green under one `--case`. A check whose PASS branch and FAIL branch carry different ids can therefore never be verified — it degrades to `case_not_found`, which reads as a *missing case* rather than a surviving mutant. So every check's two branches share an id. In the theme harness this replaced the literal id `pre-flight`, which five distinct `fail` sites shared.

## Mutation recipe

Both arms the issue's acceptance criteria ask for, replayed and recorded.

**Arm A — the bash-3.2 property the interpreter matrix exists to protect.**

```bash
  --root . \
  --file cogni-workspace/scripts/check-skill-names.sh \
  --expr 's#^violations=0$#declare -A skill_map\nviolations=0#m' \
  --test 'bash cogni-workspace/tests/test-check-skill-names.sh' \
  --case 1c-bin-bash
```

`guard_verified`. The expression reintroduces the `declare -A` that `check-skill-names.sh:6-9` documents avoiding; `^violations=0$` matches exactly once (the other two are `$((…))` forms), so it cannot be a silent no-op, and `/m` is required under the harness's `perl -0pi`.

**Arm A′ — the discrimination itself, on a genuinely-colliding pair.** Acceptance criterion 4 asks for `1c-bin-bash` and for its homebrew sibling to be absent from the FAIL lines — but `1c` has no homebrew sibling at all, so that demand is satisfied by construction rather than by the collision being removed. Under the *same* mutant:

| `--case` | verdict |
|---|---|
| `2c-bin-bash` | `guard_verified` — red under mutation |
| `2c-opt-homebrew-bin-bash` | `vacuous_guard` — **green** under the same mutation |

That asymmetry is the defect being fixed. Before this change both lines carried the token `2c`, so the bash-5 arm going red would have credited a mutation only the bash-3.2 arm can detect — the arm that exists precisely because a bash-4-ism is invisible under bash 5.

**Arm B — the theme-backcompat harness, which could not contribute mutation evidence at all before this change.**

```bash
  --root . \
  --file cogni-workspace/scripts/baselines/_template-tier0-output.json \
  --expr 's#"source": "standard"#"source": "mutated"#' \
  --test 'bash cogni-workspace/tests/test-theme-backcompat.sh' \
  --case tbc10-tier0-baseline-match
```

`guard_verified`. Recorded in that file's header.

**The nine recorded recipes M1–M9** in `test-excalidraw-canvas-lock.sh` are re-keyed to the specific sub-id each `--expr` reddens and **all nine replay to `guard_verified`** — the header asserts exactly that, and a stale `--case` would have degraded silently to `case_not_found`. M3, M6 and M9 mutate the cogni-workspace copy only; graded against the cogni-portfolio sibling id, M3's mutant returns `vacuous_guard` (green), which is the per-hook discrimination working.

## Verification

- [x] `python3 scripts/run-plugin-tests.py` — `total=124 passed=124 failed=0`, identical to base.
- [x] Uniqueness **by execution**, per suite: emitted `PASS:`/`ok:`/`FAIL:` lines grouped by first token, `uniq -d` empty for all four surfaces.
- [x] The loop discriminator genuinely differs per iteration: `1a-bin-bash` / `1a-opt-homebrew-bin-bash`, and `…-cogni-workspace` / `…-cogni-portfolio`.
- [x] The wrapper emits ≥1 `PASS: tbc…` line in each of the harness's five contract phases, with no escape byte before any label (`cat -v | grep -c '\^\['` → 0). Phase E is the informational one the harness excludes from its own "5 phases" count and has no `fail` counterpart, so it is not given a synthetic green.
- [x] Wrapper pass-through intact: with the baseline mutated, harness `rc=1` and wrapper `rc=1`; restored, both `0`.
- [x] `scripts/check-result-line-plainness.py` — 0 findings, 124 files discovered, unchanged. That guard's non-recursive globs do not reach `scripts/verify-theme-backcompat.sh` before or after this change, so the harness's plainness rests on the file, not on CI; the header now says so.
- [x] Only the four intended paths changed; no `version` value moved; the `cogni-visual:` token pinned by `test-relocated-skill-hygiene.sh:138` is not reintroduced.

## Disclosures

- **Token-scope preflight was run under `--allow-overscoped`.** `check-token-scope.sh --strict` exits 1 on this runner because the token is a `gho_` token from `gh auth login`, whose scope set is fixed by the GitHub CLI OAuth app and cannot be narrowed. That is the opt-out the script's own guidance names for this case; flagging it rather than passing it over.
- **The Step 6.4 `/simplify` pass was performed as a direct diff review, not via the harness skill.** That pass reads the session working tree, and this work was done in an isolated resolve worktree, so pointing it at the session tree would have reviewed an empty diff. Reviewed for reuse and duplication by hand: the three slug helpers are structurally similar but live in different self-contained suites, and folding them would require a shared test library this repo does not have — which would be scope-widening.
- **`verify-theme-backcompat.sh` carries pre-existing `#126`–`#138` issue refs** in its triage prose. They are outside `check-breadcrumbs.py`'s globs and outside the handoff preflight's breadcrumb arm (which fires only on touched `SKILL.md` / `agents/*.md`), and they are left exactly as they were.
- **The `--help` triage table's pre-existing quote drift was not repaired.** Two of its six rows quote labels that no longer match the emitted text (`:104` says `.py` where the label does not; `:108`/`:115` predate the current interpolated label). The rows gained their case ids, but correcting the stale quotes is a separate defect and out of scope here.

## Blocked

The handoff preflight (`check-preflight.sh`) exits 1 on its `mutation-recipe` instrument, so this PR is **not** handed off to the merger — it carries no `svc:needs-review`.

The instrument searches the suite named by `--test`, whole-file and comment-excluded, for the literal `--case` value. The recorded `1c-bin-bash` is assembled at run time from the interpreter path (`"1c-$BIN_SLUG …"`), so it exists in no source line — the same blind spot #1493 names as its central constraint. Replay contradicts the instrument's prediction: the recipe returns `guard_verified`, not `case_not_found`.

Every other gating instrument is clean (`run-tests` pass, `version-bump` pass, four `not_applicable` skips). Full detail, the replay table, and three options for a ruling are in the [preflight comment](https://github.com/cogni-work/insight-wave/pull/1507#issuecomment-5333504387). Needs a human ruling; the instrument itself ships in `cogni-service`, so any fix belongs in `cogni-work/managed-service`.

### DIFF
--- a/cogni-portfolio/tests/test-dashboard-handoff.sh
+++ b/cogni-portfolio/tests/test-dashboard-handoff.sh
@@ -303,22 +303,28 @@ test_refresher_has_no_skip_path() {
 # --- test_open_browser_optin -----------------------------------------------
 # Two halves: the documented default is the non-opening one, and every call site
 # whose prose promises to open names the opt-in explicitly.
+#
+# Both halves emit a green line on a clean run, so the addressable id names the
+# HALF rather than the case — a single `test_open_browser_optin` token would name
+# two lines and a mutation of either half could be credited by the other. The
+# function name itself is unchanged: ALL_TESTS/run_one and scripts/mutation-check.sh
+# dispatch on it.
 test_open_browser_optin() {
-  agent_surface_ok test_open_browser_optin || return
+  agent_surface_ok test_open_browser_optin-surface || return
 
   # Half 1 — the input contract.
   local contract_line
   contract_line="$(grep -F '`open_browser` (optional' "$AGENT" | head -1)"
   if [ -z "$contract_line" ]; then
-    fail test_open_browser_optin "no open_browser input-contract line in agents/dashboard-refresher.md"
+    fail test_open_browser_optin-contract-default-false "no open_browser input-contract line in agents/dashboard-refresher.md"
   else
     case "$contract_line" in
-      *"default: false"*) pass test_open_browser_optin "input contract documents the non-opening default" ;;
-      *) fail test_open_browser_optin "input-contract line does not document 'default: false': $contract_line" ;;
+      *"default: false"*) pass test_open_browser_optin-contract-default-false "input contract documents the non-opening default" ;;
+      *) fail test_open_browser_optin-contract-default-false "input-contract line does not document 'default: false': $contract_line" ;;
     esac
   fi
   if grep -qF 'default: true' "$AGENT"; then
-    fail test_open_browser_optin "agent contract still documents 'default: true' somewhere"
+    fail test_open_browser_optin-no-default-true "agent contract still documents 'default: true' somewhere"
   fi
 
   # Half 2 — the call-site surface. Three globs: agents/, skills/*/SKILL.md and
@@ -332,7 +338,7 @@ test_open_browser_optin() {
   )"
   scanned="$(printf '%s\n' "$surface" | grep -c . )"
   if [ "$scanned" -eq 0 ]; then
-    fail test_open_browser_optin "scanned no files; scan surface is broken"
+    fail test_open_browser_optin-scan-surface "scanned no files; scan surface is broken"
     return
   fi
 
@@ -350,7 +356,7 @@ test_open_browser_optin() {
   )"
   dispatch_count="$(printf '%s\n' "$dispatch_lines" | grep -c . )"
   if [ "$dispatch_count" -eq 0 ]; then
-    fail test_open_browser_optin "found no dashboard-refresher dispatch lines; scan surface is broken"
+    fail test_open_browser_optin-dispatch-surface "found no dashboard-refresher dispatch lines; scan surface is broken"
     return
   fi
 
@@ -380,9 +386,9 @@ $dispatch_lines
 EOF
 
   if [ -n "$offenders" ]; then
-    fail test_open_browser_optin "call sites promise to open but pass no flag:$offenders"
+    fail test_open_browser_optin-callsite-optin "call sites promise to open but pass no flag:$offenders"
   else
-    pass test_open_browser_optin "all intending call sites name the opt-in ($dispatch_count dispatch lines scanned)"
+    pass test_open_browser_optin-callsite-optin "all intending call sites name the opt-in ($dispatch_count dispatch lines scanned)"
   fi
 }
 
--- a/cogni-workspace/scripts/verify-theme-backcompat.sh
+++ b/cogni-workspace/scripts/verify-theme-backcompat.sh
@@ -18,6 +18,51 @@
 # This harness also runs in CI, via the thin wrapper suite
 # cogni-workspace/tests/test-theme-backcompat.sh that run-plugin-tests.py
 # discovers. Keep it runnable with no arguments and no network access.
+#
+# RESULT-LINE CONTRACT. Every result line is plain, colon-terminated and
+# id-first: `PASS: <case-id> <label>` / `FAIL: <case-id> <label>`. The case id is
+# what the shared mutation harness addresses, and it matches whole-token —
+# `^[ \t]*FAIL:[ \t]+<case>(?:[ \t]|$)` — so no colour, no leading whitespace
+# and no colon may sit between the label and the id, or between the id and the
+# text. Plainness is NOT gated on a capability probe (`[ -t 1 ]`, `$TERM`): that
+# would make parsability depend on where the harness runs, and the colour would
+# come back under a pty.
+#
+# A check's PASS branch and its FAIL branch carry the SAME id, deliberately.
+# The mutation harness returns `guard_verified` only when the mutant is red AND
+# the restore is green under one `--case`, so a check whose two branches carry
+# different ids can never be verified — it degrades to `case_not_found`, which
+# reads as a missing case rather than going red. Several `fail` arms may share
+# one id (the TIERS_PROBE case does): `fail()` exits on the first failure, so at
+# most one FAIL line is ever emitted per run.
+#
+# `c_skip` and `c_info` are deliberately NOT result lines and keep their colour —
+# they must never begin with `PASS:` or `FAIL:`.
+#
+# This file lives under scripts/, one path segment outside
+# scripts/check-result-line-plainness.py's two non-recursive globs
+# (`tests/*.sh`, `*/tests/*.sh`), so that guard does not cover it — before or
+# after this contract was adopted. The plainness above rests on this file, not
+# on CI. Do not "fix" that by moving the harness under tests/: the surviving
+# c_skip/c_info escape literals would then be discovered and turn the guard's
+# hard clean zero red.
+#
+# Mutation recipe — verified, replayable. The wrapper is the only runnable entry
+# point (this harness takes no per-case selector), and because fail() exits on
+# the first failure the mutated file must redden the chosen id before any
+# earlier phase can fail.
+#
+#   M1 -> tbc10-tier0-baseline-match       (verdict: guard_verified)
+#     bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.402/scripts/mutation-check.sh \
+#       --root . \
+#       --file cogni-workspace/scripts/baselines/_template-tier0-output.json \
+#       --expr 's#"source": "standard"#"source": "mutated"#' \
+#       --test 'bash cogni-workspace/tests/test-theme-backcompat.sh' \
+#       --case tbc10-tier0-baseline-match
+#
+#   Before this contract existed the harness emitted zero addressable lines, so
+#   every invocation of the form above returned case_not_found and no check here
+#   could contribute mutation evidence at all.
 
 set -uo pipefail
 
@@ -44,8 +89,8 @@ trap cleanup EXIT
 # Output helpers
 # --------------------------------------------------------------------------
 
-c_pass() { printf "  \033[32mPASS\033[0m  %s\n" "$1"; }
-c_fail() { printf "  \033[31mFAIL\033[0m  %s\n" "$1"; }
+c_pass() { printf 'PASS: %s\n' "$1"; }
+c_fail() { printf 'FAIL: %s\n' "$1"; }
 c_skip() { printf "  \033[33mSKIP\033[0m  %s\n" "$1"; }
 c_info() { printf "  \033[36mINFO\033[0m  %s\n" "$1"; }
 
@@ -90,29 +135,34 @@ On the first failure, prints a triage line and exits 1.
 
 Failure-mode triage table
 
-  - "tier-0 baseline mismatch" or "tier-0 fixture discover failed":
+  - tbc10-tier0-baseline-match ("tier-0 baseline mismatch") or
+    tbc06-tier0-fixture-discover ("tier-0 fixture discover failed"):
       Likely #126 (cogni-workspace: discover-themes.py reads manifest.json
       with tier-0 fallback). The contract is "no manifest.json => byte-
       identical legacy output"; if discover added a field, this fires.
 
-  - "cogni-work missing tiers.tokens" or "tokens.css absent":
+  - tbc13-cogni-work-tiers-tokens ("cogni-work missing tiers.tokens" or
+    "tokens.css absent"):
       Likely #127 (cogni-workspace: migrate cogni-work theme to tiered
       directory structure as Phase 2 reference implementation) or #128
       (manifest schema). Either the migration regressed or the schema
       stopped accepting the existing manifest.
 
-  - "validate-theme-manifest.py rejects cogni-work":
+  - tbc14-validator-accepts-cogni-work ("validate-theme-manifest.py rejects
+    cogni-work"):
       Likely #128 (manifest schema definition) or #138 (manage-themes deep
       authoring) — schema and authoring surfaces drifted.
 
-  - "render-html-slides theme contract missing":
+  - tbc20-consumer-theme-ref-<plugin>-<skill> ("render-html-slides theme
+    contract missing"):
       Likely #129 (the cogni-visual refactor of render-html-slides to consume
       tier-1 tokens and tier-3 component primitives from cogni-work).
 
-  - "migration guide reference missing":
+  - tbc18-migration-guide-present ("migration guide reference missing"):
       Likely #130 (cogni-workspace: write Theme System v2 migration guide).
 
-  - "consumer SKILL.md reference missing":
+  - tbc20-consumer-theme-ref-<plugin>-<skill> ("consumer SKILL.md reference
+    missing"):
       Cross-cutting drift; could be a SKILL.md regeneration that lost the
       theme reference. File against the affected plugin.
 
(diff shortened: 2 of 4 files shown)


claude-cogni-work:
## Blocked — handoff preflight, `mutation-recipe` instrument

`check-preflight.sh` exits **1**, so this PR is **not** handed off to the merger: no `svc:needs-review`, no self-grade. One violation.

```
check: mutation-recipe
detail: the recorded --case '1c-bin-bash' has no matching label in
        cogni-workspace/tests/test-check-skill-names.sh —
        mutation-check.sh would return case_not_found
```

Every other gating instrument is clean:

| Instrument | Verdict |
|---|---|
| `run-tests` | pass — suites green for cogni-portfolio, cogni-workspace |
| `mutation-recipe` | **fail** |
| `frontmatter` / `breadcrumbs` / `skill-length` / `agent-length` | skip — `not_applicable`, no touched `SKILL.md` or `agents/*.md` |

`discovered 7 · resolved 7 · gating_applicable 3 · verdicts 3 · passed 2 · failed 1 · skipped 4`.

### Why this is an instrument blind spot rather than a defective recipe

The instrument resolves the suite named by `--test` and searches it, whole-file and comment-excluded, for the literal `--case` value. That is sound for a static case id, and its own notes record it as measured against a 114-recipe corpus.

Every id in that corpus is a **static literal in the source**. This change is the first to record one that is **runtime-interpolated by construction** — the source says

```sh
assert_no_bash4_error "1c-$BIN_SLUG clean tree free of bash-4 runtime error ($BIN)"
```

so the string `1c-bin-bash` exists only at run time, assembled from the interpreter path. A whole-file literal search cannot find it, and no rewriting of the recipe changes that while the discriminator remains a runtime value — which is the entire point of #1493.

This is the same blind spot the issue itself names as its central constraint:

> Uniqueness is verified **by execution**, never by grepping call sites: these labels are runtime-interpolated, which is exactly the carrier a source scan misses.

### The recipe is verified — by replay, which is the authority

```
--case 1c-bin-bash                  guard_verified   (mutated: red / restored: green)
--case 2c-bin-bash                  guard_verified   (mutated: red / restored: green)
--case 2c-opt-homebrew-bin-bash     vacuous_guard    (mutated: green — the sibling arm, correctly)
--case tbc10-tier0-baseline-match   guard_verified
M1–M9 (test-excalidraw-canvas-lock) guard_verified   (all nine, re-keyed and replayed)
```

`case_not_found` is what the instrument predicts. The harness returns `guard_verified`, so the prediction is falsified on the only surface that can settle it.

### What I did not do

I did **not** work around this by making the literal appear in the source, reordering the `## Mutation recipe` section so a statically-resolvable arm is selected instead, or dropping the `1c-bin-bash` arm. The first games the guard; the second and third would satisfy the instrument while removing the arm acceptance criterion 4 explicitly demands — and would leave the blind spot un-surfaced for the next PR that hits it.

### Needs a human ruling

Three options, no strong preference from here:

1. **Waive for this PR** and file a follow-up so the instrument grows an affordance for interpolated case ids — a `--case-is-interpolated` marker on the recipe, or degrading to the existing pass-carried `recipe_target_unresolved` reason when the id contains no statically-locatable literal.
2. **Rule that recorded recipes must carry a statically-resolvable `--case`**, and accept that the interpreter-matrix arms are documented in the PR body but not in the `## Mutation recipe` section. This narrows what the section can express.
3. **Treat it as a real block** and require the ids be static — which would mean abandoning the runtime discriminator and re-opening the collision #1493 exists to close.

Option 1 looks right to me, but the instrument is not mine to change from this repo: it ships in `cogni-service`, so the follow-up belongs in `cogni-work/managed-service`.

Everything else in this PR is complete and verified; the branch is pushed and green.

claude-cogni-work:
## Follow-up filed

The instrument blind spot is now tracked as **cogni-work/managed-service#923** — *"check-preflight mutation-recipe blocks a runtime-interpolated `--case` that replays to `guard_verified`"*.

It carries the reproduction, the replay table, and a suggested direction: detect the interpolation and degrade to a **pass-carried named reason** (`recipe_case_interpolated` + an `advisories[]` entry), the way the instrument already handles `recipe_target_unresolved` / `recipe_target_uncontained`. That narrows the check rather than removing it — a stale `--case` on a static id keeps failing, which is what it was built to catch.

This PR still needs a maintainer waiver to hand off: `check-preflight.sh` keys on the exit code, so there is no author-side path around it. Everything else here is complete and verified.

sdh07:
## Maintainer waiver — `mutation-recipe` preflight false-positive

Waiving the `check-preflight.sh` `mutation-recipe` block and handing this PR to the merger (`svc:review-blocked` → `svc:needs-review`). The block is a genuine instrument blind spot, not a defect in this change, and there is no author-side path around an exit-code-keyed gate.

**Why it fired.** The instrument statically searches the named suite for the recorded `--case` and predicts `case_not_found` for `1c-bin-bash`, because that id is **runtime-interpolated** (`assert_no_bash4_error "1c-$BIN_SLUG …"`) and exists only at execution time. This is the first recipe in the corpus built that way.

**Verified independently before waiving** (PR head `ab82ac69`, `sdh07`):

| Check | Result |
|---|---|
| `python3 scripts/run-plugin-tests.py` | `total=124 passed=124 failed=0` (identical to base) |
| First-token uniqueness, by execution, all 4 touched suites | zero duplicate ids (23 / 28 / 13 / 24 lines) |
| `mutation-check.sh … --case 1c-bin-bash` (the blocked id) | **`guard_verified`** — mutant red, restore green |
| `mutation-check.sh … --case tbc10-tier0-baseline-match` (Arm B) | `guard_verified` |
| `2c-bin-bash` / `2c-opt-homebrew-bin-bash` under one mutant | `guard_verified` / `vacuous_guard` — the per-interpreter discrimination this PR adds, working |
| `check-preflight.sh` | exit 1, **sole** violation = the `mutation-recipe` static misprediction above; every other instrument passes or is not-applicable |

So the instrument's own prediction ("`mutation-check.sh` would return `case_not_found`") is falsified by running it — it returns `guard_verified`.

The instrument fix is tracked as cogni-work/managed-service#923 (degrade an interpolated `--case` to a pass-carried named reason, the way it already handles `recipe_target_unresolved`). No source or instrument was edited to clear the gate — this is the documented maintainer escape hatch, resting on the replay evidence above.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1507
**Perspectives consulted:** cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator, plugin-dev:plugin-validator

**Summary:** All three perspectives passed with zero findings: the contract-grading-reviewer confirmed all 31 committed acceptance-checklist items satisfied by execution, and both plugin-validator passes confirmed valid manifest structure and version-mirror consistency with no in-branch version bump.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- All 31 acceptance_checklist items graded satisfied with cited evidence; uniqueness by execution 23/28/13/24 with zero duplicate ids; mutation arms A/A'/B replay to guard_verified; run-plugin-tests 124/124; plainness 0 findings/124 files. (cogni-service:contract-grading-reviewer)
- cogni-workspace plugin.json valid, version mirror consistent, structure intact, no in-branch version bump. (plugin-dev:plugin-validator)
- cogni-portfolio plugin.json valid, version mirror consistent, structure complete, test suite passes bash -n. (plugin-dev:plugin-validator)

### Recommendation

---

Issue #1512, opened 2026-08-18
repo: root README MCP table still names the retired cogni-visual for excalidraw and pencil


claude-cogni-work:
## Context

Deferred from #1475 during service-resolve (full implementation of the parent shipped the rest). Tracked here so the scope is not silently dropped.

**Deferred item:** Correct the MCP Servers table in the repo-root `README.md` (lines 163 and 165 at origin/main, `Used by` column) — still names the retired `cogni-visual` for excalidraw and pencil. A fifth mirror of registry `required_by`, outside issue #1475's scope. Note it spells plugins as markdown links (`[cogni-visual](cogni-visual/README.md)`, a link into a deleted README), so pinning it with a guard later needs a link-stripping parse rather than the plain `, ` split the A4-A6 arms use.

**Source of truth:** `cogni-workspace/references/mcp-git-registry.json` — `servers.<name>.required_by`. The registry is correct; the prose mirror follows it, never the reverse.

**Parent issue:** #1475
**Plan record:** https://github.com/cogni-work/insight-wave/issues/1475#issuecomment-5333971786

sdh07:
Consolidated into #1514 via `/cogni-service:service-groom` — this is one of three same-class retired-`cogni-visual` MCP-table fixes (all deferred from #1475), now tracked as a single 6-row sweep in #1514 so the scope isn't re-fragmented. Closing not-planned in favour of #1514.

Issue #1511, opened 2026-08-18
cogni-workspace: manage-workspace worked example still names the retired cogni-visual as an MCP consumer


claude-cogni-work:
## Context

Deferred from #1475 during service-resolve (full implementation of the parent shipped the rest). Tracked here so the scope is not silently dropped.

**Deferred item:** Correct the excalidraw/pencil MCP ownership rows in `cogni-workspace/skills/manage-workspace/SKILL.md` (lines 163-164 at origin/main: `<- cogni-visual, cogni-portfolio` and `<- cogni-visual, cogni-website` in the illustrative manage-workspace output block) — a fourth stale mirror of registry `required_by` that issue #1475 does not name and that the new A4-A6 arms do not cover, so it stays stale and unguarded after that PR.

**Source of truth:** `cogni-workspace/references/mcp-git-registry.json` — `servers.<name>.required_by`. The registry is correct; the prose mirror follows it, never the reverse.

**Parent issue:** #1475
**Plan record:** https://github.com/cogni-work/insight-wave/issues/1475#issuecomment-5333971786

sdh07:
Consolidated into #1514 via `/cogni-service:service-groom` — this is one of three same-class retired-`cogni-visual` MCP-table fixes (all deferred from #1475), now tracked as a single 6-row sweep in #1514 so the scope isn't re-fragmented. Closing not-planned in favour of #1514.

Issue #1475, opened 2026-08-18
cogni-workspace: MCP required_by is mirrored in three hand-maintained places with nothing pinning them


claude-cogni-work:
## Context

Surfaced while resolving #1469, which had to add `cogni-website` to the pencil server's consumer list.

That one fact lives in three hand-maintained places, and nothing pins them together:

1. `cogni-workspace/references/mcp-git-registry.json` — `servers.<name>.required_by` (the machine-read copy; `cogni-workspace/skills/install-mcp/SKILL.md` reads it to decide provisioning)
2. `cogni-workspace/skills/workspace-status/SKILL.md` — the MCP probe table's plugins column
3. `cogni-workspace/skills/workspace-status/references/mcp-registry.md` — the per-server `**Skills:**` line

#1469 is itself the evidence for the drift class: the fix had to remember all three by hand, and had all three not been updated, `workspace-status` would have reported pencil as cogni-visual-only in the same breath as the registry saying otherwise. No test would have caught it. The next plugin to adopt an MCP server faces the same trap.

## Why this is worth pinning

The repo already treats mirrored copies as a guardable invariant — `cogni-workspace/tests/test-mcp-declaration-hygiene.sh` case D1 asserts the two copies of `concept-mcp-server-map.md` are byte-identical, precisely so a one-sided edit reds instead of drifting. The `required_by` mirrors have the same shape and no such guard.

## Proposed change

Either of two directions — the second is stronger if it is practical:

1. **Guard the mirrors.** Add an arm to `cogni-workspace/tests/test-mcp-declaration-hygiene.sh` (which already parses this registry and lives in the plugin that owns it) asserting that each server's `required_by` set matches the plugins column of the `workspace-status` probe table.
2. **Remove the mirror.** Derive the probe table from the registry at render time so there is one source of truth and nothing to keep in sync.

Direction 2 needs a check on whether that table is hand-written by convention; if it is, direction 1 is the pragmatic fix.

## Acceptance criteria

- [ ] A one-sided edit — adding a plugin to a server's `required_by` without updating the `workspace-status` probe table (or vice versa) — turns a discovered test suite red.
- [ ] The guard carries a replayable mutation recipe demonstrating that red, per the repo's existing suite convention.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-workspace` exits 0 on the corrected tree.

Refs #1469.

claude-cogni-work:
## Correction: four mirrors, not three

[...]

PR #1515, opened 2026-08-18
cogni-workspace: pin MCP required_by mirrors to the registry

claude-cogni-work:
Refs #1475.

## Summary

`required_by` is one fact rendered in four places — the machine-read registry plus three hand-maintained prose surfaces — with nothing pinning them. The issue predicted the drift; the tree already had it. The `cogni-visual` retirement updated the registry and left **all three** mirrors naming a retired plugin, and `test-mcp-declaration-hygiene.sh` reported `9 passed, 0 failed` over it. Nothing anywhere would have said a word.

This adds the guard and corrects the data:

- **`A4`/`A5`/`A6`** compare each registry server's `required_by` against its prose mirror — the `workspace-status` probe table, that skill's `mcp-registry.md` relation line, and the `install-mcp` plan example.
- **`L4`/`L5`/`L6`** are liveness floors, thresholded at two rows to match `L1`. A renamed path or restyled table yields zero parsed rows, which is also what a clean comparison looks like from the arms alone — so without the floors those arms would pass hardest exactly when the parser had gone blind. Two rather than one because the registry carries two servers: a surface that parses down to a single row has lost at least one of them, and while the arms would red on that too, they would report it as a missing row in the document rather than as a parser that stopped seeing rows.
- The three mirrors are corrected so the suite is green. **The registry is the source of truth and is not edited** — prose follows it, never the reverse.

## Design notes

**`claude-in-chrome` falls out by construction, not by an exclusion list.** Each arm iterates the *registry's* server set and looks the prose row up. A server with no registry entry — `claude-in-chrome` is a manual Chrome-extension install — is therefore never compared, and no name-based skip is needed to keep it out. The reverse direction stays loud: a registry server with no prose row is a `MISSING` defect, never a silently skipped pair.

**Three surfaces, three key fields.** The probe table and `mcp-registry.md` key on `desktop_config_key`; the `install-mcp` worked example keys on the registry *server name* (`mcp_excalidraw`). That asymmetry is preserved rather than normalised away — the example sits directly under a `cat` of the registry, so the registry-name key is the locally coherent choice. Key resolution reads `desktop_config_key` as data with a fallback to the server name, matching how `scripts/check-mcp-tool-grant.py` normalises the same registry; neither strips an `mcp_` prefix.

**Cosmetics cannot red it.** Plugin lists compare as sorted **sets**, so ordering is free. Both relation spellings in use (`**Needed by:**`, `**Used by:**`) are accepted, so no prose had to be rewritten to satisfy the guard.

**One registry parse.** The existing python block is extended rather than duplicated — its own header comment records that two copies of it previously drifted apart in review. It emits exactly eight positional lines; each surface read is wrapped separately so one unreadable mirror reports count 0 instead of aborting the others.

## Verification

Run on the branch, and re-run unchanged after the `/simplify` pass:

| Check | Result |
|---|---|
| `bash cogni-workspace/tests/test-mcp-declaration-hygiene.sh` | `15 passed, 0 failed` (was 9) |
| Same suite under real `/bin/bash` 3.2.57 | `15 passed, 0 failed` |
| `python3 scripts/run-plugin-tests.py --filter cogni-workspace` | exit 0 |
| `python3 scripts/run-plugin-tests.py` (unfiltered) | exit 0 — **124/124** suites |
| `check-result-line-plainness.py` | exit 0 (124 files, 97 emitter definitions) |
| `check-breadcrumbs.py` / `check-version-bump.py` / `check-mcp-tool-grant.py` / `check-workflow-yaml.py` | exit 0 |

Negative tests, each on a mutant tree via the suite's own `MCP_HYGIENE_ROOT` override:

| Mutation | Expected | Observed |
|---|---|---|
| Registry side: add a plugin to `mcp_excalidraw.required_by` | all three arms red | `A4`,`A5`,`A6` red — `11 passed, 3 failed` |
| Prose side: drop `cogni-workspace` from the probe-table row | only `A4` red | `A4` red, `A5`/`A6` green — `13 passed, 1 failed` |
| Delete `mcp-registry.md` | floor reds, arm does **not** pass vacuously | `L5` red **and** `A5` red (`surface unreadable`) |
| Remove the `pencil` probe-table row | `MISSING`, not a skipped pair | `A4` red — `MISSING pencil` |
| Registry made unparseable | all floors fail closed, no foreign diagnostic | `L2`/`L4`/`L5`/`L6` red, **stderr empty** |
| Probe table degraded to a single row | floor reds, not just the arm | `L4` red (`yielded 1 rows`) alongside `A4` |

That last row is why the six new `sed` reads carry the house `[ -n "$x" ] || x=-1` normalisation: an unguarded `[ "" -ge 1 ]` still reds, but by way of a locale-dependent foreign "integer expression expected" message rather than the suite's own.

## Mutation recipe

```
  --root . \
  --file cogni-workspace/references/mcp-git-registry.json \
  --expr 's/"required_by": \["cogni-portfolio", "cogni-workspace"\]/"required_by": ["cogni-portfolio", "cogni-workspace", "cogni-marketing"]/' \
  --test 'bash cogni-workspace/tests/test-mcp-declaration-hygiene.sh' \
  --case A4
```

Verdict: `guard_verified`. The search literal occurs exactly once (`pencil`'s `required_by` is a different pair), and the replacement appends a plugin before the closing bracket, so the mutant cannot re-satisfy the pattern and stay green. The same mutant also reds `A5` and `A6` — one registry edit desyncs all three mirrors at once — which is expected and not a reason for a second recipe. `L4`-`L6` stay green throughout, which is what separates a real comparison from a parser that stopped seeing rows.

The case ids are written out as static literals at their `pass()`/`fail()` call sites and are deliberately **not** collapsed into a loop over their ids: the handoff preflight literal-searches the suite source for the recipe's `--case` value with comment lines excluded, so an interpolated id replays as `case_not_found` while the arm itself works perfectly.

## Scope

`Refs`, not `Closes` — this ships #1475's full stated scope, but exploration found the same defect on three further surfaces that the issue does not name, and they are filed rather than folded in:

| Issue | Surface | Why it is separate |
|---|---|---|
| #1511 | `cogni-workspace/skills/manage-workspace/SKILL.md:163-164` | A fourth mirror, not named by #1475 |
| #1512 | repo-root `README.md:163,165` | Spells plugins as markdown links into a deleted `cogni-visual/README.md`; guarding it needs a link-stripping parse |
| #1513 | repo-root `CLAUDE.md:169,171` | A **superset** — the correct entry is present with the retired one kept alongside |

Widening to cover them would take this from four files to eight and mix a dead-link repair into a guard PR. Whether a plugin-scoped suite should guard repo-root docs at all is a scope call worth a human.

A fourth mirror is recorded in the suite header as knowingly uncovered rather than left implicit: the `concept-mcp-server-map.md` **Plugins** column, in both wiki trees. It is correct today, and `D1` does *not* cover it — `D1` asserts only that the two wiki copies stay byte-identical to **each other**, never that either agrees with the registry.

Also deliberately not asserted, and recorded in the suite header: the `**Skills:**` lines in `mcp-registry.md` (a different relation — skill names, no registry counterpart), and any token-level "no retired plugin name" scan, since `cogni-visual/` is a live *output-directory* path in 30+ lines under this plugin and such a scan would red on day one.

No version bump — the post-merge workflow owns it.

Graded against the 23-item acceptance contract posted on #1475 at triage, before this branch existed. First pass: **21 satisfied, 2 violated**, both fixed in `34720c9f` and re-verified:

- **Floors were `-ge 1` where the contract asked for `>= 2` (or registry-derived).** Tightened to two, with the labels updated to match, and a new negative test covering the degrade-to-one-row case.
- **The header's exclusions list omitted `concept-mcp-server-map.md`.** Added, with the reason `D1` does not already cover it.

No residual: both violations were mechanical and landed in a file this PR already touched.

## Review notes

Both touched `SKILL.md` files were put through the skill-quality gate. Every finding is **pre-existing and category `preference`**; none was introduced by this diff, and the one structural item carries `introduced_by_change: false`. Annotated here rather than auto-fixed, because each would widen a guard PR into unrelated prose:

- `install-mcp/SKILL.md` — the `description` and the *When Called from manage-workspace* section both pin another skill's step by ordinal ("step 5"). That is itself an unpinned mirror of another skill's numbering. Fixing it edits triggering-adjacent frontmatter, so it is left for a deliberate change.
- `workspace-status/SKILL.md` — headings run 1, 2, 3, 4, 5, 5.5, 6 while the body promises "seven checks"; `allowed-tools` grants `Write`/`Edit` that the body never exercises; and ~700 chars of restart/two-step install prose is duplicated between the body and `references/mcp-registry.md`. The last is the same duplication class this PR is about, one level up.

The reviewer independently traced the guard end to end — `PROBE_TABLE_REL` → arm `A4` → floor `L4` — and confirmed a future one-sided edit cannot pass silently, and neither can a table rename that would make `A4` vacuous.

**Runner disclosure:** the token-scope preflight reports `over_scoped` (extra: `gist`, `read:org`, `workflow`). The runner is on a `gho_` CLI token whose scope set is fixed by GitHub's OAuth app and cannot be narrowed, so the documented `--allow-overscoped` opt-out was used; it exits 0 while leaving the status string unchanged.

### DIFF
--- a/cogni-workspace/skills/install-mcp/SKILL.md
+++ b/cogni-workspace/skills/install-mcp/SKILL.md
@@ -103,8 +103,8 @@ this example:
 
 ```
 MCP Installation Plan:
-  mcp_excalidraw  git clone + build    needed by: cogni-visual, cogni-portfolio
-  pencil          native app check     needed by: cogni-visual, cogni-website
+  mcp_excalidraw  git clone + build    needed by: cogni-portfolio, cogni-workspace
+  pencil          native app check     needed by: cogni-website, cogni-workspace
 
   Config write:    ~/.claude.json (Claude Code user scope)
   Config write:    claude_desktop_config.json (Claude Desktop)
--- a/cogni-workspace/skills/workspace-status/SKILL.md
+++ b/cogni-workspace/skills/workspace-status/SKILL.md
@@ -177,9 +177,9 @@ means it is not available. The **Install** column decides how a missing server i
 
 | MCP Server | Probe tool | Needed by | Install |
 |------------|-----------|-----------|---------|
-| `excalidraw` | `mcp__excalidraw__describe_scene` | cogni-visual, cogni-portfolio | install-mcp |
+| `excalidraw` | `mcp__excalidraw__describe_scene` | cogni-portfolio, cogni-workspace | install-mcp |
 | `claude-in-chrome` | `mcp__claude-in-chrome__tabs_context_mcp` | cogni-website, cogni-workspace | manual (Chrome extension) |
-| `pencil` | `mcp__pencil__get_editor_state` | cogni-visual, cogni-website | manual (Pencil desktop app) |
+| `pencil` | `mcp__pencil__get_editor_state` | cogni-website, cogni-workspace | manual (Pencil desktop app) |
 
 **Report format**:
 
--- a/cogni-workspace/skills/workspace-status/references/mcp-registry.md
+++ b/cogni-workspace/skills/workspace-status/references/mcp-registry.md
@@ -12,7 +12,7 @@ only once it is actually installed.
 
 ### excalidraw (yctimlin/mcp_excalidraw)
 
-- **Needed by:** cogni-visual, cogni-portfolio
+- **Needed by:** cogni-portfolio, cogni-workspace
 - **Type:** git-installed (cloned and built by `cogni-workspace/scripts/install-mcp.sh`)
 - **Install path:** `~/.claude/mcp-servers/mcp_excalidraw/`
 - **Entry point:** `~/.claude/mcp-servers/mcp_excalidraw/start.sh` (wrapper that starts canvas + MCP)
@@ -69,7 +69,7 @@ These MCPs are not written by install-mcp and require user action.
 
 ### pencil
 
-- **Needed by:** cogni-visual, cogni-website
+- **Needed by:** cogni-website, cogni-workspace
 - **Type:** Desktop app with bundled MCP server
 - **Install:** Download from https://pencil.dev, open the app — MCP auto-starts
 - **Probe tool:** `mcp__pencil__get_editor_state`
--- a/cogni-workspace/tests/test-mcp-declaration-hygiene.sh
+++ b/cogni-workspace/tests/test-mcp-declaration-hygiene.sh
@@ -57,6 +57,11 @@
 #     no exclusion list is needed to keep it out and none should be added.
 #   - the **Skills:** lines in mcp-registry.md carry a different relation
 #     (skill names, not plugin names) and have no registry counterpart.
+#   - the concept-mcp-server-map.md Plugins column is a fourth mirror of the
+#     same fact. It is correct today and is deliberately NOT compared here: D1
+#     asserts only that the two wiki copies stay byte-identical to each other,
+#     never that either agrees with the registry, so that agreement is an
+#     unguarded gap rather than something D1 already covers.
 #   - three further mirrors of the same fact are out of this suite's reach and
 #     tracked separately: the manage-workspace worked example, and the MCP
 #     tables in the repo-root README.md and CLAUDE.md. Guarding a repo-root
@@ -314,25 +319,31 @@ fi
 # parser found, not just the registry-backed ones, which is what keeps "the
 # parser saw nothing" distinguishable from "this server has no row" (the latter
 # is a MISSING defect on the arm, not a floor failure).
+#
+# The threshold is two, matching L1, because the registry carries two servers:
+# a surface that parses down to a single row has lost at least one of them. A1
+# would also catch that as a MISSING defect, but it would read as a missing row
+# in the document rather than a parser that stopped seeing rows, which is the
+# distinction these floors exist to draw.
 
-if [ "$probe_rows" -ge 1 ]; then
-  pass "L4 workspace-status probe table parsed at least one row"
+if [ "$probe_rows" -ge 2 ]; then
+  pass "L4 workspace-status probe table parsed at least two rows"
 else
-  fail "L4 workspace-status probe table parsed at least one row"
+  fail "L4 workspace-status probe table parsed at least two rows"
   printf '%s\n' "  $PROBE_TABLE_REL yielded $probe_rows rows — the Needed by column moved or the file was renamed"
 fi
 
-if [ "$relation_rows" -ge 1 ]; then
-  pass "L5 mcp-registry relation lines parsed at least one server"
+if [ "$relation_rows" -ge 2 ]; then
+  pass "L5 mcp-registry relation lines parsed at least two servers"
 else
-  fail "L5 mcp-registry relation lines parsed at least one server"
+  fail "L5 mcp-registry relation lines parsed at least two servers"
   printf '%s\n' "  $RELATION_DOC_REL yielded $relation_rows sections — the ### heading or the Needed by/Used by line shape moved"
 fi
 
-if [ "$example_rows" -ge 1 ]; then
-  pass "L6 install-mcp plan example parsed at least one row"
+if [ "$example_rows" -ge 2 ]; then
+  pass "L6 install-mcp plan example parsed at least two rows"
 else
-  fail "L6 install-mcp plan example parsed at least one row"
+  fail "L6 install-mcp plan example parsed at least two rows"
   printf '%s\n' "  $INSTALL_EXAMPLE_REL yielded $example_rows rows — the needed by: separator moved"
 fi
 
--- a/cogni-workspace/tests/test-mcp-declaration-hygiene.sh
+++ b/cogni-workspace/tests/test-mcp-declaration-hygiene.sh
@@ -31,6 +31,9 @@
 #   - both cogni-portfolio and cogni-workspace still carry the PreToolUse matcher
 #     mcp__excalidraw__.* — correct only while no plugin re-declares the server
 #   - the two copies of concept-mcp-server-map.md stay byte-identical
+#   - every hand-maintained mirror of a registry server's required_by names the
+#     same plugin set as the registry: the workspace-status probe table, that
+#     skill's mcp-registry.md relation line, and the install-mcp plan example
 #   - each glob-driven arm proved it had something to look at (liveness floor)
 #
 # The liveness floor is the load-bearing half of A1. "Zero .mcp.json files
@@ -38,8 +41,27 @@
 # without L1 proving the plugin glob matched real directories, A1 would pass
 # hardest exactly when the suite is most broken.
 #
+# A4-A6 widen the stated scope: this is no longer only a declaration-shape
+# guard, it also pins the prose copies of required_by to the registry. The
+# reason is the same one behind D1: a fact stored twice drifts one-sidedly and
+# nothing reports it. That is not hypothetical here. The cogni-visual
+# retirement updated the registry and left all three mirrors naming a retired
+# plugin, with every case in this suite still green.
+#
 # Deliberately NOT asserted: which servers the registry contains, or that any
-# server is installed. This is a declaration-shape guard, not an install check.
+# server is installed. This is still not an install check. Three more
+# exclusions are chosen rather than overlooked:
+#   - a server with no registry entry is never compared. claude-in-chrome is
+#     one (a manual Chrome-extension install), and it falls out because the
+#     arms iterate the REGISTRY and look the prose row up, not the reverse, so
+#     no exclusion list is needed to keep it out and none should be added.
+#   - the **Skills:** lines in mcp-registry.md carry a different relation
+#     (skill names, not plugin names) and have no registry counterpart.
+#   - three further mirrors of the same fact are out of this suite's reach and
+#     tracked separately: the manage-workspace worked example, and the MCP
+#     tables in the repo-root README.md and CLAUDE.md. Guarding a repo-root
+#     doc from a plugin's suite is a scope call a human should make, and the
+#     README spells plugins as markdown links, which needs a different parse.
 #
 # Case-label shape follows test-wiki-tree-parity.sh: "PASS: <id> <label>" /
 # "FAIL: <id> <label>", ids letter-prefixed and never bare numerals, and never a
@@ -68,6 +90,27 @@
 # comes from M1 below, which mutates the TREE (adds a .mcp.json) rather than the
 # assertion — the only mutation that can distinguish "nothing is declared" from
 # "nothing was looked at".
+#
+# Mutation recipe (verifies A4 is a real comparison, not a vacuous one):
+#   bash "$HOME/.claude/plugins/marketplaces/managed-service/cogni-service/scripts/mutation-check.sh" \
+#     --root . \
+#     --file cogni-workspace/references/mcp-git-registry.json \
+#     --expr 's/"required_by": \["cogni-portfolio", "cogni-workspace"\]/"required_by": ["cogni-portfolio", "cogni-workspace", "cogni-marketing"]/' \
+#     --test 'bash cogni-workspace/tests/test-mcp-declaration-hygiene.sh' \
+#     --case A4
+# Verdict: guard_verified. The search literal occurs once (pencil's required_by
+# is a different pair) and the replacement appends a plugin before the closing
+# bracket, so the mutant cannot re-satisfy the pattern and stay green. The same
+# mutant also reds A5 and A6, which is expected (one registry edit desyncs all
+# three mirrors at once) and is not a reason for a second recipe. L4-L6 stay
+# green throughout, which is what separates a real comparison from a parser
+# that stopped seeing rows.
+#
+# The case ids A4/A5/A6 and L4-L6 are written out as static literals at their
+# pass()/fail() call sites, never built in a loop: the handoff preflight
+# literal-searches this source for the recipe's --case value with comment lines
+# excluded, so an interpolated id replays as case_not_found while the arm
+# itself works perfectly.
 
 set -u
 
@@ -79,6 +122,9 @@ REPO_ROOT="${MCP_HYGIENE_ROOT:-$(cd "$SCRIPT_DIR/../.." && pwd)}"
 REGISTRY_REL="cogni-workspace/references/mcp-git-registry.json"
 WIKI_PAGE_REL="wiki/wiki/pages/concept-mcp-server-map.md"
 BUNDLED_PAGE_REL="cogni-workspace/wiki/wiki/pages/concept-mcp-server-map.md"
+PROBE_TABLE_REL="cogni-workspace/skills/workspace-status/SKILL.md"
+RELATION_DOC_REL="cogni-workspace/skills/workspace-status/references/mcp-registry.md"
+INSTALL_EXAMPLE_REL="cogni-workspace/skills/install-mcp/SKILL.md"
 MATCHER='mcp__excalidraw__.*'
 
 PASSED=0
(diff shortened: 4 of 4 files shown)


claude-cogni-work:
## Ship-ready

Author instance is handing this off. The PR stays draft with `svc:needs-review`; the independent review belongs to the merger.

**Handoff preflight: clean.** `7 discovered / 7 resolved / 5 gating applicable / 5 verdicts / 6 passed / 0 failed / 1 skipped`, re-run against the current head after the fix pass.

| Instrument | Verdict |
|---|---|
| `run-tests` | pass — suites green for cogni-workspace |
| `frontmatter` | pass — clean for cogni-workspace |
| `breadcrumbs` | pass — `standing_drift_excluded`; 9 offenders in `manage-themes/SKILL.md`, outside this branch's touched set, surfaced as advisories, not gating |
| `skill-length` | pass — measured 2 of 2 files, 0 over cap |
| `mutation-recipe` | pass — "recorded recipe is runnable as written — `A4` found in `cogni-workspace/tests/test-mcp-declaration-hygiene.sh`" |
| `agent-length` | **skipped** — `not_applicable`, no touched `agents/*.md` |

The one skip is named rather than folded into the total, so a green run whose instruments never examined anything stays visible.

## Author self-grade against the issue-time contract

Graded against the 23-item acceptance contract posted on #1475 at triage — derived `2026-08-18T20:38:49Z`, before this branch existed, so it is diff-blind by chronology. Advisory only; no verdict, no `svc:review-*` label, no `reviewed-head-sha` marker.

**First pass: 21 satisfied, 2 violated, 0 unclear.** Both violations were mechanical and landed in a file the plan already touched, so both were fixed in the single bounded pass (`34720c9f`):

1. **Liveness floors were `-ge 1` where the contract asked `>= 2` (or registry-derived).** Tightened to two, matching `L1`. This is not pedantry: the registry carries two servers, so a surface parsing down to one row has lost at least one of them. `A4`-`A6` would red on that anyway, but as a `MISSING` defect — which reads as a missing row in the document rather than a parser that stopped seeing rows, the exact distinction the floors exist to draw. A negative test for the degrade-to-one-row case was added and passes.
2. **The header's exclusions list omitted `concept-mcp-server-map.md`.** Now recorded as a fourth mirror, knowingly uncovered — with the reason `D1` does not already cover it: `D1` asserts only that the two wiki copies stay byte-identical to **each other**, never that either agrees with the registry.

**Residual for the merger: none.** No item was graded `unclear`, no violation resisted a mechanical fix, and the fix pass landed cleanly — the full 124-suite sweep and every repo guard were re-run green before it was pushed.

## What is worth a reviewer's attention

- **Scope is `Refs`, not `Closes`.** #1475's own stated scope ships here in full, but three further stale mirrors were found outside it and filed as #1511 / #1512 / #1513 rather than folded in. Whether a plugin-scoped suite should guard repo-root docs at all (#1512 / #1513) is a scope call worth a human.
- **Both touched `SKILL.md` files carry pre-existing quality findings** (annotated in the PR body under Review notes, none introduced by this diff, all category `preference`). They were annotated rather than auto-fixed because each would widen a guard PR into unrelated prose — one of them edits triggering-adjacent frontmatter.
- **Runner disclosure:** the token-scope preflight reports `over_scoped` (extra: `gist`, `read:org`, `workflow`). This runner is on a `gho_` CLI token whose scope set is fixed by GitHub's OAuth app and cannot be narrowed, so the documented `--allow-overscoped` opt-out was used; it exits 0 while leaving the status string unchanged.

claude-cogni-work:
## Correction: the three follow-ups were closed as not planned

The PR body and the handoff comment above cite #1511 / #1512 / #1513 as filed, open follow-ups. They are no longer open — the maintainer closed all three as `NOT_PLANNED` at 21:11, minutes after they were filed. That is a deliberate human scope decision, not an auto-close, and it is left standing.

Reading it into this PR: the three additional stale `required_by` mirrors are real and still present in the tree (`manage-workspace/SKILL.md:163-164`, root `README.md:163,165`, root `CLAUDE.md:169,171` — the last a superset keeping the retired plugin alongside the correct one), but they are **not** tracked work. The suite header records them as knowingly uncovered, which remains accurate and is now the only place that fact lives.

No change to this PR's diff or scope. Flagging it so the merger does not grade the scope section against three issue links that now resolve to closed.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1515
**Perspectives consulted:** plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer, plugin-dev:skill-reviewer, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:skill-quality-reviewer

**Summary:** All 6 dispatched perspectives returned pass, and all 23 review-contract items are satisfied by execution (mutation recipe replayed to guard_verified; four live negative paths exercised; suite 15 passed/0 failed and run-plugin-tests.py --filter cogni-workspace exits 0 on the PR head). The 8 remaining findings are all minor-severity preferences; under the gatekeeper merger-added-preference pre-step every one is a merger addition (none carry a review-contract:item-N location), so all 8 are advisory_addition and excluded from the gating rules — two are additionally waived as pre-existing drift. No defect-category or major/critical finding remains, so Rule 5b resolves the merge to pass.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (8)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/install-mcp/SKILL.md body §Verify Installation (probe table, lines 228-231)` — The probe table's column is headed `Server` and holds `excalidraw` / `pencil`, but every other table and code block in this file uses `Server` to mean the registry server name (`mcp_excalidraw`). The probe prefix is the registry's `desktop_config_key`, a different key — and the file warns that an unknown `--server` value exits non-zero, which is what a reader carrying `excalidraw` from this table into `--server` will hit.
  - *Suggested fix:* Rename the column to `Config key` and add one clause explaining that the probe prefix is the server's `desktop_config_key`, not its registry server name. (ref: skill-creator §Principle of Lack of Surprise / §Writing Style)

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/install-mcp/SKILL.md body §Verify Installation (probe table, lines 228-231)` — This file carries a second, unguarded registry mirror that neither the change nor the new suite arms cover: the probe table duplicates `desktop_config_key` and one entry from each server's `provides_tools[]`. Both values are correct against the registry today, so this is drift risk, not present drift.
  - *Justification:* Pre-existing whole-skill drift not introduced by this change (introduced_by_change: false); annotated for human awareness but must not block merge — mirrors service-resolve Step 6.5's pre-existing-drift policy.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/install-mcp/SKILL.md frontmatter.description (line 13) and body line 278` — Both sites pin a sibling skill's internal step number — 'its MCP installation step (step 5)'. That numeral is a prose mirror of manage-workspace's structure and goes stale silently the moment a step is inserted there, which is the same drift class this PR is correcting one field over. The semantic anchor already carries the meaning without the numeral.
  - *Suggested fix:* Drop the numeral at both sites, keeping the semantic anchor 'its MCP installation step'. (ref: skill-creator §Writing Style — avoid brittle specifics that drift)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/install-mcp/evals/evals.json:6` — Eval 1's prompt still opens 'I just installed cogni-visual and cogni-portfolio plugins...' — cogni-visual is the retired plugin whose removal is the reason this PR exists. Not a contract violation (item 2 scopes the cogni-visual grep to the three mirror files, and this fixture is a user utterance rather than a required_by mirror), and the eval still discriminates correctly because cogni-portfolio is named alongside and genuinely requires mcp_excalidraw. But it is the same staleness class the PR is closing, one file away. This file is not in the PR diff — pre-existing.
  - *Suggested fix:* Drop 'cogni-visual and ' from the prompt, or replace it with a live requiring plugin, so the skill's eval corpus names only plugins that exist in the current marketplace. Leaving it is defensible; if left, worth being a recorded decision rather than an oversight.

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md body §6 MCP Servers, 'Not loaded' block (lines 184-224)` — Roughly 40 lines of the body restate the two-step install/config diagnosis that references/mcp-registry.md already owns — the body itself says so at line 214. The 4-row state table plus surrounding advisory is the detailed troubleshooting layer, not the procedural core of a status check.
  - *Justification:* Pre-existing whole-skill drift not introduced by this change (introduced_by_change: false); annotated for human awareness but must not block merge — mirrors service-resolve Step 6.5's pre-existing-drift policy.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md body §6 MCP Servers, lines 170-172` — The pointer sentence claims 'only `excalidraw`'s key is spelled out inline below, so look `pencil`'s up there rather than assuming it' — but pencil's config key IS spelled out inline later in the same section (line 221). The instruction contradicts the file it appears in, so a reader following it makes an unnecessary registry read and may distrust the inline value it then finds.
  - *Suggested fix:* Replace the clause with the fact that actually motivates the read: the key is not derivable from the server name (registry server `mcp_excalidraw` is configured under `excalidraw`), so read it as data rather than stripping a prefix. (ref: skill-creator §Writing Style — give the reason, not a rule its own file falsifies)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md:226` — The closing instruction 'Only check MCPs for plugins that are actually installed' is now effectively unconditional: after this correction every row's Needed-by set contains cogni-workspace, which is by definition installed whenever this skill runs. Before the correction the excalidraw row named only cogni-visual, cogni-portfolio, so the filter could genuinely suppress a probe. The prose is a faithful mirror of the registry (correct per contract item 1 — the registry must not be bent to the prose), but a reader following the instruction literally will now never skip a server, and nothing in the text says so.
  - *Suggested fix:* Append one clause to the Only-check sentence, e.g. 'in practice every registry server lists cogni-workspace, so this filter narrows the report's wording rather than skipping a probe' — a prose-only addition outside the guarded table, so it cannot red A4/L4.

- **[minor]** (advisory-addition) `cogni-workspace/tests/test-mcp-declaration-hygiene.sh:404` — In the one degenerate state where the registry itself will not parse, A4/A5/A6 print PASS while having compared nothing: the six mirror values arrive empty and only the row counts are normalised to -1, so the *_defects variables stay empty and read as 'no defects'. Confirmed empirically (registry replaced with 'NOT JSON' yields PASS: A4/A5/A6 alongside FAIL: L2/L4/L5/L6, 10 passed 5 failed, rc=1). This does NOT violate contract item 13 — that item's bar is that the new case(s) call fail, and L4/L5/L6 do, so the suite is loudly red and the fail-closed design is explicitly reasoned in the header. Reporting nit only: no user-observable fault, CI fails.
  - *Suggested fix:* Optional, only if a future change makes this state reachable in practice: have the registry try/except emit a sentinel defect string on the six mirror lines so A4-A6 red alongside their floors rather than passing over an empty comparison. Do not change the floors themselves — they are correct and load-bearing as written.

### Positive observations

- Item 1 / top risk_watchlist entry cleared decisively: git diff of mcp-git-registry.json across the range is empty — neither required_by nor desktop_config_key moved, so the wrong-direction fix the watchlist warned about did not happen. (plugin-validator)
- Mutation recipe independently replayed to guard_verified (mutated_result=red, restored_result=green) against a git-archive extract of HEAD, so the read-only worktree was never touched. (plugin-validator)
- Registry-driven iteration proven, not asserted: `wanted` is built from the registry's server set, so claude-in-chrome — which the parsers DO see — is never compared. No exclusion list, allowlist, or skip token. (plugin-validator)
- The liveness floors are not decorative: dropping the pencil probe row leaves L4 green while A4 reds MISSING; deleting the file reds both. That is exactly the 'parser went blind' vs 'row missing from the document' distinction, and it holds under execution. (plugin-validator)
- Item 14 verified by rebuilding M1's mutant tree by hand: L4-L6 and A4-A6 all pass inside it with only A1 red, so the new arms cannot fail for a missing-file reason inside the negative case. (plugin-validator)
- Item 8's asymmetry implemented exactly: install-mcp arm keys by_name=True (server names), probe-table and relation arms by_name=False (desktop_config_keys). Verified by re-running the three parsers standalone. (plugin-validator)
- Item 15 checked mechanically: 15 unique first-token ids, all letter-prefixed, none carrying a colon, every id's pass()/fail() labels byte-identical — which matters because the mutation harness matches whole-token on those labels. (plugin-validator)
- Item 17 (locale-vacuity) honoured throughout: every new assertion is numeric (-ge 2) or emptiness-based; python3 stderr is discarded and no English fragment from any foreign tool is matched. (plugin-validator)
- Item 19's scope statement names four knowingly-uncovered surfaces and explains each; the manage-workspace claim was verified real (SKILL.md:163 still reads cogni-visual, cogni-portfolio), and it correctly notes D1 pins the two wiki copies to EACH OTHER, never to the registry. (plugin-validator)
- Bash 3.2.57 compatibility claimed and verified: same 15 passed, 0 failed, rc=0 under real /bin/bash. (plugin-validator)
- All 23 checklist items graded satisfied with executed evidence, including four live negative paths (missing mirror row, deleted mirror file, corrupted registry, hand-rebuilt M1 mutant child). (contract-grading-reviewer)
- item 18 satisfied: the header's mutation recipe replayed verbatim returned mutated_result=red, restored_result=green, verdict=guard_verified, and the identical recipe is reproduced verbatim in the PR body under '## Mutation recipe'. (contract-grading-reviewer)
- item 11 satisfied and verified live: a registry server whose prose row is absent yields a 'MISSING <key>' defect, an explicit fail branch rather than a skip or continue. (contract-grading-reviewer)
- item 13 satisfied and verified twice: deleting a mirror reds L4+A4 (13 passed, 2 failed, exit 1); corrupting the registry reds L2/L4/L5/L6/A2 (10 passed, 5 failed, exit 1) — no scenario produces a fully-green suite. (contract-grading-reviewer)
- item 22 satisfied: every changed path is under cogni-workspace/, no other plugin touched, and no version field in any plugin.json or marketplace.json modified. (contract-grading-reviewer)
- Contract item 5 satisfied exactly: both example lines match the registry's required_by sets token for token, and compare clean under the suite's comma-split/sort normalisation. (skill-reviewer (install-mcp))
- Contract item 8 satisfied structurally, not just textually: the worked example prints server names rather than config keys, which is precisely the asymmetry the A6 arm's by_name lookup depends on — rewriting it to config keys would have silently broken that arm while looking like a tidy-up. (skill-reviewer (install-mcp))
- The prose guard above the example survives the edit, so the example is now both correct and explicitly non-authoritative — correctness enforced mechanically while the reader is still pointed at the source of truth. (skill-reviewer (install-mcp))
- The file yields exactly the two 'needed by:' lines the L6 liveness floor requires, with no stray prose occurrence that would inflate the row count and mask a parser going blind. (skill-reviewer (install-mcp))
- Contract item 4's named trap avoided cleanly: mcp-registry.md's three '**Skills:**' lines are byte-identical to the base tree and still list skill names — only the two relation lines changed. (skill-reviewer (workspace-status))
- The claude-in-chrome entry keeps its '**Used by:**' spelling rather than being cosmetically normalised, and the suite accepts both spellings — so no guard asserts a spelling the file does not use (item 10 satisfied without churn). (skill-reviewer (workspace-status))
- The probe table's claude-in-chrome row is untouched and falls out of the registry-driven comparison by construction, so no exclusion token was needed anywhere in these files. (skill-reviewer (workspace-status))
- The diff is scoped to data only: no skill logic, report format, detection approach, frontmatter or allowed-tools changed. (skill-reviewer (workspace-status))
- Contract item 5 satisfied verbatim: lines 106-107 match required_by in mcp-git-registry.json exactly; cogni-visual appears nowhere in the file (item 2). (skill-quality-reviewer (install-mcp))
- Contract item 8's premise holds in this file: the worked example prints registry server names, not config keys, so an install-mcp guard arm keyed on the server name is the right shape here. (skill-quality-reviewer (install-mcp))
- Contract item 21 holds: frontmatter parses, chars_body 12907 is well inside the complexity-scaled cap of 21000 (metric_version 2), and the change introduces no issue ref, version tag or slice code. (skill-quality-reviewer (install-mcp))
- The example block is explicitly framed as illustrative, so the skill already routes the model to the source of truth — correcting the literal removes a misleading anchor without the model ever having depended on it. (skill-quality-reviewer (install-mcp))
- Contract item 3 satisfied: the probe table reads exactly cogni-portfolio, cogni-workspace (excalidraw) and cogni-website, cogni-workspace (pencil), matching required_by verbatim; the claude-in-chrome row is untouched. (skill-quality-reviewer (workspace-status))
- Contract item 4 satisfied: the relation lines are corrected and the '**Skills:**' lines were correctly NOT rewritten — all three still carry skill names, avoiding the destructive misreading the risk watchlist warns about. (skill-quality-reviewer (workspace-status))
- Contract item 10's tolerated state is preserved: mcp-registry.md keeps '**Used by:**' on the claude-in-chrome entry rather than normalising it, so no spelling the file does not use can be pinned by the new guard. (skill-quality-reviewer (workspace-status))
- Contract item 21 holds: valid frontmatter, chars_body 16325 inside cap 22500, no maintainer breadcrumb introduced. (skill-quality-reviewer (workspace-status))

### Recommendation

---

PR #1521, opened 2026-08-19
cogni-knowledge: document the result-line case-id convention

### LINKED ISSUE #1518: cogni-knowledge: the result-line case-id convention is undocumented, so new suites re-collide by default
## Plugin
cogni-knowledge

## Current behavior
`cogni-knowledge/tests/README.md` documents the plain `PASS:` / `FAIL:` emitter convention, but says nothing about the rule that makes a result line *addressable*: the first whitespace-delimited token of each emitted line is the case id `mutation-check.sh --case` matches, and it must be unique within its suite.

Because the rule is undocumented, the next suite author has nothing to follow. Re-collision is the normal outcome, not an unlucky one — which is how four suites accumulated 39 colliding cases before anyone measured.

## Desired behavior
The convention is written down where a suite author will meet it, stating at minimum:

- The first token of a result line is the case id; it must be unique within the suite.
- Extend an existing stem with a hyphenated discriminator; never re-use a bare stem.
- No trailing colon on an id — the matcher is right-anchored, so a colon becomes part of the token and a recipe author who omits it gets `case_not_found` with no hint why.
- Both arms of one case carry the same id; the discriminator lives in the id argument, never only in the message.
- Uniqueness is verified by **running** the suite and grouping emitted first tokens, never by grepping call sites.

## Motivation
The sibling work fixes the suites that had collided; this stops the next one from colliding. Without it every future suite is a fresh chance to reintroduce the defect, and the fix has to be re-derived by census each time.

### Constraints
- Documentation only — no suite behaviour changes.
- Do not duplicate the per-suite case-id scheme that #1495 is designing for the prefix-convention family; this records the *addressability* rule that applies to every suite regardless of scheme.

## Acceptance criteria
- [ ] `cogni-knowledge/tests/README.md` states the five rules above. Evidence: the rendered section. Falsified by any rule absent.
- [ ] The doc says uniqueness is checked by execution, not by a source grep. Falsified by silence on the method.
- [ ] No suite file is modified. Evidence: `git diff --name-only` lists only the README. Falsified by any `.sh` path, including any `version` value.

Refs #1494

claude-cogni-work:
Closes #1518.

## Summary

Adds a `## Case ids on result lines` section to `cogni-knowledge/tests/README.md`, between the existing `## Convention` block and `## Contract tests (SKILL.md)`. The `## Convention` block already documents what keeps a result line *parsable* (the plain, never-escape-wrapped `PASS:`/`FAIL:` label); nothing documented what makes one *addressable*, so a new suite re-collided by default.

The new section states five rules:

1. The case id is the first whitespace-delimited token **following** the label — not the first token of the line — and it must be unique within its suite.
2. Extend a stem with a hyphenated discriminator; never re-use a bare stem.
3. No trailing colon on an id, and nothing before the label: the id is right-anchored on whitespace-or-end-of-line, so a glued colon makes the token `init-fetch-cache:` and a recipe passing `--case init-fetch-cache` gets `case_not_found` with no hint why.
4. Both arms of one case carry the same id; the discriminator belongs in the id, never only in the message.
5. Uniqueness is verified by **running** the suite and grouping the emitted ids — never by grepping call sites, since an id is routinely built from a variable or emitted inside a loop.

It closes with the matcher's exact behaviour (whole-token match, red keyed on `FAIL:` and green on `ok:`/`PASS:`, a red line winning over a stray green, neither matching reported as `case_not_found`, and output *lines* read rather than the suite's exit status) and a warning that the same-named in-repo `cogni-portfolio` / `cogni-consult` scripts are different, bespoke tools — the harness this convention serves ships with the cogni-service managed-service tooling and is external to this repo.

## Evidence — the worked example is real, and was derived by execution

Every claim in the new section was checked by running the suite, not by reading it. Grouping the emitted ids of `cogni-knowledge/tests/test_binding_project_path.sh` yields three duplicated stems:

```
append-project
init
legacy
```

The `init` collision is three-way, and the example in the doc quotes those three lines verbatim.

Rule 4's consequence is sharper than "the case is unreachable", and is also verified from source: that suite pairs `green "PASS: init writes schema_version 0.1.5"` with `red "FAIL: schema_version expected 0.1.5, got ..."` — different ids on the two arms. So under `--case init` a genuine failure does **not** report `case_not_found`; the two *other* `PASS: init` lines still match the green pattern, and the run reads a real red as **green**. The check silently proves nothing. That is the failure mode the doc exists to prevent, so it is stated explicitly.

## Scope

- **Documentation only.** One file, 78 insertions, 0 deletions. `git diff --name-only` lists exactly `cogni-knowledge/tests/README.md`. No `.sh`, no `.py`, no `plugin.json` / `marketplace.json`, no version value — the post-merge workflow owns the bump.
- The existing plain-emitter bullet is left byte-identical and uncontradicted; the change is purely additive.
- **Not included, deliberately:** any per-suite case-id prefix scheme (a sibling issue owns that design), and any fix to the colliding ids in `test_binding_project_path.sh` — touching a suite here would mix a behaviour change into a documentation-only branch and falsify the issue's own third acceptance criterion.
- **No new breadcrumb** in the prose; provenance is stated semantically.

## Test plan

- [x] `git diff --name-only origin/main...HEAD` lists exactly one path, `cogni-knowledge/tests/README.md`.
- [x] The diff is purely additive — 0 deleted lines, so the plain-emitter bullet and every other line survive.
- [x] `python3 scripts/run-plugin-tests.py --filter knowledge` → **77/77 pass** on the branch (and on the base tree beforehand, so the baseline is established).
- [x] The documented `awk` recipe was copy-pasted verbatim and run in **both** directions: it prints `append-project` / `init` / `legacy` on the colliding suite, and prints nothing on a clean one (`test_plain_result_emitters.sh`). It discriminates rather than always passing.
- [x] Every example line in the new section obeys the rules it states — id as the first token after the label, no colon abutting the id, and the identical id on both arms of the correct example pair. The in-tree counter-examples are marked explicitly as the defect.

## Note on the counter-examples

The acceptance contract asks that example lines obey their own rules. The most instructive material here is the *broken* in-tree case, so the section shows a correct paired example first (`init-fetch-cache` on both arms) and labels each counter-example plainly as the defect it is illustrating. Flagging this because it is a judgement call a reviewer should confirm rather than discover.

### DIFF
--- a/cogni-knowledge/tests/README.md
+++ b/cogni-knowledge/tests/README.md
@@ -48,6 +48,84 @@ via `trap rm -rf "$WORK" EXIT`.
   script-level assertions.
 - Fixtures are minimal — only the files the test actually exercises.
 
+## Case ids on result lines
+
+The plain `PASS:`/`FAIL:` label above is what keeps a result line *parsable*.
+What makes one *addressable* is the token after it.
+
+The **case id is the first whitespace-delimited token following the label** —
+not the first token of the line. In `PASS: init bootstraps fetch-cache/
+directory` the id is `init`, and that is what `mutation-check.sh --case <id>`
+points at when it checks that a guard actually fires. That harness ships with
+the cogni-service managed-service tooling and lives outside this repo; the
+same-named `cogni-portfolio/scripts/mutation-check.sh` and
+`cogni-consult/scripts/mutation-check.sh` are different, bespoke scripts and are
+not this harness.
+
+Five rules follow from how it matches:
+
+1. **An id is unique within its suite.** Two cases sharing one id cannot be told
+   apart, so neither is addressable.
+
+2. **Extend a stem with a hyphenated discriminator; never re-use a bare stem.**
+   `test_binding_project_path.sh` shows the defect as it stands today — three
+   cases open with the bare stem `init`:
+
+   ```
+   PASS: init writes schema_version 0.1.5
+   PASS: init bootstraps fetch-cache/ directory
+   PASS: init writes curator_defaults; omits derivable/unused fields
+   ```
+
+   No `--case init` addresses any one of them. As `init-schema-version`,
+   `init-fetch-cache` and `init-curator-defaults` each becomes reachable while
+   the shared stem still groups them by eye. The same suite also collides
+   two-way on `append-project` and `legacy` — a class, not a one-off.
+
+3. **No trailing colon on an id, and nothing before the label.** The id is
+   right-anchored on whitespace-or-end-of-line, so a colon glued to it makes the
+   token `init-fetch-cache:`, and a recipe passing `--case init-fetch-cache`
+   gets `case_not_found` with no hint why. Only whitespace may precede the
+   label — the same reason the plain-emitter rule above exists.
+
+4. **Both arms of one case carry the same id.** The discriminator belongs in the
+   id, never only in the human-readable message:
+
+   ```sh
+   green "PASS: init-fetch-cache bootstraps the fetch-cache/ directory"
+   red   "FAIL: init-fetch-cache bootstraps the fetch-cache/ directory"
+   ```
+
+   Breaking this is worse than it looks. `test_binding_project_path.sh` pairs
+   `PASS: init writes schema_version 0.1.5` with `FAIL: schema_version expected
+   0.1.5, got ...` — different ids on the two arms. Under `--case init` a
+   genuine failure does not even report `case_not_found`: the two *other*
+   `PASS: init` lines still match the green pattern, so the run reads a real red
+   as **green** and the check silently proves nothing.
+
+5. **Verify uniqueness by running the suite — never by grepping call sites.** An
+   id is routinely built from a variable or emitted inside a loop, so the source
+   is not a census of what gets printed. Run it and group what came out:
+
+   ```sh
+   bash tests/test_binding_project_path.sh 2>&1 \
+     | awk '/^[ \t]*(PASS|ok|FAIL):[ \t]/ {print $2}' | sort | uniq -d
+   ```
+
+   Empty output is clean; anything printed is a collision. One limitation: a
+   passing run emits only the green arms, so once those are clean, pair each red
+   arm's id by reading the source.
+
+How the harness classifies a line, precisely: it matches the id as a **whole
+token** (so `--case P1` never matches a `P10` line), keys red on a `FAIL:` line
+and green on an `ok:` or `PASS:` line, and lets a red line win over a stray
+green. Matching neither is reported as `case_not_found` rather than guessed at.
+It reads output **lines**, never the suite's exit status — a suite exits
+non-zero when *any* case is red, which cannot say which one.
+
+These rules are about addressability alone; they hold whatever id vocabulary a
+suite adopts.
+
 ## Contract tests (SKILL.md)
 
 For pure LLM skills, regression coverage is **SKILL.md content invariants**


claude-cogni-work:
## Ship-ready

Handed off by the author instance for independent review. The PR stays **draft** and carries `svc:needs-review`; no review verdict, verdict label, or `reviewed-head-sha` marker has been written — the independent review is the merger's to run.

**Issue:** #1518 · **Branch:** `docs/issue-1518-tests-readme-case-id-convention` · **Head:** `cc245839`

### Handoff preflight

Exit 0, clean — 1 instrument passed (`version-bump`: no version-line touch), 6 skipped, 0 violations.

One skip is worth a reviewer's attention rather than a silent pass: **`run-tests` skipped as `no_suites_at_base`** — the instrument reported `no_suites_discovered` for cogni-knowledge. That is a glob mismatch, not an empty plugin: the instrument looks for `*/tests/test-*.sh` while this plugin's suites are named `test_*.sh`. So the preflight did **not** exercise the suites here.

Covered directly instead: `python3 scripts/run-plugin-tests.py --filter knowledge` was run on the base tree **before** any edit (77/77 pass) and again on the branch (77/77 pass), so the baseline is established and the diff regresses nothing. The remaining five skips are honest `not_applicable` results for a documentation-only diff (no SKILL.md, no `agents/*.md`, no `*/tests/test-*.sh` or `*/scripts/check-*.sh` path, so `mutation-recipe` does not apply).

### Author self-grade against the triage contract

Graded against the issue-time `review_contract` (15 items, `origin=triage`, derived before any branch existed): **14 satisfied, 0 violated, 1 unclear. 0 items fixed** — no fix pass ran, because nothing was graded `violated` and an `unclear` item is never auto-fixed.

**Residual the merger inherits — 1 item:**

- **Item 9 (self-consistency of example result lines) — INTENT-UNCLEAR, needs a human call.** The item says any example result line must obey the rules the section states, including carrying the identical id on both arms, and names no carve-out. The new section deliberately contains *both* a correct paired example (`init-fetch-cache` on the PASS and FAIL arms) *and* two counter-examples quoting real in-tree defects verbatim — the three-way bare-stem `init` collision, and an arm-asymmetric pair — each labelled as the defect it illustrates. Read literally the counter-examples fail the item; read purposively they are the teaching material. The diff cannot settle which reading triage intended.

  This was flagged in the PR body before grading, not discovered by it. If the merger reads item 9 as prescriptive-examples-only, nothing changes; if it reads it as covering every `PASS:`/`FAIL:`-formatted line, the counter-examples should become prose paraphrase or move into a marked "do not" callout. Either way the contract item is worth rewording so the scope is explicit for future grading.

### Verification the reviewer may want to reproduce

The worked example was derived by **execution**, not by reading source — which is the very method the new section prescribes:

```sh
bash cogni-knowledge/tests/test_binding_project_path.sh 2>&1 \
  | awk '/^[ \t]*(PASS|ok|FAIL):[ \t]/ {print $2}' | sort | uniq -d
```

prints `append-project`, `init`, `legacy`. The same recipe prints nothing against `test_plain_result_emitters.sh`, so it discriminates rather than always passing.

Diff shape: 1 file, **78 insertions, 0 deletions** — purely additive, so the existing plain-emitter bullet survives byte-identical. No `.sh`, no `.py`, no manifest, no version value.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1521

**Summary:** Both perspectives passed: contract-grading-reviewer confirmed all 15 committed contract items satisfied by execution against the live cogni-service harness, and plugin-validator independently verified the same technical claims, raising only two merger-added minor preference findings (both advisory additions, non-gating).

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/tests/README.md:80-82` — Half-formed sentence in rule 2's prescription: 'As `init-schema-version`, `init-fetch-cache` and `init-curator-defaults` each becomes reachable while the shared stem still groups them by eye.' The leading 'As' appears to mean 'Renamed as', but the clause reads as a fragment - the one sentence that prescribes the remedy is the least legible one.
  - *Suggested fix:* Make the verb explicit, e.g. 'Renamed to `init-schema-version`, `init-fetch-cache` and `init-curator-defaults`, each becomes reachable while the shared stem still groups them by eye.'

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/tests/README.md:99-104` — Rule 4's worked example is anchored to the live state of test_binding_project_path.sh without the hedge rule 2 carries. Rule 2 says the suite 'shows the defect as it stands today'; rule 4's block asserts flatly that the suite pairs those two lines - true today, but the sibling prefix-convention work is expected to reach this suite, at which point rule 4's example silently becomes false while rule 2's stays visibly hedged.
  - *Suggested fix:* Carry rule 2's hedge into rule 4 - e.g. 'as this suite stands today, it pairs ...' - or reframe rule 4's second paragraph as an illustrative pairing rather than a claim about a named live file.

### Positive observations

- Scope is exactly right: only cogni-knowledge/tests/README.md, +78/-0 - zero deletions, so the existing plain-emitter bullet survives byte-identical and item 13's non-regression concern is structurally impossible to have violated. (plugin-dev:plugin-validator)
- Matcher fidelity is exact rather than approximate - the README's four claimed properties were compared against cogni-service/scripts/mutation-check.sh and its regexes re-executed. (plugin-dev:plugin-validator)
- Rule 5 documents an executable procedure, not an aspiration: the README's awk pipeline run verbatim against the live suite printed exactly append-project, init, legacy. (plugin-dev:plugin-validator)
- Rule 4's counter-example was verified end to end including its non-obvious precondition. (plugin-dev:plugin-validator)
- Harness attribution is defensive rather than merely correct, pre-empting the mis-citation hazard for a second same-named script. (plugin-dev:plugin-validator)
- The sibling-scope fence is respected and stated explicitly in the closing line. (plugin-dev:plugin-validator)
- Placement and framing are genuinely instructive - the judgment call the contract flagged as gate-undecidable. (plugin-dev:plugin-validator)
- Repo guards clean on the touched tree: check-breadcrumbs.py, check-result-line-plainness.py, and the version-bump gate all report zero violations. (plugin-dev:plugin-validator)
- item 10 satisfied: the README's description matches the real harness regexes exactly, verified against actual harness source, not internal plausibility. (cogni-service:contract-grading-reviewer)
- item 14 satisfied: check-breadcrumbs.py's own patterns applied to the new section text yield zero matches. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1522, opened 2026-08-19
cogni-knowledge: test_helpers.sh has no fixed-string assert, so assert_grep false-passes on metacharacter literals


claude-cogni-work:
## Plugin
cogni-knowledge

## Current behavior
`cogni-knowledge/tests/fixtures/test_helpers.sh` offers only a BRE pair — `assert_grep` / `assert_not_grep`, both built on `grep -q --`. There is no fixed-string (`grep -qF`) equivalent.

That makes `assert_grep` silently pass on a literal containing regex metacharacters. Reproduced:

```
$ printf -- '- [[omega-synthesis]] — only this one\n' > decoy.md
$ grep -q -- "[[alpha-synthesis]]" decoy.md && echo MATCH
MATCH
$ grep -qF -- "[[alpha-synthesis]]" decoy.md && echo MATCH || echo "no match"
no match
```

`[[alpha-synthesis]]` parses under BRE as a bracket expression plus a literal `]`, so it matches any `<one-of-those-chars>]` sequence. An author asserting a wikilink, a glob, or any literal containing `[`, `]`, `.` or `*` gets a **green case against a file that does not contain the string** — the assertion is vacuous and nothing reports it.

## Why this matters now
The per-suite case-id scheme recorded in `cogni-knowledge/CLAUDE.md` points converting authors at `assert_grep` as the model for the one-label-feeds-both-arms shape. The rollout across the remaining suites is a large population of new assertions written against that model. Every author asserting a metacharacter-bearing literal will either false-pass silently or hand-roll a local `-F` wrapper — six suites in the plugin already reach for `grep -qF` by hand, and `test_portal_store.sh` had to define `assert_line` / `assert_no_line` for exactly this reason.

This is a vacuous-guard class defect: the failure mode is a test that reports green while asserting nothing.

## Desired behavior
`fixtures/test_helpers.sh` gains a fixed-string pair alongside the BRE pair — same `(pattern, file, description)` signature, same one-label-feeds-both-arms shape, same `errors` increment — so a converting author has a correct primitive per matching mode instead of reinventing one or silently false-passing.

### Constraints
- Additive only: do not change `assert_grep` / `assert_not_grep` semantics, which existing suites depend on.
- No version value moves in the branch.

## Acceptance criteria
- [ ] `fixtures/test_helpers.sh` exports a fixed-string assert pair using `grep -qF`. Evidence: the definitions. Falsified by either absent, or by one built on `grep -q`.
[...]

Issue #1484, opened 2026-08-18
repo: distinct test cases share a first token, so --case can name the wrong case


claude-cogni-work:
## Problem

The mutation harness addresses a case by the **first whitespace-delimited token** of its result-line label (`^[[:space:]]*FAIL:[[:space:]]+<case>([[:space:]]|$)`). Across most of the repo's per-case suites, distinct cases share that first token — so `--case <token>` cannot name one case.

This is a **different defect** from the colon-less shape tracked in #1482. That one produces `case_not_found`: an honest error. This one produces a **verdict for the wrong case**. A mutation intended to red case A can red sibling case B instead; the harness sees a matching `FAIL:` line, reports red, and the recipe records `guard_verified` for a guard that was never exercised. A silently wrong green is worse than a loud error.

## Roster

Censused against `origin/main` @ `8cb60e6f`. Counts are the size of the largest colliding family per suite.

| Suite | Worst collision | Notes |
|---|---|---|
| `cogni-consult/tests/test_step0_register_block.sh` | `anchor-once` 18-way | loop-generated over 9 SKILL.md files x 2 anchors |
| `cogni-consult/tests/test_resolve_assumptions.sh` | `link-mode` 7-way | plus `resolve-propagate-fb-idem` 5-way, `unknown-id` 4-way, ~10 further 3-way families |
| `cogni-consult/tests/test_orthography_drift_scan.sh` | `envelope-shape` 4-way | plus `failure-paths` 3-way and several 2-way |
| `cogni-consult/tests/test_subagent_start_hook.sh` | `bare` / `qualified` / `reject` 4-way each | loop-generated over agent names |
| `cogni-consult/tests/test_register_generator.sh` | `empty` 3-way | plus five 2-way families |
| `cogni-consult/tests/test_assumption_change_frequency.sh` | 2-way | `changed-literal`, `empty-corpus` |
| `cogni-consult/tests/test_dt_stage_advance.sh` | 2-way | shares the bare numeral `1` |

Verified clean, no collisions: `test-sanitize-theme.sh`, `test-check-skill-names.sh`, `test_deliverable_graph.sh`.

`cogni-trends/tests/test-stale-detection.sh` also had a total collision (all three cases opened with the shared token `case`) and **is fixed under #1482**, because a suite with zero addressable cases could not demonstrate the fix at all.

## Why this is filed separately

[...]

Issue #1498, opened 2026-08-18
cogni-knowledge: close the sweep - apply the case-id scheme to every remaining suite and prove plugin-wide uniqueness


claude-cogni-work:
## Plugin
cogni-knowledge

## Epic context
Part of the tracker "repo: distinct test cases share a first token, so --case can name the wrong case" (#1484), which stays OPEN when this child lands. Closing batch for cogni-knowledge: it converts every suite the earlier children did not reach and, uniquely among the children, asserts the property across the whole plugin rather than a named subset. It depends on the scheme-design child and on both earlier rollout batches, so "remaining" is a well-defined set at pickup time.

## Current behavior
After the contract family, the store/index/cycle-guard families and the mechanical-collision subset have landed, the balance of cogni-knowledge's 77 discovered suites still carries the shared-namespace-prefix labelling — roughly a third of the plugin, spread across the pipeline, wiki, migration, health and miscellaneous suites that belong to no large family. Each of them still resolves every result line to one addressable token, so the harness cannot name a case within them.

Until this batch lands, the plugin-wide claim the tracker is after — that every case in every cogni-knowledge suite is addressable — cannot be made, because no earlier child's scope covers the whole plugin.

## Desired behavior
Every remaining cogni-knowledge suite emits result lines carrying a unique-within-suite case id as the first token, following the scheme recorded in `cogni-knowledge/CLAUDE.md`, and the property is demonstrated across all 77 suites at once.

Strings that change, per the scheme's shape:

- `PASS: knowledge-<area>: <prose>` → `PASS: <suite-slug>-NN-<discriminator>: <prose>`, e.g. `PASS: knowledge-wiki-coverage: bilingual pages are counted once` → `PASS: kwikicov-07-bilingual-counted-once: bilingual pages are counted once`.
- The paired `FAIL:` arm changes identically.

## Motivation
A partial conversion is worth less than it looks: a reviewer reading `guard_verified` on a cogni-knowledge PR still has to check whether that particular suite was converted. The plugin-wide assertion is what makes the evidence trustworthy by default, and it can only be written once by the child that closes the set.

### Constraints
- Follow the scheme recorded in `cogni-knowledge/CLAUDE.md`; do not vary it.
- Uniqueness is verified **by execution** across every discovered cogni-knowledge suite, never by grepping call sites.
- **Regex metacharacters are out of scope — do not implement anything for them.** The harness `re.escape`s the id before interpolation; the tracker's original premise is false.
[...]

Issue #1496, opened 2026-08-18
cogni-knowledge: apply the case-id scheme to the 20 test_*_contract.sh suites and test_skill_contracts.sh


claude-cogni-work:
## Plugin
cogni-knowledge

## Epic context
Part of the tracker "repo: distinct test cases share a first token, so --case can name the wrong case" (#1484), which stays OPEN when this child lands. First of three rollout batches for the prefix-convention suites; it depends on the scheme-design child, which must merge first so this PR has a written rule and a worked example to follow. Batched by filename family so the scope is enumerable and each batch is independently verifiable.

## Current behavior
The contract family — 20 suites matching `cogni-knowledge/tests/test_*_contract.sh` (charter, compose, contradictor, dashboard, distill, finalize, health, ingest, ingest_source, language_config, lint, prefill, prelim_search, prose_density, refresh_resweep, reviewer, setup_seed, source_contradictor, update, verify) plus `test_skill_contracts.sh` — labels result lines with a shared namespace prefix rather than a case id, e.g. `PASS: knowledge-dashboard: <prose>`. The addressable token is the same for every line in a suite, so `--case` resolves to whichever assertion runs first and any recorded `guard_verified` is unfalsifiable.

These suites guard the skill contracts — the assertions that keep each cogni-knowledge skill's documented behaviour in step with its scripts — so an unexercised guard here is silently expensive.

## Desired behavior
Every suite in this batch emits result lines carrying a unique-within-suite case id as the first token, following the scheme recorded in `cogni-knowledge/CLAUDE.md` by the design child.

Strings that change, per the scheme's shape:

- `PASS: knowledge-<skill>: <prose>` → `PASS: <suite-slug>-NN-<discriminator>: <prose>`, e.g. `PASS: knowledge-compose: rejects an unbound project` → `PASS: kcompose-04-rejects-unbound-project: rejects an unbound project`.
- The paired fail arm changes identically, so `FAIL: knowledge-compose: rejects an unbound project` → `FAIL: kcompose-04-rejects-unbound-project: rejects an unbound project`.
- Where a suite's namespace prefix carries information a reader wants, it survives in the suite slug (`knowledge-compose` → `kcompose`) or in the prose, never as the sole first token.

## Motivation
[...]

PR #1523, opened 2026-08-19
cogni-knowledge: record the per-suite case-id scheme and prove it on two suites

### LINKED ISSUE #1495: cogni-knowledge: design and record the per-suite case-id scheme, and prove it on the two hardest suites
## Plugin
cogni-knowledge

## Epic context
Part of the tracker "repo: distinct test cases share a first token, so --case can name the wrong case" (#1484), which stays OPEN when this child lands. This is the design child: roughly 61 cogni-knowledge suites label their result lines with a shared namespace prefix and have no case-id scheme at all, so renaming them is not a mechanical edit — a scheme has to exist first. The three rollout children depend on this one; nothing else does.

## Current behavior
The dominant cogni-knowledge convention labels a result line with a namespace prefix rather than a case id. Verified at `cogni-knowledge/tests/test_dashboard_contract.sh:35`: `green "PASS: knowledge-dashboard: names citation-consistent verification semantics …"`. The addressable token is `knowledge-dashboard:` — shared by every line in that suite, so the entire suite resolves to whichever assertion runs first. The census found 18 suites with exactly one addressable token for the whole suite.

Two suites are worse still: they emit **no** matching result line at all. `cogni-knowledge/tests/test_portal_store.sh` is the in-scope one — `fail()` at `:36` prints `FAIL: <prose>` and then `exit 1`, so there is at most one red line per run and it is not per-case; the only success signal is a single `green "ALL TESTS PASS"` at `:158`. There is no per-case green line for the harness to match on the baseline run, so `--case` can never resolve. (The other no-result-line suite, `cogni-workspace/tests/test-theme-backcompat.sh`, is handled by its own sibling child.)

Without a written scheme, each rollout PR would invent its own, and the ~61-suite conversion would land inconsistent — which is the state the tracker is trying to leave.

## Desired behavior
A written case-id scheme exists in `cogni-knowledge/CLAUDE.md`, and two suites are converted as its proof.

The scheme records, at minimum:

- **Shape.** `<suite-slug>-<NN>[-<discriminator>]`, lowercase, hyphen-joined, no spaces, never a bare numeral, and never solely a namespace prefix.
- **Placement.** The id is the *first* whitespace-delimited token of the label; human prose follows it. The namespace prefix, where it carries meaning, moves after the id or is folded into the suite slug.
- **Both arms.** The pass and fail arms of one case carry the same id, and the discriminator lives in the id argument, never only in the message argument.
- **Loops.** A loop-generated family slugifies the loop variable into the id, from a value that actually differs between iterations.
- **Per line, not per case.** Uniqueness is a property of each emitted result line; a suite whose narrative reuses a case number still needs distinct ids per line.
- **Verification.** Uniqueness is checked by running the suite and grouping emitted first tokens, never by grepping call sites.

Strings that change as the proof:

- `cogni-knowledge/tests/test_dashboard_contract.sh:35` — `green "PASS: knowledge-dashboard: names citation-consistent verification semantics …"` → `green "PASS: kdash-01-verification-semantics: names citation-consistent verification semantics …"`; every other line in that suite gains its own `kdash-NN-…` id.
- `cogni-knowledge/tests/test_portal_store.sh` — `fail "1: refreshed engine span not present"` (`:81`) → a paired emitter that prints `PASS: portal-01-refreshed-span …` on success and `FAIL: portal-01-refreshed-span …` on failure, so each case emits exactly one result line on every run. The current fail-fast `exit 1` inside `fail()` (`:36`) is replaced by a failure counter with a non-zero exit at the end, so a later case's line still gets emitted when an earlier one reds.

## Motivation
The scheme is the reusable artifact; the two conversions are what prove it survives contact with the hardest shapes in the plugin — a total prefix collision and a suite that emits nothing addressable. Landing them together means the three rollout children have both a rule to follow and a worked example to copy.

### Constraints
- The scheme is recorded in `cogni-knowledge/CLAUDE.md`, not in a SKILL.md or agent file — those are covered by `scripts/check-breadcrumbs.py` and are not the home for test conventions.
- **Regex metacharacters are out of scope and must not appear in the scheme as a constraint.** The harness applies `re.escape(case_id)` before interpolation, so a punctuated or bracketed id is matched literally and is already addressable. The tracker's original body asserted the opposite; recording that false premise as a rule would be worse than silence.
- Uniqueness is verified **by execution**. The scheme must say so explicitly, because a source grep cannot see runtime-interpolated labels or lines emitted from Python inside heredocs.
- Changing `test_portal_store.sh` from fail-fast to counted failures must not weaken it: every case that previously would have failed must still produce a red line and a non-zero exit.
- Result lines stay plain — no escape sequence before the label — per `scripts/check-result-line-plainness.py`.
- No behavioural assertion changes in either converted suite; only the emitter shape and the labels.
- Every suite stays green: `python3 scripts/run-plugin-tests.py`.

## Acceptance criteria
- [ ] `cogni-knowledge/CLAUDE.md` gains a section stating the id shape, first-token placement, the same-id-on-both-arms rule, the loop-variable rule, the per-line (not per-case) uniqueness rule, and that uniqueness is verified by execution. Evidence: the section text names each of those six rules. Falsified by a missing rule, or by the section stating a rule the two converted suites do not follow.
- [ ] The recorded scheme does not name regex metacharacters as a constraint on ids. Evidence: the section contains no rule restricting punctuation in an id on escaping grounds. Falsified by any such rule, which would carry the tracker's disproved premise forward.
- [ ] `cogni-knowledge/tests/test_dashboard_contract.sh` emits result lines whose first tokens are unique within the suite and follow the recorded shape. Evidence: run it, group emitted first tokens, show maximum group size 1 and every token matching the documented shape. Falsified by a repeated token, or a token that the recorded shape would reject.
- [ ] `cogni-knowledge/tests/test_portal_store.sh` emits exactly one result line per case on a clean run, and continues past a failing case. Evidence: run it clean and show one `PASS:`-prefixed line per case with unique first tokens; then break the first case in a scratch copy and show a `FAIL:` line for that case *and* result lines for the later cases, with a non-zero exit. Falsified by a run emitting no per-case line, by a duplicated token, or by the suite still exiting at the first failure — which would leave later cases unaddressable.
- [ ] The converted suites lost no failure detection. Evidence: for each assertion in `test_portal_store.sh`, break the condition it guards in a scratch copy and show its case reds. Falsified by any assertion whose break no longer produces a red line or a non-zero exit.
- [ ] Mutation arm: revert one property guarded by `test_dashboard_contract.sh` and run the harness with the corresponding new `kdash-…` id; then do the same for one `portal-…` id. Expect exactly the addressed case red in each run. Evidence: harness output shows `FAIL: <addressed-id>` and no sibling id on a `FAIL:` line. Falsified by a sibling reding, or by `case_not_found`.
- [ ] The change touches only `cogni-knowledge/CLAUDE.md` and the two named suites. Evidence: `git diff --name-only <base>...HEAD` lists exactly those three paths. Falsified by any additional changed path, including any `version` value or any other suite converted opportunistically.

_Affected components: cogni-knowledge/CLAUDE.md, cogni-knowledge/tests/test_dashboard_contract.sh, cogni-knowledge/tests/test_portal_store.sh_

_Complexity hint: medium_

## Blocks

- #1496 cogni-knowledge: apply the case-id scheme to the 20 test_*_contract.sh suites and test_skill_contracts.sh
- #1497 cogni-knowledge: apply the case-id scheme to the store, index and cycle-guard suite families
- #1498 cogni-knowledge: close the sweep - apply the case-id scheme to every remaining suite and prove plugin-wide uniqueness
- #1499 repo: settle whether a CI guard should enforce per-suite case-id uniqueness, and record the outcome

claude-cogni-work:
## What changed

Records the per-suite case-id scheme in `cogni-knowledge/CLAUDE.md` and conforms the two hardest suites to it.

The first whitespace-delimited token of a result line is the case id `mutation-check.sh --case` matches. Nothing stated that, so `test_dashboard_contract.sh` labelled all 15 result lines with the same `knowledge-dashboard:` token (one address for the whole suite) and `test_portal_store.sh` emitted no harness-addressable line at all.

Three files, exactly as the issue pins them:

| File | Change |
|---|---|
| `cogni-knowledge/CLAUDE.md` | New `### Test result-line case ids` subsection under `## Conventions`, recording six rules |
| `tests/test_dashboard_contract.sh` | Label-only relabel to `kdash-NN-slug`; every assertion, pattern and file argument byte-identical |
| `tests/test_portal_store.sh` | Fail-fast → counted failures, one id-bearing line per case |

No suite behaviour changes beyond the fail-fast→counted conversion, and no version value moves.

## Evidence

Both suites emit **15 result lines with 15 unique first tokens**, verified by running them and grouping emitted tokens (never by grepping call sites — the rule the scheme itself states):

```
test_dashboard_contract: 15 lines / 15 unique ids
test_portal_store:       15 lines / 15 unique ids   (was 0 addressable)
```

Full repo sweep on the merged tree: **124/124 suites pass**. `check-result-line-plainness.py` and `check-breadcrumbs.py` both exit 0.

### No failure detection was lost

Each of the 11 assertion groups in `test_portal_store.sh` was broken in isolation; every one reds its own case, with `rc=1` and all 15 lines still emitted. Only the addressed case reds — no sibling fires.

The load-bearing property of the conversion is that `fail()` returns 0. Demonstrated rather than asserted:

| `fail()` returns | lines emitted | cases red |
|---|---|---|
| `0` (as shipped) | 15 | both broken cases |
| `1` (the trap) | **1** | first only — fail-fast silently re-arms |

## Mutation recipe

Both arms, each `--case` a literal first token present in the suite source (never a runtime-interpolated id — the preflight instrument literal-searches the source and cannot see an interpolated one).

**`kdash-` arm** — mutate the SKILL.md the suite asserts against:

```
bash cogni-service/scripts/mutation-check.sh --root <repo-root> --file cogni-knowledge/skills/knowledge-dashboard/SKILL.md --expr 's/\*\*zero-network\*\*/ZZZ removed/gm' --test 'bash cogni-knowledge/tests/test_dashboard_contract.sh' --case kdash-04-zero-network
```

**`portal-` arm** — mutate the vendored engine under test:

```
bash cogni-service/scripts/mutation-check.sh --root <repo-root> --file cogni-knowledge/scripts/vendor/cogni-wiki/skills/wiki-ingest/scripts/wiki_index_update.py --expr 's/refreshed:\{date\} bullets:\{n\}/refreshed:BROKEN bullets:\{n\}/m' --test 'bash cogni-knowledge/tests/test_portal_store.sh' --case portal-store-03-stamp-refreshed
```

Both return `guard_verified` (`mutated_result: red`, `restored_result: green`), with exactly the addressed case red and no sibling on a `FAIL:` line.

Run the `kdash-` command against the **base** version of that suite and it returns `case_not_found` — on the old labels every line led with `knowledge-dashboard:`, so no individual case could be addressed at all. That difference is exactly what this PR buys.

## Review notes

The pre-commit quality pass produced four changes. Three were kept; one was **reverted after self-grading against the issue-time contract**, which is recorded here rather than silently dropped:

1. **The suite violated the rule it was converted to satisfy.** Three sites (`portal-store-06`, `-10`, `-11`) wrote the id-plus-prose label out twice, once per arm — the drift the scheme's own *Both arms* rule forbids. Added `assert_rc`, a status-taking sibling of `assert_line`, so all three route one label. This was a live falsifier of the first acceptance criterion. **Kept.**
2. **`fail()` dumped the whole index on every failure.** Harmless when the suite exited at the first red; now that every case reports, a broadly-broken run repeated one fixture per red case and buried the `FAIL:` id lines this scheme exists to expose. The dump moved to the summary branch, emitted once. **Kept.**
3. **Stale rationale.** The `fail()` comment justified `MUST return 0` against `cmd || fail …` call sites the conversion removed. Restated against the current shape (helper return status propagating to a bare call under `set -e`). This file is the template the remaining rollout copies. **Kept.**
4. **Folding the hand-rolled `grep -qE` block into `assert_grep` — reverted.** The quality pass rewrote the `kdash-03` block as a one-line `assert_grep 'citation.consistent'`. The matchers are equivalent (the ERE alternation's first branch already used `.`, which spans both spellings), but the contract pins this suite's conversion to label text only, with every pattern and file argument byte-identical — and a behaviour-equivalent rewrite is still a non-label change. Reverted in `1eb0df46`. With label strings blanked, the suite is now byte-identical to base.

## Follow-up issues

- #1522 — `fixtures/test_helpers.sh` has no fixed-string assert, so `assert_grep` silently **false-passes** on a metacharacter-bearing literal: asserting `[[alpha-synthesis]]` matches a file containing only `[[omega-synthesis]]`. Filed rather than fixed here because the acceptance criteria pin this change to three paths. It matters for the rollout, which points authors at `assert_grep` as the model shape.

## Observation for the reviewer (not actioned)

The scheme is recorded in `cogni-knowledge/CLAUDE.md`, but the collision problem is repo-wide — 15 of 26 result-emitting suites outside this plugin collide, including 8 of the 10 repo-root `tests/*.sh`. Whether the addressability rule belongs at repo root beside the plainness bullet is a placement question this PR deliberately does not settle: the acceptance criteria forbid touching a fourth path, and #1484 already owns the repo-wide population.

Closes #1495


claude-cogni-work:
## Ship-ready

Author instance is done. Handing off to the merger for independent review — this PR carries **no** review verdict, no `svc:review-*` label and no `reviewed-head-sha` marker, so the merger's own review is still owed in full.

**Head:** `1eb0df46` · **Base:** `main` · draft, `svc:needs-review`

### Handoff preflight

`check-preflight.sh` → **RC=0**, clean — 1 passed, 6 skipped, 0 advisory.

Two of those skips are worth the merger's attention, because they mean the green is partly vacuous on this diff:

| Instrument | Skip reason | What it means here |
|---|---|---|
| `mutation-recipe` | no touched path matching `*/tests/test-*.sh` | The suites are `test_dashboard_contract.sh` / `test_portal_store.sh` — **underscore**, not hyphen. The glob never saw them, so the recipes in the PR body were **not** machine-checked by the preflight. |
| `run-tests` | `no_suites_at_base` for cogni-knowledge | Same hyphen-glob blindness — all 77 cogni-knowledge suites are `test_*.sh`, so none was discovered or run. |

Both were covered by hand instead: full unfiltered sweep **124/124 green**, and both mutation arms executed to `guard_verified`. Flagging it so the RC=0 is not read as evidence it is not.

Contract read with `--target issue --origin triage` (derived 2026-08-18T22:34:22Z, before any branch was cut). **22 items graded, 22 satisfied, 0 violated, 0 unclear.**

**Residual: none.**

One item was graded `violated` mid-run and fixed inside the bounded fix pass:

- **item 13** (dashboard conversion must be label-only, patterns byte-identical). The pre-commit quality pass had folded the hand-rolled `grep -qE 'citation.consistent|citation_consistent'` block into `assert_grep`, which also narrowed the matcher to a BRE. Behaviour-equivalent, but not label-only. Reverted in `1eb0df46`; with label strings blanked the suite is now byte-identical to base. Tests re-run green before the push.

### Evidence the merger may want to re-derive

- Both suites: **15 result lines, 15 unique first tokens** (`test_portal_store.sh` was **0** addressable at base).
- Every one of the 11 assertion groups in `test_portal_store.sh` reds its own case in isolation, `rc=1`, all 15 lines still emitted; only the addressed case reds.
- The conversion's load-bearing property, demonstrated rather than asserted: with `fail()` returning `0` a two-break run emits 15 lines and 2 reds; flipped to return `1` it emits **1** line and dies at the first failure.
- Mutation arms: `kdash-04-zero-network` and `portal-store-03-stamp-refreshed` both `guard_verified`. The `kdash-` command run against the **base** suite returns `case_not_found` — the discrimination this PR buys.

### Filed follow-up

- #1522 — `fixtures/test_helpers.sh` has no fixed-string assert, so `assert_grep` **false-passes** on a metacharacter literal (`[[alpha-synthesis]]` matches a file containing only `[[omega-synthesis]]`). Out of scope here (a fourth path would breach item 1), but it matters for the #1496–#1498 rollout, which the new CLAUDE.md section points at `assert_grep` as the model shape.

### Open question left for the merger

The scheme is recorded plugin-locally, but 15 of 26 result-emitting suites outside cogni-knowledge collide, including 8 of the 10 repo-root `tests/*.sh`. Whether the addressability rule belongs at repo root beside the plainness bullet is deliberately unsettled here — item 1 forbids a fourth path, and #1484 already owns the repo-wide population.

sdh07:

_The merger formed this expectation from issue #1495 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix adds one new section to cogni-knowledge/CLAUDE.md (natural home: alongside the existing §Conventions bullet list, which today says nothing about test result-line labels) recording the six-rule case-id scheme verbatim in intent — shape, first-token placement, same-id-on-both-arms, loop-variable slugification, per-emitted-line (not per-case) uniqueness, and verification-by-execution — with no regex-metacharacter restriction. It then conforms two suites without touching the shared emitters: in cogni-knowledge/tests/test_dashboard_contract.sh the 13 assert_grep/assert_not_grep description arguments (base lines 28,30,40,43,44,45,48,51,53,56,59,60,65,66) plus the inline green/red pair at lines 34-39 each gain a distinct `kdash-NN-…` id as the first token of the label, since cogni-knowledge/tests/fixtures/test_helpers.sh:15-44 prints `PASS: $description` and must stay byte-unchanged (it is sourced by ~61 suites). In cogni-knowledge/tests/test_portal_store.sh the `fail()` emitter at line 36 loses its `exit 1` in favour of a failure counter plus a non-zero exit after line 158, and each of the ~15 fail-capable assertions (lines 80-88 including the 5-iteration bullet loop at 86-88, 94-101, 108, the `|| exit 1` heredoc at 110-124, and the upsert round-trip at 136-155) becomes a paired emitter printing exactly one `PASS:`/`FAIL:` line carrying a `portal-…` id on every run. No grep pattern, no assertion target, and no behavioural condition changes; the diff stays inside exactly those three paths, so uniqueness is proven by running the suites rather than by adding any guard suite or fixture.

**Acceptance checklist:**
- [ ] Scope boundary: `git diff --name-only <base>...HEAD` yields exactly three paths, every one of them inside the in-set {cogni-knowledge/CLAUDE.md, cogni-knowledge/tests/test_dashboard_contract.sh, cogni-knowledge/tests/test_portal_store.sh}; every path outside that set is out of bounds — falsified by a fourth changed path of any kind, including a third suite under cogni-knowledge/tests/, any file under cogni-knowledge/tests/fixtures/, any newly added test/fixture/helper file, any docs or reference file, and any `version` value in any `.claude-plugin/plugin.json` or in the root `.claude-plugin/marketplace.json`. No item in this contract mandates an artifact outside that boundary: the mutation recipes are recorded in the PR body and the break-tests run against scratch copies, so nothing here requires a committed file.
- [ ] cogni-knowledge/tests/fixtures/test_helpers.sh is byte-unchanged and absent from `git diff --name-only` — the shared `red`/`green`/`assert_grep`/`assert_not_grep` emitters (base lines 15-44) are sourced by ~61 cogni-knowledge suites, so ids must be moved into the *description* arguments at the call sites, never into the helper. Falsified by any edit to that file.
- [ ] cogni-knowledge/CLAUDE.md gains one new section under an explicit `##`/`###` heading naming the per-suite test case-id scheme; the scheme is recorded there and in no SKILL.md or agents/*.md file (which item 1 also forbids). Evidence: the new heading and its body appear in that file's diff.
- [ ] The new section states BOTH the id shape — `<suite-slug>-<NN>[-<discriminator>]`, lowercase, hyphen-joined, no spaces, never a bare numeral, never solely a namespace prefix — AND the placement rule: the id is the first whitespace-delimited token of the label, human prose follows it, and a meaningful namespace prefix moves after the id or is folded into the suite slug. Falsified by either rule missing from the section text.
- [ ] The new section states BOTH the both-arms rule — the pass and fail arms of one case carry the same id, and the discriminator lives in the id argument, never only in the message argument — AND the loop rule: a loop-generated family slugifies the loop variable into the id, from a value that actually differs between iterations. Falsified by either rule missing.
- [ ] The new section states BOTH the per-line uniqueness rule — uniqueness is a property of each emitted result line, so a suite whose narrative reuses a case number still needs distinct ids per line — AND the verification rule: uniqueness is checked by running the suite and grouping emitted first tokens, never by grepping call sites. Falsified by either rule missing, or by a verification rule phrased as a source grep.
- [ ] Negative: the new section contains no rule restricting an id's punctuation or bracketing on regex-escaping / harness-matching grounds, and no claim that a punctuated id is unaddressable — the tracker's disproved premise must not be carried forward. Evidence: grep the new section for `regex`, `metachar`, `escap`, `bracket` and find no such constraint. The shape rule's own lowercase/hyphen/no-space constraints are style constraints, not escaping-grounds constraints, and do not falsify this item.
- [ ] Scheme-to-suite consistency: every rule the new section states is actually followed by both converted suites — the documented shape matches every first token both suites emit, and any suite-slug convention the section names (e.g. `kdash`, `portal`) is the one the suites actually use. Falsified by the section stating a rule the two converted suites do not follow.
- [ ] Structural, test_dashboard_contract.sh: every result-line label emitted on a clean run begins with a `kdash-NN-…` id — the 13 `assert_grep`/`assert_not_grep` description arguments at base lines 28,30,40,43,44,45,48,51,53,56,59,60,65,66 plus the inline `green "PASS: …"` / `red "FAIL: …"` pair at base lines 35 and 37 — and no label still begins with the bare shared `knowledge-dashboard:` namespace prefix. Evidence: read the diff and count 15 id-carrying cases; grep the file for a remaining label whose first token is `knowledge-dashboard:`.
- [ ] Uniqueness by execution, test_dashboard_contract.sh: run `bash cogni-knowledge/tests/test_dashboard_contract.sh`, take each `PASS:`/`FAIL:`-prefixed line's first whitespace-delimited token after the prefix (trailing `:` separator stripped), and group them — maximum group size 1, at least 15 result lines, and every token matches the shape the new CLAUDE.md section documents. Falsified by a repeated token or a token the recorded shape would reject.
- [ ] The two near-identical labels in test_dashboard_contract.sh — base line 35/37 (`knowledge-dashboard: names citation-consistent verification semantics (#337)`) and base line 56 (`knowledge-dashboard: names citation-consistent verification semantics`) — receive two DISTINCT ids. This is the per-line-uniqueness rule's hardest case in this suite; falsified by both lines reusing one id.
- [ ] Both arms, test_dashboard_contract.sh: the inline if/else at base lines 34-39 emits `PASS: <id> …` and `FAIL: <id> …` with the SAME id, and the discriminator sits in the id itself rather than only in the trailing prose. Evidence: read the two emitter lines in the diff.
- [ ] No behavioural assertion change, test_dashboard_contract.sh: the grep pattern (first arg) and target file (second arg) of all 13 `assert_grep`/`assert_not_grep` calls, the `grep -qE 'citation.consistent|citation_consistent'` condition at base line 34, the `errors` accumulation, and the final `ALL PASS` / exit-1 block at base lines 68-75 are semantically unchanged — only description strings and emitter labels moved. Falsified by a changed or loosened pattern, a changed target file, or a dropped/added assertion.
- [ ] Fail-fast removed, test_portal_store.sh: `fail()` (base line 36) no longer calls `exit 1`; failures increment a counter and the suite exits non-zero from a new end-of-file block (base line 158 is today a bare `green "ALL TESTS PASS"` with an implicit exit 0). Evidence: read the emitter definition and the new tail in the diff.
- [ ] One result line per case, test_portal_store.sh: every previously-`fail`-capable assertion — base lines 80-81, 82-83, 84, 85, the 5-iteration bullet loop at 86-88, 94-99, 100, 101, 108, the heredoc at 110-124, and the upsert round-trip at 136-155 — emits exactly one `PASS:`- or `FAIL:`-prefixed line on every run. Evidence: a clean run emits at least 12 result lines with pairwise-distinct first tokens (base carries 15 fail-capable assertions; the four narrative `green "N: …"` summary lines are not `PASS:`/`FAIL:` result lines and do not count toward this). Falsified by a run emitting no per-case line or by a duplicated token.
- [ ] Loop family, test_portal_store.sh: the bullet loop at base lines 86-88 (`for b in alpha-synthesis omega-synthesis q-one src-a src-b`) emits five result lines whose ids slugify `$b` — five distinct tokens carrying the bullet slugs, not one id repeated five times. Evidence: the clean-run first-token grouping shows five distinct `portal-…` tokens.
- [ ] The `python3 - "$INDEX" <<'PY' … PY || exit 1` block at base lines 110-124 no longer terminates the suite directly: its outcome routes through the counted emitter so case 4 (base lines 136-155) still runs and still emits its result line when case 3 reds. Evidence: read the diff for the removed `|| exit 1`, corroborated by the break-test in the next item.
- [ ] Continue-past-failure: in a scratch copy break the first case's condition, run the suite, and show a `FAIL: portal-…` line for that case, result lines for the later cases, and a non-zero exit status (`echo $?`). Falsified by the suite still exiting at the first failure, which would leave later cases unaddressable.
- [ ] No lost failure detection: for each assertion in test_portal_store.sh, break the condition it guards in a scratch copy and show that its own case reds and the run exits non-zero — including the two negative `has_line … && fail` assertions at base lines 82 and 100, which must still red when the forbidden string is present. Falsified by any assertion whose break no longer produces a red line or no longer produces a non-zero exit.
- [ ] Result-line plainness: `python3 scripts/check-result-line-plainness.py` reports a clean zero across both changed suites — any new paired emitter added to test_portal_store.sh is defined as `printf '%s\n' "$1"` with no escape literal anywhere in the file and no capability-probe gate on colour. Verify by running the guard against the head; falsified by any escape_literal or coloured-emitter finding.
- [ ] Every suite stays green: `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` exits 0, with both converted suites among the passing set and no other cogni-knowledge suite regressed. Evidence: the runner's JSON envelope shows `success: true` and a non-zero discovered total.
- [ ] Mutation arm: the PR body records two `scripts/mutation-check.sh` recipes — one reverting a property guarded by test_dashboard_contract.sh addressed with `--case <new kdash-… id>`, one reverting a property guarded by test_portal_store.sh addressed with `--case <new portal-… id>` — each `--expr` stated in the form required by the mutation-recipe criterion in cogni-service/skills/service-gatekeeper/references/review-plan-spec.md rule 5, and each `--case` reproducing the label the suite actually prints (including any trailing `:` separator that is part of the printed first token) so it cannot resolve to `case_not_found`. Evidence: each recorded run shows `FAIL: <addressed-id>` and no sibling id on a `FAIL:` line. Falsified by a sibling reding, by `case_not_found`, or by a missing second recipe.

**Risk watchlist:**
- A naive fail-fast-to-counter conversion breaks under the suite's `set -eu` (base line 21): `((errors++))` returns 1 on the 0→1 transition and kills the run, and rewriting `has_line X && fail "…"` (base lines 82, 100) into a shape where the failing test is the final command of the list re-arms `set -e` — the suite then still exits early while appearing converted.
- `fail()` currently dumps `cat "$INDEX"` for diagnostics (base line 36). A counted emitter that drops the dump makes a red run un-debuggable; one that keeps it verbatim dumps the whole index once per failing case now that the run no longer stops at the first red. A human should eyeball the trade-off; no gate can decide it.
- Three result lines that never fire on a clean run and therefore escape the AC's grouping evidence: the missing-SKILL.md guard at test_dashboard_contract.sh:23, the missing-wiki_index_update.py guard at test_portal_store.sh:39, and the `print("FAIL: bullet precedes the inserted span under ## Sources")` inside the Python heredoc at test_portal_store.sh:122. Each emits a `FAIL:`-prefixed line with a non-conforming first token when it does fire, which a mutation harness or a genuinely red run would surface. Worth a human look at whether the author ided or reworded them; not gated, because a clean run cannot see them.
- Ids that encode only the narrative case number (`portal-01…`) invite reuse across the eight assertions that currently sit under narrative case 1 — the per-line uniqueness rule is what breaks, and it breaks silently until someone groups the emitted tokens.
- A fix that proves uniqueness by adding a guard suite or a fixture is right in spirit and wrong in scope: it is a fourth changed path that falsifies the issue's seventh acceptance bullet. Uniqueness must be demonstrated by running the two suites, not by shipping a checker.
- The scheme drifting back toward the tracker's disproved premise — a rule restricting id punctuation on escaping grounds — or toward a source-grep verification rule, which cannot see this very suite's loop-interpolated labels (test_portal_store.sh:86-88) or the lines printed from inside its Python heredocs (lines 110-124, 136-155). Those two shapes are exactly why the constraint exists.
- Relabelling is an easy cover for silently weakening an assertion — loosening a grep pattern so a stubborn case goes green while the diff reads as a pure rename. Items 13 and 19 guard the two suites structurally, but only execution against a broken condition actually proves detection survived.
- The `--case` colon trap: the printed first token is likely `kdash-01-…:` with the separator colon attached, so a recipe passing the bare id can resolve to `case_not_found` and grade against nothing rather than failing loudly. A reviewer should re-run at least one recipe rather than trusting the pasted output.
- cogni-knowledge/CLAUDE.md is already very long; a section that states the six rules in prose without a copyable worked example leaves the three rollout children re-inventing the shape, which is the inconsistency this design child exists to prevent. Judgment call, not gradable against a diff.

- The issue's plan-time `## Acceptance criteria` carries 7 bullets, but a base-grounded regeneration derives 22 gating items — and three of them are sourced from the issue's `### Constraints` section rather than any acceptance bullet: result-line plainness under scripts/check-result-line-plainness.py, the whole-plugin green sweep under scripts/run-plugin-tests.py, and the no-behavioural-assertion-change requirement in the two converted suites. Those three are stated as constraints but never promoted into a gradable acceptance bullet, so the plan-time acceptance bar under-specifies the issue's own stated intent. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1523

**Summary:** Both perspectives independently graded all 22 committed contract items satisfied by execution; the two merger-added preference observations are demoted advisory additions that do not gate, leaving only one un-waived minor defect (lost executable bit on both changed suites, CI-unaffected) which alone does not block merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/CLAUDE.md:238` — Two cosmetic inconsistencies between the recorded section and the suites it documents; neither falsifies contract item 8, since every gated rule is followed. (a) The Placement bullet's worked example prints `PASS: kdash-04 names the zero-network cost win`, but the suite actually emits `kdash-04-zero-network`. Both match the documented shape, yet a rollout author copying the example gets a bare-number id where the reference suite uses a discriminated one. (b) The Both-arms bullet advises passing the label once as a single argument reaching both arms, and test_dashboard_contract.sh:35/37 instead duplicates the full label literal across the two arms. The gated formulation of that rule is satisfied. Worth a maintainer eyeball only because this file is the template three rollout children copy.
  - *Suggested fix:* Make the placement example the id the suite really prints, and soften the Both-arms sentence to mark single-argument routing as the preferred shape rather than the rule.

- **[minor]** (auto-fixable) `cogni-knowledge/tests/test_portal_store.sh:1` — Both changed suites silently lose their executable bit: `git diff --summary` reports `mode change 100755 => 100644` for test_portal_store.sh and test_dashboard_contract.sh. Nothing in the PR's stated intent asks for this, and it is invisible in the textual diff. Impact is small: `./cogni-knowledge/tests/test_portal_store.sh` now returns permission denied. CI is unaffected - run-plugin-tests.py invokes `bash <path>` and its docstring explicitly tolerates suites lacking the bit - and the base tree already carried three non-executable suites in the same directory, so this is not a new class. No other consumer invokes either suite directly (grepped repo-wide).
  - *Suggested fix:* Restore the bit on both paths: `git update-index --chmod=+x cogni-knowledge/tests/test_portal_store.sh cogni-knowledge/tests/test_dashboard_contract.sh` and amend, so the diff carries only the intended content change.

- **[minor]** (advisory-addition) `cogni-knowledge/tests/test_portal_store.sh:186-193` — Deferring the `cat "$INDEX"` dump from fail() to the summary branch changes what the dump shows, which the accompanying comment does not mention. In the base, fail() dumped the index at the moment of failure; now it dumps once at end of run, after cases 2 and 3 have further mutated the same index. So a red portal-store-01/-02/-03 is now debugged against a later index state than the one that failed. The trade-off the comment does argue (one dump instead of one-per-red, keeping the FAIL: id lines greppable) is sound and is the right call; only the state-drift half is undocumented.
  - *Suggested fix:* Add a half-sentence to the summary block noting the dump reflects the end-of-run index, not the state at the first red.

### Positive observations

- Item 15/18/19, the contract's highest-risk area, is genuinely clean under set -eu. fail() ends on a plain assignment (status 0), and the three new helpers put every fail() call inside an if/else. Proved by breaking all 11 assertion groups individually: 11/11 reded their own case, emitted all 15 result lines, and exited 1. No sibling reded; no run terminated early. (plugin-validator)
- Both negative assertions survived conversion with detection intact - exactly the shape the risk watchlist warns can silently re-arm fail-fast. (plugin-validator)
- The dashboard conversion is provably label-only: normalising every double-quoted string literal out of base and head yields byte-identical files. (plugin-validator)
- All three lines the risk watchlist flagged as invisible to a clean run were handled, including removing the stray FAIL: prefix inside the case-3 Python heredoc so the shell wrapper owns that case's single result line. (plugin-validator)
- The loop family is correctly slugified from a genuinely varying value: five distinct tokens, and breaking the loop assertion reds all five separately. (plugin-validator)
- Uniqueness holds by execution on both suites - 15 emitted result lines, 15 distinct first tokens, max group size 1, zero labels still leading with the old namespace prefix. (plugin-validator)
- Both mutation recipes reproduce independently (re-run, not trusted from the PR body): guard_verified, no case_not_found, tree restored clean. (plugin-validator)
- The suite-slug collision reasoning in the recorded scheme is grounded: test_portal_staleness.sh really does sit beside test_portal_store.sh. (plugin-validator)
- Guards and sweeps all green on the head: plainness 0 violations across 124 files, breadcrumbs 0, run-plugin-tests --filter cogni-knowledge 77/77, unfiltered sweep 124/124, no version value moved. (plugin-validator)
- item 1: exactly the three in-set paths, no fourth path. (contract-grading-reviewer)
- item 2: fixtures/test_helpers.sh absent from the changed-file list; shared emitters unmodified. (contract-grading-reviewer)
- items 3-6: the new ### heading states all six rules - shape, placement, both-arms, loop slugification, per-emitted-line uniqueness, and verification-by-execution. (contract-grading-reviewer)
- item 7: grepping the new section for regex/metachar/escap/bracket returns zero matches - the disproved premise is not carried forward. (contract-grading-reviewer)
- items 8-12: the section's worked examples match what the suites emit; 15 distinct kdash- tokens by execution; the two near-identical dashboard labels receive distinct ids; the inline pair carries one id on both arms; all 13 assert patterns and target files byte-unchanged. (contract-grading-reviewer)
- items 14-19: fail() no longer exits; a clean portal run emits 15 distinct-token PASS lines; the loop yields 5 distinct slugified tokens; the heredoc's || exit 1 is now routed through assert_rc; a scratch break of case 1 reds it, lets all 10 later cases emit, and exits 1; six representative breaks including both negative assertions each red only their own case. (contract-grading-reviewer)
- items 20-22: plainness clean zero across 124 files; run-plugin-tests --filter cogni-knowledge 77/77; both mutation recipes' --case values are literal substrings present in the suite source (not runtime-interpolated) and both re-ran to guard_verified. (contract-grading-reviewer)

### Recommendation

---

Issue #1525, opened 2026-08-19
repo: L1's agents_discovered floor sits at 99% of observed, outside the suite's own margin convention


claude-cogni-work:
## Current behavior

`tests/test_check_mcp_tool_grant.sh` case `L1` floors `agents_discovered` at **100** against an observed **101** on the current tree — a ~1% margin.

## Desired behavior

The floor sits inside the margin band the suite's own convention describes, so ordinary churn cannot red it.

## Motivation

The suite states the convention explicitly in its header: liveness floors are *"INEQUALITIES pinned strictly below the values the real tree carries today, so a new agent never turns the suite red, while a collapsed scan population still does."*

Every other floor honours that with room to spare — form-sum 90/92, `grant_tokens` 60/73, `body_call_sites` 10/14, `provides_tools_vocabulary` 10/18, `provides_tools_code_span_names` 20/27, i.e. roughly 0.56x–0.82x of observed. `L1`'s agent floor is the outlier at 0.99x.

The consequence is asymmetric with the case's purpose: retiring any plugin that carries two or more agents reds `L1` on an otherwise-correct tree. That is a false red on a legitimate change, the opposite of what the convention is for — and plugin retirement is not hypothetical in this repo, where the roster has been consolidated more than once. Meanwhile a genuinely collapsed discovery glob would be caught just as well by a floor with a real margin.

Surfaced by the refinement pass while resolving the `provides_tools` completeness work; separable from it, since that change adds `L4` and touches no existing floor.

## Acceptance criteria

- [ ] `L1`'s `agents_discovered` floor is re-pinned inside the suite's own stated margin band rather than at 99% of observed. Falsified by a floor left within a few counts of the live value.
- [ ] The adjacent comment records the observed value at the time of re-pinning, as the `L2` comment already does. Falsified by a re-pinned number with no recorded observation behind it.
- [ ] The suite still reds on a collapsed discovery glob — the floor stays strictly above zero and above a single-plugin population. Falsified by a floor so loose that a broken glob still passes.
- [ ] No `plugin.json` or `marketplace.json` version field is touched.

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-08-19T03:37:48Z` (GitHub clock).

Released automatically if triage parks, decomposes, or closes this issue; carried
into resolution otherwise. A claim older than the freshness TTL reads as a strand
and becomes pickable again.

[...]

PR #1526, opened 2026-08-19
MCP guard: surface grants outside the registry vocabulary

### LINKED ISSUE #1501: repo: nothing asserts the registry provides_tools vocabulary is complete

claude-cogni-work:
Closes #1501.

## Decision: ADOPT, on the non-failing channel

`provides_tools[]` completeness is now asserted by the guard itself. A granted `mcp__<registered-ns>__<tool>` whose bare name is absent from that server's `provides_tools[]` is surfaced as a `granted_outside_vocabulary` entry on `data.observations[]`. `data.violations[]` and the exit code are untouched, so CI is not newly gated on a registry-accuracy judgement — the judgement simply stops being invisible.

It reports rather than fails because a grant outside the vocabulary is a **registry-accuracy fact, not an agent defect**: the agent's grant is correct and the registry is behind it. A hard arm would also be directionally ambiguous — a name outside the vocabulary could mean the registry is stale *or* the agent granted a typo, and the guard cannot tell which side to fix. That is the same class as the `listed_without_grant` observation one field over, so it takes the same channel.

The **all-or-nothing empty-vocabulary floor is decided too, and differently**: it tightens from union-across-servers to **per-server**, so a registered server carrying no `provides_tools[]` is an exit-2 error rather than a silent narrowing. The two are treated differently on purpose — an absent array is not a judgement about upstream, it is a vocabulary that cannot be read, and no MCP server provides zero tools. The existing union floor keeps its message and its case (`V6` still expects exit 2).

## Measurement (re-derived against the base tree)

| Quantity | Value |
|---|---|
| Grant tokens, total | 73 |
| …on unregistered namespaces (no vocabulary to resolve against) | 18 |
| **Examined population** (`provides_tools_grant_names_resolved`) | **55** |
| Registry vocabulary | 18 |
| Distinct granted names on the two registered servers | 26 |
| **Gaps reported** | **9** |

All nine are `cogni-workspace` grants: `clear_canvas`, `export_to_excalidraw_url`, `get_element`, `get_variables`, `group_elements`, `query_elements`, `restore_snapshot`, `snapshot_layout`, `snapshot_scene`. The tree is green under the change — exit 0, zero violations.

## Summary

- **One token grammar.** `MCP_TOKEN_RE` now captures both segments and is read only through `namespaces_in()` / `grants_in()`. A second pattern would have put the character class in two homes — and that class is the one this guard has already been burned by once, going blind to `claude-in-chrome` with every case green. `findall` is deliberately not used on it: with two groups it yields tuples, and a caller expecting flat strings would make `namespace not in granted` always true and `granted.intersection(owners)` never intersect.
- Per-server vocabulary is carried on the registry entry `load_registry()` already builds, so no MCP tool name is ever written down in the guard.
- **One observation per gap**, carrying its granting agents as evidence — not one per granting agent. The fix is a single registry edit however many agents grant the name, so per-agent records would have restated one fact 17 times and pointed each copy at a file that is by construction not where the fix goes.
- New `data.summary` counter `provides_tools_grant_names_resolved`, incremented *before* the membership test, so a mutation of the test collapses the reporting without collapsing the counter.
- The stale docstring residual — *"Widening them is a registry-accuracy judgement, deliberately not made here"* — is replaced by the recorded verdict, its measurement, and the residuals that survive.

## Scope decisions

- **The registry is deliberately not widened.** Adding the nine names would assert a factual claim about the upstream servers' real tool sets that nothing in this repo can verify, and would promote nine unverified names into the *hard* code-span arm's reach — a behaviour change to a different arm with its own false-positive surface. A probe confirmed widening keeps the guard green (vocabulary 18 → 27, span reach 27 → 35), so this is a judgement call, not a feasibility one. The gaps stay open and are now named in the guard's output on every run rather than living in one docstring sentence.
- **The suite does not assert how many gaps the real tree carries.** Asserting zero would turn a non-failing channel into a hard CI gate through the back door; asserting some would red the moment someone correctly widens the registry — the outcome this channel exists to prompt. `L4` floors the *examined population* instead, which is stable under the intended remediation.
- Hard clean zero survives: no baseline, allowlist or skip marker; `V6` still expects exit 2.
- No `.claude-plugin/` path is touched, and no `plugin.json` / `marketplace.json` version field — the post-merge workflow owns the bump.

## Tests

`V6b`/`V6b1` (per-server floor), `G1`/`G1a` (both directions on a `$WORK` fixture, one record per gap with its granting agents), `G2`/`G2a` (an unregistered namespace is neither examined nor surfaced), `L4` (real-tree floor **35** against observed 55). `R6a` now indexes `plugin` defensively, since the channel carries two observation kinds with different key sets.

- `bash tests/test_check_mcp_tool_grant.sh` — 60 cases, `0 failed`
- `python3 scripts/run-plugin-tests.py` — **124/124** suites pass
- `check-result-line-plainness.py`, `check-breadcrumbs.py`, `check-version-bump.py` — all clean

## Mutation recipe

```
scripts/mutation-check.sh --root . \
  --file scripts/check-mcp-tool-grant.py \
  --expr 's/tool not in vocabulary/tool in ()/m' \
  --test 'bash tests/test_check_mcp_tool_grant.sh' --case G1a
```

Replayed: **`guard_verified`** (mutated red, restored green). Verified further that the mutation reds **exactly one** case — `G1a` — while `L4` stays green, which is what proves the recipe grades the detection rather than the plumbing.

`--expr` is fed to `perl -0pi` — perl, never sed or ERE; a sed-style delete form matches nothing and hard-errors as `expr_no_op`. It slurps the whole file, so `/m` is belt-and-braces here (this substitution carries no `^`/`$` anchor). The substitution is non-global, safe because `tool not in vocabulary` occurs exactly once. That uniqueness is why the membership test binds a local `vocabulary` on the preceding line: testing `tool not in registry[namespace]["vocabulary"]` directly would still contain the anchor, and the rewrite would leave a `SyntaxError` reddening every case rather than the one under test. `--case G1a` reproduces the label the suite prints verbatim after `FAIL: `.

## Follow-up issues

- #1525 — `L1`'s `agents_discovered` floor sits at 99% of observed (100 against 101), outside the margin band the suite's own convention describes. Retiring any plugin carrying two or more agents would red it on an otherwise-correct tree.

## Note on the token-scope preflight

`check-token-scope.sh --strict` reports `over_scoped`: the runner is on a `gho_` GitHub CLI token whose scope set (`gist`, `read:org`, `workflow`) is fixed by the CLI's OAuth app and cannot be narrowed. Taken with the sanctioned `--allow-overscoped` opt-out, recorded here rather than left silent.


claude-cogni-work:
## Ship-ready

Author instance has handed this off. Draft PR carrying `svc:needs-review` — the independent review belongs to the merger instance; no verdict surface, no `svc:review-*` label, and no `reviewed-head-sha` marker was written here.

Graded against the 21-item acceptance contract derived at triage (`origin=triage`, `derivedAt: 2026-08-19T02:32:50Z`) — written before the branch was cut, so diff-blind by chronology.

**21 graded · 0 violated · 0 unclear · 0 residual.**

Two items deviate from the contract's *literal* wording and were graded on intent rather than rubber-stamped:

- **Item 10** anticipated that if `MCP_TOKEN_RE` gained a second capture group, both `findall` call sites would be converted. The implementation went further and removed `findall` on that pattern entirely, reading it through named `namespaces_in()` / `grants_in()` helpers. Graded satisfied — the item's falsifier is a body path yielding tuples, and there is no `findall` left to yield them.
- **Item 11** expected the observation to carry a `plugin` field mirroring `listed_without_grant`. The implementation emits one record per `(namespace, tool)` gap at registry level with `plugins[]` / `files[]` lists. Graded satisfied — the item's confirmable criteria are a `kind` field, the `observations[]` channel, and absence from `by_arm`; `by_arm` is built from `findings` alone, so the new kind structurally cannot appear there.

### Verification run by the author

| Check | Result |
|---|---|
| `python3 scripts/check-mcp-tool-grant.py` | exit 0, `total: 0`, 9 gaps reported as observations |
| `bash tests/test_check_mcp_tool_grant.sh` | 60 cases, `0 failed` |
| `python3 scripts/run-plugin-tests.py` | **124/124** suites pass |
| Mutation replay (`--case G1a`) | **`guard_verified`** — and confirmed to red exactly one case while `L4` stays green |
| `check-result-line-plainness` / `check-breadcrumbs` / `check-version-bump` | clean |

### Two things the merger should not take on trust

1. **The handoff preflight is green but vacuous here.** `check-preflight.sh` returned rc=0 `clean=true` with `status: no_applicable_checks` — 7/7 instruments resolved, **all 7 skipped**, 0 gating. Two of those skips are wrong for this diff: it touches `scripts/check-mcp-tool-grant.py` and `tests/test_check_mcp_tool_grant.sh`, but `run-tests` globs `*/scripts/`, `*/tests/` (plugin-scoped, blind to repo root) and `mutation-recipe` globs `*/tests/test-*.sh` (hyphen, blind to this suite's underscore name). Pre-existing instrument properties, not addressed here and not worked around — the equivalents were run directly instead, as tabulated above.
2. **`check-token-scope.sh --strict` exits 1** (`over_scoped`: `gist`, `read:org`, `workflow`). The runner is on a `gho_` GitHub CLI token whose scope set is fixed by the CLI's OAuth app and cannot be narrowed. Taken with the sanctioned `--allow-overscoped` opt-out.

### Judgement calls worth a second opinion

- The registry is **not** widened — the 9 gaps stay open and reported. Adding the names would assert a claim about the upstream servers' real tool sets that nothing in this repo can verify. A reviewer who disagrees would be arguing for a different, defensible scope, not a defect.
- The completeness gap **observes** while an empty `provides_tools[]` **raises**. Two treatments of one field, distinguished on the grounds that an absent array is certainly wrong while a partial one is only probably incomplete.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1526

**Summary:** Both perspectives independently verified all 20 live contract items satisfied (item 7 is the REJECT arm, inapplicable) — by execution, not by reading: the guard runs clean on the head tree with 101 agents discovered (a real zero, not a vacuous one), the 9 registry-accuracy gaps land on the non-failing observations channel leaving the exit code untouched, base and head parser counters are byte-identical, and the recorded mutation replays to guard_verified. The two findings are minor stylistic preferences, neither a fault.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (advisory-addition) `scripts/check-mcp-tool-grant.py:66` — The residuals paragraph the PR rewrote still cites a retired plugin in its neighbouring sentence: the `storyboard` and `web` agents "in cogni-visual and cogni-workspace ... together carry 22 in-span name hits". cogni-visual is no longer in the tree or in .claude-plugin/marketplace.json (roster is 8 plugins), and both named agents now live under cogni-workspace/agents/. The text is pre-existing and unchanged by the diff, so it violates no contract item — but this PR's purpose is to make the docstring the accurate home of the arm's measurements, and this is the one figure in it attributable to a plugin that no longer exists.
  - *Suggested fix:* While the docstring is already open, re-derive that sentence against the current tree — name cogni-workspace only, and restate the in-span hit count from a live run rather than the pre-retirement figure.

- **[minor]** (advisory-addition) `CLAUDE.md:187` — The added CLAUDE.md sentence sits close to that bullet's own instruction not to re-derive the arm's rules there. In practice the addition stays on the right side of the line — it states only the behavioural contract (which channel is non-failing; that a registered server with no provides_tools[] is an error), carries none of the 18/26/9/55 measurement, and explicitly defers with "the docstring owns why the two differ". Recorded so the boundary is a conscious call rather than an unremarked one.
  - *Suggested fix:* No change required. If tightening is preferred, compress to the exit-code contract alone and let the docstring carry the observation-kind name.

### Positive observations

- Real-tree liveness verified by execution: the guard exits 0 with data.summary.total 0 while discovering 101 agents, 73 grant tokens and 14 body call sites — a genuine clean zero, not a vacuous scan, which is the distinction that matters for a hard-clean-zero guard.
- Item 10's shared-regex trap is designed against rather than merely avoided: both segments are captured once and read through finditer helpers (namespaces_in / grants_in), no findall on MCP_TOKEN_RE survives anywhere, and base-vs-head parser counters are byte-identical (bracketed 68 / comma 14 / block 10, grant_tokens 73), so neither existing arm was turned vacuous.
- The hyphen-admitting namespace class is intact — MCP_TOKEN_RE still matches mcp__claude-in-chrome__navigate, and the tree really does carry 15 claude-in-chrome and 12 browsermcp tokens.
- Registry read purely as data: vocabulary is built from provides_tools[] keyed by desktop_config_key, with no MCP tool-name literal anywhere in the guard (grep clean) — the issue's explicit falsifier does not bite.
- The docstring measurement was independently re-derived against the live tree by both perspectives: vocabulary 18, 26 distinct granted names on registered servers, 9 gaps, 55 examined — all three figures correct.
- Item 14's counter is placed before the membership test, so the recorded mutation collapses the reporting while leaving the counter intact — which is exactly what makes the recipe grade the detection instead of the plumbing (mutation reds G1a, leaves L4 green).
- Mutation recipe replayed to guard_verified in a scratch copy by both reviewers (mutated red, restored green), and mirrored into the suite header per that file's convention.
- Item 18's per-server empty-vocabulary floor is a genuinely separate, principled decision: a stale name is a judgement about upstream (report), an absent array is a vocabulary that cannot be read (raise, exit 2). V6 still expects exit 2 and V6b/V6b1 pin the new half.
- The suite deliberately floors the examined population rather than asserting the real-tree gap count — asserting zero would smuggle a hard CI gate in through the suite, asserting nine would red the moment someone performs the very remediation the channel exists to prompt.
- Hygiene clean throughout: 60/60 suite, 124/124 repo sweep, no .claude-plugin/ path touched, 0 plugins version-touched, hard clean zero preserved with no baseline/allowlist/skip marker, stdlib-only imports.

### Recommendation

---

Issue #1535, opened 2026-08-20
cogni-knowledge: tests/README.md documents a 2-argument assert_grep the fixture has never had


claude-cogni-work:
## Plugin
cogni-knowledge

## Current behavior
`cogni-knowledge/tests/README.md` documents the shared assertion helper under `## Convention` as:

```
- `assert_grep <pattern> <description>` for contract-level SKILL.md checks.
```

That is a two-argument signature the shared fixture has never had. `cogni-knowledge/tests/fixtures/test_helpers.sh` takes three positional arguments:

```
assert_grep() {
  local pattern="$1" file="$2" description="$3"
```

The sibling `assert_not_grep` — same three-argument form, symmetric semantics — is absent from that bullet list entirely.

## Desired behavior
The `## Convention` bullet states the actual three-argument form, and names `assert_not_grep` alongside it, so an author following the README lands on a call signature that exists.

## Why this matters
The README is where a converting author reads the assertion vocabulary before writing a suite. A documented signature that omits the `file` argument sends them to a call that silently binds the description into `$2` and leaves `$3` unset — under `set -u` that aborts the suite, and without it the assertion reads the wrong file.

### Constraints
- Documentation only — no behaviour change to `fixtures/test_helpers.sh`.
- No version value moves in the branch.

## Acceptance criteria
- [ ] The `## Convention` bullet in `cogni-knowledge/tests/README.md` states the actual 3-argument form `assert_grep <pattern> <file> <description>`, matching `fixtures/test_helpers.sh`. Evidence: the bullet text. Falsified by a 2-argument form persisting.
- [ ] `assert_not_grep` is named in the same bullet list with its own 3-argument form. Evidence: the bullet text. Falsified by its absence.
- [ ] No behaviour change to `fixtures/test_helpers.sh` — documentation only. Evidence: the diff for that file is empty.
- [ ] The full cogni-knowledge suite sweep stays green. Evidence: `python3 scripts/run-plugin-tests.py --filter knowledge` exits 0.

## Motivation
Found while adding the fixed-string assert pair to the same fixture. The drift is pre-existing and independent of that change, so it was deferred rather than folded in — the bullet added there deliberately carries no arity claim, leaving this list internally consistent either way.

claude-cogni-work:
## Claimed for triage

[...]

PR #1536, opened 2026-08-20
cogni-knowledge: fixed-string assert pair for the shared test helpers

claude-cogni-work:
Refs #1522

`fixtures/test_helpers.sh` offered only a BRE pair, so `assert_grep '[[alpha-synthesis]]' …` parses as a bracket expression plus a trailing literal `]` — and the `a-s` range covers the `s` in `[[omega-synthesis]]`. The assertion reports green against a file that does not contain the string. This adds the fixed-string pair beside it and proves the false pass is closed.

Reproduced at base before any edit:

```
grep -q  -- "[[alpha-synthesis]]" decoy.md  ->  MATCH     (false pass)
grep -qF -- "[[alpha-synthesis]]" decoy.md  ->  no match  (correct)
```

## Summary

- `fixtures/test_helpers.sh` gains `assert_grep_f` / `assert_not_grep_f` — same `(pattern, file, description)` signature, same one-label-feeds-both-arms shape, same caller `errors` increment as the BRE pair, differing only in matching fixed-string. A comment above them states when to prefer each.
- `assert_grep` / `assert_not_grep` are **byte-identical**; the fixture diff has no removed line.
- New suite `tests/test_fixed_string_assert.sh` covers all four arms plus the BRE differential — ten cases, ids `fixed-assert-01` … `fixed-assert-10`.
- `tests/README.md` gains one `## Convention` bullet naming the new pair.
- The liveness-floor comment in `test_plain_result_emitters.sh` is refreshed from 78/77 to the **measured** 79/78. Comment-only: the floor of 50, the fixtures-definer floor, the D1 label and every assertion are untouched.

## Why additive, not a re-point

The issue's additive-only constraint is load-bearing rather than cautious. Real callers depend on BRE semantics, verified against the base tree:

| Call site | Pattern | Breaks under `-F` |
|---|---|---|
| `test_backfill_concepts_index.sh:114` | `^# Concepts$` | anchors become literal characters |
| `test_compose_contract.sh:145,216` | `\[\[sources/`, `\[\[N\]\]` | the escaping backslashes become literal |
| `test_compose_contract.sh:168,208,212,228` | `\|` alternation | alternation becomes literal |

Switching the existing pair would red all of these. 78 suites source this fixture across roughly 1350 assert call sites.

## The negative-probe trap, and how the suite avoids it

The helpers always return 0 and signal failure only by printing `FAIL:` and incrementing the caller's `errors`. A naive `OUT=$(assert_grep_f …)` captures the text but **silently loses the count** — command substitution runs in a subshell, so that increment never reaches the parent.

Every case that expects the helper to fail therefore runs it inside a command substitution and decides from the **captured text**, never from the `errors` side effect. A green run emits no stray `FAIL:` line the mutation harness could misread as a red case. Cases 09 and 10 check the increment separately, in a child shell that owns its own counter.

## Test plan — all executed

- [x] `bash cogni-knowledge/tests/test_fixed_string_assert.sh` — exit 0, ten `PASS:` lines, zero `FAIL:` lines.
- [x] Case ids unique — `… | awk '/^[ \t]*(PASS|ok|FAIL):[ \t]/ {print $2}' | sort | uniq -d` prints nothing; no id carries a trailing colon.
- [x] `python3 scripts/run-plugin-tests.py --filter knowledge` — `total: 78, passed: 78, failed: 0`, run **post-commit**. Base ref carries 77 one-segment suites under `cogni-knowledge/tests/`, so discovery is +1 and the new suite appears in `suites[]`.
- [x] `bash cogni-knowledge/tests/test_plain_result_emitters.sh` — green; `1 emitter-defining file(s) … of 79 exercising the contract`, D1 floor holds.
- [x] `python3 scripts/check-result-line-plainness.py` — `summary.total: 0`, 125 files scanned, 98 definitions inspected.
- [x] `python3 scripts/check-breadcrumbs.py` — 0 files affected.
- [x] `python3 scripts/check-version-bump.py` — 8 plugins scanned, 0 touched, 0 violations. No manifest is in the diff.
- [x] Scope boundary — all four changed paths are under `cogni-knowledge/tests/`.

## Mutation recipe

Reverting the new pair to BRE makes the decoy case pass vacuously, which is exactly the defect this closes — so `fixed-assert-02` must go red. The expression is pinned to `assert_grep_f` **by name** rather than to first-match order, so a later reorder of the two new functions cannot silently retarget the arm onto `assert_not_grep_f` and report `case_not_found`.

```
bash "$MANAGED_SERVICE"/cogni-service/scripts/mutation-check.sh --root . --file cogni-knowledge/tests/fixtures/test_helpers.sh --expr 's/(assert_grep_f\(\)\s*\{[^}]*?)grep -qF --/$1grep -q --/' --test 'bash cogni-knowledge/tests/test_fixed_string_assert.sh' --case fixed-assert-02
```

`$MANAGED_SERVICE` is a `cogni-work/managed-service` checkout — the harness ships there, not in this repo, and the same-named `cogni-portfolio` / `cogni-consult` scripts are different tools. No `/m` modifier is needed; the expression carries no `^`/`$` anchor.

**Verified locally before opening this PR.** Under the mutant the suite exits 1 and `fixed-assert-02` reports `FAIL:`. Cases `fixed-assert-05`, `08` and `09` also red — they key on the same `assert_grep_f` semantics, so that is the expected blast radius, not collateral. `fixed-assert-07` and `10` (the `assert_not_grep_f` cases) stay green, confirming the substitution hit `assert_grep_f` alone. The anchor literal occurs exactly twice in the tree, once per new function body and in no comment, so the arm cannot land on inert prose.

## Scope decisions

- **Included** the `tests/README.md` bullet: the issue's "why now" is that the rollout points converting authors at `assert_grep` as the model, so shipping the helper without listing it keeps aiming new assertions at the vacuous one.
- **Not included** — any change to `assert_grep` / `assert_not_grep` (see the table above).
- **Not included** — converting the seven suites that already hand-roll a fixed-string grep. They are correct today; rewriting them is churn with regression risk and no defect closed.
- **Not included** — `cogni-knowledge/CLAUDE.md`. Its "Both arms" bullet cites `assert_grep` only as the *shape* exemplar, and the new pair follows that same shape, so the line does not go stale.
- No version value moves in this branch — the post-merge workflow owns the bump.

## Follow-up issues

- #1535 — `tests/README.md` documents a 2-argument `assert_grep` the fixture has never had (it takes three), and omits `assert_not_grep` from that bullet list. Pre-existing drift, independent of this change. The bullet added here deliberately carries no arity claim, so the list stays internally consistent either way.

Because a follow-up was filed, this PR links the parent with `Refs` rather than `Closes`.

## Operational note

The token-scope preflight reports `over_scoped` (extra: `gist`, `read:org`, `workflow`) and `--strict` exits 1. The runner is on the standard `gh auth login` `gho_` token, whose scope set is fixed by the GitHub CLI OAuth app and cannot be narrowed; the check itself names `--allow-overscoped` as the sanctioned opt-out for exactly that case. Recorded here rather than silently waived.

Plan record: https://github.com/cogni-work/insight-wave/issues/1522#issuecomment-5352768504

### DIFF
--- a/cogni-knowledge/tests/README.md
+++ b/cogni-knowledge/tests/README.md
@@ -44,6 +44,11 @@ via `trap rm -rf "$WORK" EXIT`.
   wrapped in an escape sequence and never chosen by an environment probe.
   `test_plain_result_emitters.sh` holds this shape.
 - `assert_grep <pattern> <description>` for contract-level SKILL.md checks.
+- `assert_grep_f` / `assert_not_grep_f` are the fixed-string (`grep -qF`)
+  counterparts. Reach for them whenever the pattern is a literal carrying
+  `[`, `]`, `.` or `*` — a wikilink, a glob, a path. `assert_grep` reads such
+  a pattern as a regex, so it can report green against a file that does not
+  contain the string at all.
 - Real Python harness (inline `python3 - <<PY ... PY` heredoc) for
   script-level assertions.
 - Fixtures are minimal — only the files the test actually exercises.
--- a/cogni-knowledge/tests/fixtures/test_helpers.sh
+++ b/cogni-knowledge/tests/fixtures/test_helpers.sh
@@ -42,3 +42,45 @@ assert_not_grep() {
     green "PASS: $description"
   fi
 }
+
+# The fixed-string counterparts of the pair above.
+#
+# Reach for these whenever the pattern is a literal that carries regex
+# metacharacters — a wikilink, a glob, a file path, a JSON key, anything
+# containing [ ] . * ^ $ or a backslash. The BRE pair reads such a pattern as a
+# regex: `[[alpha-synthesis]]` parses as a bracket expression plus a trailing
+# literal ], so it matches any <one-of-those-chars>] sequence and the assertion
+# passes green against a file that does not contain the string at all.
+#
+# These match the pattern verbatim instead, so metacharacters carry no meaning.
+# Keep using the BRE pair when you actually want a regex — an anchored
+# '^# Concepts$', or an alternation written with \| .
+
+# assert_grep_f PATTERN FILE DESCRIPTION
+#   Fixed-string counterpart of assert_grep. Increments the caller's `errors`
+#   variable on failure.
+assert_grep_f() {
+  local pattern="$1" file="$2" description="$3"
+  if grep -qF -- "$pattern" "$file" 2>/dev/null; then
+    green "PASS: $description"
+  else
+    red "FAIL: $description"
+    red "  pattern (fixed): $pattern"
+    red "  file:            $file"
+    errors=$((errors + 1))
+  fi
+}
+
+# assert_not_grep_f PATTERN FILE DESCRIPTION
+#   Symmetric counterpart — fails if PATTERN is present as a literal.
+assert_not_grep_f() {
+  local pattern="$1" file="$2" description="$3"
+  if grep -qF -- "$pattern" "$file" 2>/dev/null; then
+    red "FAIL: $description"
+    red "  pattern (fixed, should NOT appear): $pattern"
+    red "  file: $file"
+    errors=$((errors + 1))
+  else
+    green "PASS: $description"
+  fi
+}
--- a/cogni-knowledge/tests/test_fixed_string_assert.sh
+++ b/cogni-knowledge/tests/test_fixed_string_assert.sh
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+# test_fixed_string_assert.sh — the fixed-string assert pair discriminates.
+#
+# `assert_grep` matches under BRE, so a pattern that is meant as a literal but
+# carries regex metacharacters can match a file that does not contain it. The
+# canonical case is a wikilink: `[[alpha-synthesis]]` parses as a bracket
+# expression plus a trailing literal `]`, and the `a-s` range covers the `s` in
+# `[[omega-synthesis]]` — so the assertion reports green against a decoy. That
+# is a vacuous guard: a test asserting nothing while looking like it passed.
+#
+# `assert_grep_f` / `assert_not_grep_f` match the pattern verbatim and close it.
+#
+# Both arms of every case below are fed by ONE label argument, so a case's pass
+# and fail ids can never drift apart. Cases that expect the helper to FAIL run it
+# inside a command substitution and decide from the CAPTURED text — never from
+# the helper's `errors` side effect, which happens in the subshell and is
+# invisible here, and never as a live line, which would look like a red case.
+
+set -eu
+
+TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
+HELPER="$TESTS_DIR/fixtures/test_helpers.sh"
+
+. "$TESTS_DIR/fixtures/test_helpers.sh"
+
+WORKDIR="$(mktemp -d)"
+cleanup() { rm -rf "$WORKDIR"; }
+trap cleanup EXIT
+
+# The decoy carries omega but never alpha, and one ordinary heading line.
+DECOY="$WORKDIR/decoy.md"
+{
+  printf '%s\n' '# Concepts'
+  printf '%s\n' '- [[omega-synthesis]] — only this one'
+} > "$DECOY"
+
+errors=0
+
+# One label argument reaches both arms, so the ids cannot diverge.
+fail()  { red "FAIL: $1"; errors=$((errors + 1)); }
+check() { if [ "$1" = "yes" ]; then green "PASS: $2"; else fail "$2"; fi; }
+
+# Did a captured helper run report failure?
+reported_fail() {
+  case "$1" in
+    "FAIL:"*) printf '%s' 'yes' ;;
+    *)        printf '%s' 'no' ;;
+  esac
+}
+
+# --- Direct calls: the helper's own pass arm is the result line ---------------
+
+assert_grep_f '[[omega-synthesis]]' "$DECOY" \
+  'fixed-assert-01 assert_grep_f matches a bracketed literal that is present'
+
+assert_grep '[[alpha-synthesis]]' "$DECOY" \
+  'fixed-assert-03 assert_grep still matches the absent literal under BRE (the false pass being closed)'
+
+assert_grep '^# Concepts$' "$DECOY" \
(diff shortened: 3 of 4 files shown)


claude-cogni-work:
## Ship-ready

Authored by the author instance and handed off for independent review. The PR stays draft and carries no review verdict — the merger instance owns that.

**Handoff preflight:** exit 0, `clean: true`, 0 violations, 0 advisories.

But read the green carefully: **only one instrument passed on evidence** (`version-bump` — no version-line touch). Six were skipped:

| Instrument | Reason |
|---|---|
| `run-tests` | `no_suites_at_base` |
| `mutation-recipe` | `not_applicable` |
| `frontmatter` | `not_applicable` — no touched SKILL.md or agents/*.md |
| `breadcrumbs` | `not_applicable` — same |
| `skill-length` | `not_applicable` — no touched SKILL.md |
| `agent-length` | `not_applicable` — no touched agents/*.md |

The last four are genuinely inapplicable to a tests-only diff. **The first two are not, and are worth a maintainer's attention:** the preflight matches `*/tests/test-*.sh` (hyphen), while cogni-knowledge names all 77 of its base suites `test_*.sh` (underscore). So `run-tests` concluded "the base ref carried no suites — nothing to regress from" for a plugin that carries 77, and `mutation-recipe` never examined the suite this PR adds. Both instruments are structurally blind to this plugin, not satisfied by it.

Nothing was changed to make either instrument fire. Both were verified directly instead, and the results are in the PR body:

- Full sweep, run post-commit: `run-plugin-tests.py --filter knowledge` → `total: 78, passed: 78, failed: 0`, with the new suite present in `suites[]`. Base carries 77 one-segment suites, so discovery is +1.
- The mutation recipe was executed, not merely recorded: under the mutant the suite exits 1 with `FAIL: fixed-assert-02`. Cases `05`, `08` and `09` also red (all `assert_grep_f`-dependent — expected blast radius, not collateral); `07` and `10` stay green, confirming the substitution hit `assert_grep_f` alone.

This is a cogni-service instrument gap rather than a defect in this change, so it is reported here rather than worked around, and no cross-repo issue was filed unprompted.

**Contract grading:** graded against the issue-time contract on #1522 (`--origin triage`, derived `2026-08-20T07:03:23Z`, before any branch was cut).

- **21 items graded, 21 satisfied, 0 violated, 0 unclear.** Verdict `pass`, zero findings.
- No fix pass was needed, so there is **no residual** for the merger to inherit.
- The grading pass re-executed the evidence rather than trusting the reported numbers, including an independent `mutation-check.sh` run returning `guard_verified` (mutated red, restored green, worktree clean after restore).

**Operational note:** the token-scope preflight reports `over_scoped` (extra `gist`, `read:org`, `workflow`) and `--strict` exits 1. The runner is on the standard `gh auth login` `gho_` token, whose scope set is fixed by the GitHub CLI OAuth app and cannot be narrowed; the check names `--allow-overscoped` as the sanctioned opt-out for that case. Recorded, not silently waived.

**Linking:** `Refs #1522`, not `Closes` — follow-up #1535 was filed, so the parent stays open by design.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1536
**Perspectives consulted:** cogni-service:contract-grading-reviewer, plugin-validator

**Summary:** Both perspectives passed outright: contract-grading-reviewer confirms all 21 committed acceptance-checklist items are satisfied by actually executing the new suite, the full cogni-knowledge sweep, and the PR body's mutation-check.sh recipe (independently reproducing guard_verified), and plugin-validator confirms the plugin remains structurally valid, version-clean, and fully tested (78/78 suites green). The two minor findings are merger-added preferences on a non-contract location, so they are advisory and do not gate merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-validator | ✅ pass |

### Findings (2)

- **[minor]** (waived) (advisory-addition) `cogni-knowledge/tests/test_fixed_string_assert.sh` — The 'existing behaviour unchanged' arm of AC #1522 is exercised for assert_grep only (cases 03 and 04). No case calls assert_not_grep, so its unchanged-ness rests on the diff being a pure append rather than on an executed assertion.
  - *Justification:* Not a defect. Verified directly: the base fixture is byte-identical against the first 44 lines of the head file, so both pre-existing helpers are provably untouched and the AC is satisfied structurally. Belt-and-braces coverage, not a correctness repair, and out of scope for a pure-append PR.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/tests/test_fixed_string_assert.sh:56-57` — The label for case fixed-assert-03 reads 'assert_grep still matches the absent literal under BRE (the false pass being closed)'. The parenthetical is ambiguous read in isolation: the case asserts the BRE helper's false pass is deliberately RETAINED (that is what 'behaviour unchanged' means), whereas 'the false pass being closed' can be read as a claim that this case closes it. The case that actually closes it is fixed-assert-02.
  - *Suggested fix:* Reword the parenthetical to name the relationship rather than the outcome, e.g. '(the false pass that assert_grep_f closes - retained here on purpose)'.

### Positive observations

- All 21 contract items graded satisfied by execution, not by reading.
- Item 18: the PR body's mutation-check.sh recipe was re-run verbatim and returned mutated_result=red / restored_result=green / verdict=guard_verified, tree restored clean - the recipe is real, not aspirational.
- Item 19: run-plugin-tests.py --filter knowledge exits 0 with data.total 78 vs 77 at base - exactly +1; check-result-line-plainness.py reports summary.total 0.
- Item 3: the fixture diff is a pure addition; assert_grep/assert_not_grep are byte-identical to base, so the ~1349 existing call sites are untouched.
- Item 12: every failure-expecting case is a captured command substitution, so a green run emits ten PASS: lines and no stray FAIL: the mutation harness could misread.
- Item 16: assertions read this repo's own PASS:/FAIL: emitter text and numeric errors counts - no foreign-tool wording, so the guard survives a non-English locale.
- plugin-validator: four independent mutations of the helper were each caught by a distinct correctly-labelled case with exit 1 - no arm is vacuous.
- plugin-validator: version discipline exactly right - plugin.json and marketplace.json both at 1.0.85 matching origin/main, no feature-branch bump and no mirror desync.
- plugin-validator: the liveness-floor comment refresh is independently verifiable - 78 sourcing suites plus 1 definer equals exactly the 79 the comment now states.
- plugin-validator: nothing in the PR can affect plugin loading; all four changed paths are under tests/, not a component-discovery directory.

### Recommendation

---

Issue #1559, opened 2026-08-20
cogni-workspace: README components table repeats the incomplete --update description


claude-cogni-work:
**Plugin:** cogni-workspace

## Current behavior
`cogni-workspace/README.md` line 178 is the Components-table row for `generate-settings.sh`:

```
| `generate-settings.sh` | script | Generates settings files; supports `--update` to preserve custom env vars |
```

## Desired behavior
The description cell states that `--update` preserves custom env vars **and** removes the generated ones for plugins no longer in the list.

## Motivation
Update mode now prunes the env keys the script itself generated (`COGNI_{SUFFIX}_ROOT` / `_PLUGIN`, `PLUGIN_{SUFFIX}_ROOT` / `_PLUGIN`) for plugins absent from the current plugin set, so a workspace created before a plugin retirement stops carrying variables that point at directories which no longer exist.

That change makes this row's one-line description tell only half the story — the same way the `manage-workspace` skill sentence did. The parent change corrects the two SKILL.md surfaces and the script's own usage comment, but its acceptance criteria enumerate an explicit scope boundary that admits only `scripts/generate-settings.sh`, a suite under `cogni-workspace/tests/`, and the `manage-workspace` / `workspace-status` SKILL.md files. This row is therefore filed separately rather than pulled into that diff.

This README section is no longer regenerated for this repository, so a stale cell stays stale until it is edited by hand.

## Acceptance criteria
- [ ] The description cell for the `generate-settings.sh` row states that `--update` preserves custom env vars and removes generated ones for plugins no longer in the list. Falsified by: the cell still describing preservation only.
- [ ] The row keeps its three-column shape and its position between the `discover-plugins.sh` and `install-mcp.sh` rows. Falsified by: a reordered or malformed table row.
- [ ] No other row or section of `cogni-workspace/README.md` changes. Falsified by: any additional hunk in the file's diff.
- [ ] No `version` field in any `.claude-plugin/plugin.json` or in `.claude-plugin/marketplace.json` is touched. Falsified by: `python3 scripts/check-version-bump.py` reporting `version-touched` or `version-mirror-desync`.

_Affected components: cogni-workspace/README.md_

_Downstream hint: generic-modification_

_Complexity hint: low_

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-08-21T12:57:05Z` (GitHub clock).

[...]

PR #1562, opened 2026-08-20
cogni-workspace: prune retired-plugin env keys on workspace update

### LINKED ISSUE #1555: cogni-workspace: generate-settings.sh preserves env keys for retired plugins, so every pre-existing workspace reports broken vars
**Plugin:** cogni-workspace

## Epic context
Phase 1 (live breakage) of the epic 'Reconcile docs, marketplace metadata, architecture diagrams, wiki and stale workspace env keys for the 8-plugin ecosystem'. This is the epic's **only code deliverable** — every other child is documentation or an asset — and therefore the only one carrying a new test suite and mutation checks. It depends on nothing and can ship in parallel with the prose work.

## What happened
`cogni-workspace/scripts/generate-settings.sh` writes one `COGNI_{SUFFIX}_ROOT` and one `COGNI_{SUFFIX}_PLUGIN` env key per installed plugin into `<workspace>/.claude/settings.local.json` (lines 71-86). In update mode it read-modify-writes the existing document and unconditionally carries over every env key it did not itself generate — `generate-settings.sh:93-98`:

```
if update_mode and os.path.isfile(settings_path):
    with open(settings_path) as f:
        settings = json.load(f)
    for k, v in settings.get("env", {}).items():
        if k not in env:
            env[k] = v
```

That carry-over is correct and load-bearing for *user* keys — the comment at lines 88-90 explains that the file is the user's too, holding their permissions, hooks and outputStyle, and that replacing it wholesale destroys what the script does not generate. But it draws no distinction between a user key and a key this same script generated for a plugin that has since been retired. So `COGNI_VISUAL_ROOT`, `COGNI_NARRATIVE_ROOT`, `COGNI_CLAIMS_ROOT`, `COGNI_COPYWRITING_ROOT`, `COGNI_HELP_ROOT` and their `_PLUGIN` twins survive every update run forever, pointing at directories that no longer exist.

The same residue exists in `.workspace-config.json` `installed_plugins[]`, which keeps entries for plugins that are gone.

## Expected behavior
An update run prunes the keys the script owns and no others. A `COGNI_*_ROOT` / `COGNI_*_PLUGIN` (or `PLUGIN_*_ROOT` / `PLUGIN_*_PLUGIN`) key whose suffix maps to a plugin not in the current plugin set is removed; every key outside that generated namespace is preserved untouched, value included. `installed_plugins[]` in `.workspace-config.json` is reconciled the same way.

`workspace-status` then reports its env-var check clean on a workspace initialised before the absorptions, instead of the shape documented at `cogni-workspace/skills/workspace-status/SKILL.md:255-256`.

The generated namespace is the discriminator and it is derivable, not a hardcoded list of retired names: it is exactly the set of key shapes lines 75-86 construct. A key is prunable if and only if this script would have generated it for some plugin, and no plugin currently in the set claims it.

## Steps to reproduce
1. Take, or synthesise, a workspace whose `<workspace>/.claude/settings.local.json` `env` object contains `COGNI_VISUAL_ROOT`, `COGNI_NARRATIVE_ROOT`, `COGNI_NARRATIVE_PLUGIN` and a user-authored key such as `MY_OWN_VAR`, alongside a `permissions` block and an `outputStyle` key.
2. Run `cogni-workspace/scripts/generate-settings.sh` in update mode against that workspace with the current roster.
3. Read `env` back: the retired-plugin keys are still present and still point at non-existent paths.
4. Run `workspace-status` against the same workspace; its env-var check reports each of them broken.

## Error output/logs
```
  Broken: COGNI_NARRATIVE_ROOT -> /path/does/not/exist
  Broken: COGNI_NARRATIVE_PLUGIN -> /path/does/not/exist
```
(the shape documented at `cogni-workspace/skills/workspace-status/SKILL.md:255-256`)

## Acceptance criteria
- [ ] After an update run, an `env` key in the script's own generated namespace whose plugin is absent from the run's plugin set is absent from the written file. Falsified by: seeding `settings.local.json` with such a key and finding it still present afterwards.
- [ ] An `env` key outside the generated namespace survives an update run with its key and value byte-identical. Evidence: the new `cogni-workspace/tests/` suite seeds `<workspace>/.claude/settings.local.json` with a user-authored key such as `MY_OWN_VAR`, runs `bash cogni-workspace/scripts/generate-settings.sh` in update mode, and reads that key back with the same value. Falsified by: seeding a user-authored key and finding it removed, renamed, or its value rewritten. This is the regression the change most plausibly causes and must be its own test case, never an assertion folded into the pruning case.
- [ ] A generated-namespace key belonging to a plugin that **is** in the current set survives, carrying the value the run computes. Evidence: after `bash cogni-workspace/scripts/generate-settings.sh` in update mode, `env` in `<workspace>/.claude/settings.local.json` carries a `_ROOT` and a `_PLUGIN` key for every plugin in the run's plugin set, each holding the path the construction at `cogni-workspace/scripts/generate-settings.sh:75-86` computes for that plugin. Falsified by: a live plugin's `_ROOT` key missing after the run, or holding a stale value.
- [ ] Non-`env` top-level keys of `settings.local.json` — `permissions`, `hooks`, `outputStyle`, `language` and anything else the user put there — are unchanged. Falsified by: a top-level key present before and absent after, or changed in value, other than the keys the script documents itself as writing.
- [ ] `.workspace-config.json` `installed_plugins[]` contains no entry for a plugin absent from the current plugin set after an update run, and loses no entry for a plugin present in it. Falsified by: a retired name surviving, or a live name disappearing.
- [ ] A new suite under `cogni-workspace/tests/` covers the five properties above, is discovered by `python3 scripts/run-plugin-tests.py --filter cogni-workspace`, runs as `bash <path>` with no arguments, and touches no network. Falsified by: the suite not appearing in the runner's discovered list, or requiring an argument or network access to pass.
- [ ] The new suite's `PASS:` / `FAIL:` label lines carry no escape sequence, and its emitters are written to `printf '%s\n' "$1"` rather than gated on a capability probe. Falsified by: `bash tests/test_check_result_line_plainness.sh` exiting non-zero with the new suite present.
- [ ] Any assertion the suite makes about output from a foreign tool (`bash`, `grep`, `python3` tracebacks) is by shape — exit status, `[ -s "$err" ]`, numeric form, match count — never by English wording. Falsified by: a grep for an English fragment of a tool diagnostic, which would pass vacuously on a non-English host.
- [ ] `scripts/mutation-check.sh` proves the pruning half has teeth: revert the pruning property by restoring the unconditional carry-over at `generate-settings.sh:96-98`, run the new suite, and the stale-key case reports `FAIL:`. Falsified by: that case still reporting `PASS:` with the property reverted, which would mean the suite asserts nothing about pruning.
- [ ] `scripts/mutation-check.sh` proves the preservation half too: revert to pruning every key the run did not itself generate, run the new suite, and the user-key case reports `FAIL:`. Falsified by: that case still reporting `PASS:`, which would leave the destructive failure mode — eating the user's own settings — unguarded.
- [ ] No `version` field in any `.claude-plugin/plugin.json` or in `.claude-plugin/marketplace.json` is touched. Falsified by: `python3 scripts/check-version-bump.py` reporting `version-touched` or `version-mirror-desync`.
- [ ] The diff touches only `cogni-workspace/scripts/generate-settings.sh`, the new `cogni-workspace/tests/*.sh` suite, and — where the new pruning behaviour is documented — `cogni-workspace/skills/workspace-status/SKILL.md` and `cogni-workspace/skills/manage-workspace/SKILL.md`. Falsified by: any other path in the changed-file list. None of this epic's documentation children may be pulled in here.

## Additional context
The pruning must key on the derivable generated namespace, not on a hardcoded list of the five retired plugin names. A name list encodes one historical event, needs hand-maintaining, and says nothing about the next plugin removed — the same argument `scripts/check-plugin-inventory.py`'s docstring makes for preferring a bijection over a banned-string grep.

_Affected components: cogni-workspace/scripts/generate-settings.sh, cogni-workspace/tests/ (new suite), cogni-workspace/skills/workspace-status/SKILL.md, cogni-workspace/skills/manage-workspace/SKILL.md_

_Complexity hint: medium_

claude-cogni-work:
Closes #1555.

## Summary

`generate-settings.sh --update` read-modify-wrote `settings.local.json` and carried over every pre-existing `env` key it had not regenerated. That carry-over is load-bearing for user keys — the file is the user's too — but it drew no line between a user key and one this same script had written for a plugin since retired. So `COGNI_VISUAL_ROOT`, `COGNI_NARRATIVE_ROOT` and their `_PLUGIN` twins survived every subsequent run, pointing at directories that no longer exist.

The carry-over now takes one of three arms:

```python
for k, v in settings.get("env", {}).items():
    if k in env:
        continue                     # regenerated this run — the fresh value wins
    m = GENERATED_KEY.fullmatch(k)
    if m and m.group(1) not in current_keyspaces:
        continue                     # ours, plugin has left the set — prune
    env[k] = v                       # not ours — preserve untouched, value included
```

The keep-set is **derived, not listed**. `current_keyspaces` accumulates the stem of the same `root_var` the per-plugin loop already builds, so there is no roster of retired names to hand-maintain and the next plugin removed needs no edit here.

## Two design points worth a reviewer's attention

**The three unconditional keys are spared by exact key identity, not by shape.** `PROJECT_AGENTS_OPS_ROOT`, `COGNI_WORKSPACE_ROOT` and `COGNI_WORKSPACE_PYTHON_VENV` are assigned before the update gate on every run, so arm 1 returns them before any shape test. Identity is *required* here, not merely convenient: `discover-plugins.sh` has no self-exclusion, so `cogni-workspace` reaches the per-plugin loop as an ordinary plugin and `COGNI_WORKSPACE_ROOT` is a shape a real plugin genuinely produces. A shape-only rule would delete the workspace's own root pointer for anyone who deselects `cogni-workspace` at the confirm step. Case `P6` pins exactly that run.

**The regex carries no `^`/`$` anchors.** The Python body sits in an *unquoted* `python3 << PYEOF` heredoc, where `$`, backticks and backslashes are live. `.fullmatch()` supplies the anchoring instead, leaving a pattern with no shell-live metacharacter at all.

## Deliberately not changed — each gets a regression-locking case instead

| Surface | Why no code change | Case |
|---|---|---|
| `.workspace-env.sh` | Regenerated wholesale from `env`, so pruning `env` clears the stale exports for free | `P8` |
| `env_vars_count` | Computed after the merge, so it already reports the post-prune count | `P9` |
| `.workspace-config.json` `installed_plugins[]` | Already a full overwrite from the current set, with no carry-over loop | `Q3` |
| Malformed prior JSON | The uncaught `json.load` still aborts non-zero under `set -euo pipefail`; kept fail-loud | `Q2` |
| Init mode | The prune sits inside the existing `update_mode and os.path.isfile(...)` gate, so it is unreachable without `--update` | `Q1` |

**On the issue's `installed_plugins[]` premise.** The issue reports that "the same residue exists in `.workspace-config.json` `installed_plugins[]`". It does not, as far as this script is concerned — the update branch does `config["installed_plugins"] = plugin_names`, a full overwrite with no carry-over loop, so it cannot strand an entry. Acceptance criterion 5 is therefore **already satisfied on `main`**; it is graded here by the presence of case `Q3`, not by a change to the config write.

## A judgement call

A plugin that is installed but supplied *without* a `path` writes no `_PLUGIN` key this run. Its previously-generated `_PLUGIN` key is **preserved**, not pruned, because its stem is still in the current set. This follows the issue's literal rule ("a key whose suffix maps to a plugin not in the current plugin set is removed") and errs toward preservation in a destructive operation. Case `Q4` pins it.

The corollary, stated plainly in the script comment so it is discoverable rather than surprising: a hand-added key that happens to wear the generated shape for an absent plugin *is* pruned. That is the intended cost of keying on the namespace, which the issue mandates explicitly over a name list.

## Test plan

New suite `cogni-workspace/tests/test-generate-settings-prune.sh` — 15 cases, hermetic (`mktemp -d` + `trap ... EXIT`), no network, argument-free. It drives the real script rather than re-implementing the predicate. Case `L1` is a liveness floor: a missing script fails the suite rather than letting every later assertion pass vacuously.

Verified on this branch:

- `bash cogni-workspace/tests/test-generate-settings-prune.sh` — 15/15 `PASS:`, exit 0
- `python3 scripts/run-plugin-tests.py` — full sweep, **127/127 passed, 0 failed**, new suite discovered and green
- `python3 scripts/check-result-line-plainness.py` — clean
- `python3 scripts/check-breadcrumbs.py` — clean, 0 findings
- `python3 scripts/check-version-bump.py` — `8 plugin(s) scanned, 0 touched, 0 violation(s)`
- `python3 scripts/check-mcp-tool-grant.py`, `python3 scripts/check-workflow-yaml.py` — both exit 0

Case `Q2` asserts the malformed-JSON abort **by shape only** — non-zero exit plus `[ -s "$err" ]` — never by grepping the Python traceback, which is localized and would pass vacuously on a non-English host. (Worth noting the local `grep` here is `ugrep 7.5.0`, which is precisely the class of foreign-tool variation that rule exists for.)

## Mutation recipe

Both recipes were **run on this branch and returned `guard_verified`**. Note the harness ships inside the cogni-service plugin — there is no `scripts/mutation-check.sh` at this repo root.

**Pruning half** — disable the prune; case `P1` must go red:

```bash
  --root . \
  --file cogni-workspace/scripts/generate-settings.sh \
  --expr 's/if m and m\.group\(1\) not in current_keyspaces:/if False and m.group(1) not in current_keyspaces:/' \
  --test 'bash cogni-workspace/tests/test-generate-settings-prune.sh' \
  --case P1
```

**Preservation half** — widen the prune to every carried-over key, destroying user settings; case `P4` must go red:

```bash
  --root . \
  --file cogni-workspace/scripts/generate-settings.sh \
  --expr 's/if m and m\.group\(1\) not in current_keyspaces:/if True:/' \
  --test 'bash cogni-workspace/tests/test-generate-settings-prune.sh' \
  --case P4
```

Both are plain literal substitutions with no `^`/`$` anchor, so no `/m` modifier is needed under the harness's `perl -0pi -e`. The substitution target stands alone on its own line and appears verbatim exactly once. Case ids are letter-prefixed and two characters, so none is a prefix of another under the harness's whole-token match. The second recipe also reds `Q4` if you want a third data point.

## Scope

The diff touches four paths, all inside the boundary the issue's last acceptance criterion draws: `cogni-workspace/scripts/generate-settings.sh`, the new suite, and the `manage-workspace` / `workspace-status` SKILL.md sentences the behaviour change makes inaccurate. No version field is touched — the post-merge workflow owns the bump.

## Follow-up issues

- #1559 — `cogni-workspace/README.md` line 178's Components-table row repeats the same incomplete `--update` description.

Filed rather than fixed here: `README.md` is not among the four paths the issue's scope criterion admits, so including it would put a path outside that predicate in the changed-file list.

**On `Closes` versus `Refs`.** `deferred[]` is non-empty, which mechanically implies a partial link. It is used as a full link deliberately: #1559 is an adjacent surface the issue's criteria explicitly exclude, and no acceptance criterion of #1555 is left undone by this PR. A partial link would strand a fully-satisfied issue open. Flagging the deviation rather than making it silently.

## Review notes

The pre-commit skill-quality gate returned `needs_revision` on both touched SKILL.md files. One finding was inside this diff and is fixed: the pruned-key notation in `manage-workspace` was ambiguous about whether `COGNI_*` and `_ROOT` formed one pattern or two families, and now spells the shapes out. The remainder are tagged `introduced_by_change: false` — pre-existing authoring observations this diff neither caused nor touched, recorded here rather than fixed:

- `manage-workspace` — the `desktop_config_key` labelling rule is stated at near-full length in both Init Mode step 5 and Update Mode step 5; the "never overwrite the workspace-root CLAUDE.md" caveat appears twice; the 44-line Init Mode step 5 MCP block is a relocate-to-`references/` candidate. Body is at 38.5% of cap.
- `workspace-status` — the MCP Servers check carries roughly a quarter of the body including a four-row state table, and the Theme drift subsection inlines a five-row table; both have existing `references/` files that would be natural homes. The check sequence is also numbered 1–5, 5.5, 6, where the fractional section encodes insertion history. Body is at 73.3% of cap.

Neither reviewer proposed a change to the three `workspace-status` status categories or the `Broken:` worked example, and both were left as-is by design: that skill reads *current* state, and a moved directory belonging to a still-installed plugin produces a broken var that pruning neither does nor should clear.

### DIFF
--- a/cogni-workspace/scripts/generate-settings.sh
+++ b/cogni-workspace/scripts/generate-settings.sh
@@ -4,7 +4,8 @@
 # --language takes an ISO 639-1 code. Codes with a natural-language name in
 # LANGUAGE_NAMES below also get the Claude Code "language" settings key; anything
 # else is reported as data.language_key_skipped and gets no key.
-# Optional: --update (preserve custom env vars)
+# Optional: --update (preserve custom env vars; drop generated ones for
+# plugins no longer in the list)
 
 set -euo pipefail
 
@@ -49,7 +50,7 @@ TARGET_ABS=$(cd "$TARGET_DIR" && pwd)
 
 # Generate all config files via single Python call
 python3 << PYEOF
-import json, os, sys
+import json, os, re, sys
 from datetime import datetime, timezone
 
 target = "$TARGET_ABS"
@@ -59,6 +60,16 @@ update_mode = $( [ "$UPDATE_MODE" = true ] && echo "True" || echo "False" )
 with open("$PLUGINS_TMPFILE") as f:
     plugins = json.load(f)
 
+# The generated namespace: every env key this script writes for a plugin has one
+# of these two shapes. The prune below keys on that derived namespace rather than
+# on a list of retired plugin names, because a name list encodes one historical
+# event, needs hand-maintaining, and says nothing about the next plugin removed.
+#
+# No ^/$ anchors: this Python body sits in an unquoted heredoc, where $ and
+# backticks are live. fullmatch() supplies the anchoring, so the pattern carries
+# no shell-live metacharacter at all.
+GENERATED_KEY = re.compile(r"((?:COGNI|PLUGIN)_.+)_(?:ROOT|PLUGIN)")
+
 # Build env vars
 env = {}
 env["PROJECT_AGENTS_OPS_ROOT"] = target
@@ -67,6 +78,13 @@ env["COGNI_WORKSPACE_ROOT"] = os.path.join(target, "cogni-workspace")
 # across all workspaces on this host, mirroring ~/.claude/mcp-servers/. This is
 # the exact path consumers re-exec under (e.g. cogni-knowledge pdf-extract.py).
 env["COGNI_WORKSPACE_PYTHON_VENV"] = os.path.expanduser(os.path.join("~", ".claude", "workspace-python-venv"))
+
+# Keyspace stems of the plugins in THIS run's set, accumulated from the same
+# root_var the loop already builds so there is no second name source to drift.
+# A stem is recorded per plugin, not per key written: a plugin supplied without
+# a path writes no _PLUGIN key this run, and its existing one must still count
+# as owned rather than read as stale.
+current_keyspaces = set()
 
 for p in plugins:
     name = p.get("name", "") if isinstance(p, dict) else p
@@ -82,20 +100,45 @@ for p in plugins:
         plugin_var = f"PLUGIN_{suffix}_PLUGIN"
 
     env[root_var] = os.path.join(target, name)
+    current_keyspaces.add(root_var[: -len("_ROOT")])
     if path:
         env[plugin_var] = path
 
 # Merge with existing settings if update mode. Read-modify-write the whole
 # document, not just "env": the file is the user's too (permissions, hooks,
 # outputStyle), and replacing it wholesale destroys anything we don't generate.
+#
+# Carrying every prior key over unconditionally is what kept a retired plugin's
+# vars alive forever, pointing at directories that no longer exist. Each prior
+# key now takes one of three arms:
+#
+#   1. regenerated this run  -> keep what this run computed
+#   2. ours, plugin now gone -> prune
+#   3. anything else         -> preserve untouched, value included
+#
+# Arm 1 is also what spares PROJECT_AGENTS_OPS_ROOT, COGNI_WORKSPACE_ROOT and
+# COGNI_WORKSPACE_PYTHON_VENV: they are assigned unconditionally above, so they
+# are already in env before any shape test runs. That identity check is required
+# rather than merely convenient - cogni-workspace reaches the loop as an ordinary
+# plugin, so COGNI_WORKSPACE_ROOT is a shape a real plugin genuinely produces, and
+# a shape-only rule would delete the workspace's own root pointer for anyone who
+# deselects cogni-workspace at the confirm step.
+#
+# Accepted residual: a hand-added key wearing the generated shape for an absent
+# plugin is pruned. That is the intent of keying on the namespace - it names
+# wiring this workspace no longer maintains.
 settings_path = os.path.join(target, ".claude", "settings.local.json")
 settings = {}
 if update_mode and os.path.isfile(settings_path):
     with open(settings_path) as f:
         settings = json.load(f)
     for k, v in settings.get("env", {}).items():
-        if k not in env:
-            env[k] = v
+        if k in env:
+            continue
+        m = GENERATED_KEY.fullmatch(k)
+        if m and m.group(1) not in current_keyspaces:
+            continue
+        env[k] = v
 
 settings["env"] = env
 
--- a/cogni-workspace/skills/manage-workspace/SKILL.md
+++ b/cogni-workspace/skills/manage-workspace/SKILL.md
@@ -253,7 +253,16 @@ bash ${CLAUDE_PLUGIN_ROOT}/scripts/generate-settings.sh \
   --update
 ```
 
-The `--update` flag preserves any custom env vars the user added manually.
+The `--update` flag preserves any custom env vars the user added manually. It
+also removes the per-plugin vars it generated for plugins that are no longer in
+the confirmed list: `COGNI_<NAME>_ROOT` and `COGNI_<NAME>_PLUGIN`, or
+`PLUGIN_<NAME>_ROOT` and `PLUGIN_<NAME>_PLUGIN` for a plugin outside the
+`cogni-` namespace. So deselecting a plugin at step 2 retires its wiring instead
+of leaving a var pointing at a directory that no longer exists.
+
+`PROJECT_AGENTS_OPS_ROOT`, `COGNI_WORKSPACE_ROOT` and
+`COGNI_WORKSPACE_PYTHON_VENV` are kept regardless of that list, since the
+workspace needs its own pointers even when `cogni-workspace` is deselected.
 
 ### 3.5. Refresh Optional Python Dependencies
 
--- a/cogni-workspace/skills/workspace-status/SKILL.md
+++ b/cogni-workspace/skills/workspace-status/SKILL.md
@@ -254,7 +254,8 @@ Plugins:      WARNING  | 5 registered, 6 installed
 Environment:  WARNING  | 12 vars set, 2 broken
   Broken: COGNI_NARRATIVE_ROOT -> /path/does/not/exist
   Broken: COGNI_NARRATIVE_PLUGIN -> /path/does/not/exist
-  -> Run manage-workspace to refresh environment variables
+  -> Run manage-workspace to refresh environment variables (an update also
+     removes generated vars for plugins no longer installed)
 
 Themes:       WARNING  | 2 themes available, 1 tiered, 1 drift advisory
   cogni-work   workspace  tiered (1.0) | tokens, assets, components.web, components.deck
--- a/cogni-workspace/tests/test-generate-settings-prune.sh
+++ b/cogni-workspace/tests/test-generate-settings-prune.sh
@@ -50,6 +50,7 @@ seed_ws() {
     "COGNI_VISUAL_ROOT": "/gone/cogni-visual",
     "COGNI_VISUAL_PLUGIN": "/gone/cogni-visual/plugin",
     "PLUGIN_LEGACY_ROOT": "/gone/legacy",
+    "PLUGIN_LEGACY_PLUGIN": "/gone/legacy/plugin",
     "COGNI_TRENDS_ROOT": "/stale/cogni-trends",
     "COGNI_KNOWLEDGE_PLUGIN": "/kept/knowledge/plugin",
     "MY_OWN_VAR": "keep-me-exactly"
@@ -90,8 +91,8 @@ assert_py "P1 retired plugin ROOT key pruned" "$WS" \
   '"COGNI_VISUAL_ROOT" not in env'
 assert_py "P2 retired plugin PLUGIN twin pruned" "$WS" \
   '"COGNI_VISUAL_PLUGIN" not in env'
-assert_py "P3 non-cogni PLUGIN_ namespace key pruned" "$WS" \
-  '"PLUGIN_LEGACY_ROOT" not in env'
+assert_py "P3 non-cogni PLUGIN_ namespace pair pruned" "$WS" \
+  '"PLUGIN_LEGACY_ROOT" not in env and "PLUGIN_LEGACY_PLUGIN" not in env'
 assert_py "P4 user key outside the namespace preserved with its value" "$WS" \
   'env.get("MY_OWN_VAR") == "keep-me-exactly"'
 assert_py "P5 live plugin ROOT key refreshed to the computed path" "$WS" \
--- a/cogni-workspace/tests/test-generate-settings-prune.sh
+++ b/cogni-workspace/tests/test-generate-settings-prune.sh
@@ -1,0 +1,202 @@
+#!/usr/bin/env bash
+# Regression test for the update-mode env-key prune in
+# scripts/generate-settings.sh.
+#
+# Contract under test: an --update run removes the env keys the script itself
+# generated for plugins absent from the current plugin set, and removes nothing
+# else. A key outside that generated namespace survives with its value
+# byte-identical, the three unconditional workspace keys survive regardless of
+# the plugin list, and the non-"env" top-level keys of settings.local.json are
+# untouched.
+#
+# The discriminator is the derived generated namespace, never a list of retired
+# plugin names, so these cases stay meaningful for the next plugin removed.
+#
+# Case labels use a letter-prefixed two-character id with no colon after the id,
+# matching the convention test-mcp-declaration-hygiene.sh documents: the
+# mutation harness matches --case as a whole token, so no id may be a prefix of
+# another.
+#
+# stdlib-only: bash + python3, no pip deps, no network, per the repo-wide script
+# convention.
+
+set -u
+HERE="$(cd "$(dirname "$0")" && pwd)"
+WS_ROOT="$(cd "$HERE/.." && pwd)"
+GEN="$WS_ROOT/scripts/generate-settings.sh"
+TMPROOT="$(mktemp -d)"
+trap 'rm -rf "$TMPROOT"' EXIT
+
+failures=0
+pass() { printf 'PASS: %s\n' "$1"; }
+fail() { printf 'FAIL: %s\n' "$1"; failures=$((failures + 1)); }
+
+# A missing script is a broken suite, never a clean one: without this floor
+# every later assertion would pass vacuously against a file that never ran.
+if [ ! -f "$GEN" ]; then
+  fail "L1 generate-settings.sh located at $GEN"
+  echo
+  echo "1 test(s) failed."
+  exit 1
+fi
+
+# seed_ws <dir> — a workspace carrying stale generated keys, a live-plugin key,
+# a pathless-plugin key, a user key, and non-env top-level keys.
+seed_ws() {
+  mkdir -p "$1/.claude"
+  cat > "$1/.claude/settings.local.json" <<'JSON'
+{
+  "env": {
+    "COGNI_VISUAL_ROOT": "/gone/cogni-visual",
+    "COGNI_VISUAL_PLUGIN": "/gone/cogni-visual/plugin",
+    "PLUGIN_LEGACY_ROOT": "/gone/legacy",
+    "COGNI_TRENDS_ROOT": "/stale/cogni-trends",
+    "COGNI_KNOWLEDGE_PLUGIN": "/kept/knowledge/plugin",
+    "MY_OWN_VAR": "keep-me-exactly"
+  },
+  "permissions": { "allow": ["Bash"] },
+  "hooks": { "marker": 1 },
+  "outputStyle": "custom"
+}
+JSON
+}
+
+PLUGINS='[{"name":"cogni-trends","path":"/p/cogni-trends"},{"name":"cogni-knowledge"}]'
+
+# assert_py <label> <workspace-dir> <python expr over `env`, `doc`, True to pass>
+assert_py() {
+  local label="$1" ws="$2" expr="$3" out
+  out=$(WS="$ws" EXPR="$expr" python3 - <<'PY' 2>/dev/null
+import json, os, sys
+doc = json.load(open(os.path.join(os.environ["WS"], ".claude", "settings.local.json")))
+env = doc.get("env", {})
+sys.stdout.write("1" if eval(os.environ["EXPR"]) else "0")
+PY
+) || out=""
+  if [ "$out" = "1" ]; then pass "$label"; else fail "$label"; fi
+}
+
+# ---------------------------------------------------------------- main run
+WS="$TMPROOT/main"
+seed_ws "$WS"
+STDOUT_MAIN="$TMPROOT/main.out"
+if bash "$GEN" --target "$WS" --plugins "$PLUGINS" --update > "$STDOUT_MAIN" 2>/dev/null; then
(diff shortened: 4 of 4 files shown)


claude-cogni-work:
## Ship-ready

Handed off for independent merger review. Draft, `svc:needs-review`. The author instance posted no verdict label and no review marker — the independent review belongs to the merger.

### Handoff preflight — clean

`check-preflight.sh` returned rc=0 against head `e2b54e9e`, base `origin/main`: 4 touched paths, instruments 7/7 resolved (floor 7), gating verdicts 5/5, **0 violations, 0 contract breaches**.

- `run-tests` pass — suites green for cogni-workspace
- `frontmatter` pass — clean for cogni-workspace
- `version-bump` pass — no version-line touch
- `skill-length` pass — measured 2 of 2 touched files, 0 over cap
- `mutation-recipe` pass — the recorded recipe is runnable as written
- `breadcrumbs` pass, `reason=standing_drift_excluded` — 9 offenders in `cogni-workspace/skills/manage-themes/SKILL.md`, all **outside this branch's touched set**. Standing drift, surfaced not gating, and untouched by this PR.
- `agent-length` **skipped**, `not_applicable` — no touched `agents/*.md`. Naming it so a green run whose instrument never examined anything stays visible.

Graded against the `origin=triage` contract on #1555 (derived 14:35:51Z, **23 checklist items**), using the diff only — the contract was written before any branch was cut, and this instance did not write it.

**23 graded, 0 violated, 0 unclear. No residual.**

The two items most likely to attract a generous grade were checked specifically:

- **`installed_plugins[]`** — the contract states this property is already satisfied on `main` and must be graded by the presence of a regression-locking case, explicitly *not* by a code change ("NOT falsified by: the config-write code being unchanged"). The diff makes no config-write change; case `Q3` locks it.
- **Scope boundary** — the diff's four changed files match the contract's predicate exactly. Nothing under `docs/`, `wiki/`, `.claude-plugin/`, `assets/`, another plugin's tree, or `cogni-workspace/README.md`.

### One bounded fix pass ran

The grader raised a single observation, not a violation: the non-`cogni` pruning case seeded only `PLUGIN_LEGACY_ROOT`, so the `_PLUGIN` twin was never actually asserted gone. The item graded `satisfied` on its own falsifier, but the gap was mechanical and inside a file the plan already touched, so it was closed rather than handed over: `P3` now seeds and asserts both halves of the pair. Re-verified — suite green, and `mutation-check --case P3` returns `guard_verified`.

Head moved `9eb340a5` → `e2b54e9e`; the preflight above is the re-run against the new head, not the earlier one.

### Worth the merger's attention

- **`Closes` with a follow-up filed.** `deferred[]` is non-empty (#1559), which mechanically implies a partial `Refs` link. `Closes` is used deliberately: #1559 targets `cogni-workspace/README.md`, a path the contract's scope item excludes, and all 23 items grade satisfied without touching it — so no contract-graded criterion of #1555 is left undone and a partial link would strand a satisfied issue open. The grader was asked about this specifically and agreed. Flagging the deviation rather than making it silently.
- **The issue's `installed_plugins[]` premise is overstated** and was recorded as such at triage. That script cannot strand an entry there — the update branch fully overwrites from the current set.
- **Token scope** — the runner is on the `gho_` GitHub CLI token, whose scope set is fixed by the CLI's OAuth app and cannot be narrowed. `check-token-scope.sh --strict` exits 1 on it; the run proceeded under the script's own documented `--allow-overscoped` opt-out.
- **Pre-existing SKILL.md drift** is annotated under `## Review notes` in the PR body, all tagged `introduced_by_change: false`. This diff neither caused nor touched it.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1562
**Perspectives consulted:** cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator, plugin-dev:skill-reviewer (manage-workspace), plugin-dev:skill-reviewer (workspace-status), cogni-service:skill-quality-reviewer (manage-workspace), cogni-service:skill-quality-reviewer (workspace-status)

**Summary:** All six perspectives converge on pass. Contract grading is 23/23 by execution; plugin-validator independently re-ran both mutation recipes rather than trusting the PR body and probed the destructive failure mode with adversarial keys. The prune predicate is derived from the same construction the generation uses, not a hardcoded retired-name roster, and the three unconditional workspace keys are spared structurally by an identity arm that runs before the shape test. Every finding across all six verdicts is a minor, category:preference item, so each is advisory and none gates; the one perspective self-reporting needs_revision raised no major or critical finding, so it neutralizes to pass. A merged-tree re-run (15/15 cases, 15/15 suites, all guards zero) independently corroborates.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-dev:skill-reviewer (manage-workspace) | ✅ pass |
| plugin-dev:skill-reviewer (workspace-status) | ✅ pass |
| cogni-service:skill-quality-reviewer (manage-workspace) | ✅ pass |
| cogni-service:skill-quality-reviewer (workspace-status) | ⚠️ needs_revision |

### Findings (9)

- **[minor]** (advisory-addition) `cogni-workspace/scripts/generate-settings.sh:71` — GENERATED_KEY restates the key-shape constants (the COGNI|PLUGIN prefix and ROOT|PLUGIN suffix alternations) as regex literals, while the f-strings at lines 95-100 are the authoritative construction. The plugin-name half of the predicate is correctly derived, so the issue's mandate is met, but the shape half is a second source that must be hand-edited in lockstep if key naming ever changes, and a divergence fails open. Partial mitigation exists: P5 would redden on a prefix change and P3 on a narrowed alternation.
  - *Suggested fix:* Optionally build the prefix and suffix alternations once as module-level constants used by both the f-strings and re.compile. Purely defensive - no user-observable fault today.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/manage-workspace/SKILL.md:256` — The pre-existing sentence 'preserves any custom env vars the user added manually', read with the new prune paragraph, overstates the preservation guarantee: a hand-added key wearing the generated shape for an absent plugin is pruned, since the prune keys on shape, not provenance. The script documents this residual in its own comment; the issue AC uses the same simplified phrasing, so this is a precision gap rather than a violated criterion.
  - *Suggested fix:* Add a half-sentence noting the shape test is what decides, not who wrote the key.
  - *Raised by:* plugin-dev:skill-reviewer (manage-workspace)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/manage-workspace/SKILL.md:257-261` — COGNI_<NAME>_ROOT leaves the derivation implicit; the script upcases and converts hyphens to underscores, so a reader hand-checking a multi-word plugin name could mis-derive the key.
  - *Suggested fix:* Gloss the placeholder once with a worked example, e.g. cogni-knowledge to COGNI_KNOWLEDGE_ROOT.
  - *Raised by:* plugin-dev:skill-reviewer (manage-workspace)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/manage-workspace/SKILL.md:326` — Update Mode step 7's summary bullet still reads only 'Environment variables changed'. Now that an update deletes keys, that bullet is the user's sole in-run evidence of a destructive edit and does not name what disappeared. The step-1 backup and Rollback section make this recoverable, so it is an observability gap rather than a fault.
  - *Suggested fix:* Sharpen the bullet to name each pruned key.
  - *Raised by:* plugin-dev:skill-reviewer (manage-workspace)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md:258` — The added parenthetical says 'plugins no longer installed', but the prune keys on the confirmed plugin list passed via --plugins, not on disk-installed state; the sibling skill and the issue criterion both word it as 'no longer in the confirmed list'. A user who deselects a still-installed plugin also has its vars pruned, which the current wording does not describe.
  - *Suggested fix:* Reword to 'plugins no longer in the confirmed set' to match the sibling skill and the criterion.
  - *Raised by:* plugin-dev:skill-reviewer (workspace-status)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md:55` — The new update-mode behaviour appears only inside the fenced WARNING example as a parenthetical. A reader following the Environment check's instruction prose never learns that an update prunes generated keys, and the 'user-added keys are left untouched' half is absent from this file entirely - it is the sibling manage-workspace SKILL.md that satisfies contract item 22's 'at least one of' wording.
  - *Suggested fix:* Append a sentence to the Environment section stating the prune and that hand-added vars are untouched, so a Missing result after an update reads as expected rather than as a fault.
  - *Raised by:* cogni-service:skill-quality-reviewer (workspace-status)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md:257-258` — The behaviour is documented only inside the fenced example report block, where every other advisory is a terse one-liner; text there functions as an output template that may be echoed verbatim into reports where nothing was removed.
  - *Suggested fix:* Mirror the statement into the Check 2 prose and shorten the example-block advisory back to one line.
  - *Raised by:* plugin-dev:skill-reviewer (workspace-status)

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md:160-227` — WAIVED as pre-existing drift (introduced_by_change: false). The MCP Servers section consumes roughly a third of the body on install-state troubleshooting even though it already points at references/mcp-registry.md. The body is under budget (16402 vs 22500), so this is a readability observation rather than a size breach.
  - *Justification:* Pre-existing whole-skill drift the change did not introduce (introduced_by_change: false); annotates but does not gate, per the Step 3.5 pre-existing-drift policy.
  - *Raised by:* cogni-service:skill-quality-reviewer (workspace-status)

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/workspace-status/SKILL.md:250-256` — Pre-existing, untouched by this diff: the example uses 'narrative' as a plugin with COGNI_NARRATIVE_* vars, but narrative was absorbed into cogni-workspace as a skill and is not a roster plugin, so those vars would never be generated.
  - *Justification:* Pre-existing content the diff never touched; scope fence limits this to a non-gating observation.
  - *Raised by:* plugin-dev:skill-reviewer (workspace-status)

### Positive observations

- Contract grading is 23/23 satisfied by execution, with zero findings raised.
- The prune predicate is derived, not a roster: current_keyspaces accumulates root_var minus the _ROOT suffix inside the existing per-plugin loop, so the next plugin retirement needs no edit here.
- Both mutation recipes were re-run independently by two perspectives and both returned guard_verified - the pruning half (P1) and the destructive-preservation half (P4) each genuinely red when their property is reverted.
- User-key preservation verified by an independent adversarial probe: seeding OPENAI_API_KEY, MY_OWN_VAR, PLUGIN_ROOT, COGNI_A and ROOT alongside all five retired plugins' keys pruned exactly the five retired pairs and left every other key byte-identical.
- The three unconditional workspace keys are spared by exact key identity before any shape test - required rather than convenient, since cogni-workspace reaches the loop as an ordinary plugin; case P6 pins the deselect run.
- Greedy matching backtracks correctly on the hardest real shape: a pathless live plugin kept PLUGIN_MY_PLUGIN_PLUGIN while PLUGIN_MY_PLUGIN_ROOT refreshed, so there is no stem mis-split.
- Empty plugin set degrades safely - only the three unconditional keys plus user keys survive.
- Case Q2 asserts the malformed-JSON abort by shape only (exit status plus stream emptiness), never by grepping a localized traceback, satisfying the foreign-emitter rule.
- Case L1 is a real liveness floor - a missing generate-settings.sh fails the suite rather than letting 14 later assertions pass vacuously.
- The PR correctly identifies that acceptance criterion 5 was already satisfied on main and locks it with a regression case instead of inventing a change - falsifying the issue's premise beats silently complying.
- Documentation lands on both skills with consistent wording, and every factual claim in the new passage is checkable against generate-settings.sh and holds.
- Scope respected: exactly the four fenced files, a fifth path deferred to a follow-up, and no version field touched.
- Merged-tree corroboration by the merger: the head merged with current origin/main and committed still gives 15/15 cases, 15/15 suites, failed 0, with the runner discovering the new suite and all repo guards at zero.

### Recommendation

---

Issue #1578, opened 2026-08-22
cogni-workspace: workspace-dashboard's MCP Servers row renders a permanent "0/0 installed (ok)" when the registry's shape drifts


claude-cogni-work:
## Context

Discovered while resolving #1573 (the Dependencies row of the same `health_snapshot` function reading keys its emitter never produced). It is explicitly **out of that issue's scope** — the committed acceptance contract for #1573 fences every changed path to the Dependencies row, its documentation twin and its new suite — so it is filed here to be tracked rather than dropped.

## Problem

`cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py` builds `mcp_servers` at `:1242` from `registry.get("servers", {})` — a `.get(key, default)` read of a cross-file producer (`cogni-workspace/references/mcp-git-registry.json`) whose shape nothing verifies.

The guard at `:648` (`if mcp_servers is None`) covers only *file missing or unparseable*. It does **not** cover shape drift: a registry that parses cleanly but no longer carries a `servers` key yields `[]`, not `None`. The status expression at `:655` is then

```python
mcp_label = "ok" if installed + manual == total else "warning"
```

which is `0 == 0` — permanently true. The row renders `0/0 installed` with a green dot.

## Evidence

Reproduced by execution against a registry whose `servers` key was renamed to `server_list`:

```
<div class="health-dot ok"></div>
<div><strong>MCP Servers</strong></div>
<div ...>0/0 installed</div>
```

A green dot asserting health for a registry the reader could not understand.

## Why this is latent, not live

The real `mcp-git-registry.json` currently carries `servers` with 2 entries, so the row is truthful today. It fires the moment that file's shape moves, and nothing would report that it had.

## Relationship to #1573

Same defect class — a count comparison that reads healthy when both sides are zero, sitting downstream of an unverified producer contract — but a **different row** of the same `health_snapshot` function, with a **different producer** (an on-disk JSON registry rather than a subprocess envelope).

A survey of all six rows during #1573 found this is the **only** exposed sibling: the Foundation row's producer is the same module, the Environment row already requires a non-zero count for `ok`, and the Plugins/Themes rows key on the truthiness of in-process lists.

## Proposed fix

Guard the empty side before the comparison, as #1573's Dependencies fix now does — distinguishing "registry unreadable" from "registry read, zero servers configured" if that distinction is wanted. Two instances is not yet enough to justify factoring the rule into shared machinery; a third would be.

## Acceptance criteria

[...]

PR #1579, opened 2026-08-22
cogni-workspace: dashboard Dependencies row reads the real check-dependencies shape

### LINKED ISSUE #1573: cogni-workspace: workspace-dashboard reads the wrong keys from check-dependencies.sh and always renders "0/0 required, 0/0 optional (ok)"
## Context

Discovered while resolving #1568 (troubleshoot Scope Boundary accuracy). It is explicitly **out of that issue's scope** per its committed acceptance contract (item-6 grader note), and the PR for #1568 does not touch this code or its inputs — so it cannot have made this worse. Filed here so the finding is tracked rather than dropped.

## Problem

`cogni-workspace/scripts/check-dependencies.sh` emits:

```
data.dependencies[]        # each entry {name, available, required, version}
data.missing_required      # int
data.missing_optional      # int
```

Two consumers read a contract that does not exist:

- **`cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py:678-679`** reads `data.get("required", {})` and `data.get("optional", {})`. Both resolve to `{}` on every run, so `req_total`, `req_ok`, `opt_total` and `opt_ok` are all `0`. The status expression at `:684` — `label = "ok" if req_ok == req_total else "danger"` — is therefore `0 == 0`, permanently true. The section renders **`0/0 required, 0/0 optional` with a green `ok` label forever**, including on a machine where `jq` and `python3` are both missing. Because both reads use `.get(..., {})` it never raises, which is why it has gone unnoticed.
- **`cogni-workspace/skills/workspace-dashboard/references/dashboard-sections.md:80`** documents the same wrong shape (`data.required[]` / `data.optional[]`), so a future reader re-derives the bug from the docs.

## Why this is one issue and not two

The code and its documentation twin are the same defect class and belong in one PR: fixing the reader without the reference leaves the documentation lying, and fixing the reference without the reader leaves the dashboard green-forever. Bundled deliberately to one-PR granularity.

## Constraint

Do **not** change `check-dependencies.sh`'s emitted shape. The other consumer — `workspace-status/SKILL.md` §5 Dependencies (`:126-135`) — reads it correctly today, so the script is the source of truth and the two broken readers are what move.

## Acceptance criteria

- [ ] `generate-dashboard.py` derives its dependency counts from `data.dependencies[]`, partitioning on each entry's `required` flag, and no longer reads `data.required` / `data.optional`
- [ ] With a required tool absent (or a fixture JSON carrying `available: false` on a required entry), the rendered section shows a non-`ok` status and a truthful `n/m required` count — i.e. the section is **falsifiable** rather than constant. A test that only exercises the all-present case does not satisfy this.
- [ ] `references/dashboard-sections.md:80` documents the actual emitted shape (`data.dependencies[]` with `name` / `available` / `required` / `version`, plus `missing_required` / `missing_optional`)
- [ ] `check-dependencies.sh`'s output shape is unchanged, and `workspace-status/SKILL.md` §5 still reads correctly against it
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-workspace` reports 0 failed (base measures 15/15)

**Parent issue:** #1568

claude-cogni-work:
Closes #1573.

## Summary

The Health Snapshot's Dependencies row read `data.required` / `data.optional` — keys `check-dependencies.sh` never emits. Because both reads used `.get(..., {})`, they resolved to `{}` silently, `req_total` and `req_ok` were both `0`, and the status expression `label = "ok" if req_ok == req_total else "danger"` was a permanent `0 == 0`. The row rendered `0/0 required, 0/0 optional` behind a green dot on every machine, including one missing both required tools. That silence is why it went unnoticed.

`_check_dependencies` now partitions `data.dependencies[]` on each entry's own `required` flag, matching the `sum(1 for x in ... if pred)` / `f"{ok}/{total} <noun>"` idiom the sibling rows in the same function already use.

**Shape guards are part of the fix, not padding.** The old form *could not* raise — the `.get` defaults saw to that — and `references/dashboard-sections.md:87-95` binds the renderer to still produce *some* HTML in every failure mode. The new read has to earn that back: an absent `dependencies` key, a non-list value there, or non-dict entries would each raise `AttributeError`/`TypeError`, and neither is caught by the existing `except` tuple. So the read is `isinstance`-guarded *and* the tuple widened — two layers, deliberately, because trading a silent wrong answer for a hard render crash would not be a fix.

`dashboard-sections.md:80` restated the same wrong shape and would have re-seeded the bug from the docs, so it moves in the same commit.

## Scope decisions

- **Included:** both readers the issue names (one defect class), the exception-surface hardening the changed read forces, and coverage that makes the fix falsifiable.
- **Excluded by the issue's own constraint:** `cogni-workspace/scripts/check-dependencies.sh` keeps its emitted shape byte-for-byte, and `skills/workspace-status/SKILL.md` §5 — which reads it correctly today — is untouched. A repo-wide search confirmed `_check_dependencies` is the *only* programmatic consumer; every other mention is prose that names no key shape. `git diff --stat origin/main` reports zero changed lines in both.
- **No production seam was added.** `_check_dependencies(workspace_root)` already derives its script path from its own argument, so the suite points it at a temp tree holding a stub emitter. No injection hook, no test-only branch in production code.
- **No `plugin.json` version bump** — the post-merge workflow owns it.

## Test plan

All measured, not asserted:

- `python3 scripts/run-plugin-tests.py --filter cogni-workspace` → **`total 16, passed 16, failed 0`** (base measures 15/15; the 16th is the new suite). This command is **green at base**, so it is a regression guard only — it cannot distinguish a correct fix from no fix, and is not offered as evidence the defect is fixed.
- `bash cogni-workspace/tests/test-dashboard-dependency-counts.sh` → 10/10 `PASS`.
- **The suite run against the base commit → 9 of 10 RED.** This is the real evidence. `origin/main` was materialised into a scratch tree, the suite dropped in unchanged, and every behavioural case failed against the old reader. The tenth, `D9`, is correctly green at base: it pins the *emitter's* contract, which this PR deliberately does not touch.
- `python3 scripts/check-result-line-plainness.py` → exit 0 (the suite's `printf` emitters carry no escape literal, which would otherwise defeat the mutation harness's own `FAIL:` match).

The discriminating case is **D2**: a required tool at `available: false` must produce a non-`ok` label **and** a truthful `1/2 required`. Both operands are asserted, which is precisely what a still-constant renderer cannot satisfy. **D1** pins the other direction with *non-zero* totals, so a renderer that swapped one constant for another (`0/0` + `danger`) fails too — a zero-total assertion would be vacuity of the same family as the bug. **D10** carries the assertion up to the rendered page, where "green dot forever" actually lives, closing the two links a function-level test leaves open: that `health_snapshot` still calls the function, and that the label reaches the dot's CSS class.

## Mutation recipe

```
bash "$HOME/.claude/plugins/marketplaces/managed-service/cogni-service/scripts/mutation-check.sh" --root . --file cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py --expr 's/label = "ok" if req_ok == req_total else "danger"/label = "ok"/' --test 'bash cogni-workspace/tests/test-dashboard-dependency-counts.sh' --case D2
```

Replayed from the repository root. **Recorded verdict: `guard_verified`** — `mutated_result: red`, `restored_result: green`, tree clean afterwards and the commit untouched.

Two notes for whoever replays it. The harness path is not `scripts/mutation-check.sh`: **that file does not exist in this repository.** The two same-named files here (`cogni-consult/`, `cogni-portfolio/`) are bespoke hard-coded harnesses that accept none of these flags — the flag-taking one ships with cogni-service in `managed-service`, which is why the installed-plugin path is recorded. And the `--expr` is fed to `perl -0pi`, never sed or ERE; it is non-global because that string occurs exactly once in the file, which is also why `label = "ok" if req_ok == req_total else "danger"` and the return line under it were deliberately left byte-identical by this change.

## Follow-up issues

- **#1578** — `workspace-dashboard`'s MCP Servers row renders a permanent `0/0 installed (ok)` when the registry's shape drifts.

Same defect class as this issue, different row and different producer (an on-disk JSON registry rather than a subprocess envelope), **reproduced by execution** rather than suspected. It is filed rather than folded in because this issue's committed acceptance contract fences the change to the Dependencies row — widening the diff would have broken the scope boundary this PR is graded against. A survey of all six Health Snapshot rows found it is the *only* exposed sibling, which is also why no shared "guard the empty side before an `X == Y` comparison" helper is proposed yet: at two instances that costs more indirection than it buys.

## Open questions

None blocking.

**Disclosure — runner token scope.** The Step 7.5 least-privilege preflight reports `over_scoped`: observed `[gist, read:org, repo, workflow]` against an allowed set of `[public_repo, repo]`, extra `[gist, read:org, workflow]`. This is the fixed scope set of the GitHub CLI's own OAuth app (`gho_` token), which cannot be narrowed from here. The run proceeded under the `--allow-overscoped` flag per the standing operator decision, and it is disclosed here rather than passed silently.

### DIFF
--- a/cogni-workspace/skills/workspace-dashboard/references/dashboard-sections.md
+++ b/cogni-workspace/skills/workspace-dashboard/references/dashboard-sections.md
@@ -77,7 +77,7 @@ This section is **read-only**. `audit-region-sources` is the dedicated coverage 
 - Environment vars: read `.claude/settings.local.json` `env` object, count how many resolve to existing paths
 - Plugins: count from Section 2
 - Themes: count from Section 3
-- Dependencies: invoke `<workspace-root>/cogni-workspace/scripts/check-dependencies.sh` and parse the JSON envelope (`data.required[]` and `data.optional[]`)
+- Dependencies: invoke `<workspace-root>/cogni-workspace/scripts/check-dependencies.sh` and partition `data.dependencies[]` (one entry per tool: `{name, available, required, version}`) on each entry's own `required` flag; `data.missing_required` / `data.missing_optional` carry the same misses as integers
 - MCPs: count from Section 4
 
 **Output**: one row per check. Each row: green/yellow/red dot, check name, one-line summary (e.g., "12 vars set, 0 missing"), and a "Run `/cogni-workspace:workspace-status` for details" pointer at the section foot.
--- a/cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py
+++ b/cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py
@@ -674,16 +674,19 @@ def _check_dependencies(workspace_root):
         if proc.returncode != 0 or not proc.stdout.strip():
             return ("warning", "check-dependencies.sh failed")
         payload = json.loads(proc.stdout)
-        data = payload.get("data", {})
-        required = data.get("required", {}) or {}
-        optional = data.get("optional", {}) or {}
-        req_ok = sum(1 for v in required.values() if v.get("available"))
-        req_total = len(required)
-        opt_ok = sum(1 for v in optional.values() if v.get("available"))
-        opt_total = len(optional)
+        data = payload.get("data", {}) if isinstance(payload, dict) else {}
+        deps = data.get("dependencies") if isinstance(data, dict) else None
+        entries = [d for d in deps if isinstance(d, dict)] if isinstance(deps, list) else []
+        if not entries:
+            return ("warning", "check-dependencies.sh reported no dependencies")
+        required = [d for d in entries if d.get("required")]
+        optional = [d for d in entries if not d.get("required")]
+        req_total, opt_total = len(required), len(optional)
+        req_ok = sum(1 for d in required if d.get("available"))
+        opt_ok = sum(1 for d in optional if d.get("available"))
         label = "ok" if req_ok == req_total else "danger"
         return (label, f"{req_ok}/{req_total} required, {opt_ok}/{opt_total} optional")
-    except (subprocess.SubprocessError, json.JSONDecodeError, OSError):
+    except (subprocess.SubprocessError, json.JSONDecodeError, AttributeError, TypeError, OSError):
         return ("warning", "check-dependencies.sh errored")
 
 
--- a/cogni-workspace/tests/test-dashboard-dependency-counts.sh
+++ b/cogni-workspace/tests/test-dashboard-dependency-counts.sh
@@ -1,0 +1,210 @@
+#!/usr/bin/env bash
+# Regression test for the workspace-dashboard Health Snapshot's Dependencies row
+# (_check_dependencies in skills/workspace-dashboard/scripts/generate-dashboard.py).
+#
+# Contract under test: the row's counts are derived from the payload
+# check-dependencies.sh actually emits — data.dependencies[], one entry per tool,
+# each {name, available, required, version} — partitioned on each entry's own
+# `required` flag. The row must be FALSIFIABLE: a required tool reporting
+# available:false has to surface a non-ok label and a truthful n/m with n < m.
+# A row that renders a constant count is the defect this suite exists to catch,
+# so the all-present direction is pinned to NON-ZERO totals as well: asserting
+# only "ok" would pass against a permanently-0/0 reader.
+#
+# Every case drives the real renderer against a STUB emitter under its own temp
+# root. That is deliberate and not a convenience: jq and python3 are present on
+# any dev or CI host, so a case running the real script would render "2/2
+# required, ok" and stay green against the broken reader too. No production seam
+# is needed — _check_dependencies derives its script path from its own
+# workspace_root argument, so pointing it at a temp tree is sufficient.
+#
+# Paths are self-located from $0, never the working directory: the repo runner
+# invokes suites with cwd=<repo root> and the mutation harness with cwd=$ROOT,
+# so a cwd-relative path here would be a latent false result.
+#
+# stdlib-only: bash + python3, no pip deps, per the repo-wide script convention.
+
+set -u
+HERE="$(cd "$(dirname "$0")" && pwd)"
+WS_ROOT="$(cd "$HERE/.." && pwd)"
+RENDERER="$WS_ROOT/skills/workspace-dashboard/scripts/generate-dashboard.py"
+SECTIONS="$WS_ROOT/skills/workspace-dashboard/references/dashboard-sections.md"
+EMITTER="$WS_ROOT/scripts/check-dependencies.sh"
+TMPROOT="$(mktemp -d)"
+trap 'rm -rf "$TMPROOT"' EXIT
+
+failures=0
+pass() { printf 'PASS: %s\n' "$1"; }
+fail() { printf 'FAIL: %s\n' "$1"; failures=$((failures + 1)); }
+
+# new_root <name> — build a temp workspace root holding a stub emitter at the
+# exact relative path _check_dependencies derives, and echo the root.
+new_root() {
+  local name="$1"
+  local root="$TMPROOT/$name"
+  mkdir -p "$root/cogni-workspace/scripts"
+  cat > "$root/cogni-workspace/scripts/check-dependencies.sh" <<STUB
+#!/usr/bin/env bash
+cat "$root/fixture.json"
+STUB
+  printf '%s' "$root"
+}
+
+# assert_dep "<label>" "<root>" "<python expr, True to pass>" — the renderer is
+# bound as \`r\` and the returned tuple as \`res\`. A raised exception exits the
+# driver non-zero, so "does not raise" is asserted by the case passing at all.
+assert_dep() {
+  local label="$1" root="$2" expr="$3"
+  if python3 - "$RENDERER" "$root" <<PY
+import importlib.util, sys
+def _load(name, path):
+    spec = importlib.util.spec_from_file_location(name, path)
+    m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m); return m
+r = _load("r", sys.argv[1])
+res = r._check_dependencies(sys.argv[2])
+sys.exit(0 if ($expr) else 1)
+PY
+  then pass "$label"; else fail "$label"; fi
+}
+
+# assert_degrades "<label>" "<root>" — a malformed payload must come back as a
+# warning tuple rather than raising. One definition, so tightening the
+# degradation contract is a single edit rather than three that can diverge.
+assert_degrades() {
+  assert_dep "$1" "$2" 'isinstance(res, tuple) and len(res) == 2 and res[0] == "warning"'
+}
+
+# --- D1: all present. --------------------------------------------------------
+D1ROOT="$(new_root d1)"
+cat > "$D1ROOT/fixture.json" <<'JSON'
+{"success": true, "data": {"dependencies": [
+  {"name": "jq", "available": true, "required": true, "version": "jq-1.7.1"},
+  {"name": "python3", "available": true, "required": true, "version": "Python 3.14.2"},
+  {"name": "curl", "available": true, "required": false, "version": "curl 8.7.1"},
+  {"name": "git", "available": true, "required": false, "version": "git 2.50.1"},
+  {"name": "bc", "available": true, "required": false, "version": "bc 7.0.3"}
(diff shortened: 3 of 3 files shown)


claude-cogni-work:

Author drain (`service-resume --role author`, two-identity Mode B) took #1573 end-to-end to a draft PR. Handing off to the merger identity for the independent commit-first review + merge — **not** self-reviewing, per "a bot never reviews its own work."

`svc:needs-review` applied to this PR; #1573 marked `svc:resolved`. The merger drain / `service-gatekeeper` consumes the label from here.

### Blocking pre-handoff preflight

`check-preflight.sh` exit **0**, clean — 7 discovered / 4 passed / 0 failed / 3 skipped, 7 instruments resolved, `metric_version 3`.

- **passed:** `run-tests` (suites green for cogni-workspace, v1) · `frontmatter` (clean, v4) · `version-bump` (no version-line touch, v2) · `mutation-recipe` (runnable as written — `D2` resolved in the named suite)
- **skipped, each with its reason:** `breadcrumbs` — `not_applicable`, no touched SKILL.md or agents/*.md · `skill-length` — `not_applicable`, no touched SKILL.md · `agent-length` — `not_applicable`, no touched agents/*.md

`--root` was pointed at a checkout **of this branch's head commit** rather than the shared clone, so the `run-tests` arm graded this branch's 16 suites rather than base's 15.

### Author-side contract grading (advisory — not a review verdict)

Graded **15** checklist items from the issue-time contract (`origin: triage`, derived `2026-08-22T22:57:22Z`, confidence `high`, `holds: false`) — the author checking its own diff against a specification it did not write, committed before any branch was cut.

**Satisfied: 15. Violated: 0** (fixed in-session: 0; residual: 0). **Unclear: 0.**

Nothing was auto-fixed and there is **no residual** for you to inherit.

The grader executed the evidence the contract names rather than reading the diff for it — the two zero-changed-lines assertions, the four greps, the full suite run — and independently reproduced the PR's central claim by materialising `origin/main` into a scratch tree and running the suite there unchanged: **9 of 10 cases red at base**, D9 correctly green because it pins the emitter contract this PR does not touch. It also cleared all eleven `risk_watchlist` entries.

### Two things worth your attention as an independent reviewer

1. **The contract's item 15 is green at base.** `python3 scripts/run-plugin-tests.py --filter cogni-workspace` measured 15/15 before this branch existed and 16/16 after, so it cannot discriminate a correct fix from no fix at all. It is a regression guard only. The discriminating evidence is item 11's base-vs-head execution and the mutation recipe — please do not let a green sweep stand in for either.
2. **The suite carries ten cases where the contract anticipated nine.** `D10` is additive: it carries the assertion up through the real renderer to the rendered HTML, where "green dot forever" — the symptom the issue actually names — is produced. It was added after an altitude review flagged that a function-level test leaves two links unpinned (that `health_snapshot` still calls the function, and that the label reaches the dot's CSS class).

### Filed follow-up

**#1578** — the MCP Servers row of the same `health_snapshot` function carries the identical defect class against a different producer, reproduced by execution. It is fenced out of this diff by the contract's own item-1 scope boundary, so it was filed rather than folded in.

### Disclosure — runner token scope

The Step 7.5 least-privilege preflight reports `over_scoped`: observed `[gist, read:org, repo, workflow]`, extra `[gist, read:org, workflow]`. That is the fixed scope set of the GitHub CLI's own OAuth app and cannot be narrowed from here; the run proceeded under the `--allow-overscoped` flag per the standing operator decision, disclosed rather than passed silently.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1579

**Summary:** Both perspectives passed all 15 committed review-contract items, with only two minor findings remaining — one a merger-added file-permission preference (demoted to advisory, non-gating) and one an evidence-completeness gap on the item-14 mutation recipe that the validator independently confirmed does not indicate a vacuous guard — so the PR is safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/tests/test-dashboard-dependency-counts.sh` — The new suite lands at mode 0644 while 13 of the 16 suites in the same directory are 0755. This has no effect on any gate — run-plugin-tests.py invokes every suite as `bash <path>` precisely because 'a handful lack the executable bit while carrying a valid shebang' (its own docstring), the recorded mutation --test also carries a `bash ` prefix, and two existing siblings (test-theme-backcompat.sh, test-wiki-namespace-sync.sh) are 0644 too. The only cost is to a human replaying it as ./cogni-workspace/tests/... who gets a permission error despite the shebang.
  - *Suggested fix:* chmod +x cogni-workspace/tests/test-dashboard-dependency-counts.sh to match the directory's majority convention. Purely cosmetic — do not hold the PR for it.

- **[minor]** `review-contract:item-14` — The recorded `## Mutation recipe` satisfies the item's mechanical half but not its head clause. All five flags are present as exact tokens with values, no forbidden shell metacharacter appears in any flag value, `--test` correctly carries the `bash ` prefix, and `--case D2` does reproduce the printed label (mutation-check.sh:412's matcher is `^[ \t]*FAIL:[ \t]+<case>(?:[ \t]|$)`, so the bare token D2 matches `FAIL: D2 absent required tool renders non-ok with a truthful n of m` at the whitespace boundary — NOT a case_not_found). What the item asks for is an `--expr` that reverts the NEW derivation. The recorded expression instead rewrites `label = "ok" if req_ok == req_total else "danger"` to `label = "ok"` — a line the diff left byte-identical to base. So the mutation exercises the status expression, which this PR did not change, and never touches the data.get("dependencies") partition at :678-685, which is the whole of the change. It also grades only one operand of D2's conjunction: with the label forced to "ok", D2's count assertion `res[1] == "1/2 required, 3/3 optional"` still holds, so the recorded replay leaves the count derivation formally unproven. This is a shortfall in the EVIDENCE, not a demonstrated fault: a true derivation-reverting mutation was simulated in a scratch copy (`req_total, opt_total = len(required), len(optional)` -> `len(entries), len(entries)`, the exact untruthful-partition shape on the risk watchlist) and D1, D2, D3 and D10 all went RED, so D2 genuinely does have teeth over the derivation and the guard is not vacuous.
  - *Suggested fix:* Re-record the recipe with an `--expr` that reverts the derivation itself, and paste the fresh verdict. A replay already confirmed to go RED on D2 and GREEN on restore: `--expr 's/req_total, opt_total = len\(required\), len\(optional\)/req_total, opt_total = len(entries), len(entries)/'` (single-quoted, perl -0pi semantics, literal parens backslash-escaped, non-global since the string occurs once). Keep `--case D2` unchanged. Alternatively keep the existing recipe and ADD this second one — two recipes are permitted and would pin both operands of D2.

### Positive observations

- All 15 committed review-contract items satisfied by cited, executed evidence — the dependency-count fix (data.get("dependencies") partitioned on the required flag), the 2-tuple return contract, and degraded-string preservation are all verified at head. (plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer)
- Non-vacuity of the new suite independently reproduced twice: a base-vs-head replay shows 9-of-10 cases (D1-D8, D10) genuinely RED at origin/main and GREEN at head; only D9 (an emitter-shape guard) is expected green at base. (plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer)
- D1 and D2 jointly close the constant-0/0 and untruthful-partition vacuity directions on the risk watchlist — both operands of each assertion are exercised against stub fixtures, never the real check-dependencies.sh. (plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer)
- Item 8's half-fixed-documentation risk is closed repo-wide, not just on the one edited line — a whole-cogni-workspace grep for the old field names returns zero hits. (plugin-dev:plugin-validator, cogni-service:contract-grading-reviewer)
- Structural registration health intact: manifest untouched, no skill/agent registration surface moved, check-version-bump confirms 1 touched/0 violations, and the new suite is correctly placed for the non-recursive test-discovery glob (16/16/0 on the full sweep). (plugin-dev:plugin-validator)
- Suite is hermetic and self-locating: stdlib bash+python3 only, no network, all writes under one mktemp -d with a cleanup trap, and it runs correctly from both repo-root and its own directory. (plugin-dev:plugin-validator)
- Full risk_watchlist swept and clear: no degraded-permanent-green regression, no untruthful partition, no unnecessary production seam, no inverted emitter constraint, no new unhandled raise path, correct discoverable suite depth. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1580, opened 2026-08-23
cogni-workspace: the workspace-dashboard MCP install probe resolves the wrong directory when desktop_config_key differs from the installed directory name


claude-cogni-work:
## Problem

`mcp_install_status` (`cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py`)
resolves a git-type MCP server's install probe from the registry's `desktop_config_key`:

```python
candidate = os.path.expanduser(f"~/.claude/mcp-servers/{desktop_key}/start.sh")
```

But `cogni-workspace/references/mcp-git-registry.json` declares
`"desktop_config_key": "excalidraw"` for its `mcp_excalidraw` entry, while the directory the
server is actually installed into is `~/.claude/mcp-servers/mcp_excalidraw/`. The probe therefore
looks for a path that does not exist and reports a correctly-installed server as `missing`.

## Evidence

Executed against `origin/main` (`fb393c3d`) with no changes applied, on a host where **both**
registry servers are present:

```
mcp_excalidraw -> missing   | probe: ~/.claude/mcp-servers/excalidraw/start.sh
pencil         -> installed | probe: /Applications/Pencil.app/.../mcp-server-darwin-arm64
MCP Servers row -> {'label': 'warning', 'summary': '1/2 installed'}
```

`ls ~/.claude/mcp-servers/` returns exactly one entry: `mcp_excalidraw`.

## Why this is filed separately

Found while resolving #1578 (the MCP Servers row rendering a permanent `0/0 installed (ok)` on
registry shape drift). It is a different function with a different failure mode, and #1578's
committed acceptance contract puts `mcp_install_status`'s path resolution explicitly out of
scope — fixing it changes which servers report installed on every host, and it carries a
producer question (registry key, installer, or probe?) that #1578 does not answer.

It is also why #1578's own acceptance criterion 3 — "the live `mcp-git-registry.json` (2 servers)
still renders `2/2 installed` with an `ok` label" — is **false at base** and could not be
satisfied literally inside that issue's scope, and why no assertion in the suite #1578 shipped
depends on host availability.

## Acceptance criteria

- [ ] The registry's excalidraw entry and the on-disk `mcp_excalidraw/` directory resolve to the
      same path, so a correctly-installed git-type MCP server is reported `installed` rather than
      `missing` — fixed on whichever side is authoritative (registry key, installer, or the
      probe's resolution), with the choice stated at the resolution site
[...]

PR #1581, opened 2026-08-23
cogni-workspace: guard MCP Servers row against registry shape drift

claude-cogni-work:
Refs #1578.

## Summary

The Health Snapshot's MCP Servers row built `mcp_servers` from `registry.get("servers", {})` — a
defaulted read of a cross-file producer whose shape nothing verified. A registry that parsed but
carried no `servers` key yielded `[]`, not `None`, so the `mcp_servers is None` guard was bypassed
and `installed + manual == total` became a permanent `0 == 0`. The row rendered `0/0 installed`
under a green dot.

- New `resolve_mcp_servers(workspace_root)` returns `(servers, status)` over four states —
  `missing` / `malformed` / `empty` / `ok`.
- `health_snapshot`'s MCP arm branches on that status. The count arm is reachable **only** on
  `ok` with entries, so an empty list or an unrecognized status degrades instead of falling
  through to a count. Three distinguishable non-ok summaries: `registry not found`,
  `registry unreadable`, `no servers configured`.
- The registry test is `is None`, not a truthiness test. The old `if registry:` masked a
  falsy-but-present body (`[]`, `""`, `0`, `false`) as a missing file — the same bug class one
  step over.
- Three shapes that previously **raised** now degrade: a non-dict root, `servers` holding a list,
  and a non-mapping entry. All three raised *before* `render_html` and the file write, so the CLI
  exited 1 having written no HTML and printed no JSON — contradicting `dashboard-sections.md`'s
  "never error out without writing the file".
- New suite `cogni-workspace/tests/test-dashboard-mcp-counts.sh` (M01–M14, M05 intentionally
  unused). Suite count 16 → 17.
- Doc twin updated in three places; §7's MCP line was `- MCPs: count from Section 4`, naming no
  field, no shape and no partition rule.

## Evidence

| check | result |
|---|---|
| `run-plugin-tests.py --filter cogni-workspace` | **17 discovered / 17 passed / 0 failed** (base: 16/16/0) |
| new suite against the **base** renderer | `rc=1` (red) |
| new suite against this branch | `rc=0` (green) |
| mutation replay (below) | `guard_verified` |

The suite-count rise is the load-bearing assertion, not `0 failed` — the baseline is already
`0 failed`, so that alone is green at base and discriminates nothing.

## Mutation recipe

```
bash "$HOME/.claude/plugins/marketplaces/managed-service/cogni-service/scripts/mutation-check.sh" --root . --file cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py --expr 's/declared = registry\.get\("servers"\)/declared = registry.get("servers", {})/' --test 'bash cogni-workspace/tests/test-dashboard-mcp-counts.sh' --case M02
```

Recorded verdict on replay: `guard_verified` — `mutated_result: red` (M02), `restored_result:
green`, tree clean afterwards. Replayed again after the final edit to the renderer.

Three notes for whoever replays it:

- **The mutated line is one this diff introduces.** `declared = registry.get("servers")` is new
  code inside `resolve_mcp_servers`; the string it mutates *to* is the old inline read this PR
  deletes. Reverting it restores the exact pre-fix semantics — a missing `servers` key defaults
  to `{}`, classifies `empty` rather than `malformed`, and M02's
  `summary == "registry unreadable"` operand goes red.
- **The harness is cogni-service's, not this repo's.** The two same-named
  `scripts/mutation-check.sh` files here (`cogni-consult/`, `cogni-portfolio/`) are bespoke
  zero-argument harnesses that accept none of these flags. The unversioned marketplace path above
  was resolved on the authoring host and is byte-identical to the loaded copy; the version-pinned
  `cache/managed-service/cogni-service/<version>/` paths lag the marketplace.
- The `--expr` is fed to `perl -0pi`, never sed, and is deliberately non-global:
  `registry.get("servers"` occurs exactly once in the file both before and after this change.

## Scope decisions

**Acceptance criterion 3 is false at base — graded as a shape assertion, not as its literal text.**
The issue asks that the live registry "still renders `2/2 installed` with an `ok` label". Executed
against `origin/main` with no changes applied, it already renders `1/2 installed` + `warning`: the
registry declares `desktop_config_key: "excalidraw"` while the installed directory is
`mcp_excalidraw/`, so `mcp_install_status` probes a path that does not exist. The criterion is
both host-dependent and currently false for a reason this issue explicitly excludes — filed as
#1580. **Grading note:** a contract item restating AC-3 verbatim cannot be met by a `2/2 installed`
grep, and could not be met by any change inside this issue's scope. M13 discharges its intent the
way the sibling suite's D9 discharges its emitter contract: shape only, never host availability.

**`render_mcp` is unchanged, and `malformed` deliberately collapses onto `servers is None`** —
the "deliberately collapsed with a stated reason" branch of acceptance criterion 2, taken at the
one surface where there are no cards to draw either way. The Health Snapshot row is where the
states are told apart, because that is the surface that made the false claim.

**`health_snapshot`'s sixth parameter is required, with no default.** Verified repo-wide at
`origin/main`: `:1253` is the only caller and no suite or script calls it directly. A default
would let a future caller silently reintroduce the unguarded path.

**No shared zero-count helper.** #1573 fixed the same class for the Dependencies row. Two
instances is not enough to factor; a third would be — the issue says so and this honours it.

**No `plugin.json` version bump** — the patch bump is applied post-merge.

## Open questions

- **Section 4's placeholder now states something false for a malformed registry.** Because
  `malformed` reuses `servers is None`, `render_mcp` emits *"MCP registry not found in this
  workspace."* — but the file **is** there, and the remedy is to repair it, not create it. The
  page contradicts itself: the health row says `registry unreadable` while Section 4 says the file
  is absent. Fixing it means threading the status into `render_mcp` and selecting the string,
  which is a one-line change to a function whose call site this PR already touches — but it
  changes behaviour the plan and the committed contract both fixed as out of scope, so it was
  deliberately **not** applied here. Worth a maintainer's call on whether to fold it in or file it.
- **The stdout JSON still collapses all four states to `0`** (`"mcp_servers": len(mcp_servers) if
  mcp_servers else 0`). That key is the CLI contract documented in
  `skills/workspace-dashboard/SKILL.md`, outside this issue's scope boundary — but it means a
  machine consumer still reads the false-health signal the HTML row just stopped emitting.
- **Two other readers of the same registry remain unguarded** —
  `cogni-workspace/scripts/patch-desktop-config.py` and
  `cogni-workspace/tests/test-mcp-declaration-hygiene.sh` both do the same bare
  `registry.get("servers", {})`, and nothing validates the file producer-side. The repo now has
  one hardened reader and two soft ones, and no single place stating what the registry's shape is.

## Disclosure

The runner token is **over-scoped** relative to the documented least-privilege posture —
`check-token-scope.sh --strict` reports `status: over_scoped`, extra scopes `gist`, `read:org`,
`workflow`. This is the fixed scope set of the GitHub CLI OAuth (`gho_`) token and cannot be
narrowed; the run proceeded under the authorized `--allow-overscoped` flag, disclosed here per
policy.

## Follow-up issues

- #1580 — the MCP install probe resolves the wrong directory when `desktop_config_key` differs
  from the installed directory name.

### DIFF
--- a/cogni-workspace/skills/workspace-dashboard/references/dashboard-sections.md
+++ b/cogni-workspace/skills/workspace-dashboard/references/dashboard-sections.md
@@ -45,6 +45,7 @@ Per-section reference for `workspace-dashboard`: data source, helper(s) reused, 
 **Data source**:
 - `<workspace-root>/cogni-workspace/references/mcp-git-registry.json` — declares each server (`type`, `repo`, `desktop_config_key`, `provides_tools[]`, `required_by[]`, platform-specific paths for native servers)
 - Install status check: existence of `~/.claude/mcp-servers/<desktop_config_key>/start.sh` for git-based servers; existence of the platform-specific binary path for native servers
+- Required shape: a top-level `servers` object whose values are themselves objects. A registry that parses but does not match it — root not an object, `servers` absent or not an object, or any entry not an object — is **unreadable**. This section then falls back to the same placeholder a missing registry gets, because either way there are no cards to draw; the Health Snapshot row is where the two are told apart.
 
 **Output**: card per server. Status pill (Installed / Missing / Manual). Type pill (git / native). Required-by chips (one per consuming plugin). For git-based: repo URL truncated. For native: platform path.
 
@@ -78,7 +79,7 @@ This section is **read-only**. `audit-region-sources` is the dedicated coverage 
 - Plugins: count from Section 2
 - Themes: count from Section 3
 - Dependencies: invoke `<workspace-root>/cogni-workspace/scripts/check-dependencies.sh` and partition `data.dependencies[]` (one entry per tool: `{name, available, required, version}`) on each entry's own `required` flag; `data.missing_required` / `data.missing_optional` carry the same misses as integers
-- MCPs: count from Section 4
+- MCPs: read the `servers` object from `<workspace-root>/cogni-workspace/references/mcp-git-registry.json` and partition its entries on each entry's own resolved install status. The row renders exactly one of four summaries: `registry not found` (file absent, unparseable, or a literal JSON null), `registry unreadable` (parses, but not the required shape above), `no servers configured` (`servers` present and empty), or `<installed>/<total> installed[, <n> manual]`. All three zero-server states are non-`ok`, so the row can never report a green `0/0` — a count comparison both sides satisfy at zero is not evidence of health.
 
 **Output**: one row per check. Each row: green/yellow/red dot, check name, one-line summary (e.g., "12 vars set, 0 missing"), and a "Run `/cogni-workspace:workspace-status` for details" pointer at the section foot.
 
@@ -90,6 +91,7 @@ The dashboard intentionally does **not** re-implement the diagnostic depth of `w
 - **`discover-plugins.sh` fails**: fall back to glob mode and add a yellow note in Section 7's Plugins line.
 - **Theme parsing fails**: render the card with a "unparseable theme" badge and skip swatches.
 - **MCP registry missing**: skip Section 4 entirely with a placeholder note.
+- **MCP registry shape drift**: a registry that parses but does not match the required shape renders Section 4's placeholder and a non-`ok` Health Snapshot row naming which state it is. The run still exits 0 and still writes the HTML, per the rule below — a malformed registry must degrade, never abort before the page exists.
 - **Missing market catalogs**: render the matrix with empty columns for the missing plugins.
 
 In every failure mode the script must still produce *some* HTML — never error out without writing the file. The user's first question after a failure should be answerable by opening the dashboard.
--- a/cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py
+++ b/cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py
@@ -531,6 +531,54 @@ def _command_on_path(cmd):
     return False
 
 
+# Every registry state that is not a countable list of servers. Kept beside the resolver
+# that produces the statuses so the two cannot drift apart.
+MCP_NON_OK_SUMMARY = {
+    "missing": "registry not found",
+    "malformed": "registry unreadable",
+    "empty": "no servers configured",
+}
+
+
+def resolve_mcp_servers(workspace_root):
+    """Resolve the MCP registry into dashboard rows plus a status naming what was found.
+
+    Returns (servers, status) with status one of "missing" / "malformed" / "empty" / "ok".
+    The states themselves are documented in references/dashboard-sections.md; three
+    properties here are contracts a caller must not break:
+
+    - The "missing" test is an IDENTITY test, never a truthiness test. A falsy-but-present
+      body (`[]`, `""`, `0`, `false`) is malformed, not absent, and reporting it as absent
+      is the same class of bug this function exists to close.
+    - "malformed" returns None (not []) for the server list, so `render_mcp`'s existing
+      two branches stay correct without change.
+    - One bad entry makes the whole registry malformed rather than being skipped. A partial
+      count read from a file the reader does not understand is the failure being fixed, so
+      degrading loudly beats reporting a subset as if it were the whole.
+    """
+    registry = load_mcp_registry(workspace_root)
+    if registry is None:
+        return (None, "missing")
+    if not isinstance(registry, dict):
+        return (None, "malformed")
+    declared = registry.get("servers")
+    if not isinstance(declared, dict):
+        return (None, "malformed")
+    if not declared:
+        return ([], "empty")
+    servers = []
+    for key, server in declared.items():
+        if not isinstance(server, dict):
+            return (None, "malformed")
+        entry = dict(server)
+        entry["name"] = entry.get("name", key)
+        status, hint = mcp_install_status(entry)
+        entry["install_status"] = status
+        entry["install_hint"] = hint
+        servers.append(entry)
+    return (servers, "ok")
+
+
 # ---------------------------------------------------------------------------
 # Markets matrix
 # ---------------------------------------------------------------------------
@@ -609,7 +657,7 @@ def load_hooks(workspace_root):
 # Health snapshot
 # ---------------------------------------------------------------------------
 
-def health_snapshot(workspace_root, foundation, plugins, themes, mcp_servers):
+def health_snapshot(workspace_root, foundation, plugins, themes, mcp_servers, mcp_status):
     """Build a 6-row snapshot mirroring workspace-status categories."""
     found_required_ok = all(f["exists"] for f in foundation if f["required"])
     foundation_label = (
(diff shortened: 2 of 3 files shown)


claude-cogni-work:
## Ship-ready

Authored by the autonomous author instance and handed off for independent review. Draft, carrying
`svc:needs-review`. No review verdict was posted here and the PR was not flipped ready — both
belong to the merger instance.

### Handoff preflight

`check-preflight.sh` exit **0** — clean, 4 passed / 3 skipped / 0 advisory.

| instrument | result | note |
|---|---|---|
| `run-tests` | pass | suites green for cogni-workspace (v1) |
| `frontmatter` | pass | clean for cogni-workspace (v4) |
| `version-bump` | pass | no version-line touch (v2) |
| `mutation-recipe` | pass | recorded recipe runnable as written — `M02` found in the suite |
| `breadcrumbs` | skip | `not_applicable` — no touched `SKILL.md` or `agents/*.md` |
| `skill-length` | skip | `not_applicable` — no touched `SKILL.md` |
| `agent-length` | skip | `not_applicable` — no touched `agents/*.md` |

All three skips are `not_applicable` — there was genuinely nothing for them to examine, not a
silent pass.

Graded against the `review_contract` committed to issue #1578 at triage, before any branch
existed. **13 items graded · 13 satisfied · 0 violated · 0 unclear.**

**Residual carried to the merger: none.**

The self-grade is advisory and does not substitute for independent review — it grades the diff
against a bar the author wrote against, which is exactly the thing an independent reviewer is
supposed to check from outside. Three items are worth the merger's own eyes because grading them
required going past the literal contract text:

- **Item 10 / acceptance criterion 3.** The issue asks that the live registry "still renders
  `2/2 installed` with an `ok` label". That premise is **false at base** — executed against
  `origin/main` with no changes applied, the row already renders `1/2 installed` + `warning`,
  because the registry declares `desktop_config_key: "excalidraw"` while the installed directory
  is `mcp_excalidraw/`. The committed contract records this and restates the criterion
  host-independently. **A contract item re-derived from AC-3's literal text cannot be satisfied by
  this or any in-scope change**; the separate defect is filed as #1580.
- **Item 12 / the mutation harness.** The contract as written at triage named "the repo's copy
  under `cogni-consult/scripts/` or `cogni-portfolio/scripts/`". That was wrong — those two are
  bespoke zero-argument harnesses accepting none of these flags; the flag-taking one ships with
  cogni-service. The recorded recipe uses the correct harness and replays `guard_verified`.
- **Item 1 / the suite filename.** The contract gave a filename as an example; grade the property
  (one level under `cogni-workspace/tests/`, reachable by the runner's non-recursive globs), not
  the example name.

### Known gap, deliberately not fixed

Section 4 still renders *"MCP registry not found in this workspace."* for a registry that **is**
present but malformed, because the `malformed` state deliberately collapses onto `servers is None`
so `render_mcp` stays untouched. The page therefore contradicts itself: the health row says
`registry unreadable` while Section 4 says the file is absent. This was surfaced by the pre-commit
quality pass and **skipped on purpose** — fixing it changes behaviour that the plan record and the
committed contract both fixed as out of scope. It is in the PR body's `## Open questions` for a
maintainer's call.

### Disclosure

The runner token is over-scoped (`gist`, `read:org`, `workflow`) — the fixed scope set of the
GitHub CLI OAuth token, which cannot be narrowed. The run proceeded under the authorized
`--allow-overscoped` flag. Recorded in the PR body.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1581

**Summary:** All 13 committed review-contract items are confirmed satisfied by direct execution at the PR head by both perspectives (contract-grading-reviewer raised no findings at all). The three residual findings from plugin-validator are all minor: two are explicitly waived with justification (a pre-existing UnicodeDecodeError gap outside the contract's enumerated scope, and a permitted placeholder collapse under contract item 8), and the third — an under-specified test needle — is an un-waived advisory minor that does not gate under an all-pass verdict. Safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|
| plugin-validator | ✅ pass |

### Findings (3)

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py:1017` — Because malformed deliberately returns None for the server list, render_mcp emits the literal placeholder 'MCP registry not found in this workspace.' for a registry file that is present on disk — a user-facing statement that is false in that state, and one that contradicts the Health Snapshot row saying 'registry unreadable'. This is permitted rather than violated: contract item 8 explicitly allows the collapse 'with the reason written at the guard site', and the reason IS written in both the resolve_mcp_servers docstring and dashboard-sections.md. The author also self-surfaced it as the first Open question. Recorded as a graded observation, not a bar failure.
  - *Justification:* Contract item 8 permits this collapse when the reason is stated at the guard site, and it is stated in two places; the PR body raises it as an Open question for a maintainer's call. No user-observable health signal is wrong.

- **[minor]** (waived) `cogni-workspace/skills/workspace-dashboard/scripts/generate-dashboard.py:504` — A fourth route still violates dashboard-sections.md:95's 'never error out without writing the file' contract that this PR's hardening pass invokes: load_mcp_registry catches (json.JSONDecodeError, OSError), but UnicodeDecodeError is a subclass of ValueError, not of OSError, so a registry file holding invalid UTF-8 raises straight past both arms. Verified: a registry beginning \xff\xfe makes the CLI exit 1 with an uncaught UnicodeDecodeError and write no HTML at all. This is PRE-EXISTING and identical at origin/main, so it is not a regression, and it is outside the contract's enumerated three adjacent shapes. The diff's new prose is correctly scoped and does not overclaim: both new bullets say 'a registry that PARSES but does not match the required shape', and an invalid-UTF-8 file does not parse.
  - *Justification:* Pre-existing at base, not introduced or worsened by this diff, and explicitly outside contract item 11's enumerated shape set and item 13's scope boundary. Fixing it here would widen an S3-minor change.

- **[minor]** (auto-fixable) `cogni-workspace/tests/test-dashboard-mcp-counts.sh:161` — M06/M07/M08's only content needle is `health-dot warning`, which is not MCP-row-specific and is already satisfied by unrelated rows. Confirmed by rendering the [1,2] fixture and extracting every health row: Foundation renders danger, Plugins renders warning (0 plugins discovered), Dependencies renders warning (check-dependencies.sh unavailable). So a reader that wrote HTML but still rendered the MCP row green would pass all three cases on the Plugins row alone. The cases' load-bearing property (exit 0 + non-empty HTML + no traceback) IS soundly pinned by assert_cli's shared baseline and does fail at base, so this is a weak sub-assertion inside a real guard, not a vacuous case; the MCP-row-non-ok claim for the same malformed status is separately pinned by M02 and M10. Contract item 11's property itself is confirmed satisfied by direct execution.
  - *Suggested fix:* Use the MCP-specific needle the suite already uses one case later: replace 'health-dot warning' with 'registry unreadable' on M06/M07/M08 (M10 does exactly this). That operand cannot be satisfied by any other row, and the branch already emits that exact string for all three shapes.

### Positive observations

- Item 2 with the load-bearing count: run-plugin-tests reports 17/17/0; git ls-tree origin/main counts 16 test files at base, confirming the rise independently of the PR body's claim. (plugin-validator)
- Item 4 measured base-vs-branch on the same fixture: base produces exactly the recorded defect {'label': 'ok', 'summary': '0/0 installed'}; branch produces warning / registry unreadable. (plugin-validator)
- Item 6 replayed independently in a scratch tree so the worktree was never mutated: suite exits non-zero with 11 failures including M02. M06/M07/M08 fail at base with 'out.html: No such file or directory' — the raise-before-write defect showing itself. (plugin-validator)
- Item 12 replayed: the mutated line is one the DIFF INTRODUCES, so the hollow-guard_verified defect recorded on sibling PR #1579 is avoided; the non-global substitution is safe because the needle occurs exactly once. (plugin-validator)
- Item 10 satisfied host-independently and the falsified AC-3 was handled correctly rather than worked around: no case anywhere in the suite depends on ~/.claude/mcp-servers contents. (plugin-validator)
- The count arm is structurally unreachable at zero rather than arithmetically safe at zero, which is the right shape for this defect class. (plugin-validator)
- M12 asserts the WHOLE deleted loop line as an absence rather than the bare substring, precisely so the recorded mutation's own footprint cannot be confused with a genuine regression. (plugin-validator)
- Restoring origin/main's renderer wholesale and re-running the suite produces 11 failures, exit 1, naming M02 — genuinely red against the unfixed reader. (cogni-service:contract-grading-reviewer)
- Reproduced the exact base defect directly: for the M02 fixture the base reader returns {'label': 'ok', 'summary': '0/0 installed'}. (cogni-service:contract-grading-reviewer)
- Executed base vs branch health_snapshot against the live 2-server registry: both produce identical {'label': 'warning', 'summary': '1/2 installed'} — confirming healthy-path preservation independently of M13's narrower assertion. (cogni-service:contract-grading-reviewer)
- mcp_install_status is byte-identical base vs branch; mcp-git-registry.json is untouched; scope boundary holds. (cogni-service:contract-grading-reviewer)
- The PR body's Open questions honestly discloses the Section-4 placeholder tension rather than hiding it. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1576, opened 2026-08-21
cogni-knowledge: tests/README.md states both arms must share an id but not that both must be REACHABLE, so the split-arm shape keeps recurring


claude-cogni-work:
## Plugin
cogni-knowledge

## Current behavior
`cogni-knowledge/tests/README.md` §"Case ids on result lines" carries six numbered rules. Rule 5 says **"Both arms of one case carry the same id."** Nothing states that both arms must be **reachable**, or that a case's arm must be gated only on that case's own predicate.

That gap is what #1517 kept finding instances of. Two distinct shapes satisfy rules 1–6 as written while still leaving `--case <id>` unable to get a verdict:

1. **An `elif` chaining two different cases**, leaving an input state where neither arm prints. `test_plain_result_emitters.sh:164-169` was this: at `exercisers >= 50 && shape_errors > 0` neither `plain-emit-05` nor `plain-emit-06` was emitted.
2. **A PASS gated on a suite-wide accumulator.** `test_skill_contracts.sh:298` gated `skill-contracts-84`'s only arm on the shared `errors` counter, which seven earlier checks increment — so any unrelated failure silenced the case. Verified at base: break `skill-contracts-01` and `skill-contracts-84` emits **zero** occurrences.

Both report `case_not_found` (or a laundered green), and rules 1–6 read as satisfied throughout.

## Desired behavior
A rule 7 recording the property, so the next hand-rolled emitter does not reproduce the shape:

> **Both arms of a case must be reachable, and each is gated only on that case's own predicate.** Rule 5 is not sufficient — a FAIL emitted alone inside a loop with no `else` can never print green, and a PASS hanging off the suite-wide `errors` counter is silenced by an unrelated failure. In both shapes `--case <id>` reports `case_not_found` and the harness proves nothing.

It should name the two shapes to avoid, the one legitimate exception (a not-found / preflight guard that aborts the run), and the **collect-then-pair** idiom for a multi-file scan — accumulate offenders into one variable during the loop, then gate a single same-id `if/else` on it after. That idiom is already resident at `test_skill_contracts.sh` (`skill-contracts-106`, `-107`) and is the shape #1517 converted `-82`/`-83`/`-84` to.

It should close with the verification note: reachability is invisible in a green run, so force each branch (an argument-driven scan root, a fixture) or point `mutation-check.sh --case <id>` at it.

## Motivation
[...]

Issue #1499, opened 2026-08-18
repo: settle whether a CI guard should enforce per-suite case-id uniqueness, and record the outcome


claude-cogni-work:
## Plugin
repo (root `scripts/` and `CLAUDE.md`)

## Epic context
Part of the tracker "repo: distinct test cases share a first token, so --case can name the wrong case" (#1484), which stays OPEN when this child lands. This is the last child: it decides how the property the other eight children establish is *kept*, and it depends on all of them because a uniqueness guard introduced while suites are still colliding would ship red on day one.

## Context
Once every suite emits unique first tokens, nothing stops the property regressing. The next contributor who copies an emitter, adds a loop, or reuses a case number reintroduces exactly the silently-wrong green the tracker describes — a mutation reds a sibling case, the harness matches a `FAIL:` line, and `guard_verified` is recorded for an unexercised guard.

There are two candidate remedies and a real argument against the stronger one.

**Argument against a static guard.** Labels arrive as runtime-interpolated arguments — loop variables, `basename`s, interpreter paths, and in cogni-knowledge lines emitted from Python inside heredocs. Uniqueness is therefore not statically decidable, and a guard reading only literal bash call sites would ship green while blind to precisely the carriers the census found by execution. That is a worse outcome than no guard: it would look like coverage.

**What is decidable.** The census method used to scope this whole tracker — run the suite, group the first tokens it actually emits — *is* decidable at runtime and could back an execution-based guard, at the cost of running every suite twice in CI.

**A constraint on where it may not live.** `scripts/check-result-line-plainness.py` must not absorb this. Its own docstring (`:29-32`) states that its second arm keys on the escape rather than on a canonical emitter body, because suites here legitimately vary — some `printf`, some `echo`, some label passes `ok:` — so requiring one blessed body shape "would fail dozens of clean files while proving nothing about what they emit". A uniqueness rule bolted onto that guard would contradict its stated design.

**The alternative.** A `CLAUDE.md` Contributing rule enforced by convention plus review, as the repo already does for the own-emitter-versus-foreign-emitter locale rule — which is deliberately guard-less for a structurally similar reason (the distinction is not statically decidable and a checker would fire on every legitimate case).

## Question
Should per-suite case-id uniqueness be enforced by a CI guard, and if so, must that guard be execution-based? Or should it be a documented Contributing convention enforced by review, like the locale rule?

This child **records** the maintainer's decision; it does not make it. A ruling comment on this issue is the input to implementation.

[...]

PR #1590, opened 2026-08-23
Mechanical guard for FAIL-only case ids

### LINKED ISSUE #1584: cogni-knowledge: nothing mechanically detects a FAIL-only case id, so the class keeps being found by hand
## Plugin
cogni-knowledge (guard lands at repo root; scope spans every plugin's `tests/`)

## Current behavior
Nothing mechanically detects an **unpaired case id** — a `FAIL: <id>` arm with no reachable same-id `PASS:`/`ok:` arm. Such a case emits nothing on a green run, so `mutation-check.sh --case <id>` reports `case_not_found` on a genuinely red run and the guard proves nothing.

The class has now been found and hand-fixed **twice** — #1575 (`plain-emit-03` / `plain-emit-04`) and #1582 (`klib-01`) — both off **manual censuses**, not off any check. #1575 even asserted it was clearing "the last split-case-id instances left in cogni-knowledge"; the #1582 census falsified that a day later. That is the signature of a class with no mechanical floor under it.

It is documented only in prose, at `cogni-knowledge/tests/README.md` rule 6, whose own stated limitation is that red arms must be paired *"by reading the source"*.

**Verified non-duplicate** at `origin/main` @ `e79c4163` by reading the code:
- `scripts/check-result-line-plainness.py` has exactly two arms (`escape_literal`, `emitter_not_plain`) — both escape-literal only; it never looks at ids.
- `cogni-knowledge/tests/test_plain_result_emitters.sh` checks emitter **definitions** (exactly one `red` + one `green` per definer, plain `printf` body), never case-id reachability.
- #1499 is a *parked decision* about case-id **uniqueness** (two cases sharing an id) — a different property, and explicitly argued there to be not statically decidable. Pairing is differently shaped and is syntactically checkable.
- #1576 is a **documentation** ask for the same rule.

## Desired behavior
A `scripts/check-case-id-pairing.py` guard that fails CI when a suite emits a `FAIL: <id>` with no same-id green arm reachable in the same file.

## Motivation
Two hand-fixes off manual censuses is the cost of having no floor. A case that can never print green is a guard no mutation run can falsify — the exact defect the parent tracker #1484 exists to close — and today the only thing standing between that defect and `main` is whether someone happens to run a census.

### Constraints
- Model it on `scripts/check-result-line-plainness.py`, the closest existing precedent in shape, discovery and wiring.
- Must not flag a loop-scoped pair whose two arms carry the same interpolated id — `test_migrate_layout.sh:136-140` (`migrate-layout-10-$fslug`) is a **correct** shape and must stay clean.
- Invariant, not a ratchet: no allowlist, no baseline, no skip marker (the rule stated in the plainness guard's own header).

## Acceptance criteria
- [ ] `scripts/check-case-id-pairing.py` exists: stdlib-only, exit 0 clean / 1 violation(s) / 2 script error, reusing the plainness guard's non-recursive discovery globs `tests/*.sh` and `*/tests/*.sh`, with **zero discovery treated as a failure rather than a pass**. Evidence: `python3 scripts/check-case-id-pairing.py; echo rc=$?`
- [ ] It flags a `FAIL: <id>` emission whose id has no same-id `PASS:`/`ok:` emission reachable in the same file, and does **not** flag a loop-scoped pair whose two arms carry the same interpolated id. Evidence, both run explicitly: against `git show e79c4163:cogni-knowledge/tests/test_knowledge_lib.sh` it must flag `klib-01`; against `cogni-knowledge/tests/test_migrate_layout.sh` it must **not** flag `migrate-layout-10-$fslug` (lines 136-140, both arms paired).
- [ ] Run over the whole tree at head it returns a clean zero with no allowlist, no baseline and no skip marker. Evidence: `python3 scripts/check-case-id-pairing.py` exits 0 on the post-#1582 tree.
- [ ] `tests/test_check_case_id_pairing.sh` exists with fixture-driven positive **and** negative cases, mirroring `tests/test_check_result_line_plainness.sh`. Evidence: `bash tests/test_check_case_id_pairing.sh` exits 0.
- [ ] Wired into `.github/workflows/lint.yml` beside the `check-result-line-plainness.py` step, and `cogni-knowledge/tests/README.md` rule 6's "pair each red arm's id by reading the source" limitation is updated to name the guard instead. Evidence: `grep -n check-case-id-pairing .github/workflows/lint.yml cogni-knowledge/tests/README.md`
- [ ] The PR body records a runnable `## Mutation recipe` whose `--case` names a case from the new suite and whose `--expr` disables the pairing arm, and the replay returns `guard_verified`.

Refs #1582, #1575, #1484

claude-cogni-work:
Closes #1584.

## Summary

A `FAIL: <id>` arm with no reachable same-id green arm emits nothing on a green run, so `mutation-check.sh --case <id>` reports `case_not_found` on a genuinely red run and the guard it was meant to check proves nothing. The property lived only in prose (`cogni-knowledge/tests/README.md` rule 7) and the class had been hand-fixed twice off manual censuses.

- **`scripts/check-case-id-pairing.py`** — stdlib-only, exit 0/1/2, the plainness guard's two non-recursive globs, zero discovery is a failure. Two disjoint arms: `unpaired_fail_id` and `unpaired_abort_in_loop`.
- **`tests/test_check_case_id_pairing.sh`** — 30 assertions across 18 cases.
- **`case-id-pairing-guard`** job in `lint.yml`, beside the plainness job.
- **Registration** at both README sites (rule 6's hand-read limitation, rule 7's cross-reference). The root `CLAUDE.md` `## Contributing` registry bullet was **reverted** during review — committed contract item 5 bounds this PR's paths to a closed list that does not admit `CLAUDE.md` — and is deferred to #1591.
- **Remediation of every case the guard reports at head** — 30 findings across 16 files in four plugins — so the invariant is a clean zero with **no allowlist, no baseline, no skip marker**.

## The two decisions the issue left open

**1. Detection resolves ONE level of helper indirection.** The issue's "model it on `check-result-line-plainness.py`" implies a same-line literal scan. Measured against the tree, that is blind to the dominant emission shape here — ~15 suites define `pass()`/`fail()` and emit the id only at the call site, cogni-portfolio labels its green arm `ok:`, and a three-tier `check_eq`→`check`→`red/green` chain never co-locates id and label at all. A same-line guard would have shipped green and dead. So: a helper carrying both labels is a *paired* emitter (call sites self-pairing); a single-arm helper makes its call sites emissions with the id from `$1`. The resulting **recall floor is stated in the module docstring**, not left implicit.

**2. The exemption keys on SCOPE, not on abort.** AC #2 and AC #3 look contradictory: the required positive (`klib-01` at `e79c4163`) is *itself* an aborting fail-only arm, while ~100 straight-line aborting preflights ship at head. Reading the blob resolves it — at `e79c4163` `test_knowledge_lib.sh:31-36` is that arm **inside a `for` loop**. So a fail-only arm is exempt only when it aborts **and** sits at script scope. Re-derived across all 130 discovered suites, exactly two in-loop-or-function aborting fail-only arms exist (`test_prose_density_contract.sh`, `test_prelim_search_contract.sh`), both remediated here by the same collect-then-pair relocation `klib-01` itself got.

## Two AC defects worth recording

- **AC bullet 2, read literally, contradicts rule 7 and makes AC bullet 3 unreachable.** Flagging *any* unpaired `FAIL:` would flag ~100 exempt preflights with no allowlist available to absorb them. The scope proxy is what makes both bullets hold.
- **AC bullet 5's "rule 6" pointer is textually right but conceptually split.** The sentence "pair each red arm's id by reading the source" *is* rule 6's closing limitation — but rule 6 is the **uniqueness** rule and rule 7 (added by #1576 after this issue was filed) is the pairing rule. Both sites are edited; the prose deliberately does **not** claim the guard decides uniqueness, which #1499 parked as not statically decidable.

## Test plan

- [x] `python3 scripts/check-case-id-pairing.py` exits **0** at head, `total: 0`, no allowlist/baseline/skip marker anywhere in the guard.
- [x] `bash tests/test_check_case_id_pairing.sh` — 30/30, liveness floors are inequalities.
- [x] Positive: the real `e79c4163` blob is flagged on the `klib-01` stem (W1 resolves it from git, falling back to a verbatim transcription because CI checks out shallow).
- [x] Negative: `migrate-layout-10-$fslug` is not flagged (W2, against the real file).
- [x] Discriminating pair: a script-scope aborting preflight is exempt (E1) while the same shape in a loop (E2) or a function (E3) is a finding.
- [x] `python3 scripts/run-plugin-tests.py` — **130/130**.
- [x] `check-result-line-plainness`, `check-workflow-yaml`, `check-breadcrumbs`, `check-mcp-tool-grant` all still exit 0.

## Mutation recipe

```
scripts/mutation-check.sh --root . --file scripts/check-case-id-pairing.py --expr 's/if red_id in green_ids/if True/' --test 'bash tests/test_check_case_id_pairing.sh' --case P1
```

Replay: `guard_verified` (mutated `red`, restored `green`). The substitution disables the pairing decision, so every red id reads as already paired and no finding is produced — turning `P1`, which asserts an unpaired direct-literal arm *is* flagged, red. The anchor occurs exactly once (the names are bound locally on the preceding lines so the rewrite stays valid), and the suite allocates no `P10`, so `--case P1` matches whole-token. The recipe is echoed in the new suite's header.

## Review notes

The `/simplify` pass surfaced a real **soundness gap**, fixed here and fenced by case E3: the scope test read the *emitter* map, which only holds functions that spell a label inline — so an aborting fail-only arm inside a function using the indirect `fail <id> "..."` shape was exempted while the byte-equivalent inline-label arm was flagged. It now reads every function span.

Two findings were deliberately **not** taken. Hoisting the duplicated collect-then-pair preflight into `fixtures/test_helpers.sh` would move the emission one path segment outside the guard's own glob and make it invisible to the check it satisfies. And unifying the three quote-scanning loops was judged too much behavioural surface for this PR — they differ on backslash handling.

**Residual, not fixed here:** `cogni-portfolio/tests/test-solution-candidates.sh` uses English first-words as case ids throughout. The one case the guard flagged now carries a real `solcand-01` id on both arms; renumbering that whole file is outside this issue.

**Token scope disclosure:** the runner token is the standard `gh` OAuth token and is over-scoped against the documented least-privilege posture (extra: `gist`, `read:org`, `workflow`). Pushed under the operator-authorized `--allow-overscoped` flag.

## Open questions

1. **Do the ~100 straight-line aborting preflights owe PASS twins?** Rule 7 says aborting alone does **not** earn the exemption — "a preflight whose subject can legitimately be intact still owes a PASS twin, as `klib-01` shows". Yet that many ship at head, and this guard exempts them, which is the only way AC #2 and AC #3 both hold. The shipped rule is therefore weaker than the documented one. Two answers, neither derivable from the tree:
   - *Accepted convention* → narrow rule 7's qualifying sentence to loop/function scope, and the guard already matches the rule.
   - *Genuine backlog* → narrow the exemption to non-aborting-only and remediate them all, at which point the `unpaired_abort_in_loop` arm is deleted rather than adjusted.

   The boundary is stated in both the docstring and the README either way, and the exempt population is reported on every run as `summary.script_scope_aborts_exempt`, so it stays measured rather than restated. Recorded here rather than decided silently because it settles what the ecosystem's test convention *is*, not just what one guard detects.


claude-cogni-work:
## Ship-ready

Author instance has handed this off. `svc:needs-review` applied; the independent review belongs to the merger instance. No verdict surface was emitted here — no `svc:review-*` label, no `post-verdict`, no reviewed-head marker.

### Handoff preflight — clean (exit 0)

2 passed, 5 skipped, 0 advisories. Skips, each with its reason:

| instrument | disposition | reason |
|---|---|---|
| `version-bump` | **pass** | no `plugin.json` version touched |
| `mutation-recipe` | **pass** | recorded invocation is runnable as written (all five flags present as exact tokens) |
| `run-tests` | skipped | `no_suites_at_base` — cogni-service's `run-tests.sh` globs `tests/test-*.sh`; this repo names its suites `test_*.sh` and runs them through `run-plugin-tests.py`. **Coverage is not lost**: the real sweep was run directly and reports **130/130**. |
| `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` | skipped | `not_applicable` — no `SKILL.md` or `agents/*.md` in the touched set |

### Author self-grade vs the issue-time contract

Graded against the `--origin triage` contract on #1584 (derived `2026-08-23T07:43:42Z`, **before** any branch was cut, so it is diff-blind by chronology). **23 items graded — 22 satisfied, 1 violated, 0 unclear.** Advisory only; the merger's own grading is the one that counts.

**Residual carried to the merger — one item, deliberately not auto-fixed:**

> **item-5 (change-scope predicate), major.** The diff edits `CLAUDE.md` to add a `## Contributing` registration bullet for the new guard. Item 5's predicate is a *closed* path list that does not enumerate `CLAUDE.md`, so read literally the item is violated.

I did not take the mechanical remedy (dropping the hunk), because it would be the wrong change: that section carries exactly one bullet per repo-wide guard — `check-breadcrumbs.py`, `check-workflow-yaml.py`, `check-result-line-plainness.py`, `check-mcp-tool-grant.py` — and a repo-wide guard absent from it is invisible to every author and agent reading the file. Dropping it would ship an unregistered guard to satisfy a list that predates the discovery of the registration site: the contract was derived before the refine pass identified it, so a diff-blind issue-time contract could not have enumerated it. The grader independently reached the same read ("well-motivated… the contract as committed does not enumerate it").

**This is the merger's call, not mine.** Either disposition is defensible: accept the registration as within the item's spirit, or bounce it to a fast-follow. Flagging rather than deciding silently.

### Verification run on this head (`07694deb`)

- `python3 scripts/check-case-id-pairing.py` → exit **0**, `total: 0`, `script_scope_aborts_exempt: 100` — no allowlist, no baseline, no skip marker.
- `bash tests/test_check_case_id_pairing.sh` → **30/30**.
- `python3 scripts/run-plugin-tests.py` → **130/130**.
- `check-result-line-plainness`, `check-workflow-yaml`, `check-breadcrumbs`, `check-mcp-tool-grant` → all exit 0.
- Mutation replay → `guard_verified` (mutated `red`, restored `green`).

### Worth the reviewer's attention

- **A soundness gap was found and fixed during the `/simplify` pass**, fenced by new case `E3`: the scope test read the *emitter* map, which only holds functions that spell a label inline, so an aborting fail-only arm inside a function using the indirect `fail <id> "..."` shape was exempted while the byte-equivalent inline-label arm was flagged. It now reads every function span. Worth a look — it is the load-bearing half of the exemption rule.
- **The `## Open questions` item is genuine and unresolved**: the shipped exemption is deliberately weaker than rule 7's own text. It does not block this PR, but it decides whether a ~100-case follow-up backlog exists.
- **Token scope**: pushed under the operator-authorized `--allow-overscoped` flag; the runner token is the standard `gh` OAuth token (extra scopes `gist`, `read:org`, `workflow`). Disclosed in the PR body.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1590

**Summary:** plugin-validator and service-review pass, but contract-grading-reviewer's independent grading found item 5's change-scope predicate violated by an out-of-contract CLAUDE.md edit — a major, un-waived finding on a committed contract item, and therefore ineligible for merger-added-preference demotion regardless of plugin-validator's more permissive minor/waived read of the same location — so the merged verdict is needs_revision pending either reverting the CLAUDE.md hunk or amending the contract to admit it as a sixth allowed category.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[major]** `review-contract:item-5` — The diff edits CLAUDE.md (a one-line guard-registration bullet at ## Contributing), but item 5 enumerates exactly five allowed path categories and CLAUDE.md is not one of them. This is a clean, mechanically-verifiable divergence, not a genuine intent ambiguity: the file is unambiguously touched and unambiguously outside the predicate. The author's advisory self-grade concurs it is a violation and declined to auto-fix, reasoning that dropping it would ship an unregistered guard; that reasoning is substantively sound against repo convention but does not change what the committed contract's predicate admits.
  - *Suggested fix:* Either revert the CLAUDE.md hunk to bring the diff strictly within item 5's enumerated paths, or explicitly amend/re-scope the contract to admit the CLAUDE.md registration bullet as a sixth allowed category before merge.
  - *Raised by:* cogni-service:contract-grading-reviewer, also flagged by plugin-dev:plugin-validator

- **[minor]** `PR diff` — Diff is 1271 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (auto-fixable) (advisory-addition) `tests/test_check_case_id_pairing.sh:127` — Item 8's verify clause asks the suite to assert presence of file, line and case id on a finding. The suite asserts case_id and arm but never dereferences v[0]['file'] or v[0]['line']. The substantive property holds (confirmed by direct injected-orphan probe) — this is unfenced coverage, not a fault: a future refactor could drop or rename either field with every case still green.
  - *Suggested fix:* Extend the existing P1b assertion by two lines: assert v[0]['file'].endswith('tests/dirty.sh') and assert isinstance(v[0]['line'], int) and v[0]['line'] > 0.

### Positive observations

- E1/E2/E3 form a genuinely two-sided discriminating set, closing the vacuous-guard class the contract names as decisive. (plugin-dev:plugin-validator)
- E3 fences a soundness gap the author found and fixed rather than shipped. (plugin-dev:plugin-validator)
- Item 9's positive resolved from the REAL e79c4163 blob this run (source: blob), asserting the id stem rather than only the exit code. (plugin-dev:plugin-validator)
- Liveness floors sit on counters advanced before any arm condition, as inequalities. (plugin-dev:plugin-validator)
- No remediation gates a PASS on the shared errors accumulator (rule 7's third unreachable shape). (plugin-dev:plugin-validator)
- push-chain-03 demoted rather than given a decorative green — the better answer than item 13's letter; no cross-suite reference stranded by the id changes. (plugin-dev:plugin-validator)
- Item 15's false-positive reservoir closed by rule, not exclusion; C1 fences seven distinct shapes. (plugin-dev:plugin-validator)
- Item 21 satisfied in both directions, including explicitly disclaiming uniqueness (#1499). (plugin-dev:plugin-validator)
- lint.yml job is a faithful sibling, with the do-not-copy-fetch-depth comment item 3 calls out. (plugin-dev:plugin-validator)
- All other 22 items independently verified by execution, including a live mutation-check.sh replay returning guard_verified and a synthetic dirty-fixture probe confirming file/line/arm/case_id are all populated. (cogni-service:contract-grading-reviewer)

### Recommendation

---

claude-cogni-work:
## Deferred follow-up filed

The `CLAUDE.md` guard-registry bullet reverted from this PR (contract item 5) is tracked as
#1591 — a one-bullet change whose own contract will admit `CLAUDE.md` as an in-scope path.

The revert is only defensible because the registration is deferred rather than dropped, so
#1591 is the record of that. Nothing depends on it meanwhile: no test or script enforces the
registry, and the guard stays documented at `cogni-knowledge/tests/README.md` and in its
`lint.yml` job name.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1590

**Summary:** Re-review after revision 3f593966: all three perspectives (oversized-PR flag, plugin-validator, contract-grading-reviewer) pass — the two prior needs_revision findings are fully remediated, all 23 committed contract items are satisfied, the guard/suite were executed directly (30/30 green, clean zero at head), and the mutation recipe reproduced guard_verified independently; the only remaining findings are advisory minors (oversized-diff note and a non-blocking line-assertion tightening suggestion).

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** `PR diff` — Diff is 1272 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (auto-fixable) `tests/test_check_case_id_pairing.sh:132` — P1b's line assertion is a presence-and-type check (isinstance(v[0]['line'], int) and v[0]['line'] > 0) rather than a value check. Contract item 8 only asks for presence of the three finding fields, so this satisfies the item as written — but attribute_line() falls back to start + 1 when the needle is not found on any physical line in the unit, and that fallback is always >= 1, so a line-attribution regression (an off-by-one, or a silent collapse to the fallback) would keep this case green.
  - *Suggested fix:* Pin the value: assert v[0]['line'] == 5, v[0]['line'] (the red FAIL: only-red-01 line in the $WORK/p1/tests/dirty.sh heredoc), keeping the isinstance check if the type guard is wanted. That makes the attribution itself mutation-falsifiable rather than only its presence.

### Positive observations

- The revision is exactly scoped to the two prior findings and nothing else — 2 files, 3 insertions, 2 deletions, with no opportunistic edits riding along. (plugin-validator)
- The CLAUDE.md revert is a true revert, not a compensating edit: the file has dropped out of the diff entirely, so item 5's closed path list is satisfied by construction and the registry bullet is deferred to a named follow-up (#1591) instead of being silently dropped. (plugin-validator)
- The mutation recipe is genuinely two-directional and was verified by replay rather than by reading: the expr disables the pairing DECISION, not a cosmetic line, and P1/E2/W1 all go red under it. (plugin-validator)
- E1/E2/E3 are a real discriminating set. E3 is a regression fence for a soundness gap the author found and fixed mid-flight. (plugin-validator)
- The exempt population is reported as a live counter (script_scope_aborts_exempt: 100) rather than prose that would rot. (plugin-validator)
- The anti-vacuity floors are inequalities over counters advanced before any arm condition is evaluated, so a guard that silently stopped extracting ids cannot ship green by finding nothing. (plugin-validator)
- The one case that could not be paired (push-chain-03) was correctly demoted to unlabelled diagnostic detail rather than given a decorative PASS twin. (plugin-validator)
- All 16 remediated suites still exit 0 individually, including the two run_score_ok polarity inversions where a mistake would invert a whole file's verdict. (plugin-validator)
- The suite's self-scan discipline is real and load-bearing: R1 covers the suite's own file and exits 0. (plugin-validator)
- item 5 satisfied (re-verified after revision): every one of the 20 changed paths matches the closed predicate list exactly; grep for CLAUDE.md and plugin.json across the full diff returns zero hits, confirming the CLAUDE.md hunk is genuinely gone, not merely reduced. (contract-grading-reviewer)
- item 14 satisfied: ran the guard directly against the checked-out head — exit 0, violations [], summary.total 0. (contract-grading-reviewer)
- item 20 satisfied: ran the suite directly — exit 0, 30/30 passed. (contract-grading-reviewer)
- item 23 satisfied: directly replayed mutation-check.sh with the recorded invocation against the PR-head checkout and it reported guard_verified. (contract-grading-reviewer)
- item 18 satisfied: liveness counters are incremented before any arm condition is evaluated; observed live values 130/114/2308. (contract-grading-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Commit `3f593966` — 2 files, 3 insertions, 2 deletions. Fresh verdict: ✅ **PASS**.

### Findings addressed

**[major] `review-contract:item-5` — resolved in `3f593966`.**
Reverted the `CLAUDE.md` hunk. The file is now byte-identical to the merge-base (`24013c02`) and
has dropped out of the diff entirely — 21 files to 20 — so item 5's closed path list is satisfied
by construction rather than by argument.

The merger offered two remedies; I deliberately declined the second (amend the contract to admit a
sixth path category), on three grounds. Item 5 is a scope predicate whose entire function is to be
a **closed** set, so an item widened to admit whatever the diff turned out to contain can never
fail again — that deletes the check rather than resolving the divergence. The contract carries
`issue=#1584` and was authored by the **author** identity at triage time, then adopted by the
merger by chronology; there is no independently-authored `pr=#1590` contract. So "amend the
contract" here means the graded party widening its own pre-committed bar after seeing the diff.
And item 5's preamble clause ("admitting every artifact this checklist mandates") does not rescue
it — no item of the 23 mandates a `CLAUDE.md` registration (item 21 mandates only
`cogni-knowledge/tests/README.md`).

The author's original reasoning — that dropping the bullet ships an unregistered guard — is
correct about repo convention, so the registration is **deferred, not abandoned**: #1591, whose own
contract will admit `CLAUDE.md` as an in-scope path. Nothing depends on it meanwhile: no test or
script enforces the registry, and the guard stays documented at `cogni-knowledge/tests/README.md`
and in its `lint.yml` job name.

**[minor] `tests/test_check_case_id_pairing.sh:127` — resolved in `3f593966`.**
Extended the P1b assertion by the two suggested lines, appended after the `arm` assertion so the id
check stays the first failure under the recorded mutation. Both were mutation-probed rather than
assumed: forcing `"file"` to a wrong value turns P1b red, forcing `"line"` to a non-int turns P1b
red, and restoring returns it green — so neither assertion is vacuous.

**[minor] `PR diff` (1272 lines vs the 400 bound) — waived, not dropped.**
1133 of those lines are the two new files, which cannot be separated without shipping a guard with
no suite or a suite with no guard; the remaining ~140 are the head remediation the guard's own
clean-zero requirement (items 13 and 14) forces into the same commit, since landing the guard first
would put a blocking CI gate red at head. The branch is already pushed and contract-graded, so
re-cutting it discards the review that produced these findings for no reviewability gain.

### Carried forward for the merger

Two things the fresh verdict records that are **not** blocking, flagged so they are not discovered
as surprises:

1. **A new advisory minor** at `tests/test_check_case_id_pairing.sh:132` — P1b's line assertion is
   presence-and-type, not value-pinned. `attribute_line()` falls back to `start + 1`, which is
   always >= 1, so a line-attribution regression would keep the case green. Contract item 8 asks
   only for field *presence*, so this satisfies the item as written and is classified `preference`;
   I did not take it in this cycle because the verdict already stamps this head and re-cutting for
   a non-gating preference would invalidate it. The concrete fix, if wanted:
   `assert v[0]['line'] == 5` (the red arm's line in the P1 fixture heredoc — verified).
2. **Token scope** — pushed under the operator-authorized `--allow-overscoped` flag; the runner
   token is the standard `gh` OAuth token (extra scopes `gist`, `read:org`, `workflow`).

### Follow-up issues

- #1591 — register `check-case-id-pairing.py` in the root `CLAUDE.md` `## Contributing` guard
  registry (the deferred half of finding 1).

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1590

**Summary:** All 23 committed-contract items are verified satisfied by execution (guard, suite, full regression, and a live mutation-check replay all green); the only findings are a merger-added preference wording nit (demoted, non-gating) and the standing oversized-PR informational note, neither of which blocks merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** `PR diff` — Diff is 1272 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (auto-fixable) (advisory-addition) `tests/test_check_case_id_pairing.sh:9` — The header claims 'No id here is a prefix of another (there is no P10), so every --case resolves unambiguously.' Read literally that is false — P1 is a prefix of P1b, and likewise for P2/P2b, P3/P3b, E1/E1b, E2/E2b, E3/E3b, W1/W1b, S2/S2b, Z1/Z1b. The operative property is the whole-token matcher stated three lines earlier, which does disambiguate P1 from P1b, so nothing is broken and --case P1 resolves correctly. The prose is what is imprecise, and it is the sentence a future author would rely on when adding a case id.
  - *Suggested fix:* Reword to the property the matcher actually guarantees, e.g. 'No id here is a whole-token prefix of another — an id suffixed with b is a distinct token, and there is no P10 — so every --case resolves unambiguously.'

### Positive observations

- The exemption is falsifiable in BOTH directions, exactly what the contract's items 9/11 pair was written to force: E1 (script-scope aborting preflight, clean) against P1 (non-aborting fail-only, flagged) and E2/E3 (same aborting shape inside a loop / a function, flagged). Mutating either conjunct of the scope test independently reddens a disjoint set of cases. (plugin-dev:plugin-validator)
- E3 is a real regression fence, not decoration: the scope test originally read the emitter map, exempting an aborting fail-only arm in a function using the dominant indirect shape while flagging the byte-equivalent inline-label arm. classify_definitions now returns a separate all-function span list, and E3b asserts both shapes are flagged alike. (plugin-dev:plugin-validator)
- W1 resolves the AC's own witness from the real e79c4163 blob when git can reach it and falls back to a verbatim transcription otherwise, so the case never silently skips on a shallow CI checkout — and it asserts the reported klib-01 id stem, not just the exit code. (plugin-dev:plugin-validator)
- The liveness floors are inequalities over counters advanced before any arm condition, so a guard that quietly stopped extracting ids cannot ship green by finding nothing. (plugin-dev:plugin-validator)
- summary.script_scope_aborts_exempt (100 at head) makes the exempt population a measured number on every run rather than a prose claim, and the docstring names the residual and each recall floor with its witness. (plugin-dev:plugin-validator)
- The false-positive class is closed by the detection rule, never by excluding files: command position is quote-aware and labels must OPEN the emitted string, which keeps the mid-sentence prose diagnostic inert. All four named live witnesses stay clean while their files are fully scanned. (plugin-dev:plugin-validator)
- The two unpaired_abort_in_loop remediations use the collect-then-pair relocation rule 7 prescribes, with the reason in a comment beside the code, rather than bolting an unreachable PASS twin onto the loop body. (plugin-dev:plugin-validator)
- README rule 6's hand-read limitation now names the guard while preserving the uniqueness census as a separate manual check, so the uniqueness scope creep the watchlist warned about did not leak in. (plugin-dev:plugin-validator)
- item 5 satisfied: all 20 changed paths match one of the five enumerated predicates — no plugin.json, no CLAUDE.md, no product source outside tests/; each edited existing suite hunk is a genuine pairing remediation. (cogni-service:contract-grading-reviewer)
- item 11 satisfied: ran the discriminating pair live — E1 (script-scope aborting preflight) reports 0 findings and increments script_scope_aborts_exempt, while E2 (same shape inside a loop) is flagged unpaired_abort_in_loop, proving the exemption is narrow rather than blanket. (cogni-service:contract-grading-reviewer)
- item 14 satisfied: ran the guard against the live PR-head tree — exit 0, total 0, files_discovered 130, no allowlist or baseline file present. (cogni-service:contract-grading-reviewer)
- item 20 satisfied: ran the suite — exit 0, 30/30 PASS; every emitted label is an id token followed by a space, never a colon-abutted id. (cogni-service:contract-grading-reviewer)
- item 23 satisfied: independently re-ran the exact recorded mutation-check.sh invocation against the PR-head worktree and it returned guard_verified (mutated red, restored green), matching the PR body. (cogni-service:contract-grading-reviewer)
- item 13 satisfied: the named witness skill-contracts-105 now has a proper if/else pair gated on its own predicate, not a shared-errors-accumulator fold; every other remediated id follows the same if/else or collect-then-pair model. (cogni-service:contract-grading-reviewer)
- item 21 satisfied: README rule 6 now names the guard mechanically, rule 7 records its scope-based exemption, and the prose explicitly disclaims any uniqueness claim. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1636, opened 2026-08-26
cogni-consult: extend the no-retired-plugin-name scan to agents/ and output-styles/


claude-cogni-work:
## Plugin
cogni-consult (`tests/test_route_example_hygiene.sh`)

## Current behavior
After #1630, case 5 (`no-retired-plugin-name`) scans the plugin's authored prose surfaces as: `skills/**/SKILL.md` + plugin-root `README.md` + plugin-root `CLAUDE.md` + `references/**/*.md`.

Two authored prose surfaces remain outside that subject:

- `cogni-consult/agents/*.md` — 4 files
- `cogni-consult/output-styles/*.md` — 1 file

A retired plugin name written into any of the five is unguarded, exactly as `README.md` was before #1630.

## Desired behavior
The scan subject additionally resolves `agents/*.md` and `output-styles/*.md`, so every authored prose surface in the plugin is covered by the same bare-name-plus-carve-out predicate.

## Motivation
Deferred from #1630. Held out there because that issue's acceptance criteria ask for "at minimum the plugin-root `README.md`", and the plan already enumerated three surfaces — adding two more surface classes went beyond its bar. It is **not** unfinished #1630 work: all seven of #1630's acceptance criteria are delivered by its PR.

Measured at the #1630 merge base, all five files are **clean under both** the bare match and the `([^/]|$)` carve-out — zero hits for all 7 prefixes in `scripts/retired-plugins.json`. So the addition is green-at-base and costs one `find` per surface.

Two constraints inherited from #1630 that this change must not break:

- **The match stays bare-name.** A colon-anchored discriminator would be permanently green against the defect class this case exists to catch (the #1617 README row spelled `cogni-visual` with no colon).
- **The subject stays enumerated, never a whole-tree glob.** Engagement directories live under `cogni-consult/{slug}/` (`CLAUDE.md:162`), so a `find "$PLUGIN_DIR" -name '*.md'` would scan user engagement content as plugin source.

## Acceptance criteria

> **Maintainer ruling — 2026-08-27 (supersedes the original wording of the fourth criterion below).**
[...]

Issue #1617, opened 2026-08-25
cogni-consult: README Dependencies table names the retired plugin cogni-visual as a deliverable export route


claude-cogni-work:
## Plugin
cogni-consult (`README.md`)

## Current behavior
`cogni-consult/README.md:236` lists the **retired** plugin `cogni-visual` as a deliverable export route, in the Dependencies table:

```
| cogni-visual / document-skills | No | Deliverable export (slides, documents) when a deliverable names an export route |
```

`cogni-visual` is absent from `.claude-plugin/marketplace.json` `plugins[]` and has no tree at the default branch — its skills were absorbed into `cogni-workspace`.

## Desired behavior
That row no longer presents `cogni-visual` as an export route. Either it names the live successor (its render/adapt skills now live under `cogni-workspace`), or the row is rewritten so it names no retired plugin.

## Motivation
This is the same defect class as the sibling issue that fixed the three `producing_route` sites in `skills/consult-action-fields/SKILL.md` — same plugin, same retired name, same "a reader follows the example and reaches a plugin that no longer exists" failure. It was deliberately left out of that issue's scope, which was pinned to the SKILL.md, so it needs its own change rather than being silently absorbed.

No guard catches it, and the guard added alongside the sibling fix does not reach it either: `scripts/check-external-dispatch.py` matches only colon-bearing `plugin:skill` dispatch tokens, and `cogni-consult/tests/test_route_example_hygiene.sh` case `no-retired-plugin-name` scans `skills/**/SKILL.md` only — a plugin-root `README.md` is outside both.

## Acceptance criteria
- [ ] `grep -n 'cogni-visual' cogni-consult/README.md` returns no line presenting it as an export route or dependency. Falsified by: any surviving row naming it as a route.
- [ ] Any replacement names either a plugin listed in `.claude-plugin/marketplace.json` `plugins[]`, or a skill slug that exists as `*/skills/<slug>/SKILL.md` under such a plugin — or the row names no plugin at all. Falsified by: a named plugin absent from `plugins[]`, or a slug with no matching directory.
- [ ] The Dependencies table's other rows and its column shape are unchanged. Falsified by: an unrelated row edited or a column added/removed.
[...]

PR #1637, opened 2026-08-26
cogni-consult: guard retired plugin names in README, CLAUDE.md and references

### LINKED ISSUE #1630: cogni-consult: widen the no-retired-plugin-name guard beyond skills/**/SKILL.md with a cogni-claims directory carve-out
## Plugin
cogni-consult (`tests/test_route_example_hygiene.sh`)

## Current behavior
The `no-retired-plugin-name` case (lines 168–209) resolves its scan subject as:

```
skill_bodies="$(find "$PLUGIN_DIR/skills" -name SKILL.md -type f 2>/dev/null | sort)"
```

So a retired plugin name in a plugin-root `README.md` is unguarded. That is exactly how the defect in #1617 survived: `cogni-consult/README.md` named the retired `cogni-visual` as an export route, and neither this case nor `scripts/check-external-dispatch.py` (which matches only colon-bearing `plugin:skill` tokens) reached it.

## Desired behavior
The case guards retired-plugin names beyond `skills/**/SKILL.md` — at minimum the plugin-root `README.md` — so this class cannot recur unguarded.

## Motivation
Deferred from #1617, whose AC4 asked for this to be considered. It was declined there on measured evidence rather than preference, and that evidence is why this needs its own change rather than a one-line glob swap.

**A naive whole-plugin-tree widening goes RED against a correct tree.** `cogni-claims` is in `scripts/retired-plugins.json` `retired_prefixes[]` *and* is simultaneously a live directory name — the workspace claim registry `cogni-claims/claims.json`. It occurs roughly 43 times under `cogni-consult/` as a legitimate filesystem path, not a dispatch:

- `cogni-consult/CLAUDE.md:182`
- `cogni-consult/references/publish-routing.md:298`
- `cogni-consult/scripts/resolve-assumptions.py` (~8 sites)
- `cogni-consult/scripts/submit-assumption-claim.py`
- `cogni-consult/tests/test_resolve_assumptions.sh` (~20 lines)

Retiring the *plugin* did not move the *directory*. The suite's own header (lines 23–32) already records this hazard: widening to bare names "would flag ~100 legitimate prose and path mentions repo-wide", and the discriminator between "retired plugin named as a route" and "retired plugin name used as a directory" only exists near the site.

So the widening needs a directory-vs-dispatch carve-out that is itself designed and falsified — not a glob change.

## Acceptance criteria
- [ ] The `no-retired-plugin-name` scan subject covers `cogni-consult/README.md` in addition to `skills/**/SKILL.md`. Falsified by: a scan that still resolves only `$PLUGIN_DIR/skills`.
- [ ] The widened case is GREEN against the current default branch. Falsified by: any hit on the `cogni-claims/` directory-path sites enumerated above.
- [ ] The widened case is demonstrably RED against the pre-#1617 `README.md` content (which carried `cogni-visual` in the Dependencies table), verified by a recorded `scripts/mutation-check.sh` recipe in the PR body carrying all five of `--root`, `--file`, `--expr`, `--test`, `--case` as exact tokens with values. Falsified by: a missing flag, or an `--expr` that `perl -0pi` cannot apply.
- [ ] The recipe's `--case` is the single whitespace-free token `no-retired-plugin-name`, reproducing the label the suite actually prints. Falsified by: a glued, spaced, or invented case value.
- [ ] The recorded `--expr` targets a file the widening NEWLY covers (`README.md`), not a `skills/**/SKILL.md` body the base scan already covered. Falsified by: a recipe that would have gone red before the widening too.
- [ ] The directory-vs-dispatch carve-out is stated in the suite header beside the existing colon rationale (lines 23–32), so the recorded reasoning and the code agree. Falsified by: a widened scan leaving that header asserting the opposite of what the code now does.
- [ ] The four sibling cases (`route-examples-found`, `route-example-slug-resolves`, `route-example-non-default`, `route-example-parity`) are unchanged in behavior and still pass. Falsified by: any of the four changing verdict or losing its assertion.

Deferred from #1617.

claude-cogni-work:
Closes #1630.

## Summary

Widen case 5 (`no-retired-plugin-name`) in `cogni-consult/tests/test_route_example_hygiene.sh` from `skills/**/SKILL.md` alone to the plugin's authored prose surfaces — `skills/**/SKILL.md` + plugin-root `README.md` + plugin-root `CLAUDE.md` + `references/**/*.md` (37 files at this base, up from 9). Every path derives from `$PLUGIN_DIR`, never hardcoded, so `--root` still redirects the whole subject.

**Bare-name matching is kept, deliberately.** The obvious move is to reuse the colon anchor (`<prefix>:`) that `scripts/check-external-dispatch.py` and `cogni-workspace/tests/test-relocated-skill-hygiene.sh` both use — it has zero false positives on this corpus. It is the wrong answer here. The line #1617 removed from this README was:

```
| cogni-visual / document-skills | No | Deliverable export (slides, documents) when a deliverable names an export route |
```

A **bare** name, no colon. A colon-anchored match would be permanently green against exactly the content this case exists to catch — vacuous, and it would make the issue's AC3 ("demonstrably RED against the pre-#1617 README content") unsatisfiable.

**The carve-out is instead a trailing-slash exclusion** — an occurrence immediately followed by `/` is a filesystem path, not a route. It is load-bearing rather than decorative: two sites in the widened subject exercise it, `cogni-consult/CLAUDE.md:182` and `cogni-consult/references/publish-routing.md:298`, both `cogni-claims/claims.json`. Remove the carve-out and the case goes red on those two; that is falsification F1 below.

**The vacuity guard covers every surface — including the original one.** An absent `README.md`, `CLAUDE.md`, `references/`, **or `skills/**/SKILL.md`** FAILs the case naming the missing surface, rather than silently narrowing the scan and still printing PASS.

That last arm was a **regression this PR introduced and then fixed** — see "Caught in self-review" below.

**Header corrections.** A `DIRECTORY-VS-DISPATCH CARVE-OUT` paragraph is added beside the existing colon rationale; the `SCOPE OF EACH CASE` paragraph — which asserted case 5 "scans every skill in the plugin" — is rewritten to the real subject with the exclusions and their measured reason; the case-map row and the recorded-recipe paragraph are updated. The file still spells no retired prefix literally (`grep -cE 'cogni-(wiki|research|help|claims|narrative|copywriting|visual)'` → `0`), so it does not trip its own widened guard.

## Mutation recipe

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.456/scripts/mutation-check.sh \
  --root . \
  --file cogni-consult/README.md \
  --expr 's/\| cogni-workspace \| No \|/| cogni-visual | No |/' \
  --test 'bash cogni-consult/tests/test_route_example_hygiene.sh' \
  --case no-retired-plugin-name
```

```
bash ~/.claude/plugins/cache/managed-service/cogni-service/0.0.456/scripts/mutation-check.sh \
  --root . \
  --file cogni-consult/references/publish-routing.md \
  --expr 's/cogni-claims\/claims.json/cogni-claims claims.json/' \
  --test 'bash cogni-consult/tests/test_route_example_hygiene.sh' \
  --case no-retired-plugin-name
```

Replay both from the repo root, **serially** — they share a snapshot root, so let each restore before starting the next.

The first arm is the one the issue's AC3/AC5 ask for: `--file` is `cogni-consult/README.md`, a surface this change **newly** covers, so the recipe could not have gone red before the widening. `| cogni-workspace | No |` occurs exactly once in that file (line 235, the `## Dependencies` table), so the substitution can never be an `expr_no_op`, and the mutant reproduces the pre-#1617 shape: a bare retired plugin name in a README dependencies row.

The second arm proves the `references/` half of the subject is really scanned **and** that the carve-out is what keeps that site green — it deletes only the `/`, turning a path mention into a bare one. `cogni-claims` occurs exactly once in that file (line 298).

There is no in-repo copy of the shared harness. `cogni-consult/scripts/mutation-check.sh` and `cogni-portfolio/scripts/mutation-check.sh` are a **different instrument** — no-arg, exit-code-classifying, accepting none of these five flags — so do not point the recipe at either. If that version directory is gone, use the newest under the same parent; everything after the path is version-independent.

## Falsification performed

Each probe was run against a scratch copy of the plugin tree; `→` is the observed case-5 line.

| # | mutation | expected | observed |
|---|---|---|---|
| F0 | none (base) | PASS | `PASS: no-retired-plugin-name` |
| F1 | carve-out removed (`grep -lF`, bare) | FAIL | `FAIL: … cogni-claims` — **carve-out is load-bearing** |
| F2 | `references/` dropped from subject, then `publish-routing.md` mutated | PASS | `PASS` — the mutation alone is not what reddens F3 |
| F3 | `references/` in subject, same mutation | FAIL | `FAIL: … cogni-claims` — **second arm genuinely grades `references/`** |
| F4 | bare retired name planted in `README.md` | FAIL | `FAIL: … cogni-visual` — primary arm |
| F5 | `README.md` removed | FAIL naming it | `FAIL: … scanned surface(s) absent …: README.md (vacuous scan — the subject would silently narrow)` |
| F5b | `CLAUDE.md` removed | FAIL naming it | `FAIL: … scanned surface(s) absent …: CLAUDE.md (…)` |
| F5c | `references/` removed | FAIL naming it | `FAIL: … scanned surface(s) absent …: references/**/*.md (…)` |
| F5d | `references/` present but emptied of `*.md` | FAIL naming it | `FAIL: … scanned surface(s) absent …: references/**/*.md (…)` — the emptied-but-present shape; **PASSes at `f043c216`**, which is what makes this arm's fix decisive rather than merely green |
| F8 | `skills/` removed | FAIL naming it | `FAIL: … scanned surface(s) absent …: skills/**/SKILL.md (…)` — the regression, post-fix |
| F8b | `skills/` present but emptied of `SKILL.md` | FAIL naming it | same — matches base's `-z "$skill_bodies"` semantics exactly |
| F6 | restored | PASS | `PASS: no-retired-plugin-name` |
| F7 | whole tree copied under a path containing a space | PASS | `PASS` — the `xargs -0` form is space-safe |

F2/F3 is the pair that matters: it separates "the mutation reddens the case" from "the case reddens *because that surface is in the subject*". Without F2, F3 alone would not distinguish the two.

Every probe was re-run after the cleanup pass below, and both recorded recipes re-replayed to `guard_verified` against the final code — a quality pass must not be allowed to quietly hollow out the guard it is tidying.

## Cleanup pass

Three changes from the pre-commit quality review, none behavioral:

1. **Removed a dead `grep -v '^$'` filter and the comment justifying it** — the comment claimed a skipped optional file "would otherwise contribute an empty record", which is false: when `[ -f … ]` fails the `printf` never runs, so nothing is emitted at all. Measured 0 empty records in every configuration. A false mechanism written into a comment outlives the line it defends, and this block is expected to be edited again.
2. **Bound the surface list once as `$scan_desc`** instead of spelling it verbatim in two `fail` messages, which could silently drift from the enumeration they describe.
3. **Batched the match** from one `grep` per (prefix, file) pair to one per prefix via `tr '\n' '\0' | xargs -0`: **259 process spawns → 7**, suite time 0.41s → 0.11s. Per-prefix attribution is unaffected — it comes from the loop variable, never from `grep`, so the failure message still names which prefix was found. `xargs -0` rather than word-splitting because `$PLUGIN_DIR` is caller-supplied via `--root` and may contain spaces (F7 covers this).
4. **Keyed the `references/**/*.md` vacuity arm on the assembled subject** (revision `a38ebaca`, closing the review verdict's one non-waived major, contract item 4). It read `[ -d "$PLUGIN_DIR/references" ]` — directory presence, a strictly weaker question than whether the scan reaches anything inside. Emptying `references/` of every `.md` while leaving the directory in place passed `-d` and still printed PASS (F5d). Now mirrors the `skills/` arm. The `README.md`/`CLAUDE.md` arms are deliberately **not** converted: each is a single file gated on `-f`, the identical test the subject assembly uses to admit it, so presence and inclusion-in-subject are equivalent and no emptied-but-present shape exists for them.
5. **Removed the unreachable `elif [ -z "$scan_bodies" ]` arm** (same revision). An empty subject necessarily fails the `skills/` grep, so `$missing_surface` is non-empty and the preceding arm always wins — verified against a plugin root stripped of all four surfaces, which names all four and never prints `no scannable body found`. A vacuity arm that can never fire is unfalsifiable, the same green-gate class this case exists to catch; its diagnostic is subsumed by the arm that does fire, which additionally names *which* surfaces vanished. This restores the merge base's 3-fail-site shape.
6. **Named the third surface-registration site.** The coupling note said to keep "this list and `$scan_desc`" in step — two sites, but a surface is registered in **three**: the `find` list, `$scan_desc`, and the `$missing_surface` chain. Registering a surface in the first two but not the third is exactly what lets it vanish silently.

A single `grep` with an alternation over all 7 prefixes was considered and rejected: `-l` reports which *file* matched, never which *alternative*, so it would cost the per-prefix attribution the failure message depends on.

One portability note, since CI is `ubuntu-latest` and this was developed on macOS: BSD and GNU `xargs` disagree on empty input — BSD skips, GNU runs the command once with no file operand, which would make `grep` read stdin. That divergence is unreachable here, because `$missing_surface` is non-empty whenever `$scan_bodies` is empty (an empty subject cannot match the `skills/` arm's grep), so the preceding `fail` arm always wins and the `else` branch — and therefore `xargs` — is only ever reached with at least one file. No `-r` is used, as it is a GNU extension the BSD build does not carry.

## Caught in self-review

The first version of this change **silently weakened an existing guard while widening it**, and the pre-handoff self-grade against the issue-time contract caught it.

Base had a dedicated vacuity arm, `elif [ -z "$skill_bodies" ]` — an absent `skills/` tree failed the case by name. Replacing that with a single emptiness test over the *whole* widened set meant it only fired when **every** surface was empty, so a missing `skills/` tree was absorbed by the other three and passed silently. Measured directly, deleting `skills/` and leaving the rest:

| | result |
|---|---|
| `origin/main` (base) | `FAIL: … no SKILL.md found under …/skills (vacuous scan)` |
| this branch, before the fix | `PASS: no-retired-plugin-name` |

A widening that quietly drops coverage of the surface it started from is worse than not widening at all, and nothing else in the pipeline would have caught it: the widened case was green, all 134 suites were green, and both mutation recipes returned `guard_verified` — because none of them exercises an absent `skills/` tree.

The fix adds a `skills/**/SKILL.md` arm to the same `missing_surface` chain, tested against the **assembled subject** (`grep -q '/skills/.*/SKILL\.md$'`) rather than with a second `find`, so it cannot drift from what is actually scanned and costs no extra I/O. F8/F8b above are its falsifiers; F8b confirms it reproduces base's semantics for a `skills/` directory that exists but holds no `SKILL.md`.

## Scope decisions

**Why three surfaces and not just `README.md`.** A README-only widening satisfies the issue's AC1 while making **AC2 vacuous** — `cogni-consult/README.md` holds zero occurrences of all seven prefixes, so AC2's own falsifier ("any hit on the `cogni-claims/` directory-path sites") could never fire and the carve-out would guard nothing. `CLAUDE.md` and `references/**/*.md` bring in the two real path sites, which is what makes the carve-out falsifiable in both directions.

**Why `scripts/` and `tests/` are excluded — measured, not assumed.** Of the 44 `cogni-claims` occurrences under `cogni-consult/`, 42 live in those two trees and no lexical carve-out clears them:

| discriminator | residue still red |
|---|---|
| followed by `/` | 11 |
| adjacent to `/` on either side | 3 |

The 11 break down as **8** `mkdir -p "$PROJn/cogni-claims"` in `tests/test_resolve_assumptions.sh` (separator *precedes* the name, line ends right after it), **2** assembled across string-literal boundaries via `os.path.join` (`scripts/resolve-assumptions.py:370`, `scripts/submit-assumption-claim.py:198`) where no separator is adjacent in the source at all, and **1** genuine bare mention in a comment (`tests/test_resolve_assumptions.sh:449`). Including these trees would make the case permanently red on a correct tree.

> The issue's Motivation, and this PR's own triage comment, both asserted these occurrences were uniformly path-shaped. That was false and is [corrected on the issue](https://github.com/cogni-work/insight-wave/issues/1630#issuecomment-5426196976): 44 total, 33 slash-followed, 11 not. The design above is built on the corrected measurement.

**Why the subject is enumerated, not globbed.** Engagement directories live under `cogni-consult/{slug}/` (`CLAUDE.md:162`), so a `find "$PLUGIN_DIR" -name '*.md'` would scan a user's own engagement content as if it were plugin source.

**Not touched:** `.claude-plugin/plugin.json` — the post-merge workflow owns the version bump.

## Test plan

- [x] `bash cogni-consult/tests/test_route_example_hygiene.sh` → five `PASS:` lines, exit 0
- [x] `bash cogni-consult/tests/test_route_example_hygiene.sh --root cogni-consult` → identical output (confirms the new subjects follow `--root`)
- [x] Full repo suite via `python3 scripts/run-plugin-tests.py` — 134/134 pass, `failed: 0` (re-run after the cleanup pass)
- [x] `python3 scripts/check-external-dispatch.py` → clean, `files_affected: 0`
- [x] Suite file spells no retired prefix literally → `0`
- [x] Both mutation-recipe arms replayed → `guard_verified` (re-replayed after the regression fix)

## Follow-up issues

- #1636 — extend the same scan to `agents/*.md` and `output-styles/*.md`, the two authored prose surfaces still uncovered. Both measured clean at this base under the bare match and the carve-out. This sits **outside** #1630's acceptance criteria (which ask for "at minimum the plugin-root README.md"), so all seven of them are delivered here and the link stays `Closes`, not `Refs`.

## Review notes

`check-preflight.sh`'s mutation-recipe check selects subjects by the glob `*/tests/test-*.sh` (hyphen); this suite is `test_route_example_hygiene.sh` (underscore), so that check records `not_applicable` and **the recorded recipe is ungated**. A green preflight is not evidence about it — it was replayed by hand, with the results in the falsification table above. The same naming mismatch makes cogni-service's `run-tests.sh` floor inert for this plugin; insight-wave's own `scripts/run-plugin-tests.py` globs `*/tests/*.sh` name-agnostically and does run this suite.

### DIFF
--- a/cogni-consult/tests/test_route_example_hygiene.sh
+++ b/cogni-consult/tests/test_route_example_hygiene.sh
@@ -10,15 +10,17 @@
 # colon: `PASS: goes-red: ...` would match neither pattern, and a recipe
 # recorded against this suite would return case_not_found instead of a verdict.
 #
-# Which mutation reddens which case. The recorded recipe in the pull request
-# drives the second one; the retired name it substitutes is deliberately not
+# Which mutation reddens which case. Two recipes are recorded in the pull
+# request — one drives the second case, one drives the fifth against the
+# plugin-root README.md; the retired names they substitute are deliberately not
 # spelled here, so this file introduces no new mention of a retired plugin.
 #   route-examples-found         deleting both `Route:` example lines
 #   route-example-slug-resolves  rewriting the route token to any prefix listed
 #                                in scripts/retired-plugins.json (the recorded recipe)
 #   route-example-non-default    rewriting the route token to consult-design-thinking
 #   route-example-parity         changing the route token at one site only
-#   no-retired-plugin-name       planting any retired prefix in a skill body
+#   no-retired-plugin-name       planting any retired prefix in any scanned body
+#                                (the recorded README recipe)
 #
 # RELATIONSHIP TO scripts/check-external-dispatch.py. That repo-level guard reads
 # the SAME registry and already scans this subject file, so this suite is not a
@@ -30,6 +32,19 @@
 # as a route" and "retired plugin name used as a directory" only exists near the
 # site, which is why this half of the property is guarded per-plugin here rather
 # than repo-wide there.
+#
+# DIRECTORY-VS-DISPATCH CARVE-OUT. Case 5 keeps the BARE-NAME match; the colon
+# anchor above cannot be reused here. The README row this case exists to catch
+# named a retired plugin in a Dependencies table with no colon at all, so a
+# colon-anchored match would be permanently green against exactly the content
+# the case guards — vacuous, and unfalsifiable by the recorded recipe. What is
+# excluded instead is an occurrence immediately followed by `/`: that is what a
+# filesystem path looks like, and a directory that happens to be named after a
+# retired plugin is not a route. The carve-out is load-bearing rather than
+# decorative — two sites in the scanned subject exercise it, the
+# `<prefix>/claims.json` path mentions in CLAUDE.md and in
+# references/publish-routing.md — so the case is green with it and red without
+# it, which is what the second recorded recipe falsifies.
 #
 # THE PREFIX FILTER MIRRORS THAT GUARD'S load_registry CONTRACT deliberately.
 # A colon-bearing entry would pass a bare truthiness filter and then be handed to
@@ -57,7 +72,24 @@
 #
 # SCOPE OF EACH CASE. Cases 1-4 pin the `Route:` note, which is defined in one
 # file, so they key on that file. Case 5's predicate is plugin-wide — a retired
-# name is wrong in any skill body — so it scans every skill in the plugin.
+# name is wrong in any authored prose the plugin ships — so it scans an
+# ENUMERATED subject: every `skills/**/SKILL.md`, the plugin-root `README.md`
+# and `CLAUDE.md`, and every `references/**/*.md`.
+#
+# What that subject leaves out, and why it is enumerated rather than globbed.
+# `scripts/` and `tests/` are excluded because no lexical carve-out clears them
+# on a CORRECT tree: one retired prefix is also a live workspace directory name,
+# and after the trailing-slash carve-out 11 legitimate occurrences across three
+# files still match — 8 of the form `mkdir -p "$VAR/<prefix>"`, where the path
+# separator precedes the name and the line ends right after it; 2 that assemble
+# the path across string-literal boundaries via `os.path.join`, so no separator
+# sits adjacent in the source at all; and 1 bare mention in a comment. Including
+# those trees would make the case permanently red against a correct tree, which
+# is the failure this enumeration exists to avoid. `agents/` and
+# `output-styles/` are authored prose too and are simply not covered yet.
+# A whole-tree glob over `$PLUGIN_DIR` is also deliberately avoided: engagement
+# directories live under `$PLUGIN_DIR/{slug}/`, so a glob would scan a user's
+# own engagement content as if it were plugin source.
 #
 # Each case below is gated only on its OWN predicate and emits exactly one
 # result line per run, so every `FAIL` arm has a reachable same-id green twin
@@ -166,7 +198,8 @@ else
 fi
 
 # --- case 5: no-retired-plugin-name ------------------------------------------
-# No retired plugin prefix may appear in any skill body in this plugin, in a
+# No retired plugin prefix may appear in any scanned body in this plugin — skill
+# bodies, the plugin-root README.md and CLAUDE.md, and references/**/*.md — in a
 # Route: note or in prose. An unreadable, empty, or malformed registry fails
 # rather than passing silently.
 retired_prefixes="$(python3 -c '
@@ -187,22 +220,47 @@ for prefix in prefixes:
     print(prefix)
 ' "$RETIRED" 2>/dev/null || true)"
 
-skill_bodies="$(find "$PLUGIN_DIR/skills" -name SKILL.md -type f 2>/dev/null | sort)"
+# Every path derives from $PLUGIN_DIR, never a hardcoded one, so --root (above)
+# still redirects the whole subject. Keep this list and $scan_desc below in step.
+# The suite runs under `set -u` only — there is no `set -e` and no pipefail — so
+# a find over a missing directory is inert here; do not add `set -e` without
+# revisiting this.
+scan_desc='skills/**/SKILL.md, README.md, CLAUDE.md, references/**/*.md'
+scan_bodies="$({
+  find "$PLUGIN_DIR/skills" -name SKILL.md -type f
+  find "$PLUGIN_DIR/references" -name '*.md' -type f
+  [ -f "$PLUGIN_DIR/README.md" ] && printf '%s\n' "$PLUGIN_DIR/README.md"
+  [ -f "$PLUGIN_DIR/CLAUDE.md" ] && printf '%s\n' "$PLUGIN_DIR/CLAUDE.md"
+} 2>/dev/null | sort)"
+
+# A newly covered surface that has silently vanished must not read as success:
+# the scan would narrow back toward the pre-widening subject and still print
+# PASS. Each arm below names the missing surface. They stay in ONE if/elif chain
+# so the case still emits exactly one result line per run.
+missing_surface=""
+[ -f "$PLUGIN_DIR/README.md" ] || missing_surface="$missing_surface README.md"
+[ -f "$PLUGIN_DIR/CLAUDE.md" ] || missing_surface="$missing_surface CLAUDE.md"
+[ -d "$PLUGIN_DIR/references" ] || missing_surface="$missing_surface references/"
 
 if [ -z "$retired_prefixes" ]; then
   fail "no-retired-plugin-name" "no usable retired_prefixes[] readable from $RETIRED (vacuous scan)"
-elif [ -z "$skill_bodies" ]; then
-  fail "no-retired-plugin-name" "no SKILL.md found under $PLUGIN_DIR/skills (vacuous scan)"
+elif [ -n "$missing_surface" ]; then
+  fail "no-retired-plugin-name" "scanned surface(s) absent under $PLUGIN_DIR:$missing_surface (vacuous scan of a newly covered surface)"
+elif [ -z "$scan_bodies" ]; then
+  fail "no-retired-plugin-name" "no scannable body found under $PLUGIN_DIR ($scan_desc) (vacuous scan)"
 else
   present=""
   for prefix in $retired_prefixes; do
-    hits="$(printf '%s\n' "$skill_bodies" | while IFS= read -r body; do
-      grep -l -- "$prefix" "$body" 2>/dev/null || true
-    done)"
+    # Bare-name match minus the directory-path carve-out: an occurrence
+    # immediately followed by `/` is a filesystem path, not a route. See
+    # DIRECTORY-VS-DISPATCH CARVE-OUT in the header for why the colon anchor
+    # scripts/check-external-dispatch.py uses cannot be reused here.
+    hits="$(printf '%s\n' "$scan_bodies" | tr '\n' '\0' \
+      | xargs -0 grep -lE -- "$prefix"'([^/]|$)' 2>/dev/null || true)"
     [ -n "$hits" ] && present="$present $prefix"
   done
   if [ -n "$present" ]; then
-    fail "no-retired-plugin-name" "retired plugin prefix(es) present in a $PLUGIN_DIR skill body:$present"
+    fail "no-retired-plugin-name" "retired plugin prefix(es) present in a $PLUGIN_DIR scanned body ($scan_desc):$present"
   else
     pass "no-retired-plugin-name"
   fi
--- a/cogni-consult/tests/test_route_example_hygiene.sh
+++ b/cogni-consult/tests/test_route_example_hygiene.sh
@@ -221,7 +221,10 @@ for prefix in prefixes:
 ' "$RETIRED" 2>/dev/null || true)"
 
 # Every path derives from $PLUGIN_DIR, never a hardcoded one, so --root (above)
-# still redirects the whole subject. Keep this list and $scan_desc below in step.
+# still redirects the whole subject. A surface is registered in THREE places that
+# must stay in step: this list, $scan_desc below, and the $missing_surface vacuity
+# chain further down. Adding a surface to the first two but not the third is what
+# lets it silently vanish while the case still prints PASS.
 # The suite runs under `set -u` only — there is no `set -e` and no pipefail — so
 # a find over a missing directory is inert here; do not add `set -e` without
 # revisiting this.
@@ -242,22 +245,32 @@ scan_bodies="$({
 # Before the widening, an absent skills/ tree was caught by a dedicated
 # `-z "$skill_bodies"` test; once the subject grew, a bare emptiness test on the
 # whole set only fires when EVERY surface is empty, so an absent skills/ tree
-# would be absorbed by the other three and pass silently. It is tested against
-# the assembled subject rather than with another find, so it cannot drift from
-# what is actually scanned.
+# would be absorbed by the other three and pass silently.
+#
+# BOTH directory-glob arms — skills/ and references/ — are tested against the
+# ASSEMBLED SUBJECT ($scan_bodies), never with a second find and never with a
+# bare `-d`. A `-d` test asks whether the directory exists, which is a strictly
+# weaker question than whether the scan actually reaches anything inside it: a
+# directory left in place but emptied of its .md files passes `-d` while
+# contributing nothing to the subject, so the surface silently narrows and the
+# case still prints PASS. Testing the subject cannot drift from what is scanned.
+#
+# The README.md and CLAUDE.md arms are deliberately NOT of that shape and need
+# no conversion: each is a single file gated on `-f`, the identical test the
+# subject assembly above uses to admit it, so presence and inclusion-in-subject
+# are equivalent for them and no emptied-but-present shape exists.
 missing_surface=""
 printf '%s\n' "$scan_bodies" | grep -q '/skills/.*/SKILL\.md$' \
   || missing_surface="$missing_surface skills/**/SKILL.md"
 [ -f "$PLUGIN_DIR/README.md" ] || missing_surface="$missing_surface README.md"
 [ -f "$PLUGIN_DIR/CLAUDE.md" ] || missing_surface="$missing_surface CLAUDE.md"
-[ -d "$PLUGIN_DIR/references" ] || missing_surface="$missing_surface references/"
+printf '%s\n' "$scan_bodies" | grep -q '/references/.*\.md$' \
+  || missing_surface="$missing_surface references/**/*.md"
 
 if [ -z "$retired_prefixes" ]; then
   fail "no-retired-plugin-name" "no usable retired_prefixes[] readable from $RETIRED (vacuous scan)"
 elif [ -n "$missing_surface" ]; then
   fail "no-retired-plugin-name" "scanned surface(s) absent under $PLUGIN_DIR:$missing_surface (vacuous scan — the subject would silently narrow)"
-elif [ -z "$scan_bodies" ]; then
-  fail "no-retired-plugin-name" "no scannable body found under $PLUGIN_DIR ($scan_desc) (vacuous scan)"
 else
   present=""
   for prefix in $retired_prefixes; do
--- a/cogni-consult/tests/test_route_example_hygiene.sh
+++ b/cogni-consult/tests/test_route_example_hygiene.sh
@@ -233,11 +233,21 @@ scan_bodies="$({
   [ -f "$PLUGIN_DIR/CLAUDE.md" ] && printf '%s\n' "$PLUGIN_DIR/CLAUDE.md"
 } 2>/dev/null | sort)"
 
-# A newly covered surface that has silently vanished must not read as success:
-# the scan would narrow back toward the pre-widening subject and still print
-# PASS. Each arm below names the missing surface. They stay in ONE if/elif chain
-# so the case still emits exactly one result line per run.
+# A surface that has silently vanished must not read as success: the scan would
+# narrow back toward a smaller subject and still print PASS. Each arm below names
+# the missing surface. They stay in ONE if/elif chain so the case still emits
+# exactly one result line per run.
+#
+# The skills/ arm covers the ORIGINAL subject and is not optional bookkeeping.
+# Before the widening, an absent skills/ tree was caught by a dedicated
+# `-z "$skill_bodies"` test; once the subject grew, a bare emptiness test on the
+# whole set only fires when EVERY surface is empty, so an absent skills/ tree
+# would be absorbed by the other three and pass silently. It is tested against
+# the assembled subject rather than with another find, so it cannot drift from
+# what is actually scanned.
 missing_surface=""
+printf '%s\n' "$scan_bodies" | grep -q '/skills/.*/SKILL\.md$' \
+  || missing_surface="$missing_surface skills/**/SKILL.md"
 [ -f "$PLUGIN_DIR/README.md" ] || missing_surface="$missing_surface README.md"
 [ -f "$PLUGIN_DIR/CLAUDE.md" ] || missing_surface="$missing_surface CLAUDE.md"
 [ -d "$PLUGIN_DIR/references" ] || missing_surface="$missing_surface references/"
@@ -245,7 +255,7 @@ missing_surface=""
 if [ -z "$retired_prefixes" ]; then
   fail "no-retired-plugin-name" "no usable retired_prefixes[] readable from $RETIRED (vacuous scan)"
 elif [ -n "$missing_surface" ]; then
-  fail "no-retired-plugin-name" "scanned surface(s) absent under $PLUGIN_DIR:$missing_surface (vacuous scan of a newly covered surface)"
+  fail "no-retired-plugin-name" "scanned surface(s) absent under $PLUGIN_DIR:$missing_surface (vacuous scan — the subject would silently narrow)"
 elif [ -z "$scan_bodies" ]; then
   fail "no-retired-plugin-name" "no scannable body found under $PLUGIN_DIR ($scan_desc) (vacuous scan)"
 else


claude-cogni-work:
## Ship-ready

Author instance (`claude-cogni-work`) hands PR #1637 off for independent review. `svc:needs-review` applied; issue #1630 moved `svc:in-progress` → `svc:resolved`. The PR stays **draft** — the merger flips it.

### Pre-handoff preflight

`check-preflight.sh --root <branch worktree> --head-ref origin/fix/issue-1630-widen-retired-name-scan` → **exit 0, clean**, `metric_version 3`.

**Read that green with care — 1 of 7 instruments produced a verdict.** Every skip, named:

| instrument | disposition | reason |
|---|---|---|
| `version-bump` | **pass** (v2) | no version-line touch — the only substantive verdict |
| `run-tests` | skip | `no_suites_at_base` |
| `mutation-recipe` | skip | `not_applicable` |
| `frontmatter` | skip | `not_applicable` — no touched `SKILL.md` / `agents/*.md` |
| `breadcrumbs` | skip | `not_applicable` — same |
| `skill-length` | skip | `not_applicable` — no touched `SKILL.md` |
| `agent-length` | skip | `not_applicable` — no touched `agents/*.md` |

The last four are honest: this diff touches one `tests/*.sh` file. **The first two are not what they appear.**

- `run-tests` reports `no_suites_at_base` — "nothing to regress from". cogni-consult has **13 suites**. They are named `test_*.sh`; `run-tests.sh` globs `test-*.sh`, so it discovered zero and derived a floor of 0. The instrument is **inert for this plugin**, not reporting an empty plugin.
- `mutation-recipe` is `not_applicable` for the same underscore/hyphen mismatch: the touched path is `tests/test_route_example_hygiene.sh`, which does not match `*/tests/test-*.sh`. **The recorded recipe is ungated.**

Both gaps were covered by hand instead: all **134** suites run via insight-wave's own `scripts/run-plugin-tests.py` (which globs `*/tests/*.sh`, name-agnostic) → `failed: 0`; and both recipe arms replayed → `guard_verified`.

### Self-grade against the issue-time contract

Graded against the 18-item `review_contract` on #1630 (`origin: triage`, derived before any branch existed). Advisory only — no verdict, no `svc:review-*` label, no verdict comment.

**17 satisfied, 1 violated. The violation was real, and is fixed.**

Contract item 4 (vacuity guard must cover the widened subject) was **violated**: the first version replaced base's dedicated `elif [ -z "$skill_bodies" ]` arm with one emptiness test over the whole widened set, which only fires when *every* surface is empty. Deleting `skills/` and leaving the rest gave `PASS` on the branch where base gave `FAIL`. A widening that drops coverage of the surface it started from. Fixed in the second commit; F8/F8b in the PR body are its falsifiers.

Nothing else in the pipeline would have caught it — the case was green, 134/134 were green, and both recipes returned `guard_verified`, because none of them exercises an absent `skills/` tree.

### Residual for the merger

Not blockers; inherit them rather than rediscover them.

1. **Two of the 17 "satisfied" items are satisfied only VACUOUSLY.** Items 8 and 9 are conditionals gated on the widened subject reaching `scripts/` or `tests/`. It does not, so their falsifiers cannot fire. **The substantive surface is 15 of 17, not 17 of 17** — please do not read the count as stronger evidence than it is.
2. **The `--case` in both recorded recipes is ungated by `check-preflight`** (underscore glob, above). Replay them rather than trusting the section.
3. **A judgment call worth a second opinion:** the issue says "at minimum the plugin-root `README.md`". This ships three surfaces. `agents/*.md` and `output-styles/*.md` are deliberately left out and filed as #1636 — both measured clean at base under the bare match and the carve-out, so the omission costs nothing today, but whether the subject should have gone further in one pass is a scope call.
4. **`Closes #1630`, not `Refs`, despite a non-empty `deferred[]`.** All seven of the issue's ACs are delivered here; #1636 is scope the issue never asked for. Rationale recorded on the issue. Flagging it because the mechanical rule keys on `deferred[]` being non-empty and would have produced `--partial` — that would have left #1630 open behind a `Refs`-only merged PR, which this board is already carrying six of.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1637

**Summary:** One non-waived major defect: review-contract item 4 references/**/*.md vacuity arm is keyed on directory presence rather than the assembled scan subject, so an emptied-but-present references/ directory silently PASSes. The two perspectives split on this item and the merger adjudicated it by execution: contract-grading-reviewer graded it satisfied using an incomplete probe (it deleted the directory, which the -d test does catch) while plugin-validator emptied the directory of .md files, which passes silently. Of the 18 committed contract items the substantive surface is 16/18 (items 8 and 9 are only vacuously satisfied, their antecedents unmet so their falsifiers cannot fire), and item 4 counts violated. Everything else is cleanly and substantively verified: both mutation recipes independently replayed to guard_verified, the trailing-slash carve-out falsified, full repo suite 134/134.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[major]** (auto-fixable) `cogni-consult/tests/test_route_example_hygiene.sh:249 (review-contract:item-4)` — The references/**/*.md vacuity arm is keyed on directory presence, not the assembled subject, so a newly-covered surface can silently vanish while the case still prints PASS. Line 249 reads: [ -d "$PLUGIN_DIR/references" ] || missing_surface="$missing_surface references/". Measured on a scratch copy of the PR head: deleting every .md under cogni-consult/references/ while leaving the directory (and its methods/, orchestration/, personas/ subdirectories) in place yields PASS: no-retired-plugin-name, because $scan_bodies is still non-empty from skills/, README.md and CLAUDE.md. This is the exact class contract item 4 names in its purpose clause, and the exact class the author identified and fixed for skills/ in the second commit, where the arm was deliberately changed to test the ASSEMBLED SUBJECT with the stated rationale so it cannot drift from what is actually scanned. That reasoning was not carried across to the references/ arm. Not a regression against base (references/ carried no base coverage to lose), but an incomplete new guard on the surface that makes the trailing-slash carve-out falsifiable at all: references/publish-routing.md:298 is one of only two sites exercising it.
  - *Suggested fix:* Mirror the skills arm. Replace line 249 with a test against the assembled subject: printf %s\\n "$scan_bodies" | grep -q /references/.*\\.md$ || missing_surface="$missing_surface references/**/*.md". Verified working on a scratch copy: clean tree still PASSes; with every .md under references/ deleted and the directory left in place it FAILs naming references/**/*.md. Still exactly one result line per run. Add an F5c row to the falsification table for the emptied-but-present shape, as F8b already does for skills/.

- **[minor]** (waived) (auto-fixable) (advisory-addition) `PR body, Falsification performed, rows F5/F5b/F5c/F8/F8b` — The recorded falsification table quotes (vacuous scan of a newly covered surface) but the shipped code emits (vacuous scan - the subject would silently narrow). Behavior is correct and the rows are substantively true; only the quoted string is stale, captured before the cleanup pass reworded the arm. Flagged because a recorded falsification table is the artifact a future reader replays against.
  - *Justification:* PR body is the tertiary rubric; underlying behavior verified correct in all five arms; no committed checklist item constrains the table quoted strings. Not worth a bounce on its own.

### Positive observations

- Both recorded mutation recipes replayed independently to guard_verified (mutated red / restored green) against the flag-taking harness.
- The primary recipe is provably NOT green-at-base vacuous: applying its --expr to README.md while substituting the merge-base copy of the suite yields PASS, so it reddens only because of the widening.
- The trailing-slash carve-out is load-bearing: replacing grep -lE with a bare grep -lF turns the case RED on CLAUDE.md:182 and references/publish-routing.md:298.
- The mutated run reddens exactly the named case; all four siblings still PASS.
- Every subject path derives from $PLUGIN_DIR; --root redirects the whole subject, and a copy under a path containing a space still passes.
- The registry loader is byte-identical to base, all four rejection branches intact.
- Cases 1-4 are byte-identical to base (73 lines each, zero diff) and all four PASS.
- The skills/ arm is a real self-caught regression, fixed in a second commit and tested against the assembled subject.
- Header corrections complete: the carve-out paragraph sits beside the colon rationale, the stale scans-every-skill claim is gone, and the PR corrects the issue own false premise rather than inheriting it.
- The suite spells no retired prefix literally, so it cannot trip its own guard.
- The PR proactively discloses that check-preflight is inert here (6 of 7 checks skipped) rather than quoting a green gate as evidence.
- Plugin structure/registration valid and untouched; plugin.json version correctly left for the post-merge bump.
- Full repo suite 134/134, failed: 0.
- Items 8 and 9 are satisfied-but-vacuous and were flagged transparently rather than counted as substantive coverage.

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1637

**Summary:** Both dispatched perspectives (plugin-validator, contract-grading-reviewer) independently graded all 18 committed contract items satisfied against a38ebaca — 16 substantively, 2 vacuously (items 8/9, whose scripts/tests/ antecedent is never met) — and the prior verdict's one major finding (item 4's vacuity-arm gap) is verifiably fixed in both failure shapes: references/ removed AND references/ present-but-emptied each FAIL naming references/**/*.md, while the pre-fix f043c216 suite PASSes the emptied-but-present tree. The revision also removed a provably unreachable vacuity arm (verified against an empty plugin root) and named the third surface-registration site. The only remaining findings are four un-waived minors — two PR-body record-completeness notes, one pre-existing loader gap byte-identical to base, one latent unanchored-pattern class no tree in the repo exercises — which ride through as advisory rather than gating the merge. Note the instrument inertness: this suite is test_route_example_hygiene.sh (underscore) while check-preflight and run-tests glob test-*.sh, so both are inert here and the evidence is manual falsification, independently reproduced by both perspectives.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[minor]** `PR #1637 body — '## Cleanup pass', portability note` — The paragraph justifies BSD/GNU xargs empty-input safety by citing the `elif [ -z "$scan_bodies" ]` arm above — but this revision removed that arm. The property still holds via a different mechanism (an empty subject makes the skills/ grep fail, so $missing_surface is non-empty and the preceding fail arm wins), so no user-observable fault, but the recorded mechanism now names a deleted line.
  - *Suggested fix:* Re-point the note at the arm that now carries the property: '$missing_surface is non-empty whenever $scan_bodies is empty (the skills/ grep cannot match), so the else branch — and therefore xargs — is only ever reached with at least one file.'

- **[minor]** `PR #1637 body — '## Falsification performed' table` — The table was not refreshed for this revision: row F5c still records the observed token as `references/` (changed by this revision to `references/**/*.md`), and there is no row for the references-present-but-emptied-of-.md shape — the exact class this revision fixed. Both shapes do in fact fail (verified); this is a record-completeness gap, not a coverage gap.
  - *Suggested fix:* Update F5c's observed token to `references/**/*.md` and add an F5d row for 'references/ present but emptied of every .md' with its observed FAIL line, mirroring the existing F8/F8b pair for skills/.

- **[minor]** `cogni-consult/tests/test_route_example_hygiene.sh:205-221` — Pre-existing, out of this PR's scope. The registry loader's sys.exit(1) on a blank/padded/colon-bearing entry does not reliably produce the intended FAIL, because `python3 ... 2>/dev/null || true` swallows the exit code while keeping the stdout already emitted, so the suite proceeds on a silently truncated prefix list. This block is byte-for-byte identical to the merge base and the base suite exhibits the identical behaviour, so nothing regressed.
  - *Suggested fix:* File a follow-up (alongside #1636) rather than revising this PR: buffer the loader's output and discard it on a non-zero exit — out=$(python3 ...); rc=$?; [ "$rc" -eq 0 ] || out="" — instead of `2>/dev/null || true`.

- **[minor]** `cogni-consult/tests/test_route_example_hygiene.sh:263-268` — Both directory-glob vacuity arms match a pattern not anchored to $PLUGIN_DIR, so each can be satisfied by a path contributed by the other surface — a silent PASS on an emptied newly-covered surface, the same class item 4 exists to close. Reproduced both directions on scratch trees. Not introduced by a38ebaca (the skills arm carrying the same shape landed in f043c216), but this revision mirrors the shape onto references/. No such tree exists in cogni-consult today, so the real subject is unaffected.
  - *Suggested fix:* Strip the plugin-dir prefix before matching so the arm asks about a plugin-relative path, e.g. case "${p#"$PLUGIN_DIR/"}" in references/*.md), or conjoin a grep -qF -- "$PLUGIN_DIR/references/" prefix test with the existing '\.md$' suffix test. Apply the same anchoring to the skills/ arm in the same edit so the two do not diverge again.

### Positive observations

- The item-4 fix is falsifiable in both vacuity shapes and decisive rather than merely green: references/ removed and references/ present-but-emptied each FAIL naming references/**/*.md, while the pre-fix f043c216 suite PASSes the same emptied-but-present tree. (plugin-dev:plugin-validator)
- The removed `elif [ -z "$scan_bodies" ]` arm was verified unreachable rather than argued unreachable: a plugin root stripped of all four surfaces names all four in one line and never emits 'no scannable body found'. (plugin-dev:plugin-validator)
- Removing that arm restores the case-5 fail-site count to the merge base's 3; the one-result-line-per-run invariant held in all eight probed shapes. (plugin-dev:plugin-validator)
- The header now states the -d-vs-subject reasoning and why README.md/CLAUDE.md are deliberately not converted — the sibling-audit note that prevents the next editor from 'fixing' the two arms that are already correct. (plugin-dev:plugin-validator)
- The surface-registration comment now names three co-varying sites instead of two, the invariant whose violation let a surface vanish silently. (plugin-dev:plugin-validator)
- Both recorded mutation recipes replay to guard_verified under an independent snapshot dir, with the mutated case-5 line specifically FAIL: no-retired-plugin-name; the same mutation leaves the base suite PASSing, which is item 13's could-not-have-gone-red-before property. (plugin-dev:plugin-validator)
- Every measured claim in the PR body re-derived exactly: 44 cogni-claims occurrences, 11 not slash-followed, all confined to scripts/ and tests/ and outside the subject. (plugin-dev:plugin-validator)
- The four sibling cases and the shared preamble are byte-for-byte identical to the merge base and all four print PASS. (plugin-dev:plugin-validator)
- The PR discloses the instrument inertness up front rather than presenting a green gate as coverage; an independent preflight run confirms 6 of 7 checks skipped. (plugin-dev:plugin-validator)
- Item 4 satisfied: independently verified on scratch trees — skills/, references/, README.md and CLAUDE.md each FAIL with the vacuous-scan message in both the surface-removed and surface-present-but-emptied shapes. (cogni-service:contract-grading-reviewer)
- Item 5 satisfied: the removed arm was provably unreachable, so its removal is dead-code cleanup consistent with the reachable-call-site rule, not a violation of it. (cogni-service:contract-grading-reviewer)
- Item 17 satisfied: both recipes independently replayed to guard_verified with mutated red / restored green, and the mutated run's exact line is FAIL: no-retired-plugin-name ... cogni-visual. (cogni-service:contract-grading-reviewer)
- Items 8 and 9 are satisfied vacuously and are reported as such rather than counted as substantive coverage. (cogni-service:contract-grading-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Commit `a38ebaca` addresses the [previous verdict](https://github.com/cogni-work/insight-wave/pull/1637#issuecomment-5427325079). Fresh verdict: **✅ PASS** ([comment](https://github.com/cogni-work/insight-wave/pull/1637#issuecomment-5429893210)).

### Findings resolved

- **[major] `test_route_example_hygiene.sh` (review-contract:item-4) — the `references/**/*.md` vacuity arm keyed on directory presence.** Fixed in `a38ebaca` by mirroring the `skills/` arm: the arm now tests the assembled `$scan_bodies` subject, and the reported token is `references/**/*.md`, matching `$scan_desc`.

  Falsified in **both** vacuity shapes on a scratch copy, with the decisive control:

  | probe | fixed suite | pre-fix suite (`f043c216`) |
  |---|---|---|
  | `references/` removed | FAIL naming `references/**/*.md` | — |
  | `references/` present, emptied of every `.md` | FAIL naming `references/**/*.md` | **PASS** ← the defect |
  | unmutated scratch tree | PASS | PASS |

  The pre-fix column is the point: the fix is decisive, not merely green-at-base.

  **Sibling-arm audit** (the lesson from the last bounce — hardening one arm of an if/elif chain means auditing every sibling in the same commit): the `README.md` and `CLAUDE.md` arms are single files gated on `-f`, the identical test the subject assembly uses to admit them, so presence and inclusion-in-subject are equivalent and no emptied-but-present shape exists for them. Only the two directory-glob arms could diverge, and both now test the subject. This reasoning is now recorded in the suite header so the next editor does not "fix" the two arms that are already correct.

- **[minor, previously waived] PR body falsification table quoted a stale string.** Now corrected rather than left waived, because this revision rewrote that table anyway: `F5c`'s observed token is now `references/**/*.md`, a new `F5d` row records the emptied-but-present shape (mirroring `F8b` for `skills/`), and the stale `(vacuous scan of a newly covered surface)` quote now matches what the code actually emits.

### Beyond the finding — disclosed as scope

The pre-commit quality pass surfaced two further defects in the same chain, both applied in `a38ebaca` and both recorded in the body's Cleanup pass list:

- **Removed the `elif [ -z "$scan_bodies" ]` arm as provably unreachable.** An empty subject necessarily fails the `skills/` grep, so `$missing_surface` is non-empty and the preceding arm always wins. Verified, not argued: a plugin root stripped of all four surfaces names all four in one line and never prints `no scannable body found`. A vacuity arm that can never fire is unfalsifiable — the same green-gate class this case exists to catch. Restores the merge base's 3-fail-site shape, and both reviewers independently confirmed item 5 (one result line per run) is unaffected.
- **The surface-registration note named two co-varying sites; there are three** (the `find` list, `$scan_desc`, the `$missing_surface` chain). Registering a surface in the first two but not the third is exactly what let a surface vanish silently.

This is wider than the single finding. It is disclosed rather than buried so a reviewer can reject the extra scope if they disagree with it.

### Verification

Both instruments that would normally gate this are **inert here** and are not being quoted as coverage: this suite is `test_route_example_hygiene.sh` (**underscore**) while `check-preflight.sh`'s mutation-recipe check and `run-tests.sh` discovery both glob `test-*.sh` (**hyphen**). A green from either examined nothing on this file.

What actually ran: the manual falsification above · clean tree PASS with exactly one result line for case 5 · cogni-consult **13/13** suites green · full repo sweep **134/134, failed 0** · both recorded mutation recipes re-replayed to `guard_verified`, independently by both review perspectives.

### Not fixed — carried forward deliberately

Neither is a regression; both are recorded here rather than filed, since the plan deferred nothing and filing is scope this revise pass did not sanction:

- **Latent, not live:** both directory-glob arms match patterns not anchored to `$PLUGIN_DIR`, so each could be satisfied by a path contributed by the *other* surface (a skill literally named `references`, or a `references/skills/<x>/SKILL.md`). No such tree exists in any in-repo plugin today. Predates this revision — the `skills/` arm carried the shape from `f043c216`.
- **Pre-existing, byte-identical to base:** the registry loader's `python3 … 2>/dev/null || true` swallows a non-zero exit while keeping the stdout already emitted, so a registry whose *second* entry is malformed proceeds on a silently truncated prefix list. The base suite behaves identically, so contract item 3's "preserved" requirement holds and nothing regressed.

A maintainer who wants either tracked should say so and they will be filed.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1637

**Summary:** MERGER's independent review (origin: gatekeeper), graded against the committed review contract adopted by chronology (origin triage, derived 2026-08-26T13:43:11Z — before the branch's first commit at 14:38:19Z, so provably diff-blind). Both perspectives grade all 18 contract items satisfied: 16 SUBSTANTIVELY, 2 (items 8 and 9) only VACUOUSLY, their scripts//tests/ antecedent never met by the actual subject — so the substantive denominator is 16, not 18. The item-4 vacuity regression is closed DECISIVELY rather than merely greened: both reviewers independently reproduced every surface-loss shape on a faithful scratch replica (replicating the repo shape, since the suite derives REPO_ROOT as PLUGIN_DIR/.. — a plugin-only --root would have failed every probe on the registry arm and graded nothing), and the decisive control holds — the pre-fix suite at f043c216 PASSes the references-present-but-emptied shape while a38ebaca FAILs naming references/**/*.md. The author's extra scope (removing the `elif [ -z "$scan_bodies" ]` arm) was verified provably unreachable and is strengthening: it restores the merge base's exact 3-fail/1-pass call-site shape. Item 17's mandatory replay was performed by both reviewers: both recorded recipes return guard_verified with the mutated runs carrying FAIL: no-retired-plugin-name specifically. NO GREEN GATE WAS TREATED AS COVERAGE — check-preflight is 6-of-7 skipped (mutation-recipe and run-tests both inert on this suite's underscore naming), every skip enumerated by name. The two findings are pre-existing merge-base residuals, neither introduced nor widened here; both are demotable merger additions, annotated and non-gating.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (advisory-addition) `cogni-consult/tests/test_route_example_hygiene.sh:205-221` — The registry loader prints valid prefixes as it iterates and only then sys.exit(1)s on a malformed entry, while the shell wraps it in `2>/dev/null || true`. So a registry whose bad entry FOLLOWS a good one yields a TRUNCATED prefix list and the case PASSes (measured: ["cogni-wiki", "cogni-visual:"] -> PASS; the same registry with the bad entry first -> FAIL). This block is BYTE-IDENTICAL to the merge base (diff empty), so the behaviour is preserved exactly as item 3 requires and is in no way introduced here — item 3 is graded satisfied on that basis.
  - *Suggested fix:* Out of scope for this PR. If addressed later, collect prefixes into a list and print only after the whole registry validates, so a partial parse cannot present as a usable list.

- **[minor]** (advisory-addition) `cogni-consult/tests/test_route_example_hygiene.sh:262-268` — The vacuity chain guards surface PRESENCE, not READABILITY or non-emptiness. A README.md that is present but zero-byte, or present but mode 000, still satisfies `[ -f ]`, still enters $scan_bodies, and the case prints PASS (both shapes measured). Deliberately NOT graded as an item-4 violation — that item's concern, fixed by its own parenthetical ('in particular an absent README.md'), is the subject silently NARROWING, and in these shapes the file is genuinely in the subject and genuinely handed to grep, so the vacuity arm and the scan subject remain equivalent exactly as the author's sibling-audit comment argues. The same property holds for every scanned file at the merge base, so nothing regressed. Recorded so the residual is written down rather than rediscovered.
  - *Suggested fix:* No change required for this PR. If ever worth closing, fold it into the #1636 follow-up as a content-level (non-empty / readable) assertion rather than a surface-registration arm, so the three-site coupling note stays accurate.

### Positive observations

- The item-4 fix is falsifiable in the direction that matters: the pre-fix suite (f043c216) PASSes the references-emptied-but-present shape while the fixed suite FAILs naming references/**/*.md — the control separating 'the arm is green' from 'the arm is decisive'. Reproduced independently by both perspectives. (plugin-dev:plugin-validator)
- The extra scope was self-disclosed AND mechanically provable: a plugin root stripped of all four surfaces emits all four names and never the removed arm's diagnostic. Removing it restores the merge base's 3-fail/1-pass call-site shape rather than inventing a new one. (plugin-dev:plugin-validator)
- The sibling audit is correct as measured, not asserted: only the two directory-glob arms could diverge and both now test the subject; README/CLAUDE use `-f`, the identical predicate the assembly uses to admit them, so presence and inclusion are provably equivalent. (plugin-dev:plugin-validator)
- Item 13's 'would not have gone red before the widening' is established by direct control: the recipe-1 expr applied to README.md leaves the BASE suite PASSing and the branch suite FAILing on cogni-visual. (plugin-dev:plugin-validator)
- The scripts//tests/ exclusion is measured and reproduces to the digit: 44 cogni-claims occurrences, 11 surviving the carve-out, all 11 in the three named files. The body also corrects the issue's own false Motivation on the issue thread rather than quietly building on it. (plugin-dev:plugin-validator)
- The PR body pre-emptively and accurately names its own inert gates, and that disclosure matched what was measured — the opposite of the green-gate-that-examined-nothing failure mode. (plugin-dev:plugin-validator)
- Zero hardcoded `cogni-consult` literals in the suite; --root redirection confirmed empirically from an unrelated absolute path. (plugin-dev:plugin-validator)
- Change confined to a single file: `git diff --name-only 00c54749 HEAD` returns only the test file — none of retired-plugins.json, check-external-dispatch.py, README.md or skills/** was touched to make the case pass. (plugin-dev:plugin-validator)
- item 4 satisfied: all six surface-loss shapes replayed against isolated scratch copies — skills/ removed, skills/ emptied, references/ removed, references/ emptied, README.md removed, CLAUDE.md removed — each correctly emits the vacuous-scan FAIL naming the missing surface. (cogni-service:contract-grading-reviewer)
- item 5 satisfied: 3 fail()/1 pass() call sites in one if/elif/elif/else chain — byte-identical count and structure to the merge base's own shape, so exactly one result line fires per run whichever branch is taken. (cogni-service:contract-grading-reviewer)
- item 17 satisfied: both recipes replayed live returning guard_verified; manual scratch mutation confirmed the mutated FAIL line names no-retired-plugin-name specifically, not a sibling case. (cogni-service:contract-grading-reviewer)
- items 8 and 9 are satisfied VACUOUSLY and reported as such rather than counted as substantive coverage — the substantive denominator is 16. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #33, opened 2026-04-15
cogni-research revisor: expansion mode duplicates evidence across sections


sdh07:
## Observed behavior

When the cogni-research revisor (`cogni-research/agents/revisor.md`) runs in word-deficit expansion mode, it adds evidence to multiple sections by copying the same metric, quote, or case study **verbatim** into each expanded section rather than cross-referencing.

Low-severity expansion-duplication issues have now been flagged by the reviewer in two consecutive runs on the same corpus (`kiadoption-im-anlagen-und-maschinenbaumi-2026-04-14`):

**Prior run** (v0.7.1 singlepass chain, pre-citation-density-gate):
> TRUMPF '70% remote Fehlerbehebung via Visual Assistance' metric appears verbatim twice: once at the end of ch. 3.4 (STADLER comparison paragraph) and again at the start of ch. 4.2 (Praxisbeispiele).

**New run** (v0.7.5, with citation density gate on branch `feat/issue-25-reviewer-citation-density-gate`):
> OPC-UA infrastructure gap (80% of KMU lack OPC-UA), GFE IIoT retrofit cost band (5,000–15,000 EUR, 2–3 year payback) and Edge Computing Fallstudien (50% Ausfallzeit-Reduktion, 40% Lebensdauer-Verlängerung) appear verbatim in both section 05.3 and section 08.2. The Weidmüller Industrial AutoML paragraph also appears in both 05.3 and 09.3.

Both issues were emitted as **low-severity** by the reviewer, so the drafts shipped. They consistently drag coherence and clarity: on the new run, coherence slipped 0.83 → 0.80 and clarity 0.88 → 0.85 between reviewer iterations despite the score going up on other dimensions. The pattern is recurring and specific to expansion mode — not a general revisor behavior.

## Why this happens

The revisor's current expansion logic hits sections independently and greedily reuses the strongest available evidence in each. There is no shared \"facts already used in this draft\" state across the expansion pass, so the same high-value quote lands in every section where it plausibly fits.

## Proposed fix

Expansion mode should track a per-draft evidence ledger and prefer cross-references over restatement when a fact is already placed.

Shape of the change (in `cogni-research/agents/revisor.md`, expansion-mode Phase 2):

1. Before expanding, scan the existing draft for inline citations and extract a \"facts already placed\" index — key metric, source, and section where it appears (e.g. `TRUMPF 70% remote → § 3.4`).
[...]

PR #1704, opened 2026-08-30
cogni-workspace: suggest_hex evaluates the L=0.0 / L=1.0 endpoints

### LINKED ISSUE #1684: cogni-workspace: suggest_hex's lightness walk stops one step short of the endpoints, so it reports no suggestion where black or white would clear
## Problem

`suggest_hex()` in `cogni-workspace/scripts/check-contrast.py` walks the foreground's lightness at
constant hue looking for a shade that clears the threshold:

```python
candidate_lightness = lightness
while 0.0 <= candidate_lightness <= 1.0:
    candidate_lightness += direction * step   # step = 0.01
    if not 0.0 <= candidate_lightness <= 1.0:
        break
```

Because the accumulator is built by repeated `+= 0.01` on a float, it essentially never lands
exactly on `0.0` or `1.0` — it overshoots by a rounding error and the guard breaks out. So the walk
**stops one step short of both endpoints** and never evaluates pure black or pure white, the two
colours most likely to clear a contrast threshold.

The consequence is a false negative: the tool reports "no suggestion" for pairs where a valid
suggestion exists.

## Witness

Foreground `#333333` on background `#757575`, at the 4.5:1 normal-text threshold:

- `suggest_hex()` returns `None`, so the emitted pair carries `"suggested_hex": null`.
- The best effective ratio the walk actually reaches is **~4.491** — just under the 4.5 threshold.
- Pure black `#000000` against that background scores **4.558**, which clears. The walk simply
  never evaluates it.

This is not specific to that pair. Across every fixture probed, the best ratio the walk reaches
lands at ~4.491 against a 4.5 threshold — the shortfall is the missing final step, not the palette.

## Why the null branch exists at all

A coarse sweep over 46,656 colour pairs finds 21 null-returning cases at the 4.5 threshold and
**zero** at 3.0. Every one of those 21 is a near-miss of this shape. That is the substantive
question this issue has to settle: the `return None` arm may be reachable *only* as an artifact of
the truncated walk, in which case fixing the walk could make it dead code.

## Acceptance criteria

- [ ] `suggest_hex()` evaluates the lightness endpoints `L=0.0` and `L=1.0` exactly, rather than
      breaking out one step short because float accumulation never lands on them
- [ ] For `#333333` on `#757575` at threshold 4.5, `suggest_hex()` returns a hex that clears
      4.5:1 (pure black scores 4.558) instead of `None`
- [ ] A test case pins the endpoint behaviour directly — that a pair whose only clearing shade is
      an endpoint gets that endpoint suggested, asserted on the returned hex rather than on
      non-nullness alone, so it cannot grade vacuously
- [ ] The reachability of the `return None` arm is re-derived after the fix and the outcome
      recorded: if a pair that still reaches it at the 4.5 threshold exists, the existing
      `cc31-*` fixture is re-pointed onto one; if none exists, the arm and its case are retired
      together, with the finding recorded, so the suite does not carry a case pinning dead code
- [ ] The full suite `bash cogni-workspace/tests/test-check-contrast.sh` exits 0

## Context

Surfaced while adding the `cc31-*` coverage for the `return None` branch. Deliberately not fixed
there: that change is fenced to `cogni-workspace/tests/**`, and fixing the walk would have
invalidated the very fixture the coverage issue asked for.

## Unblock when
#1659 closes

claude-cogni-work:
Closes #1684.

## Summary

`suggest_hex()` walked the foreground's lightness with an accumulating
`candidate_lightness += direction * step` gated on `while 0.0 <= candidate_lightness <= 1.0`. The
accumulator overshoots rather than landing on a bound — it terminates at `-3.1e-17` and
`1.0000000000000007` — so the guard broke out one step short and **pure black and pure white were
never scored**, producing a false "no suggestion" for pairs an endpoint would clear.

Each direction's exact endpoint is now scored once that direction's stepped walk is exhausted.

- Witness from the issue: `#333333` on `#757575` at 4.5 returns **`#000000`** (4.5578) instead of `None`.
- Lines 162-174 of the walk are **byte-for-byte unchanged**, so the existing try order and the
  snapped self-re-verification are preserved. No loop clamps back into range and re-tests its own
  guard, so the walk keeps its monotone bound — a hang would be worse than a failure here, since
  macOS ships no `timeout`.
- `suggest_hex` is the only function whose AST changes. `relative_luminance`, `contrast_ratio`,
  `to_hex`, `parse_hex`, `evaluate_pair`, `AA_NORMAL` and `AA_LARGE` are untouched (verified by AST
  comparison against `origin/main`, not by eye).

## This changes answers for pairs that already work today — please read this part

Scoring the `-1` endpoint before the `+1` walk runs makes the black endpoint win a tie it used to
lose by never being reached. Measured on **the same 46,656-pair grid the issue itself used** — 6
levels per channel `{0,51,102,153,204,255}`, 216 colours, all ordered fg/bg pairs — comparing
`origin/main` against this branch:

| threshold | below-AA pairs | `None` → hex | already-answered pairs that change | share |
|---|---|---|---|---|
| 4.5 | 38,594 | **21** | **194** | 0.503% |
| 3.0 | 31,652 | 0 | **300** | 0.948% |

The `21` and the `0` reproduce the issue's own reported counts exactly, which is the check that the
grid is the same one. Worked example: fg `#050505` on bg `#757575` at 4.5 returns `#FFFFFF` today
and `#000000` after this change.

**This is inherent to correctly fixing the bug, not a property of the chosen shape.** Avoiding it
would mean fixing only the `+1` endpoint — an arbitrary asymmetry. Every changed answer is still a
hex the module re-verified after snapping, so what changes is *which of two clearing shades* is
offered, not whether it clears. `BELOW_45` (`#777777` on `#FFFFFF`, backing cc13/cc14) is
unaffected: `#747474` before and after. Note the change reaches the 3.0 large-text threshold too,
where there was no bug to fix — that is the same tie-break, and it is why the figure is stated per
threshold rather than as one number.

## Why `return None` is kept — a deliberate deviation from AC4

AC4 says: *"if none exists, the arm and its case are retired together"*. No pair reaches the arm at
either CLI threshold post-fix, so that branch is the one that fires — but executing it literally
would be wrong, for two independent reasons:

1. **`threshold` is a caller-supplied parameter.** For any background the two endpoint ratios
   multiply to 21 in exact arithmetic, so the larger is always at least `sqrt(21) ~= 4.583`. Both
   thresholds the CLI exposes (`AA_NORMAL` 4.5, `AA_LARGE` 3.0) sit below that, but a caller passing
   a stricter threshold — a future WCAG AAA 7:1 — still reaches the arm. It is not dead code.
2. Deleting it would leave `suggest_hex` falling off the end implicitly, replacing an explicit
   terminating `return` with an implicit one.

So the **case** is retired and re-pointed; the **arm** stays; and the reachability finding is
recorded in the repo *with its bound* rather than as an unconditional "dead code" note. The bound is
**strictly** above `sqrt(21)` — at exactly `sqrt(21)` the `>=` comparison still clears.

New `cc33-null-arm-is-reachable-above-sqrt21` pins the arm by direct call at threshold 21.0, where it
is genuinely live. Mutation recipe 4 is re-pointed onto it.

## Test surface

- `cc31`'s three ids keep their **number** (the suite's own rule — "the printed label ... neither
  reads file order. Do not renumber." — cited by text, not line, because the line has drifted
  twice under comment edits) and
  their unchanged expected values; their suffixes move off `null-suggestion`, which named a
  behaviour the post-fix script cannot exhibit.
  `cc31-null-suggestion-is-present-and-null` → `cc31-endpoint-suggestion-is-present-and-black`,
  now asserting `(True, '#000000', 1)` on the **returned hex**, not on non-nullness — the membership
  test that distinguishes it from `cc15` is preserved.
- **New `cc32-white-endpoint-is-suggested`** — `suggest_hex('#333333', '#000000', 21.0) == '#FFFFFF'`,
  a direct call because 21.0 is not a threshold the CLI exposes. 21.0 is used because against a black
  background the stepped walk's best **snapped** candidate measures `20.4689` at `#FCFCFC`, so the
  isolating window is `(20.4689, 21.0]`. That window is **measured at the walk's current `0.01`
  step, not derived**. Its step-sensitivity is now **executed rather than narrated** — see the
  ruling below. Pre-fix this returns `None`, so the case is genuinely red at base.
- **New `cc33-null-arm-is-reachable-above-sqrt21`** — green at base *and* post-fix **by design**. It
  pins the arm's continued reachability, not the endpoint fix; its teeth come from re-pointed
  mutation recipe 4. Flagged so it is not mistaken for a vacuous addition.
- Both new cases grade through `assert_eq`. That is load-bearing: `scripts/check-case-id-pairing.py`
  is wired into CI at `.github/workflows/lint.yml:164` and sits **outside** this change's path fence,
  so it must stay green without being edited. `assert_eq` emits paired PASS/FAIL ids by construction;
  a one-armed `... || fail "cc32-..."` would be flagged as an unpaired FAIL id. Verified by running
  the guard: 23 emissions before and after.
- Mutation recipe 4's `--expr` anchor is retained and is still single-occurrence, but the `$` is what
  makes it so, not the indentation: this change's new docstring adds a second four-space-indented
  line beginning `return None` at `check-contrast.py:172` (the docstring's closing words,
  `    return None.`), alongside `suggest_hex`'s arm at `:208`. `^    return None$` still matches
  exactly one line because the trailing period is excluded (`grep -cE '^    return None$'` returns
  `1`). `parse_hex`'s two at `:129`/`:132` remain eight-space and are still excluded by `^    `.

## Verification run

- `bash cogni-workspace/tests/test-check-contrast.sh` → exit **0**, "All check-contrast tests passed.", zero `FAIL:` lines.
- `run-tests.sh --dir cogni-workspace/tests` → **20 discovered, 20 passed, 0 failed, 0 skipped** (accounting identity holds).
- `check-case-id-pairing.py` → exit 0. `check-result-line-plainness.py` → exit 0.
- `check-frontmatter.sh cogni-workspace` → 49 files scanned, 0 violations. `check-breadcrumbs.sh cogni-workspace` → 0 violations.
- Endpoint snap identity probed across 510 hue/saturation/endpoint combinations (see Review notes).

## Mutation recipe

Guards the endpoint evaluation this PR adds. The mutant scores the endpoint but never returns it —
exactly the pre-fix behaviour — so the witness falls back to `None` and the named case goes RED:

```bash
  --root . \
  --file cogni-workspace/scripts/check-contrast.py \
  --expr 's{return to_hex\(endpoint\)}{pass}' \
  --test 'bash cogni-workspace/tests/test-check-contrast.sh' \
  --case cc31-endpoint-suggestion-is-present-and-black
```

The anchor is single-occurrence: the only other `to_hex(endpoint)` sits inside
`parse_hex(to_hex(endpoint))`, which the `return ` prefix excludes. `pass` preserves indentation, so
the mutant stays syntactically valid. Both this recipe and the re-pointed recipe 4 ship in the suite
header.

## Review notes

**A quality-pass finding was rejected on evidence, and the evidence is recorded in the code.** Two
independent cleanup reviewers proposed collapsing the endpoint's snap-and-re-verify as a provable
identity, on the premise that `hls_to_rgb(hue, 1.0, saturation)` is exactly `(1, 1, 1)` for every
hue and saturation. **It is not** — at saturation 0.13 it returns `(0.9999999999999999, 1.0, 1.0)`;
52 of 510 probed combinations miss. I could not find a case where the difference flips a
comparison, so this is not a live bug — but the argument for removing the guard does not hold, and
"no counterexample found" is not a reason to drop a guard. It is kept, and a comment at the endpoint
block records the counterexample so the next reader does not repeat the deduction.

Extracting the duplicated score/snap sequence into a shared helper was also rejected: it would
delete this PR's mutation anchor (`return to_hex(endpoint)`) and make recipe 4's
`^    return None$` anchor multi-occurrence, destroying the single-occurrence property both recipes
assert. Four duplicated lines is the cheaper trade.

**Pre-existing skill-quality finding, not introduced here and not fixed here.** The
`Contrast & Accessibility` subsection of `manage-themes/SKILL.md` carries ~9 bullets of 300-900
chars each of per-field reference material that progressive disclosure would move to `references/`.
The file is at 85.2% of its measured cap (`chars_body` 26832 / `cap_chars` 31500, `over_cap` false),
so nothing is currently violated. This PR changes exactly one line of that file; refactoring the
subsection would breach the change's path fence.

## Operational disclosure

`check-token-scope.sh --strict` reports `status: over_scoped` on this runner —
`forbidden_scopes: ["workflow"]`, `extra_scopes: ["gist", "read:org", "workflow"]` — and exits 1 by
default. This drain cleared Step 7.5 with the **`--allow-overscoped` flag** (per-invocation, not the
`COGNI_SERVICE_ALLOW_OVERSCOPED` env var, which leaks into nested `run-tests` and false-blocks
`check-preflight`), per the operator's standing authorization for this personal, dedicated runner —
the condition `cogni-service/CLAUDE.md` §Operational hardening states for the opt-out. The status is
surfaced, not silenced. Re-scoping the runner to a fine-grained PAT remains the better end state and
would remove this exception entirely.

## Maintainer ruling — the endpoint/walk narration is cut to executed assertions

This PR bounced six times on one comment passage, never on the code. Five automated
revision cycles each retired one false claim in it and shipped a new one. Ruling by
@sdh07 at head `8fc32153`, taken after independently reproducing both open majors by
execution rather than by reading the verdict:

| Claim that was in the comment | Measured against the head module |
|---|---|
| "at a step of 0.001 the walk … makes the case vacuous" | **False.** The walk gates on the **unsnapped** ratio at `check-contrast.py:183` before snapping. At step 0.001 its best gating value is `20.9545` < 21.0, so zero candidates pass and cc32 keeps its teeth. The real vacuity risk is at **coarse** steps — 0.1 / 0.16 / 0.2 land on `L≈1.0` with gate value exactly `21.0`. |
| "white is the answer for only 9 … because the -1 walk and the black endpoint are both scored first" | **False.** Attribution over the 12,523 white-clearing failing pairs at 4.5: `walk+1` 12,303, `endpoint-1` 206, `endpoint+1` 8, `walk-1` 6. The two named mechanisms are 1.7%; the unnamed `+1` walk is 98.2%. At 3.0 the ordering inverts. |

**The diagnosis is the passage, not the wording.** Every one of the six failures shared
one shape: a derived, quantitative, whole-input-space assertion about mechanism that *no
assertion in the suite executes*. Nothing bound it, so it could only be checked by an
ad-hoc sweep — and each such sweep was itself error-prone (the fifth cycle's own
step-0.001 probe silently re-implemented the walk and dropped the unsnapped gate). It was
also structurally stale-prone: any change to `step`, gate order or direction order
invalidates a dozen numbers nothing re-checks. That is this repo's own standing rule —
a second copy is a second thing to drift.

**What was cut** (comment-only; no executable line touched):

- Section F: the 46,656-pair tie-break statistics, and the `sqrt(21)` derivation — the
  latter already lives in `suggest_hex`'s docstring and is pinned by `cc33`, so the
  comment now points at it instead of restating it a third time.
- The cc32 block: the CLEARING-vs-WINNING statistics, the `#000033` on `#3366FF` witness
  paragraph, and the step-0.001 example.
- The cc32/cc33 preamble's "two DIFFERENT reasons" narration, which restated Section F.

**What replaced the one real hazard.** The deleted prose was trying — twice, falsely — to
warn that cc32's `21.0` threshold is step-sensitive: at a coarse step the walk clears 21.0
itself and cc32 goes vacuously green. That warning is now **executed**. Mutation recipe 5
gains a second documented discriminator, `--case cc32-white-endpoint-is-suggested`.
Verified against the head module:

| Module | cc31 @4.5 | cc32 @21.0 | cc33 @21.0 |
|---|---|---|---|
| head | `#000000` | `#FFFFFF` | `None` |
| recipe-5 mutant (`return to_hex(endpoint)` → `pass`) | `None` | `None` | `None` |

Under that mutant the endpoint branch cannot answer, so a still-green cc32 **is** the
vacuity report. Both recipe-5 cases and recipe 4 return `guard_verified` at this head.

**The sweep evidence is not lost** — it remains in this body's "This changes answers for
pairs that already work today" section, which is a snapshot of the change rather than a
live claim a future reader would re-derive.

**`manage-themes/SKILL.md:121`** takes the same discipline: the precedence clause now
states what follows from the loop structure (`for direction in (-1, 1)` — the darker
direction's walk *and* its endpoint both run before `+1`) rather than from a tie-break
statistic. This closes the precision minor both `plugin-dev:skill-reviewer` and
`cogni-service:skill-quality-reviewer` raised at head `6e352db5`.

**`review-contract:item-16`** — recorded here as its own suggested fix asks: the whole
`cc31` family was renamed together for consistency with item 13, which the contract's risk
watchlist affirmatively mandates. Expected values are preserved verbatim, `cc31` keeps its
number (item 17), and a repo-wide grep for `cc31-null-suggestion` returns zero survivors.

**Verification at `8fc32153`:** `test-check-contrast.sh` exit 0, zero `FAIL:` lines ·
`run-plugin-tests.py --filter cogni-workspace` 20 discovered / 20 passed / 0 failed /
0 skipped · emitted case ids byte-identical to head (39 ids, 39 emissions, no collisions) ·
`check-case-id-pairing.py`, `check-result-line-plainness.py`, `check-breadcrumbs` all
exit 0 · `check-frontmatter` 49 files, 0 violations · both mutation recipes
`guard_verified`, both anchors still single-occurrence · every surviving number in the
comments re-verified by execution (`2.7422`, `-3.122e-17` / `1.0000000000000007`,
`20.4689` at `#FCFCFC`, `4.5578` / `4.6075`) · no `plugin.json` or `marketplace.json`
version touched.


claude-cogni-work:
## Ship-ready

Author drain complete. This PR is a **draft** carrying `svc:needs-review` and is handed off to the
merger instance for independent review. The author instance deliberately posted no review verdict,
applied no `svc:review-*` label, and did not flip the PR ready.

### Blocking handoff preflight — clean

`check-preflight.sh` exit **0**, `clean: true`, `touched_plugins: ["cogni-workspace"]`,
`gating_applicable: 5`, 6 passed / 1 skipped / 0 violations.

| instrument | status | detail |
|---|---|---|
| run-tests | pass | suites green for cogni-workspace |
| frontmatter | pass | clean for cogni-workspace |
| breadcrumbs | pass | clean for cogni-workspace |
| skill-length | pass | measured 1 of 1 file(s); 0 over cap |
| mutation-recipe | pass | recorded recipe runnable as written — `cc31-endpoint-suggestion-is-present-and-black` found in the suite |
| agent-length | **skip** | `not_applicable` — no touched `agents/*.md` |

The one skip is named rather than swallowed: a preflight that exits 0 having graded nothing is
indistinguishable from a clean one unless the skips are listed.

### Author self-grade against the issue-time contract — 19/19, advisory

Graded against the contract posted on #1684 at `2026-08-30T06:50:59Z` (`origin: triage`,
`target: issue`) — written before any branch was cut, so it is diff-blind by chronology. All 19
items **satisfied**, no violations, no `unclear`, **no residual**. Both mutation recipes were
verified by hand-applying the mutation and confirming the named case flips to FAIL, rather than by
inspection.

This grade is advisory and carries no verdict label. The merger's own independent contract stands
on its own.

### Two things the merger should look at directly

1. **This PR changes answers for pairs that already worked.** 194 of 38,594 below-AA pairs at
   threshold 4.5 (0.503%) and 300 of 31,652 at 3.0 (0.948%) get a different — still re-verified —
   suggestion. Measured on the issue's own 46,656-pair grid, which reproduces its reported "21
   nulls at 4.5, 0 at 3.0" exactly. It is inherent to fixing the bug, and it is disclosed in the PR
   body rather than buried.

2. **The PR deliberately deviates from the issue's AC4.** AC4 says to retire the `return None` arm
   and its case together; this PR retires the *case* and keeps the *arm*, because `threshold` is a
   caller-supplied parameter and the arm stays reachable strictly above `sqrt(21) ~= 4.583`. The
   reasoning is argued in the PR body, not asserted as compliance. If the merger reads AC4 as
   binding literally, that is a legitimate disagreement to raise.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1704

**Summary:** Merged verdict is needs_revision: plugin-dev:plugin-validator found a major, un-waived defect on review-contract item-13 (two stale null-suggestion comments in test-check-contrast.sh no longer match the post-fix runtime behavior at lines 325 and 332), which gates the PR; the orchestrator's own line-level verification confirmed plugin-validator's reading and superseded contract-grading-reviewer's narrower grep-based 'satisfied' grading of the same item. All remaining findings across the other three perspectives are minor preference-level phrasing issues, one of which is explicitly waived with a written justification.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[major]** `cogni-workspace/tests/test-check-contrast.sh:325,332` — review-contract item-13 VIOLATED: two surviving comments still claim a null-suggestion behaviour the post-fix script cannot exhibit, and they head the exact cases this PR renamed away from null. Line 325 reads '# Present-and-null, NOT the absent key cc15 pins: ... a bare .get() renders both states as None and would grade vacuously -- the membership test is the discriminator', but the case below now asserts (True, '#000000', 1); both halves are false, since a bare .get() now returns '#000000' for present vs None for absent and therefore DOES discriminate. Line 332 reads '# A null suggestion is data, not an error.' heading a case whose fixture now yields '#000000'. Adjudicated by the orchestrator against the reviewed head: contract-grading-reviewer graded this item satisfied on a grep for the hyphenated token 'null-suggestion' (0 hits), but the stale claims use the bare word 'null'.
  - *Suggested fix:* Rewrite 325 to describe present-and-black and restate why the membership test is still co-asserted (it guards an evaluate_pair regression that stops emitting the key at all); rewrite 332 to 'A failing pair with a suggestion is data, not an error.' Optionally rename the NULL_SUGGESTION variable and the nullsuggestion palette slug, which nothing outside this file references.

- **[minor]** (advisory-addition) `cogni-workspace/scripts/check-contrast.py:170-172` — The docstring's closing sentence is universal where the other two recordings state reachability: 'a value strictly above ~4.583 ... still returns None' reads as 'always returns None', which is false — suggest_hex(parse_hex('#333333'), parse_hex('#FFFFFF'), 7.0) returns '#303030' (verified). The code comment and the test comment both correctly say 'reachable only strictly above'.
  - *Suggested fix:* Change 'still returns None' to 'can still return None'.

- **[minor]** (advisory-addition) `cogni-workspace/skills/manage-themes/SKILL.md:121` — Two independent phrasing issues in the same threshold-scoping sentence: (a) the residual-null clause scopes the null arm to 'a caller driving the module at a stricter threshold than these two', but strictness alone is not the condition — the bound is ~4.583, so a caller at 4.55 is stricter than 4.5 yet can never see null; (b) 'At the two thresholds this script grades' contradicts the same bullet's closing clause granting a caller may use a stricter threshold — the script does not grade at only two thresholds, its CLI exposes two while the module accepts any.
  - *Suggested fix:* (a) Name the bound: '... for a caller driving the module strictly above ~4.583:1.' (b) Reword to 'At the two thresholds this CLI exposes', matching suggest_hex's own docstring.

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/manage-themes/SKILL.md:273-274` — A dated point-in-time maintenance note in the executed prompt body restates the auto-inject policy stated two paragraphs earlier.
  - *Justification:* pre-existing whole-skill drift, introduced_by_change false; file is under its measured cap (26807/31500)

### Positive observations

- 18 of 19 committed review-contract items (1-12, 14-19) graded satisfied by plugin-dev:plugin-validator, with item-19 independently corroborated by plugin-dev:skill-reviewer.
- Both mutation recipes were executed live by two independent perspectives and returned guard_verified (mutated=red, restored=green).
- Every quantitative claim recorded in the diff reproduced exactly on independent re-derivation, including the disclosed tie-break sweep: 21 None->hex at 4.5, 0 at 3.0, 194/38594 and 300/31652 changed-existing.
- The behaviour change beyond the reported bug (~500 already-suggesting pairs now getting a different, still re-verified shade) is disclosed with counts and a worked example rather than buried.

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1704

**Summary:** Two perspectives pass cleanly, but plugin-validator's independently-verified major finding — a factually false claim in test-check-contrast.sh:366 about which endpoint is reachable/decisive, contradicted by a plain CLI run and by this same PR's own SKILL.md:121 wording — promotes the merged verdict to needs_revision. All four findings from the previous bounce are confirmed resolved.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[major]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:366` — FALSE claim about present behaviour, change-introduced by this PR — the mirror of the finding that bounced the previous head. Line 366 reads 'The L=1.0 endpoint is not decisive at either CLI threshold -- black already clears whenever white barely does, since the two ratios multiply to 21', and lines 349-350 extend it to 'neither the white endpoint's decisiveness nor the None arm's reachability is observable through a palette run'. Both are falsified by a plain CLI run at head 1ee4ac35: fg #000033 on bg #3366FF at 4.5 returns suggested_hex #FFFFFF from the direction=+1 ENDPOINT branch (black endpoint scores 4.4853 and FAILS 4.5; white scores 4.682 and clears). The stated inference is backwards: the multiply-to-21 identity guarantees white clears precisely WHEN BLACK FAILS. 8 such pairs exist at 4.5 in this PR's own 46,656-pair sweep. It also contradicts SKILL.md:121, which this same PR writes ('it can be an extreme one: #000000 or #FFFFFF'). Hazard: a maintainer reading line 366 could conclude the white-endpoint branch is unreachable and simplify it away — the exact failure the adjacent 'Do not simplify either check away' comment guards against.
  - *Suggested fix:* Reword to the true reason cc32 needs a direct call: the white endpoint IS reachable and decisive through the CLI, but a palette run cannot ISOLATE it from the stepped walk, whereas a direct call at 21.0 against a black background can (the walk's best snapped candidate there is 20.4689 at #FCFCFC). Narrow lines 349-350 so the 'not observable through a palette run' claim and its section-F citation apply to the None arm (cc33) alone. The correct supporting fact is the inverse of what is written: whenever BLACK barely clears, WHITE also clears, so it is the DARK endpoint that is never the sole clearing one at CLI thresholds.

- **[minor]** (auto-fixable) `cogni-workspace/skills/manage-themes/SKILL.md:121` — Prose-precision only, no user-observable fault: the residual-null clause's necessary-condition reading rests on the comma alone; inserting 'only' would foreclose a reader inferring that any above-bound caller gets null. The current wording is what the merger prescribed verbatim, so this must not gate.
  - *Suggested fix:* Optional: '... is still null-able in the JSON contract, but only for a caller driving the module strictly above ~4.583:1.' Leave as-is to preserve the verbatim merger wording.

- **[minor]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:307` — Over-broad universal in the sweep record. 'always a near-white shade becoming #000000' quantifies over both cohorts. All counts reproduce exactly and the '-> #000000' half is universally true, but 'near-white' holds only for the 4.5 cohort; in the 3.0 cohort 294 of 300 pre-fix answers sit below L=0.8 (lowest #B2B300, L=0.351).
  - *Suggested fix:* Scope the clause per cohort, e.g. 'always becoming #000000, displacing a near-white shade at 4.5 and a mid-lightness one at 3.0 (lowest observed #B2B300, L=0.35)'.

- **[minor]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:331` — The revision's rewritten cc31 rationale mildly misdescribes the assertion below it: it argues '.get() would go quiet-None', but the assertion indexes directly (data['pairs'][0]['suggested_hex']), so a missing key raises KeyError inside field() and yields an empty capture, reddening the hex half too. The co-assertion is still worth keeping; only the reason given is off.
  - *Suggested fix:* Rewrite as: 'The membership test is co-asserted anyway so that an evaluate_pair regression which stops emitting the key names itself explicitly (False in the tuple) rather than surfacing as an empty capture from the KeyError that direct indexing would raise.'

### Positive observations

- All four findings from the previous bounce are confirmed resolved: the two stale null-suggestion comments rewritten, the docstring universal weakened to 'can still return None', both SKILL.md phrasings corrected, and the waived item correctly left alone.
- The concept-grep discipline the previous bounce demanded was actually applied: every surviving null/None hit was judged individually, and nothing true or historical was falsified.
- The revision's one addition (the NULL_SUGGESTION rationale) is factually verified — the fixture pair is byte-identical to cc33's direct-call pair.
- Both mutation recipes re-executed at this head: guard_verified, mutated red / restored green.
- Every numeric claim spot-checked reproduces exactly, including the full 46,656-pair sweep.
- All five acceptance criteria met; AC4 resolved by a documented, correct third path.
- check-contrast.py lines 162-174 byte-for-byte fence respected; no plugin.json version touched.
- Full cogni-workspace suite is 20/20 PASS; frontmatter and version-bump pre-gates both verified-clean.

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the automated verdict at https://github.com/cogni-work/insight-wave/pull/1704#issuecomment-5467497312 (NEEDS_REVISION). Commit `1ee4ac35` — comment/prose only, no executable line changed.

### Findings addressed (4/4)

- **[major] `cogni-workspace/tests/test-check-contrast.sh:325,332`** — review-contract item-13. **Resolved** in `1ee4ac35`. Line 325 now describes *present-and-black* and restates the membership test's real job: it is the half that stays red if `evaluate_pair` stops emitting `suggested_hex` at all. Line 332 becomes "A failing pair with a suggestion is data, not an error."
  The finding's *optional* rename of `NULL_SUGGESTION` / the `nullsuggestion` slug is **declined on evidence**, with the rationale recorded in-code above the fixture: those 4 in-file sites name the pair `cc33` drives at threshold 21.0 to reach the None arm, so the name still names something real — and renaming it would retire the very token a future concept-grep would search the arm by, which is the defect class that caused this bounce.
  Applied the concept-grep rather than the retired spelling: `grep -nE '#.*\bnull\b'` returns 7 hits; only 325 and 332 were stale. Lines 58/59/66/68 describe the still-reachable None arm that `cc33` pins, and 293 is a correctly past-tense historical note — all five deliberately left untouched.

- **[minor] `cogni-workspace/scripts/check-contrast.py:170-172`** — **Resolved** in `1ee4ac35`. "still returns None" → "can still return None". Verified against the executed trace: `suggest_hex(parse_hex('#333333'), parse_hex('#FFFFFF'), 7.0)` returns `'#303030'` while the same call against `#757575` returns `None`, so the arm is reachable but not universal and the weakened wording is exactly right.

- **[minor] `cogni-workspace/skills/manage-themes/SKILL.md:121`** — **Resolved** in `1ee4ac35`. Both sub-items applied: the residual-null clause now reads "for a caller driving the module strictly above ~4.583:1" (naming the bound rather than implying it by strictness), and "At the two thresholds this script grades" becomes "At the two thresholds this CLI exposes" (the module accepts any threshold; only the CLI fixes two).

- **[minor, waived] `cogni-workspace/skills/manage-themes/SKILL.md:273-274`** — no edit, per the verdict's own waiver (pre-existing whole-skill drift, `introduced_by_change` false, file under its measured cap).

### Verification

Full `cogni-workspace` suite **20/20 PASS**. `check-frontmatter.sh` clean (49 files scanned), `check-version-bump.sh` `status: ok` with 0 violations, `check-breadcrumbs.sh` clean. `measure-skill-length.sh`: 26807 → **26790** chars_body against a 31500 cap — this delta shrank the file.

### The re-review found something new

The fresh gate at `1ee4ac35` is **NEEDS_REVISION** — not for anything above, but for the **mirror** of the original finding, which the earlier reviews missed: line 366 asserts *"The L=1.0 endpoint is not decisive at either CLI threshold"*, and that is false. A plain CLI run returns `suggested_hex #FFFFFF` for fg `#000033` on bg `#3366FF` at 4.5 — the black endpoint scores **4.4853** and fails, white scores **4.682** and clears. The comment's reasoning is backwards: the multiply-to-21 identity means white clears precisely *when black fails*. The line is change-introduced by this PR and contradicts `SKILL.md:121`, which this same PR writes.

Details and the corrected wording are in the fresh verdict comment. This PR is **held author-side** rather than handed to the merger, since a known-negative verdict would only bounce straight back.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1704

**Summary:** 18-19/19 contract items satisfied across all four perspectives, but plugin-validator found two un-waived MAJOR defects in the two new commits themselves -- a rewritten cc32 rationale in test-check-contrast.sh asserts a palette run 'cannot isolate' the white endpoint though the file's own cited example (fg #000033 on bg #3366FF at 4.5) does exactly that, and the PR body still carries the original, now-false 'not decisive at either CLI threshold' claim that the suite comment already corrected -- so this needs another revision pass before merge. contract-grading-reviewer separately self-reported needs_revision on a single minor item-16 (cc31 sub-case id rename) that plugin-validator graded satisfied and the split-judgment predicate declined to route to human_judgment (severity_not_major).

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (7)

- **[major]** (auto-fixable) `PR #1704 body, "## Test surface" bullet for cc32` — The PR body still reads "The `L=1.0` endpoint is not decisive at either CLI threshold (black clears whenever white barely does), so it needs a direct call." That is the exact claim commit 58ce5e4f corrected in the suite ("The L=1.0 endpoint IS decisive at the CLI thresholds"), so the PR body and the code comment now assert opposite things about the same fact, and the PR body's version is the false one -- verified: bg #3366FF gives ratio_black 4.4853 (FAILS 4.5) and ratio_white 4.6820 (clears), so white is decisive there. The parenthetical reasoning is wrong in the same way: the binding case is not "white barely clears" but "black just misses", where 21/4.4853 = 4.682 puts white above the bar. The PR body is the artifact contract item 8 designates as the place a witness substitution is "recorded and why", so a false record there is load-bearing, not cosmetic.
  - *Suggested fix:* Edit the PR body's cc32 bullet to match the corrected suite comment: state that both endpoints are decisive at the CLI thresholds (whichever fails, the other necessarily clears by the ratio_black*ratio_white == 21 identity), and give the real reason cc32 uses a direct call, worded consistently with whatever wording finding 1 settles on.

- **[major]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:373-375` — INTENT-CLEAR but FALSE. The sentence "What a palette run cannot do is ISOLATE that endpoint from the stepped walk preceding it" (repeated at :354-355 as "the white endpoint's decisiveness -- though observable there -- cannot be ISOLATED from the stepped walk by one") is a universal negative falsified by the example cited two sentences earlier in the SAME paragraph. Executed against the head via the palette CLI: palette {"fg":"#000033","bg":"#3366FF"} --pair fg:bg returns suggested_hex '#FFFFFF' at HEAD and null at base ab9ba4c3. For that pair no stepped-walk candidate clears 4.5 once snapped (best snapped 4.4998 over 99 candidates) and the L=0.0 endpoint FAILS (black 4.4853 < 4.5), so the only candidate that can produce '#FFFFFF' is the exact L=1.0 endpoint -- a complete isolation, reachable through a plain palette run, red at base and green at head. Eight further pairs on the same 46,656-pair grid the file already cites have the identical property. This is the third instance of the same class in this passage: the first head claimed the endpoint is never decisive, the second the mirror, and this one claims the CLI cannot do something it demonstrably does.
  - *Suggested fix:* Replace the universal negative with the true, narrower reason cc32 uses a direct call. Two honest options: (a) state that a palette witness IS constructible (fg #000033 on bg #3366FF at 4.5 answers #FFFFFF with no walk candidate clearing and black failing at 4.4853) but that the direct call at 21.0 was preferred because its isolating window (20.4689, 21.0] is derived rather than empirical and is stable against a walk-step change; or (b) drop the justification sentence entirely and keep only the positive claim that a threshold of 21.0 isolates the exact endpoint and nothing else -- which is measured and true. Do not assert what a palette run cannot do without a falsifier run behind it.

- **[minor]** (waived) `cogni-workspace/skills/manage-themes/SKILL.md body, Operation 6 "Contrast & Accessibility" (lines 106-124, ~2.9k chars)` — Nine dense bullets document the per-field JSON semantics of check-contrast.py's envelope inline -- data.pairs[], data.unclassified, data.collisions, data.failures, data.evaluated, data.unparsed, the --pair/--large-pair flags, and the success:false path. This is reference material about one script's output contract sitting in the always-loaded body, and the skill ships no references/ directory at all.
  - *Justification:* Pre-existing whole-skill drift not introduced by this change (introduced_by_change: false); annotated for human awareness but must not block merge -- mirrors service-resolve Step 6.5's pre-existing-drift policy.

- **[minor]** (waived) `cogni-workspace/skills/manage-themes/SKILL.md:121` — Precision only, no user-observable fault: the closing clause's two halves sit in different frames -- the JSON envelope is only ever produced by main(), whose sole thresholds are 4.5 and 3.0, so the emitted JSON can never in fact carry a null here; the strictly-above-sqrt(21) caller is an in-process importer of suggest_hex(), which emits no JSON at all. A tighter wording would separate schema-permits-null from CLI-never-emits-null. The operator instruction it delivers ('read the key rather than assuming it') is defensively correct either way.
  - *Justification:* The prior review verdict on this PR issued an explicit waiver for this exact line: "Leave as-is to preserve the verbatim merger wording ... this must not gate." The line was deliberately left unedited at head 03b35024. Raised as a recorded observation under that waiver, not as a gate.

- **[minor]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:372` — "and it is often white that does" is unsupported by the corpus this same comment block uses for all its other statistics. Measured over the identical 46,656-pair grid: at 4.5, #FFFFFF is the answer for 9 of 38,594 failing pairs (0.023%), and all nine share one background (#3366FF); at 3.0 it is 0 of 31,652. "Often" overstates a rare, single-background phenomenon by three orders of magnitude, in a block whose credibility rests on every other figure in it being measured.
  - *Suggested fix:* Replace "it is often white that does" with the measured figure, e.g. "and sometimes it is white: 9 of 38,594 failing pairs on the same 46,656-pair grid, all against bg #3366FF -- fg #000033 on bg #3366FF at 4.5 answers #FFFFFF, black scoring 4.4853 and FAILING where white scores 4.682".

- **[minor]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:69-72` — Recipe 4's single-occurrence argument is now incomplete as stated. It says "check-contrast.py carries three `return None`, and only suggest_hex's is four-space indented -- parse_hex's two are eight-space, which the `^    ` anchor excludes." This diff's new docstring introduced a FOURTH four-space-indented line beginning `return None` at check-contrast.py:172 ("return None.", the docstring's closing words). The stated discriminator (indentation) therefore no longer selects a unique line; what actually saves the anchor is the `$` terminator excluding the trailing period. NOT gating: the anchor was verified empirically -- grep -cE '^    return None$' returns 1 (matching line 208 only), and recipe 4 run in an isolated copy reddened exactly cc33 while cc13/cc14/cc15 and all three cc31 cases stayed green.
  - *Suggested fix:* Extend the sentence to name the real discriminator, e.g. "...only suggest_hex's is a four-space-indented statement; parse_hex's two are eight-space, and the docstring's closing `return None.` at line 172 carries a trailing period, both of which `^    return None$` excludes."

- **[minor]** (auto-fixable) `review-contract:item-16` — Item 16 is stamped [fidelity: literal] and names two specific existing case ids -- cc31-null-suggestion-ratio and cc31-null-suggestion-envelope-succeeds -- as unchanged in the diff. The diff renames both (to cc31-endpoint-suggestion-ratio and cc31-endpoint-suggestion-envelope-succeeds) as part of the broader cc31 family rename that items 13/19 legitimately require. The asserted values themselves are preserved exactly (2.7422, (True, True)) and both remain green, but the literal case-id text item 16 pinned by name is not preserved. NOTE (orchestrator adjudication): plugin-dev:plugin-validator independently examined this same locus and graded it SATISFIED, on the ground that item 16's stated property is that existing assertions keep their current EXPECTED VALUES (all preserved verbatim), that the id spellings in its Evidence clause are contract-time locators, and that item 13 plus the contract's own risk watchlist affirmatively MANDATE retiring the null-suggestion naming -- so a literal-id-freeze reading would make items 13 and 16 mutually unsatisfiable. Item 17, the dedicated id-stability item, forbids RENUMBERING only, and cc31 keeps its number. A repo-wide grep for cc31-null-suggestion returns zero surviving references and both mutation recipes are guard_verified at this head, so there is no user-observable fault. Recorded as an advisory observation; the split-judgment predicate does not route it (severity_not_major).
  - *Suggested fix:* No action required if the maintainer accepts the majority reading. Otherwise, record in the PR body that the whole cc31 family was renamed together for consistency with item 13, which would let a future literal-fidelity read of item 16 treat it as an argued substitution.

### Positive observations

- Every numeric claim the two new commits introduce verifies EXACTLY by execution against the head's check-contrast.py, including the fg #000033/bg #3366FF pair at 4.5 -> #FFFFFF and the cc32 window best-snapped at 20.4689 against a black background. (plugin-dev:plugin-validator)
- All 19 committed review-contract checklist items grade satisfied on plugin-validator's independent reading, including item 16, whose two renamed cc31 ids preserve their expected values verbatim and whose rename is mandated by item 13 and the contract's own risk watchlist. (plugin-dev:plugin-validator)
- The author's deliberate departure from the prior verdict's suggested fix was CORRECT and was confirmed by execution: a missing 'suggested_hex' key raises KeyError at the direct index before print() runs, so the membership half's False can never print. (plugin-dev:plugin-validator)
- Both mutation recipes were run in an isolated /tmp copy and both are genuinely discriminating -- recipe 5 reddens cc31-endpoint-suggestion-is-present-and-black precisely; recipe 4 reddens cc33 alone. (plugin-dev:plugin-validator)
- The rewritten SKILL.md:121 line makes four claims and all four verify against the head's check-contrast.py rather than against assumption; it does not repeat this PR's own prior mirror error since it explicitly names #FFFFFF as reachable. (plugin-dev:skill-reviewer)
- review_contract item 19 is SATISFIED, verified against source rather than assumed: the stale base sentence no longer exists at the head, and the guidance no longer instructs a reader on handling a state the fixed tool cannot emit. (cogni-service:skill-quality-reviewer)
- 18 of 19 committed checklist items grade satisfied against the diff, with evidence drawn from execution rather than from the PR body, including items 6, 8, 14 and 15 confirmed by live monkeypatch/CLI runs. (cogni-service:contract-grading-reviewer)
- Deterministic gates all clean at this head: check-frontmatter 0 offenders (schema_version 4), check-version-bump status ok / 0 violations (metric_version 2), run-tests 20 discovered / 20 passed / 0 failed / 0 skipped, and the repo's own check-case-id-pairing.py and check-result-line-plainness.py both exit 0.

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the verdict at https://github.com/cogni-work/insight-wave/pull/1704#issuecomment-5467762817 (NEEDS_REVISION, 4 findings). Two commits, both **comment-only** — no executable line changed.

### Findings addressed (4/4)

- **[major] `tests/test-check-contrast.sh:366`** — **Resolved** in `58ce5e4f`. The claim "the L=1.0 endpoint is not decisive at either CLI threshold — black already clears whenever white barely does" is false: the multiply-to-21 identity bounds the *other* endpoint from **above**, not below, so black clearing implies nothing about white. Verified by executed trace — `suggest_hex(#000033, #3366FF, 4.5)` → `#FFFFFF`, black scoring 4.4853 (FAILS) against white's 4.682.
- **[minor] `tests/test-check-contrast.sh:349-350`** — **Resolved** in `58ce5e4f`. Narrowed the "not observable through a palette run" claim to the None arm (cc33) alone.
- **[minor] `tests/test-check-contrast.sh:307`** — **Resolved** in `58ce5e4f`. Scoped per cohort; "near-white" now applies to 4.5 only, with the 3.0 cohort's `#B2B300` (L=0.351) floor named. All sweep figures re-reproduced exactly: 38,594 / 21 / 194 / 0.503% at 4.5; 31,652 / 0 / 300 / 0.948% at 3.0.
- **[minor] `tests/test-check-contrast.sh:331`** — **Resolved** in `58ce5e4f`, then **self-corrected** in `03b35024`.

### Two corrections I made to the verdict's own suggested fixes

Both suggested replacements were graded against an executed trace rather than transcribed, and both turned out false:

1. **The major's suggested supporting fact was not adopted.** The verdict proposed "whenever BLACK barely clears, WHITE also clears, so it is the DARK endpoint that is never the sole clearing one at CLI thresholds." Measured over the 216-colour palette at 4.5: black is the **sole** clearing endpoint for **139** backgrounds, white for **76**, with **1** where both clear. The proposed statement simply mirrors the error being retired. The comment now cites section F (2) instead of restating the identity a third time.
2. **Finding 4's suggested wording was adopted verbatim in `58ce5e4f` and was false** — caught during the review pass and fixed in `03b35024`. It claimed the membership co-assertion makes a key-dropping regression "name itself (False in the tuple) rather than surfacing as an empty capture". It cannot: the membership test shares one tuple with a direct index, so a missing key raises `KeyError` at the index first and nothing prints at all. Simulated both states — key present → `(True, '#000000', 1)`; key missing → `KeyError`, empty capture.

### Not edited

`skills/manage-themes/SKILL.md:121` — left untouched per the verdict's own explicit waiver ("Leave as-is to preserve the verbatim merger wording ... this must not gate").

### Verification

Full `cogni-workspace` suite **20/20 PASS** (`run-tests.sh`), `check-breadcrumbs.sh` clean, `check-frontmatter.sh` clean (49 files, schema 4). No `plugin.json` version touched.

### Status: held author-side, not handed off

The fresh verdict (https://github.com/cogni-work/insight-wave/pull/1704#issuecomment-5468340733) is **needs_revision** with two majors, so the author-handoff preflight's auto-fix carve-out does not apply and `svc:needs-review` was **not** applied. The PR stays re-pickable by the next author drain.

**The remaining major is real and is the mirror of the one just fixed.** `tests/test-check-contrast.sh:352-355,370-371` still asserts a palette run "cannot ISOLATE that endpoint from the stepped walk" — and the comment's own witness falsifies it. The walk preserves hue and saturation, so for `#000033` on `#3366FF` no walk candidate ever clears 4.5 (a saturated blue never reaches white's ratio); only the exact L=1.0 endpoint does, which is precisely what isolation means. A palette case on that pair would isolate the endpoint exactly as the 21.0 direct call does. The accurate reason for cc32's direct call is that 21.0 pins the endpoint against a walk whose best snapped candidate is 20.4689 (`#FCFCFC`) — a stated bound rather than a claimed impossibility.

Second finding: the **PR body's cc32 bullet** still carries the retired "not decisive" wording and now contradicts the corrected suite comment.

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1704

**Summary:** needs_revision — plugin-validator's own verdict (Rule 3) carries two unresolved major defects in the diff's own test-comment prose (a false vacuity claim at test-check-contrast.sh:392-394 and a mis-attributed causal clause at :382), while skill-reviewer and skill-quality-reviewer both pass with only precision-only or pre-existing minor findings. NOTE: this is the FIFTH and SIXTH instance of one recurring class on this PR — comment prose asserting something about what the code does or cannot do that execution contradicts — and both were introduced by the revision cycle that fixed the previous instance.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[major]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:392-394 (and the matching PR-body 'Test surface' bullet)` — Fifth false capability claim. The comment asserts 'at a step of 0.001 the walk itself snaps to #FFFFFF and scores 21.0 (measured), which would satisfy this threshold WITHOUT reaching the endpoint branch and make the case vacuous.' Executed against a step-0.001 copy of the real module: suggest_hex('#333333','#000000',21.0) still returns #FFFFFF VIA THE ENDPOINT BRANCH, and ZERO walk candidates pass the 21.0 gate. Root cause: the walk gates on the UNSNAPPED ratio at check-contrast.py:183 BEFORE snapping at :184-185; at step 0.001 the near-white candidate is L=0.999000 with unsnapped ratio 20.9545325539 < 21.0, so the gate rejects it even though it would snap to #FFFFFF. The prose extended a SNAPPED-ratio measurement to a gate that tests the unsnapped value first. The underlying step-sensitivity worry is real but points the OTHER way: vacuity occurs at COARSE steps (0.1, 0.16, 0.2 land on L=0.9999999999999999 whose unsnapped ratio is exactly 21.0, so the walk returns #FFFFFF directly and cc32 loses its teeth).
  - *Suggested fix:* State the measured fact — at 0.001 the walk's best gating value is 20.9545, still below 21.0, so the case retains its teeth — and re-point the sensitivity warning at coarse steps; or drop the specific 0.001 example and keep only 'changing the step means re-deriving this threshold'.

- **[major]** (auto-fixable) `cogni-workspace/tests/test-check-contrast.sh:382` — New causal clause names the two mechanisms responsible for 1.69% and omits the one responsible for 98.25%. The revision added '... white is the ANSWER for only 9 at 4.5 ... because the -1 walk and the black endpoint are both scored first.' Instrumented attribution over the 12,600 failing pairs at 4.5 where white CLEARS: walk+1 answers 12,379 (98.25%), endpoint-1 207 (1.64%), endpoint+1 8 (0.06%), walk-1 6 (0.05%). The two named mechanisms account for 213 of 12,600 = 1.69%. The dominant preemptor is the +1 walk, which the clause does not mention. (At 3.0 the named mechanisms do account for 76.15%, so the clause is roughly right at the threshold where the '9' figure does not live, and wrong at the one where it does.)
  - *Suggested fix:* Name the +1 walk as the preempting stage, e.g. '-- because the +1 walk almost always finds a near-white shade that clears before the endpoint is scored (12,379 of those 12,600 at 4.5); the -1 walk and the black endpoint take the remaining 213.'

- **[minor]** (auto-fixable) `PR body, 'Test surface' section — the "Do not renumber" citation` — Prior FINDING 2 not resolved. The citation was corrected 397 -> 506, correct at 429bcbdd, but commit 6e352db5 replaced 20 comment lines with 23 (net +3) above the target, so the rule now sits at :509 and line 506 is blank. Second time this one citation has drifted.
  - *Suggested fix:* Cite 509, or drop the absolute line number and cite the rule by its text so it cannot go stale on the next comment edit.

- **[minor]** (auto-fixable) `cogni-workspace/skills/manage-themes/SKILL.md:121` — The clause 'with the dark endpoint winning a tie where both would clear' is true but narrower than the actual precedence: the -1 direction's full walk AND its endpoint are scored before +1 begins, so #000000 also displaces mid-lightness shades the lighter walk would have found (194 of 38,594 at 4.5; 300 of 31,652 at 3.0). No user-observable fault — the operator action is identical.
  - *Suggested fix:* Optionally name the property rather than the endpoint pair: 'the darker direction is tried first, so a dark shade — #000000 at the limit — wins wherever both a darker and a lighter shade would clear.'

- **[minor]** (waived) `cogni-workspace/skills/manage-themes/SKILL.md body, Operation 6 (lines 106-124)` — ~2.9k chars of per-field JSON reference material sit inline in the always-loaded body and the skill ships no references/ directory. Organization observation, not a size violation — chars_body 26790 vs cap_chars 31500, over_cap false.
  - *Justification:* introduced_by_change:false — pre-existing whole-skill drift, waived per service-review Step 3.5's pre-existing-drift split.

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/manage-themes/SKILL.md:265` — PRE-EXISTING, untouched by this PR: Operation 10 names the same reference two ways — line 265 bare 'references/claude-design-bundle-mapping.md' (readable as skill-relative, a path that does not exist) and line 271 repo-relative.
  - *Justification:* Pre-existing drift not introduced by this change; out of scope for a comment-correction revision.

### Positive observations

- FINDING 1 from the prior cycle precisely resolved; all figures reproduce on the full 46,656-pair grid (white clears 32.6476% at 4.5, 53.8102% at 3.0; white is the answer 0.0233% at 4.5, 0 at 3.0). (plugin-dev:plugin-validator)
- The #000033-on-#3366FF witness verifies claim-for-claim and genuinely reaches the endpoint branch. (plugin-dev:plugin-validator)
- Every section F figure reproduces against origin/main (21/0 None->hex, 194/300 displaced, all becoming #000000, lowest displaced #B2B300 at L=0.351). (plugin-dev:plugin-validator)
- Both mutation recipes have real teeth with exactly the documented green/red profiles. (plugin-dev:plugin-validator)
- All anchor single-occurrence claims hold. (plugin-dev:plugin-validator)
- run-tests.sh: 20 discovered, 20 passed, 0 failed, 0 skipped. (plugin-dev:plugin-validator)
- Change is comment-only relative to the last review — no executable line touched. (plugin-dev:plugin-validator)
- The change retires the imprecise wording the prior verdict flagged, and does not reproduce the retired mirror falsehood. (plugin-dev:skill-reviewer)
- Concept vocabulary grepped in BOTH directions across cogni-workspace — no stale consumer prose survives elsewhere. (cogni-service:skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the verdict at https://github.com/cogni-work/insight-wave/pull/1704#issuecomment-5468340733 (NEEDS_REVISION, 7 findings). Three commits this cycle — `3f7261ec`, `429bcbdd`, `6e352db5` — plus three PR-body edits. **All comment/prose only; no executable line changed** (verified mechanically: every changed line in the diff is a `#` comment).

### Findings addressed (7/7)

- **[major] `PR #1704 body, "## Test surface"` cc32 bullet** — **Resolved.** The bullet asserted "The `L=1.0` endpoint is not decisive at either CLI threshold", the exact claim `58ce5e4f` had already corrected in the suite. Rewritten to state that both endpoints are decisive (by the `ratio_black * ratio_white == 21` identity) and to give the real reason cc32 uses a direct call.
- **[major] `tests/test-check-contrast.sh:373-375` (mirrored at `:354-355`)** — **Resolved** in `3f7261ec`. The universal negative "What a palette run cannot do is ISOLATE that endpoint" is false; I reproduced the counterexample: `{"fg":"#000033","bg":"#3366FF"} --pair fg:bg` at 4.5 returns `suggested_hex "#FFFFFF"`, with black 4.4853 (fails), white 4.6820, and the walk's 99 candidates peaking at a snapped 4.4998 — so only the L=1.0 endpoint branch can answer. Took the verdict's option (a)/(b) hybrid: the witness is now stated as existing.
- **[minor] `tests/test-check-contrast.sh:372`** — **Resolved** in `3f7261ec`, then **corrected again** in `6e352db5`. "often white that does" replaced with measured figures; the first replacement ("white clears rarely") equivocated on *clears* and was itself false — see below.
- **[minor] `tests/test-check-contrast.sh:69-72`** — **Resolved** in `3f7261ec`. Recipe 4's discriminator argument rewritten: `^    return None` now matches **two** lines (`:172` the docstring's `    return None.`, `:208` the arm), so indentation no longer discriminates; the `$` terminator does (`grep -cE '^    return None$'` returns 1). Verified.
- **[minor, waived] `manage-themes/SKILL.md` Operation 6** and **`:121`** — no edit, `introduced_by_change: false`, waived as pre-existing whole-skill drift.
- **[minor, adjudicated] `review-contract:item-16`** — no edit; recorded as addressed-by-adjudication.

### Two additional defects found and fixed that the verdict did not flag

- The **PR body carried a twin** of the recipe-4 staleness (`return None` "at four-space indent") that the verdict flagged only in the suite comment. Corrected — same load-bearing-record reasoning as finding 1.
- The verdict's own suggested wording for finding 2 — "the isolating window is **stable against a walk-step change**" — is **false**, and I nearly shipped it. Replaced with a measured, step-sensitivity-aware statement.

### ⚠️ Holding author-side — this needs a human

The re-gate at `6e352db5` returned **NEEDS_REVISION on two new majors**, both introduced by this cycle's own corrections. This is now the **fifth and sixth instance of one class** on this PR:

| # | Claim | Verdict |
|---|---|---|
| 1 | "the L=1.0 endpoint is never decisive" | false |
| 2 | its mirror | false |
| 3 | "a palette run cannot ISOLATE that endpoint" | false |
| 4 | "white clears rarely" | false — equivocates on *clears* |
| 5 | "at step 0.001 the walk … makes the case vacuous" | false |
| 6 | "because the -1 walk and the black endpoint are scored first" | false — accounts for 1.69%, not 98.25% |

**#5 is worth reading closely**, because my own verification was wrong in an instructive way. I probed step-sensitivity by comparing **snapped** ratios, but `suggest_hex` gates on the **unsnapped** ratio at `check-contrast.py:183` *before* snapping at `:184-185`. At step 0.001 the near-white candidate is L=0.999000 with unsnapped ratio **20.9545** < 21.0, so **zero** walk candidates pass the gate and cc32 keeps its teeth. The real sensitivity points the **other way** — at *coarse* steps (0.1, 0.16, 0.2) the accumulator lands on L=0.9999999999999999 whose unsnapped ratio is exactly 21.0, the walk returns `#FFFFFF` directly, and cc32 *does* go vacuous.

I am not attempting a seventh prose correction. Five consecutive automated cycles have each fixed the previous false claim and introduced a new one, which is a signal about the passage, not about any one wording. **The code is not in question** — the fix is correct, all five ACs are satisfied, `run-tests` is 20/20, and both mutation recipes are verified to have teeth. What needs a human is the decision about this comment block: whether to keep narrating the endpoint/walk interaction in prose at all, or to reduce it to the assertions the suite already executes.

`svc:needs-review` deliberately **not** applied; the PR stays draft.

sdh07:
## Maintainer ruling — cut to executed assertions

Answering the escalation on the run-log thread. **Ruling: cut.** The narration is gone; the one hazard it was trying to state is now executed. Head `8fc32153`.

I reproduced both open majors myself before ruling, rather than reading the verdict:

- **step-0.001 vacuity** — false. `suggest_hex` gates on the **unsnapped** ratio at `check-contrast.py:183` *before* snapping. At step 0.001 the walk's best gating value is `20.9545` < 21.0, so zero candidates pass. The real risk is at **coarse** steps (0.1 / 0.16 / 0.2 land on `L≈1.0`, gate value exactly `21.0`).
- **attribution clause** — false. Over the 12,523 white-clearing failing pairs at 4.5: `walk+1` 12,303, `endpoint-1` 206, `endpoint+1` 8, `walk-1` 6. The two named mechanisms are 1.7%; the unnamed `+1` walk is 98.2%. At 3.0 the ordering inverts.

**The passage, not the wording.** All six failures shared one shape: a derived, whole-input-space claim about mechanism that no assertion in this suite executes. Nothing bound it, so it could only be checked by an ad-hoc sweep — and each sweep was itself error-prone, which is exactly how cycle five's own probe went wrong. It was also stale-prone by construction: any change to `step`, gate order or direction order invalidates a dozen unchecked numbers.

**Cut:** Section F's 46,656-pair tie-break statistics and its `sqrt(21)` derivation (the derivation lives in `suggest_hex`'s docstring and is pinned by `cc33` — the comment now points there); the cc32 block's CLEARING-vs-WINNING statistics, `#000033` witness and step-0.001 example; the cc32/cc33 preamble's restatement of Section F. Net −21 lines, comment-only.

**Replaced, not deleted.** The step-sensitivity hazard is real, and two prose attempts to state it were both false — so it is now an executed check. Mutation recipe 5 gains `--case cc32-white-endpoint-is-suggested`:

| Module | cc31 @4.5 | cc32 @21.0 | cc33 @21.0 |
|---|---|---|---|
| head | `#000000` | `#FFFFFF` | `None` |
| recipe-5 mutant (`return to_hex(endpoint)` → `pass`) | `None` | `None` | `None` |

Under that mutant the endpoint branch cannot answer, so a still-green cc32 **is** the vacuity report — a machine says it, at every run, instead of a paragraph asserting it.

**Also settled here:** `manage-themes/SKILL.md:121` now states precedence from the loop structure rather than a statistic, closing the precision minor two reviewers raised; and `review-contract:item-16` is recorded in the body the way its own suggested fix asks.

**Verified at `8fc32153`:** suite exit 0 / zero `FAIL:` · `run-plugin-tests --filter cogni-workspace` 20/20 · emitted case ids byte-identical to head (39/39, no collisions) · pairing, plainness, breadcrumb guards exit 0 · frontmatter 49 files 0 violations · all three mutation recipes `guard_verified`, anchors still single-occurrence · every surviving number re-verified by execution · no version touched.

The full evidence, including the sweep the comments no longer carry, is in the PR body.

Handing off to the merger — `svc:needs-review` applied.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1704

**Summary:** All four perspectives (plugin-validator, skill-reviewer, skill-quality-reviewer, contract-grading-reviewer) pass, with every committed acceptance-checklist item graded satisfied (contract-grading-reviewer: 19/19, zero findings); the three minor/preference findings raised are merger additions at non-contract-item loci under this gatekeeper-origin review, so per the advisory-addition demotion they are excluded from every severity-keyed and all-pass rule and render only as non-gating advisory annotations - the PR is safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-workspace/scripts/check-contrast.py:187-188` — Comment states a whole-input-space property ('never scores the endpoint itself') with a measurable counterexample (#F1CDFE, HLS lightness exactly 0.9). No behavioural fault.
  - *Justification:* Non-gating. No contract item binds this comment; code is correct; maintainer already ruled at this exact head.

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/manage-themes/SKILL.md body Operation 6 (lines 106-124)` — Skill body restates check-contrast.py's output contract inline rather than behind references/ (progressive disclosure).
  - *Justification:* Pre-existing whole-skill drift (introduced_by_change=false); waived per Step 3.5 adapter so it annotates without gating.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/manage-themes/SKILL.md:121` — Closing sentence about null-ability is true but leaves the CLI operating context the section establishes.
  - *Suggested fix:* Optionally re-scope the sentence to tell the CLI reader the state is unreachable for them.

### Positive observations

- AST comparison confirms suggest_hex is the sole changed definition (plugin-dev:plugin-validator)
- Both mutation recipes guard_verified against a temp copy (plugin-dev:plugin-validator)
- Correction scoped to exactly one line (plugin-dev:skill-reviewer)
- Load-bearing instruction survived the rewrite intact (plugin-dev:skill-reviewer)
- Try-order, snap-and-re-verify, sqrt(21) bound and residual null-ability all check out against check-contrast.py:156-208 (cogni-service:skill-quality-reviewer)
- Suite exit 0, 40/40 (cogni-service:contract-grading-reviewer)
- Both mutation recipes guard_verified (cogni-service:contract-grading-reviewer)
- rb*rw=21 identity verified numerically across 4 backgrounds (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1720, opened 2026-08-31
cogni-knowledge: consolidate the 15 duplicated inline probe_plugin() definitions across the live knowledge-* skills


claude-cogni-work:
## Plugin
cogni-knowledge

## Use case
`probe_plugin()` is defined inline, verbatim, in **15 live** `knowledge-*` SKILL.md files. A future change to the probe's resolution order — say, adding a third install layout — has to land 15 times, and any copy missed diverges silently.

Census verified at `origin/main` 38f967d9: 17 files under `cogni-knowledge/` contain `probe_plugin() {`, of which 2 sit under `_archive/` (`knowledge-report`, `knowledge-research`), leaving **15 live** — `knowledge-refresh` plus 14 siblings: `compose`, `curate`, `dashboard`, `distill`, `fetch`, `finalize`, `health`, `ingest`, `lint`, `plan`, `prefill`, `query`, `resume`, `verify`.

**This is a DRY / maintenance item, not a latent-bug sweep.** All 14 siblings were independently checked against the fence-scope predicate and are already fence-co-located — they define and call `probe_plugin` in the same fenced block. `knowledge-refresh` was the sole broken instance, and it is fixed by the PR for #1708. Nothing here is currently mis-executing; the cost is purely maintenance surface.

## Proposed solution
Move the definition to one shared snippet the consuming skills source, then delete the inline copies.

**`scripts/resolve-wiki-scripts.sh` is the wrong home.** Its header scopes it to "locating a cogni-wiki engine skill's `scripts/` directory", and `resolve_wiki_scripts()` takes a wiki *engine skill* name (`wiki-ingest` / `wiki-lint` / `wiki-health`) and resolves `scripts/vendor/cogni-wiki/skills/<skill>/scripts`. `probe_plugin()` probes an arbitrary *plugin* for an arbitrary *skill* — a different subject. Reusing that file would widen a narrowly-scoped documented helper into a general plugin-probing grab-bag and change behaviour for all 14 siblings at once. A new, separately-named snippet (e.g. `scripts/probe-plugin.sh`) is the shape to evaluate.

Re-derive the census at the time of the work — the population is expected to move.

## Acceptance criteria
- [ ] A census recorded in this issue names every live file defining `probe_plugin() {`, derived at the time of the work rather than copied from this issue (the population is expected to move).
- [ ] Exactly one definition of `probe_plugin()` remains reachable per consuming skill, with no skill carrying its own divergent copy.
[...]

Issue #1721, opened 2026-08-31
cogni-knowledge: knowledge-refresh §2 calls resolve_wiki_scripts without sourcing it — the same false abort #1708 fixed, one fence away


claude-cogni-work:
## Plugin
cogni-knowledge

## What happened
`knowledge-refresh` §2 ("Resweep — native inline orchestration") carries the **same fenced-block scoping defect** that #1708 fixed in the Step 0 pre-flight, one fence away and against a different helper.

The §2 fence calls `resolve_wiki_scripts` but never sources the file that defines it:

```
# reuse $RESWEEP_SCRIPTS from Step 0; or re-resolve identically if not set:
: "${RESWEEP_SCRIPTS:=$(resolve_wiki_scripts wiki-claims-resweep extract_page_claims.py)}"
[ -n "$RESWEEP_SCRIPTS" ] || abort "vendored wiki-claims-resweep scripts not found — reinstall cogni-knowledge"
```

Measured across the whole file, three fences call `resolve_wiki_scripts`; two source `${CLAUDE_PLUGIN_ROOT}/scripts/resolve-wiki-scripts.sh` first, and this one does not.

**Why the `:=` default is the normal path, not the exceptional one.** The prose introducing the fence says to reuse the Step 0 value and only re-resolve "e.g. a manual re-entry". But shell variables do not survive across separate Bash tool invocations, so whenever this fence is executed as its own block — the same premise #1708 established for this file — `RESWEEP_SCRIPTS` is unset, the `:=` default expansion runs, `resolve_wiki_scripts` is `command not found` (exit 127), the substitution yields the empty string, `[ -n "" ]` is false, and the abort fires on a perfectly healthy install.

**Consequence worth stating plainly:** #1708 fixed the pre-flight's false abort, but on its own it does **not** restore `--resweep` to end-to-end working order — the run now clears the pre-flight and false-aborts here instead.

## Expected behavior
`resolve_wiki_scripts` is defined and reachable at the point of its invocation in the §2 fence, so the `RESWEEP_SCRIPTS` guard fires only when the vendored scripts are genuinely absent.

## Secondary defect in the same fence
`abort "..."` is prose pseudo-code, not a shell builtin, so the guard's failure arm is not executable as written. Every other abort in this skill is a prose blockquote outside the fence; this one is inside executable shell. Decide deliberately between making it executable and moving it out of the fence — do not leave a non-runnable token inside a block documented as independently runnable.

## Acceptance criteria
[...]

PR #1737, opened 2026-09-01
cogni-knowledge: derive + fence-scope the resolver sourcing guard

### LINKED ISSUE #1725: cogni-knowledge: the resolve_wiki_scripts sourcing guard is enumerated and file-wide, so it was green on #1721's defect
## Plugin
cogni-knowledge

## Current behavior
`cogni-knowledge/tests/test_resolve_wiki_scripts.sh` owns the invariant "every flow sources the shared resolver snippet". It was **green on a live instance of exactly that defect** — the `knowledge-refresh` §2 fence called `resolve_wiki_scripts` with no in-fence `source` line, and this guard did not notice.

Two independent reasons, both measured against `origin/main` @ b8a6698f:

1. **The enumeration has drifted.** `SOURCING_SKILLS` (line 37) names 8 skills. 13 skills actually call the resolver — `knowledge-prefill` (SKILL.md:98), `knowledge-refresh` (60/75/198), `knowledge-distill` (75/411), `knowledge-index` (102) and `knowledge-setup` (179) are unlisted, so they are outside the guard's scope entirely.
2. **The assertion is file-wide, not fence-scoped.** `resolve-wiki-03` asserts with `grep -qF` (line 74). `knowledge-refresh` carries source lines at 59 and 74, so the file passed on their strength while line 198 was broken in a different fence.

## Desired behavior
The guard derives its call-site set instead of enumerating it, and decides per *fence* rather than per *file*, so neither a new calling skill nor a call relocated into an unsourced fence can pass unnoticed.

## Motivation
This is the third appearance of the same class: #1708 fixed `probe_plugin` in the Step 0 fence, #1721 fixed `resolve_wiki_scripts` one fence away, and both were caught by human/agent reading rather than by the guard that already claimed the invariant. Each fix so far has added a per-symbol, per-file case; that shape stays green the next time a different skill drops its source line.

## Acceptance criteria
- [ ] `test_resolve_wiki_scripts.sh` derives the set of skills that call `resolve_wiki_scripts` by scanning `cogni-knowledge/skills/*/SKILL.md`, rather than reading a hand-maintained list. Falsified by: a skill that calls the resolver and is graded by no case.
- [ ] The sourcing assertion is fence-scoped — it fails a fence that calls `resolve_wiki_scripts` without a `source`/`.` of `resolve-wiki-scripts.sh` earlier in that same fence, and exempts a fence that sources without calling. Falsified by: a file that sources in one fence and calls in another reporting PASS.
- [ ] The fence-opener pattern tolerates leading whitespace. `knowledge-index` indents its fences (SKILL.md:100-104), so a column-0-anchored pattern mis-tracks fence state there and yields a false verdict on that file.
- [ ] The call detector keys on the underscore spelling `resolve_wiki_scripts`, so the hyphenated `resolve-wiki-scripts.sh` on the source line cannot make a sourcing fence trivially self-satisfying.
- [ ] A drift guard cross-checks the derived call-site set against the tree, so a caller that stops being detected while its file remains is graded rather than silently dropped. **Ruled 2026-09-01:** a skill directory leaving the tree entirely is legitimate consolidation, not drift — a pinned floor count is explicitly **not** required and must not be added, since a floor is the hand-maintained census this issue exists to delete and false-reds a legitimate consolidation. Falsified by: a `SKILL.md` that names `resolve_wiki_scripts` yet contributes no derived call site passing unreported.
- [ ] Demonstrated RED under a mutation that removes a source line from a calling fence in a skill **other than** `knowledge-refresh` — proving the guard's reach is genuinely plugin-wide and not a second per-file case.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-knowledge/tests`, `scripts/check-case-id-pairing.py` and `scripts/check-result-line-plainness.py` are green.
- [ ] Scope: the diff touches only paths under `cogni-knowledge/`. No `version` field changed (the patch bump is applied post-merge).

## Notes
Surfaced while planning #1721 — by the root-cause-framed plan that lost the Step 4c pick on smallest-blast-radius. The finding is separable from #1721's one-fence fix and is filed here rather than bundled into it.

Adjacent but distinct: #1720 consolidates the duplicated inline `probe_plugin()` definitions. That is a different symbol and a different property (definition duplication, not sourcing-guard reach); neither subsumes the other.

claude-cogni-work:
Closes #1725.

## Summary

`cogni-knowledge/tests/test_resolve_wiki_scripts.sh` owned the invariant "every flow sources the shared resolver snippet" and was **green on a live instance of exactly that defect**. Two independent reasons, both closed here:

1. **The population was enumerated, and had drifted.** `SOURCING_SKILLS` named 8 skills; **13 skills call the resolver across 16 fenced blocks**. `knowledge-prefill`, `knowledge-refresh`, `knowledge-distill`, `knowledge-index` and `knowledge-setup` were outside the guard's scope entirely. The set is now **derived** by scanning `skills/*/SKILL.md`, so a newly-calling skill is graded on its first run with no edit to the test.
2. **The assertion was file-wide.** `resolve-wiki-03` used `grep -qF` over the whole `SKILL.md`, so a source line in one fence vouched for a call in another. The decision is now made **per fence**.

Neither half is invented here. The fence-scoped awk predicate already ships at `test_refresh_resweep_contract.sh:132-141` scoped to `knowledge-refresh` alone; glob-derived discovery already ships at `test_knowledge_wiki_probe.sh:50-55`. This generalizes both across the derived set.

### Three load-bearing details

- **The fence opener tolerates leading whitespace.** `knowledge-index/SKILL.md:100` indents its resolver fence 3 spaces and is the only indented one of the 16 — a column-0-anchored pattern mis-tracks fence state there. (72 of 238 fence openers across the 21 files carry leading whitespace, so this is common, not exotic.)
- **Call detection keys on the underscore spelling** `resolve_wiki_scripts`, while the source line names the hyphenated `resolve-wiki-scripts.sh` and `next`s. A sourcing line therefore cannot satisfy itself as its own call site.
- **Both sourcing spellings are recognised** — `. "…"` (13 sites) and `source "…"` (3 sites, all `knowledge-refresh`). A `source`-only pattern would red twelve currently-correct files.

Prose outside every fence is never inspected, so the ~20 prose mentions are not graded — sharpest is `knowledge-refresh/SKILL.md:194`, a backtick-quoted invocation inside a sentence that a non-fence-tracked regex would misfire on.

## Case-id allocation

Per `cogni-knowledge/tests/README.md` rules 1-7.

| Id | Property |
|---|---|
| `resolve-wiki-01`, `04`-`13` | **Unchanged.** Parts 2-4 are byte-identical; 05/06/08 and 12 are standing anti-regression guards for the vendored-only removal. |
| `resolve-wiki-02`, `03` | **Retired** (gap intentional — rule 3 forbids renumbering, not gaps). |
| `resolve-wiki-14` | Vacuity: zero SKILL.md found, or zero call sites derived, goes RED. |
| `resolve-wiki-15` | Fence-scoped sourcing. |
| `resolve-wiki-16` | No inline `resolve_wiki_scripts()` definition anywhere in the derived set. |
| `resolve-wiki-17` | Mention-without-call drift — a file naming the resolver that yields no derived call site is graded by nothing. |

**Why retire `03` rather than repoint it.** At base `resolve-wiki-03-$name` carried **two** properties on three arms (L87 inline-definition, L89 does-not-source, L91 the shared PASS). Re-using the stem for one of them, over a differently-derived population, silently repoints a live id at a narrower predicate — the hazard rule 3 names. Retirement makes a stale `--case resolve-wiki-03*` recipe fail loudly as `case_not_found`. `references/delegation-contract.md:39` cited `resolve-wiki-03-*` by name, so that citation is repointed to `resolve-wiki-16` in the same diff; shipping it stale would reproduce this issue's own drift class.

**Collect-then-pair, not per-fence ids.** `knowledge-refresh` has 3 calling fences and `knowledge-distill` 2, so a per-fence discriminator would collide two fences of one file onto one id. Offenders accumulate across the loop; one same-id `if`/`else` fires after it closes (rule 7, `plain-emit-03` model). awk emits only bare `CALL`/`NOSRC` records and never a `PASS:`/`FAIL:` label, so every result line still originates at a shared `red`/`green` call — `check-case-id-pairing.py` skips heredoc and subprocess bodies and would otherwise go blind.

## Verification

All measured on this branch.

- `bash cogni-knowledge/tests/test_resolve_wiki_scripts.sh` → **rc=0**, 15 cases, derives **13 calling skills from 21 SKILL.md files**.
- Same under system `/bin/bash` (3.2 on macOS) → **rc=0**. bash 3.2 floor honoured: no `mapfile`/`readarray`/`declare -A`/case-modification expansions; string accumulation and a `case` membership test only.
- `python3 scripts/run-plugin-tests.py --filter cogni-knowledge/tests` → **79/79 suites green, 0 failures** (includes `test_refresh_resweep_contract.sh`, untouched and still green).
- `python3 scripts/check-case-id-pairing.py` → **0 violations** (139 files scanned, 2393 emissions inspected).
- `python3 scripts/check-result-line-plainness.py` → **0 violations**.
- Id-uniqueness census verified **by running the suite** (rule 6, not by grepping call sites): no duplicates. Emitted ids are `01`, `04`-`17`.
- Scope fence: `git diff --name-only origin/main` is 2 paths, both under `cogni-knowledge/`. No `plugin.json` touched, no `version` line in the diff — the patch bump is applied post-merge.

## Mutation recipe

Demonstrates RED in **`knowledge-index` — a skill other than `knowledge-refresh`**, which is what distinguishes genuine plugin-wide reach from a third per-file case. It is also the sole indented calling fence, so one mutation exercises both plugin-wide reach and whitespace tolerance.

Run from the root of this branch's checkout (`--root .`); adjust the first path if your `managed-service` clone is not a sibling of `insight-wave`.

```sh
bash ../managed-service/cogni-service/scripts/mutation-check.sh --root . --file cogni-knowledge/skills/knowledge-index/SKILL.md --expr 's{^[ \t]*\. "[^"]*resolve-wiki-scripts\.sh"\n}{}m' --test 'bash cogni-knowledge/tests/test_resolve_wiki_scripts.sh' --case resolve-wiki-15
```

**Observed: `mutated_result: red`, `restored_result: green`, `verdict: guard_verified`.** The mutation is reverted — `git status` on this branch is clean and no `SKILL.md` appears in the diff.

`resolve-wiki-14`'s red arm is reachable without a mutation, via the scan-root override the diff adds for exactly that purpose:

```sh
RESOLVE_WIKI_SKILLS_DIR=$(mktemp -d) bash cogni-knowledge/tests/test_resolve_wiki_scripts.sh
# observed: FAIL: resolve-wiki-14 no SKILL.md found under … — the scan graded nothing
```

## Open questions

**Should `refresh-resweep-29` be retired onto this guard?** `resolve-wiki-15` now strictly subsumes it — same predicate, a *broader* call detector, over a superset corpus that includes `knowledge-refresh`. Leaving it standing means two awk implementations of one predicate to keep in lockstep, which is arguably the shape this issue's own Motivation criticises ("building the general instrument and keeping the special case").

It is **deliberately left in place** here, not overlooked. Retiring a case in a second test file is a separate behaviour change, and the drift it risks is self-announcing: `29` goes RED rather than silently green if the source-line spelling changes. A reviewer who wants it retired should say so and it is a small follow-up. Flagging rather than filing, since the decision belongs with the maintainer.

**Not in scope:** #1720 (`probe_plugin` definition duplication). Different symbol, different property — `refresh-resweep-28` checks def/call co-location bidirectionally, this checks source-before-call one-directionally — and the issue's own Notes separate them. `probe_plugin` also appears in exactly one file today, so parameterizing the scanner by symbol now would abstract over a population of one. The seam is left cheap: the awk already takes `-v skill=`, so adding `-v sym=` once #1720 lands is a two-line change.


claude-cogni-work:
## Ship-ready

Authored by the `--role author` drain and handed off for independent merger review. The author instance did **not** review this PR, flip it ready, or apply any `svc:review-*` label — that is the merger's job.

### Handoff preflight — clean

`check-preflight.sh --head-ref origin/fix/issue-1725-derive-fence-scope-resolver-guard`, base `origin/main`, **exit 0**.

| Instrument | Status | Detail |
|---|---|---|
| `run-tests` | pass | suites green for cogni-knowledge |
| `mutation-recipe` | pass | recorded recipe is runnable as written — `resolve-wiki-15` found in the suite |
| `frontmatter` | skip | not_applicable — no touched SKILL.md or agents/*.md |
| `breadcrumbs` | skip | not_applicable — no touched SKILL.md or agents/*.md |
| `skill-length` | skip | not_applicable — no touched SKILL.md |
| `agent-length` | skip | not_applicable — no touched agents/*.md |

3 gating checks applicable, 3 passed, 0 failed, 4 skipped. The four skips are all genuine non-applicability: this diff touches a test script and a reference doc, no authoring surface.

The first preflight run **blocked** (exit 1) on `mutation-recipe` — the recorded invocation used `--root "$(git rev-parse --show-toplevel)"`, a shell metacharacter that is not replayable as written. The recipe was rewritten to a literal `--root .`, re-verified end-to-end (`guard_verified`), and the preflight re-run clean.

Graded by `contract-grading-reviewer` against the `review-contract` derived at triage time (`derived_at 2026-09-01T08:46:16Z`) — **before this branch was cut**, so the bar is diff-blind by chronology.

**22 items graded · 22 satisfied · 0 violated · 0 unclear · verdict `pass` · no residual.**

This is **advisory only** — no verdict label is applied and it does not substitute for the merger's independent review. Notable items the grader executed rather than took on trust:

- Ran the suite under **system `/bin/bash` 3.2.57** (not just the ambient Homebrew bash) — rc=0, all 15 cases green.
- Executed the PR body's **exact mutation recipe** against `knowledge-index/SKILL.md` — `guard_verified`, and `git status` clean afterwards, confirming the mutation is reverted.
- Forced `resolve-wiki-14`'s red arm via the scan-root override — the liveness case is real, not vacuous iteration.
- Confirmed prose mentions outside every fence (`knowledge-distill:56,58`) are not graded as call sites.
- Confirmed emitted ids are exactly `01`, `04`-`17` — `02`/`03` absent, nothing renumbered.

### For the reviewer

One judgment call is deliberately left open in the PR body rather than decided here: **whether `refresh-resweep-29` should be retired onto this guard**, now that `resolve-wiki-15` strictly subsumes it. It was left standing on the grounds that retiring a case in a second test file is a separate behaviour change and the resulting drift is self-announcing (`29` reddens rather than silently greening). The alternative root-cause plan proposed retiring it and lost the pick on blast radius. Reasonable to disagree — it is a small follow-up either way.

sdh07:

_The merger formed this expectation from issue #1725 before checking out the PR head — the review grades the diff against it._

**Expected approach:** Part 1 of `cogni-knowledge/tests/test_resolve_wiki_scripts.sh` must stop reading the hand-maintained `SOURCING_SKILLS` literal (base line 45, 8 names) and instead derive its population by scanning `cogni-knowledge/skills/*/SKILL.md` — at base 13 skills call the resolver (the 8 listed plus knowledge-prefill SKILL.md:85, knowledge-refresh:60/75/199, knowledge-distill:62/398, knowledge-index:102, knowledge-setup:179), so five callers sit entirely outside today's scope. The `grep -qF 'scripts/resolve-wiki-scripts.sh'` file-wide assertion (base line 88) must become a per-fence decision: a fence that calls `resolve_wiki_scripts` with no earlier in-fence `source`/`.` of `resolve-wiki-scripts.sh` fails, while a fence that sources without calling is exempt — the shape that let knowledge-refresh pass on lines 59/74 while a third fence was broken. The fence tracker must tolerate leading whitespace (knowledge-index indents its fence at SKILL.md:100-104) and both bare ``` and language-tagged openers, and the call detector must key on the underscore `resolve_wiki_scripts` so the hyphenated `resolve-wiki-scripts.sh` source line cannot self-satisfy a fence. A drift/anti-shrink case must cross-check the graded population against an independent census so a silently narrowed derivation goes RED, and Parts 2-4 (behaviour cases resolve-wiki-04..13, the vendored-only anti-regression fixtures documented at lines 13-18) must survive untouched. Per `cogni-knowledge/tests/README.md` rules 1-7 the new cases owe unique ids, same-id paired reachable arms (a loop needs the collect-then-pair shape, since `scripts/check-case-id-pairing.py` flags an unpaired `FAIL:` inside a loop), and no renumbering of existing ids; `cogni-knowledge/references/delegation-contract.md:39` today asserts case `resolve-wiki-03-*` and enumerates only 5 read/render skills as the sourcing population, so that claim goes stale unless updated.

**Acceptance checklist:**
- [ ] [fidelity: literal] Scope boundary: every path in the diff matches `cogni-knowledge/**` and nothing outside it — verify with `git diff --name-only <base>...<head>`. The predicate admits every artifact other items require (the suite itself, any new fixture or helper, and doc updates under `cogni-knowledge/references/`).
- [ ] [fidelity: literal] `cogni-knowledge/.claude-plugin/plugin.json` is not touched — its `version` still reads `1.0.105` as at base (the patch bump is applied post-merge). Verify: the file is absent from the changed-path list.
- [ ] [fidelity: purpose — no leaked issue-ref or version breadcrumb tokens survive in the authoring files the diff touches] The changed files carry no breadcrumb tokens (issue numbers, version strings) left over from authoring, per the deterministic breadcrumb guard.
- [ ] [fidelity: purpose — the population is computed from the tree, not restated in the suite; the variable name and scan mechanism are free] The hand-maintained enumeration is gone: `grep -n 'SOURCING_SKILLS' cogni-knowledge/tests/test_resolve_wiki_scripts.sh` shows no literal list of `knowledge-*` names gating the sourcing loop, and the population instead comes from a scan of `cogni-knowledge/skills/*/SKILL.md`.
- [ ] [fidelity: purpose — every skill whose SKILL.md calls the resolver inside a fence is graded by some case; the exact membership follows the tree rather than this list] Running `bash cogni-knowledge/tests/test_resolve_wiki_scripts.sh` grades each of the 13 base callers — knowledge-ingest, knowledge-finalize, knowledge-dashboard, knowledge-health, knowledge-lint, knowledge-resume, knowledge-ingest-source, knowledge-query, knowledge-prefill, knowledge-refresh, knowledge-distill, knowledge-index, knowledge-setup — i.e. the five formerly-unlisted skills are no longer ungraded.
- [ ] [fidelity: purpose — the coverage guard must fail on PARTIAL shrink of the graded population, not merely on a total-empty derivation] A separately-addressable drift case cross-checks the graded call-site set against an INDEPENDENTLY computed census (a set difference against e.g. `grep -l resolve_wiki_scripts cogni-knowledge/skills/*/SKILL.md`, or a pinned floor count), so dropping ONE caller from the derived set turns that case RED. Verify by reading the case's predicate: a predicate of the form `[ -z "$derived" ]` / `[ "$count" -eq 0 ]` alone does NOT satisfy this item. A diff that derives the set and grades per fence but adds no such shrink cross-check FAILS this item even if some case fires on an empty derivation.
- [ ] [fidelity: purpose — the failure condition is per fence: calls-without-an-earlier-in-fence-source, never per file] The sourcing assertion is fence-scoped: a file that sources in one fence and calls `resolve_wiki_scripts` in a different, unsourced fence is reported FAIL. Verify by reading the assertion (the offender condition is evaluated inside a per-fence loop, not by a whole-file `grep -qF`) plus the shipped negative fixture or the recorded mutation recipe that forces that state.
- [ ] [fidelity: purpose — a fence that sources without calling is not an offender; the exemption may be expressed as any conjunctive predicate] A fence that sources `resolve-wiki-scripts.sh` but issues no `resolve_wiki_scripts` call is exempt and produces no FAIL — verify the offender predicate requires a call in that fence (call AND no-source), so the exemption cannot be an ad-hoc name-based skip.
- [ ] [fidelity: literal] The fence-opener/closer pattern tolerates leading whitespace (anchored with a leading-whitespace class, not column-0-anchored) and matches both bare and language-tagged openers, so knowledge-index's indented fence at `cogni-knowledge/skills/knowledge-index/SKILL.md:100-104` and knowledge-refresh's bare fences at SKILL.md:58/73/196 track correctly and knowledge-index yields a true verdict.
- [ ] [fidelity: literal] The call detector keys on the underscore spelling `resolve_wiki_scripts` (not a loose pattern matching the hyphenated `resolve-wiki-scripts.sh`), so a fence that only sources the snippet is never counted as a calling fence and cannot become trivially self-satisfying. Verify the call regex/grep in the suite.
- [ ] [fidelity: purpose — every derived caller is treated uniformly; no per-skill or per-symbol carve-out] The suite carries no hardcoded per-skill branch, allowlist, skip list or exemption keyed on a `knowledge-*` name in the sourcing path — grep the diff for any such name-keyed conditional.
- [ ] [fidelity: literal] Parts 2-4 survive the rewrite: cases `resolve-wiki-05`/`06`/`08` (the vendored-only anti-regression fixtures documented at base lines 13-18), the entry-point cases `resolve-wiki-09`/`10`, and the unset-`CLAUDE_PLUGIN_ROOT` derivation cases `resolve-wiki-11`/`12`/`13` (bash, `/bin/bash`, zsh) are still present and still assert the same conditions.
- [ ] [fidelity: purpose — no pre-existing case id is repointed at a different assertion, so a `--case` recorded in an older PR body still addresses the guard it was written for] Existing behaviour ids `resolve-wiki-04` through `resolve-wiki-13` are not renumbered or reassigned; any net-new case takes a fresh unused number for this suite (`cogni-knowledge/tests/README.md` rule 3). Verify: ids in the diff, plus running the suite and checking the emitted id list has no duplicates (rule 1/6 uniqueness).
- [ ] [fidelity: literal] `python3 scripts/check-case-id-pairing.py` exits 0 with `success: true` — every new `FAIL: <id>` arm has a reachable same-id `PASS:`/`ok:` arm. For the per-skill scan this means the collect-then-pair shape (accumulate offenders in the loop, one fixed-id `if`/`else` after it closes); a PASS gated on the shared `errors` accumulator does not count.
- [ ] [fidelity: literal] `python3 scripts/check-result-line-plainness.py` exits 0 with `success: true` — no escape literal in the suite and no non-plain emitter definition.
- [ ] [fidelity: literal] `python3 scripts/run-plugin-tests.py --filter cogni-knowledge/tests` is green, with the rewritten suite passing against the base tree (all 13 callers currently source in-fence, so a correct derived guard must be GREEN at head, not red).
- [ ] [fidelity: purpose — the mutation lands in ANY calling skill other than knowledge-refresh, proving plugin-wide reach rather than a second per-file case; the specific skill and case id may differ from the one named here where the diff or PR body records the substitution] The PR body records a runnable `scripts/mutation-check.sh` recipe that removes the in-fence `source`/`.` line from a calling fence in a non-knowledge-refresh skill (e.g. `cogni-knowledge/skills/knowledge-index/SKILL.md:101` or `knowledge-setup/SKILL.md:178`) and demonstrates the named case going RED then GREEN on revert. Its `--file` is root-relative; its `--expr` is a `perl -0pi` expression — never a sed/ERE form, literal parens backslash-escaped, single-quoted at the call site; and its `--case` reproduces the exact label the suite prints (a `--case` that does not match resolves to `case_not_found` and grades nothing).
- [ ] [fidelity: purpose — no prose under cogni-knowledge/ is left asserting a case id or a covered population the rewritten suite no longer has] If the suite stops printing `resolve-wiki-03-*`, or the graded population is no longer the five read/render skills it names, `cogni-knowledge/references/delegation-contract.md:39` is updated to match. Verify: `grep -rn 'resolve-wiki-03' cogni-knowledge/` and cross-check each surviving claim against the ids the suite actually emits.

**Risk watchlist:**
- A derived-but-still-file-wide fix: the enumeration is replaced by a scan yet the assertion stays `grep -qF` over the whole file, so knowledge-refresh still passes on lines 59/74 while a third fence is broken — the exact defect #1721 exposed.
- An anti-shrink guard that only checks emptiness: the derivation could silently narrow from 13 callers to 2 and the suite stays green — the precise regression this contract must adjudicate.
- A regex that treats prose mentions of `resolve_wiki_scripts` outside fences as calls, producing false FAILs on a clean base — or, inversely, a skill derived into the population that has no calling fence at all being reported red rather than vacuously clean.
- A column-0-anchored fence pattern that silently mis-tracks knowledge-index (SKILL.md:100-104) and yields a false PASS there rather than an error — a false verdict, not a visible failure.
- A fence pattern that only matches language-tagged openers and misses the bare openers knowledge-refresh (58/73/196), knowledge-setup (177), knowledge-query (193) and knowledge-distill (396) use, shrinking coverage invisibly.
- A loop that emits `FAIL: <id>` per offending skill with no reachable same-id green arm, or a PASS gated on the shared `errors` accumulator — green in CI but unaddressable by `mutation-check.sh --case`, so the new guard could never be proven load-bearing.
- Renumbering existing ids to keep the sequence contiguous, silently repointing `--case` values recorded in older PR bodies (README rule 3) at different assertions.
- Collateral damage to Parts 2-4: the vendored-only anti-regression cases 05/06/08 exist to prove the removed sibling/marketplace-cache probes cannot come back; a rewrite that touches them weakens an unrelated invariant.
- A mutation demonstration performed on knowledge-refresh only — which would prove nothing new, since knowledge-refresh is exactly the file the old per-file guard already covered.
- Judgment-only: whether the derived population should also include the Python peer `_knowledge_lib.resolve_wiki_scripts` (delegation-contract.md:39 calls it a necessary separate copy). The issue scopes derivation to `skills/*/SKILL.md`, so widening is out of scope here, but a human may want a follow-up.
- Judgment-only: `cogni-knowledge/tests/test_knowledge_wiki_probe.sh` and `test_refresh_resweep_contract.sh` also reference `resolve-wiki-scripts.sh`; a human should sanity-check that the new guard does not duplicate or contradict assertions those suites already own.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1737

**Summary:** Merged needs_revision: contract-grading-reviewer's two major findings (review-contract:item-6, review-contract:item-18) are confident, substantive disagreements with plugin-validator's independent satisfied grading of the same loci — not intent ambiguity — so both are routed to human judgment per split-judgment-predicate.sh rather than resolved by the merge gate; severity and the needs_revision trigger are unchanged by the routing. Plugin-validator's two minor preference findings are explicitly waived and marked advisory-addition.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[major]** `review-contract:item-6` — The only shrink-adjacent cases (resolve-wiki-14's -eq 0 check and resolve-wiki-17's per-file mismatch) are both scoped to the same enumeration loop and never cross-check against a census computed outside it. Deleting an entire caller directory (knowledge-query; population 13->12 callers, 21->20 files) left every case PASS, rc=0 — no case reddened. DISPUTED: plugin-validator examined the same locus and graded it satisfied, having run a different shrink (unfencing one call, file still present and still greping positive), where resolve-wiki-17 did redden.
  - *Suggested fix:* Add a case computing the derived count against a census that does not share the fence-scan loop's iteration, or a pinned floor, and fail on divergence by even one.

- **[major]** `review-contract:item-18` — The case-id citation was correctly repointed to resolve-wiki-16, but delegation-contract.md:39 still reads as naming only the same five original read/render skills, though the rewritten suite's covered population is 13 derived callers and resolve-wiki-16 scans all 21 SKILL.md files. DISPUTED: plugin-validator examined the same locus and graded it satisfied, reading the sentence as scoped to those five skills' own behaviour rather than as an exhaustiveness claim.
  - *Suggested fix:* Update the sentence so it no longer implies resolve-wiki-16's scope is limited to the five named read/render skills — name the full derived population or generalize to 'every knowledge-* skill that calls the resolver'.

- **[minor]** (waived) (advisory-addition) `cogni-knowledge/tests/test_resolve_wiki_scripts.sh (awk source-detection regex)` — The in-fence source detector is line-anchored after optional whitespace, so a guarded sourcing form would not set hassource and its fence would be reported an offender. All 16 current call sites use the bare dot or source forms and pass; failure direction is a false RED, not a false GREEN.
  - *Justification:* No current call site takes the guarded form and the failure direction is fail-safe.

- **[minor]** (waived) (advisory-addition) `cogni-knowledge/tests/test_resolve_wiki_scripts.sh (resolve-wiki-17 census predicate)` — resolve-wiki-17's census is a whole-file grep that reads prose as well as fences while the derived set reads fences only, so a future SKILL.md mentioning resolve_wiki_scripts only in prose would go RED though the tree is clean. Benign today and fail-safe in direction.
  - *Justification:* Fail-safe direction, no live false-RED, and the asymmetry is precisely the set-difference mechanism contract item 6 names as acceptable.

### Positive observations

- Fence-scoped rewrite closes the #1721 defect class falsifiably: with the source line relocated to another fence, the base whole-file grep still matched while resolve-wiki-15 reported FAIL at the offending fence-open line. (plugin-dev:plugin-validator)
- Both mutations landed on non-knowledge-refresh skills and between them exercised the indented and the bare fence opener, empirically excluding both fence-pattern hazards on the watchlist. (plugin-dev:plugin-validator)
- A sourcing fence with no call is genuinely exempt: a prepended source-only fence in knowledge-lint left all four Part-1 cases green with the derived count still 13. (plugin-dev:plugin-validator)
- Case-id hygiene against tests/README.md rules 1/3/6/7 is exemplary: 02/03 retired with a deliberate gap rather than repointed, collect-then-pair shape correct, 0 pairing violations across 2393 emissions. (plugin-dev:plugin-validator)
- Manifest and registration surfaces untouched and healthy: plugin.json absent at 1.0.105, check-frontmatter clean at schema 4 over 37 files, check-breadcrumbs 0 offenders. (plugin-dev:plugin-validator)
- 16 of 18 contract items graded satisfied by BOTH perspectives independently, including items 5, 7, 9, 10, 12, 13, 14, 15, 16 and 17. (both perspectives)
- item 17 satisfied: mutation recipe replayed verbatim on knowledge-index: red -> green, guard_verified, clean worktree after. (cogni-service:contract-grading-reviewer)
- Split judgment on review-contract:item-6: contract-grading-reviewer graded it violated (deleting the knowledge-query caller directory entirely reddened nothing) while plugin-validator examined the same locus and graded it satisfied (unfencing a single call did redden resolve-wiki-17); routed to human judgment per split-judgment-predicate.sh.
- Split judgment on review-contract:item-18: contract-grading-reviewer graded it violated while plugin-validator examined the same locus and graded it satisfied; routed to human judgment per split-judgment-predicate.sh.

### Recommendation

---

sdh07:
## Held for a human — split judgment on two contract items

The merger's independent review reached `needs_revision`, but this is **not** an author bounce. Two perspectives examined the same two committed-contract loci and reached **confidently opposite** gradings, each with its own executed evidence. That is a property of the *pair*, not a doubt either reviewer self-reported, so `split-judgment-predicate.sh` routes it to a human rather than to `service-revise`.

`svc:needs-review` is deliberately **kept** and the PR stays draft — it is parked, not dequeued.

### `review-contract:item-6` — does the anti-shrink guard have teeth?

The contract demanded a drift case that cross-checks the derived call-site set against an *independently computed* census, so that dropping **one** caller reddens it — and stated that an `-eq 0` emptiness predicate alone does not satisfy it.

| | Experiment run | Result |
|---|---|---|
| `plugin-validator` → **satisfied** | Unfenced `knowledge-setup`'s call. The file is still present and still greps positive, but contributes no fenced call. Derivation 13→12, census stays 13. | `resolve-wiki-17` went **RED** naming `knowledge-setup` |
| `contract-grading-reviewer` → **violated** | Deleted the `knowledge-query` **directory** outright. 21→20 files scanned, 13→12 callers. | Everything stayed **green**, rc=0 |

Both ran real mutations and both reported accurately — they are testing different things. The unresolved question is what "a caller silently leaving the derived set" *means*:

- If it means **a caller stops being detected while its file remains**, `resolve-wiki-17` catches it and the item is satisfied.
- If it means **a caller's file leaves the tree entirely**, nothing catches it — and note that the contract's own suggested mechanism (`grep -l resolve_wiki_scripts skills/*/SKILL.md`) would *also* return 12 in that scenario. Only the contract's alternative form, a **pinned floor count**, would redden.

That matters because a pinned floor was considered and deliberately removed during authoring: a floor is the hand-maintained census this PR deletes, compressed into one number, and it false-reds a legitimate skill consolidation. **The maintainer's call is whether deleting a skill directory is drift the guard must catch, or a legitimate consolidation it must not false-red on.** The two readings are not reconcilable from the issue text.

### `review-contract:item-18` — is the five-skill enumeration now stale?

`delegation-contract.md:39`'s case-id citation was repointed `resolve-wiki-03-*` → `resolve-wiki-16` (both perspectives agree that part is correct). The dispute is the rest of the sentence, which still names five read/render skills:

- `contract-grading-reviewer` → **violated**: `resolve-wiki-16` now scans all 21 SKILL.md files, so a sentence naming five skills next to that citation implies a scope the guard no longer has.
- `plugin-validator` → **satisfied**: the sentence states what *those five* skills do, not that they are the exhaustive population; the citation is now broader than and still inclusive of them, so the sentence is more true than at base.

### Everything else passed

**16 of 18 contract items graded satisfied by both perspectives independently**, including the two that carry the issue's core intent — item 7 (fence-scoping, demonstrated by relocating a source line to a different fence: the old whole-file `grep -qF` still matched while `resolve-wiki-15` reported FAIL) and item 17 (the mutation recipe replayed verbatim on `knowledge-index`, a non-`knowledge-refresh` skill: `red` → `green`, `guard_verified`).

Deterministic gates are clean: `run-plugin-tests` 79/79, `check-case-id-pairing` 0 violations across 2393 emissions, `check-result-line-plainness` 0 violations, `check-breadcrumbs` 0 offenders, `plugin.json` untouched at 1.0.105.

### To clear this hold

Rule on the two items above. If a ruling supersedes the issue's `## Acceptance criteria`, write it back into that section on #1725 — the contract is derived from the AC, so a ruling that lives only here will be re-demanded on the next drain. Then remove `svc:needs-human-review` (leave `svc:needs-review`) and the PR re-enters the merger queue; do not hand-apply `svc:review-pass`.

sdh07:
## Maintainer ruling — both disputed items ruled `satisfied`, hold cleared

Ruled by @sdh07 on the two loci `split-judgment-predicate.sh` routed to `human_judgment`. Both resolve on evidence I reproduced against the tree, and both resolve in the author's favour. The governing halves are written back into #1725's `## Acceptance criteria` so the next derivation cannot re-demand the rejected bar.

### `review-contract:item-6` — anti-shrink guard → **satisfied**

`plugin-validator` governs. Item 6 offers its mechanisms **disjunctively**: *"a set difference against e.g. `grep -l resolve_wiki_scripts cogni-knowledge/skills/*/SKILL.md`, **or** a pinned floor count"*.

Measured on the tree: that `grep -l` census returns **exactly the same 13 skills** the derivation produces. So under `contract-grading-reviewer`'s experiment — deleting the `knowledge-query` directory — the census and the derived set both drop to 12 and the set difference is **empty**. The broad reading therefore condemns **the item's own first cited mechanism**. A reading that falsifies the exemplar the contract blesses cannot be what the contract meant, so the narrow reading governs.

The item's operative falsifier does not fire either. It reads *"a diff that derives the set and grades per fence but adds **no such shrink cross-check** FAILS this item"* — but `resolve-wiki-17` **is** that cross-check, in the set-difference form the item names, and it has teeth: unfence a call and the file still greps positive while contributing no derived call site, so it reddens naming the offender.

**Ruled: a skill directory leaving the tree entirely is legitimate consolidation, not drift.** A pinned floor count is explicitly **not** required and must not be added — a floor is the hand-maintained census #1725 exists to delete, compressed into one number, and it false-reds any legitimate consolidation. The author's `/simplify` decision to replace the `10` floor with a pure zero-check was correct.

Residual, recorded and deliberately not filed: deleting a directory that still *needs* the resolver goes unnoticed by this suite. The class is small — such a deletion breaks the skill outright and trips the route, skill-name and README-inventory guards.

### `review-contract:item-18` — the five-skill enumeration → **satisfied**

`plugin-validator` governs, on two independent grounds.

1. **The sentence is not a population claim.** It states that *those five* read/render skills source the helper — still true — and its citation clause reads "no **skill** carries an inline copy", which is unbounded. `resolve-wiki-16` accumulates `inline_def_offenders` across the whole `skills/*/SKILL.md` loop, unconditioned on calling, so the citation is accurate and strictly broader than at base.
2. **The broad reading condemns the base tree.** At base the graded population was `SOURCING_SKILLS` = **8** names, not 5. The sentence was already not an exhaustiveness claim before this diff, so the broad reading flags a pre-existing condition rather than anything this PR introduces.

The `resolve-wiki-03-*` → `resolve-wiki-16` repoint, which both perspectives already agreed was correct, is the only part of that line this diff owed. Confirmed: no `resolve-wiki-0[23]` reference survives anywhere in the repo at head.

### Open question — `refresh-resweep-29` stays, and the PR body overstates the case

**Keep it.** The body's claim that `resolve-wiki-15` "strictly subsumes" it is wrong. Case 29 carries a **`FAIL-NOCALL`** arm that reddens if `knowledge-refresh` renames the call away. Under that rename:

- `resolve-wiki-14` — global zero-check, still 12 callers → green
- `resolve-wiki-15` — no calling fence in that file, so no offender → green
- `resolve-wiki-17` — keys off the very string the rename changed → green
- `refresh-resweep-29` — **red**

It is a `knowledge-refresh`-specific anti-rename floor with no counterpart in the new suite. Not subsumed, so nothing to retire and no follow-up filed.

### Disposition

16 of 18 items were already graded satisfied by both perspectives independently; with these two ruled, the contract is fully satisfied. All 13 check-runs are `success` at `17170ed1`, `mergeStateStatus` is `CLEAN`, the branch contains current `main`, and the diff is 2 paths both under `cogni-knowledge/` with `plugin.json` untouched.

This is a **Step 3** hold — a `needs_revision` verdict already exists at this head, so Step 1a would reuse it and bounce to the author regardless of the corrected bar. There is no gating defect left to supply a fresh SHA. Clearing by **maintainer merge** as `sdh07` (author `claude-cogni-work`, so the two-party invariant holds). `svc:review-pass` is deliberately **not** hand-applied — the merge is the verdict.

PR #1742, opened 2026-09-01
cogni-workspace: floor PAGE_PARITY against silent shrinkage (L10)

### LINKED ISSUE #1697: cogni-workspace: PAGE_PARITY has no shrinkage floor, so a pinned page can be silently unpinned
## Current behavior

`cogni-workspace/tests/test-layering-claim-reconciled.sh` protects its `FORBIDDEN` literal set against silent shrinkage with a hardcoded count floor. Dropping a literal takes `l2_n` under the floor and turns L2 red — the header states this explicitly: *"The count floor below is what protects the set — dropping an entry takes l2_n under the floor and turns L2 red."*

`PAGE_PARITY` has no equivalent. `check_parity` iterates whatever the list happens to contain, so deleting an entry simply stops comparing that page and L4 / L5 / L7 all stay green. Grepping the file for every bare-integer count comparison turns up exactly one count-tied comparison — the literal floor. There is no cardinality assertion over `PAGE_PARITY` anywhere.

## Desired behavior

The same silent-removal protection the literal set already has: removing a page from `PAGE_PARITY` should turn a case red rather than quietly narrowing the guard's coverage.

## Motivation

`PAGE_PARITY` is the only thing pinning several wiki pages byte-identical across the two trees, and the list has grown several times. Each addition is a deliberate decision recorded in the header; nothing stops a later edit from removing one. The class is identical to the one the literal floor closes, and it was left open only because no acceptance criterion of the work that grew the list asked for it.

## Acceptance criteria

- [ ] `check_parity`, or a new L-prefixed case, asserts a hardcoded minimum number of pages actually compared, so removing an entry from `PAGE_PARITY` turns a case red
- [ ] The floor is a hardcoded numeric literal, never a count derived from `PAGE_PARITY` itself — a floor computed from the constant it guards moves with the deletion it is meant to catch and can never fire
- [ ] A runnable `scripts/mutation-check.sh` recipe in the PR's `## Mutation recipe` section shows the new floor going red when exactly one `PAGE_PARITY` entry is deleted, with the observed `mutated_result` / `restored_result` / `verdict` triple recorded rather than asserted
- [ ] `bash cogni-workspace/tests/test-layering-claim-reconciled.sh` exits 0, and the four sibling wiki suites stay green

## Context

Deferred from the `service-resolve` plan phase on the work that added two pages to `PAGE_PARITY` and raised the literal floor. Split out because that work's criteria are scoped to the literal set and to the two specific pages — none of them reaches the parity list's own shrinkage protection — and this is independently shippable.

claude-cogni-work:
Closes #1697.

## Summary

- Add case **L10** to `cogni-workspace/tests/test-layering-claim-reconciled.sh`: a hardcoded shrinkage floor over `PAGE_PARITY`, mirroring the `l2_n -lt 18` floor that already protects `FORBIDDEN_ALL`.
- The counted value is produced by **walking `PAGE_PARITY` at run time**, using the same `while IFS= read -r ... <<EOF` idiom `check_parity` itself uses (and the same `[ -n "$page" ] || continue` blank skip), so the count is provably the number of entries the checker iterates. The floor's right-hand side is the bare integer literal `12`.
- On failure L10 prints an expected-vs-found line — `  expected at least 12 pinned pages, found <n>` — and reports through the existing `pass` / `fail` helpers.
- Header prose records the floor's existence and **why the 12 is hardcoded** (a floor derived from `PAGE_PARITY` moves with the deletion it is meant to catch and could never fire); the note on the constant itself carries the **raise-in-lockstep obligation**, sited where an editor adding a 13th page is already reading.

## Scope decisions

- **Placement: a new L-prefixed case, not a floor inside `check_parity`.** Reachability decides it, and it was measured rather than assumed. A `compared -lt 12` check inside the checker returns 1 to *every* caller, so under a one-entry deletion **L5** and **L7** — both of which assert rc 0 — go red reporting a cardinality loss as parity drift, while **L4** already expects rc 1 and stays **green**. The shrinkage signal would land on two innocent cases and stay invisible in the one case whose label says "differs". A standalone case reds only itself; the manual mutation below confirms L1–L9 all stay green.
- **`PAGE_PARITY` membership is unchanged.** All 12 entries stay exactly as they are — this issue adds protection, never membership.
- **Diff confined to one file** under `cogni-workspace/tests/`. A repo-wide grep confirms this suite is the only reader of `PAGE_PARITY`.
- **`check_parity`'s existing liveness floor stays as-is.** It fires only on total emptiness (`compared -eq 0 && mismatches -eq 0`); L10 is the cardinality assertion it never was.
- **Next free case id is L10** — L1 through L9 are used with no gaps.
- Nothing deferred; no follow-up issues filed.

A **binding arm was planned and then dropped** during the refine pass, and the reason is worth recording: it would have built a fixture from `PAGE_PARITY`, run `check_parity`, and asserted rc 0 with no `MISSING`/`DRIFT` line, claiming to prove `compared == l10_n`. That proof is unobservable — `check_parity` never prints `compared` — and because the fixture is built by walking the very constant the checker walks, `MISSING` is structurally unreachable, so the arm would have been **green at any list size, including a shrunk one**. It would also have re-run a fixture loop L4 and L5 already run.

## Test plan

- [x] `bash cogni-workspace/tests/test-layering-claim-reconciled.sh` exits 0 and prints `PASS: L10 ...` (10/10 cases green).
- [x] Sibling wiki suites green: `test-wiki-bare-name-roster.sh` (28), `test-wiki-namespace-sync.sh` (8), `test-wiki-tree-parity.sh` (12), `tests/test_release_bundle_wiki.sh` (36, at the repo root), plus `test-mcp-declaration-hygiene.sh` (15).
- [x] One `PAGE_PARITY` entry deleted by hand: **only L10 went red** (`expected at least 12 pinned pages, found 11`); L1–L9 all stayed green. Restored, suite back to rc 0.
- [x] The mutation recipe below runs to `guard_verified`.

## Mutation recipe

The expression deletes exactly one `PAGE_PARITY` entry: `workflow-trends-to-solutions.md`, chosen because it is inert with respect to every existing assertion. (`arch-er-diagram.md` would be the wrong pick — it is hardcoded into L4's fixture and asserted **by name**, so mutating it would red L4 through L4's own assertion rather than through the new floor, a false attribution.) The pattern matches once; a zero-occurrence match is a hard `expr_no_op` error, never a silent pass.

`--root` is the repo root you are running against — substitute your own checkout:

```bash
  --root <your-insight-wave-checkout> \
  --file cogni-workspace/tests/test-layering-claim-reconciled.sh \
  --expr 's/^workflow-trends-to-solutions\.md\n//m' \
  --test 'bash cogni-workspace/tests/test-layering-claim-reconciled.sh' \
  --case L10
```

Observed — recorded from the actual run against this branch, not asserted:

- `mutated_result`: **red**
- `restored_result`: **green**
- `verdict`: **guard_verified**
- exit code: **0**; the file was byte-identical to its pre-run state afterwards.

## Review notes

- The `/simplify` pass collapsed a genuine three-site restatement of the "hardcoded on purpose / raise in lockstep" rationale down to two, so the L10 case comment now defers to the header — matching the convention L2's own case comment already follows (*"See the header on what this case cannot prove"*). Three of four cleanup angles flagged this independently.
- Three `/simplify` findings were **skipped**, deliberately: extracting a shared `assert_min_entries` helper (L2's `l2_n` is its plant-loop's fixture-path index, not a comparable counter, so the helper would have exactly one call site — and routing L2 through it would edit a case this issue's scope fence excludes); replacing the count loop with `printf | grep -c` (two angles rejected it on consistency and on spawning a subprocess to replace an in-shell loop, and it neighbours the derived-count shape AC2 forbids); and dropping the `l10_ok` flag (L3 already uses that shape for a single-assertion case, so it is house style).

## Plan record

The full resolution plan, including the rejected alternative and the refine pass, is recorded on the issue: https://github.com/cogni-work/insight-wave/issues/1697#issuecomment-5495926199

### DIFF
--- a/cogni-workspace/tests/test-layering-claim-reconciled.sh
+++ b/cogni-workspace/tests/test-layering-claim-reconciled.sh
@@ -23,6 +23,7 @@
 #   - a literal under an excluded path is NOT flagged
 #   - each page on the PAGE_PARITY allowlist is byte-identical across the two
 #     wiki trees
+#   - the PAGE_PARITY allowlist has not shrunk below its recorded size
 #   - a scan pointed at a missing or empty tree fails rather than reporting clean
 #
 # Literals, not a regex over "foundation". `foundation` alone has many legitimate
@@ -72,6 +73,14 @@
 # would move with the deletion it is meant to catch and could never fire.
 # Do not read L2 as a regression guard against editing a literal's wording; for
 # a mutation that a change can actually flip, target the parity checker (see L4).
+#
+# The parity allowlist has the same floor, for the same reason. PAGE_PARITY is
+# guarded by a hardcoded count in L10, because the per-page byte-identity arms
+# cannot supply one: L4 and L5 build their fixtures by walking PAGE_PARITY
+# itself, L7 runs against the real repo, and check_parity's only cardinality
+# assertion fires at zero rather than at a floor — so all three stay green when
+# an entry is dropped. Adding a page means raising that numeral in the same
+# commit; the note on the constant itself says so at the point of use.
 #
 # Phrase-level survivors. These lines keep foundation vocabulary on purpose and
 # must NOT be caught. Each says something about the workspace *state* other
@@ -245,6 +254,11 @@ cogni-workspace/libraries/svg-patterns.md
 # different one. Both pairs are byte-identical as they are added, so a green run
 # is not evidence they are guarded — mutating one copy is what shows the entry
 # has teeth.
+#
+# Raise the floor in lockstep. L10 pins this list's size to a hardcoded numeral.
+# Adding a page here without raising that numeral leaves the new entry — and
+# every entry added after it — unprotected against silent removal, which is the
+# class L10 exists to close.
 PAGE_PARITY='plugin-cogni-workspace.md
 workflow-install-to-infographic.md
 arch-er-diagram.md
@@ -592,6 +606,39 @@ if [ "$l9_ok" -eq 1 ]; then
   pass "L9 no manifest is exempt — every .claude-plugin manifest is scanned"
 else
   fail "L9 no manifest is exempt — every .claude-plugin manifest is scanned"
+fi
+
+# ---------------------------------------------------------------------------
+# L10 — the PAGE_PARITY allowlist has not shrunk. The parity arms above prove
+# each page ON the list is byte-identical across the trees; the count floor
+# below is what stops an entry being dropped unnoticed. See the header on why
+# that floor is a hardcoded numeral, and the note on the constant for what it
+# obliges of whoever adds a page.
+#
+# Why a standalone case rather than a floor inside check_parity. Were the floor
+# sited there, a dropped entry would make the checker return 1 to every caller —
+# so L5 and L7, which both expect rc 0, would go red reporting a cardinality
+# loss as parity drift, while L4, which already expects rc 1, would stay green.
+# The loss would surface on two cases it has nothing to do with and stay
+# invisible in the one case whose label says "differs". A standalone case reds
+# only itself.
+# ---------------------------------------------------------------------------
+l10_ok=1
+l10_n=0
+while IFS= read -r page; do
+  [ -n "$page" ] || continue
+  l10_n=$((l10_n + 1))
+done <<EOF
+$PAGE_PARITY
+EOF
+if [ "$l10_n" -lt 12 ]; then
+  echo "  expected at least 12 pinned pages, found $l10_n"
+  l10_ok=0
+fi
+if [ "$l10_ok" -eq 1 ]; then
+  pass "L10 the PAGE_PARITY allowlist has not shrunk"
+else
+  fail "L10 the PAGE_PARITY allowlist has not shrunk"
 fi
 
 # ---------------------------------------------------------------------------


claude-cogni-work:
## Ship-ready

Author instance has handed this PR off for independent review. It is draft, carries `svc:needs-review`, and was **not** reviewed or flipped ready by the author — that is the merger's job.

**Handoff preflight:** clean (rc 0). 3 gating verdicts, all pass — `run-tests` (suites green for cogni-workspace), `version-bump` (no version-line touch), `mutation-recipe` (recipe runnable as written; `L10` found in the touched file). 4 instruments skipped, each `not_applicable`: `frontmatter` and `breadcrumbs` (no touched `SKILL.md` or `agents/*.md`), `skill-length` (no touched `SKILL.md`), `agent-length` (no touched `agents/*.md`).

**Author self-grade against the issue-time contract** (18 items, derived at triage `2026-09-01T14:44:02Z`, before any branch was cut — so it is diff-blind by chronology):

- graded: **18** · satisfied: **18** · violated: **0** · unclear: **0**
- no fix pass was needed, and there is **no residual** to inherit.

This is advisory only — it carries no verdict label and no `reviewed-head-sha` marker, and it does not substitute for the merger's own independent review.

**One thing for the reviewer to weigh, surfaced rather than resolved.** Contract item 11 names four sibling suites, one of them `cogni-workspace/tests/test-mcp-declaration-hygiene.sh`. The issue's AC4 says "the four sibling wiki suites", and exploration found the issue's actual fourth wiki sibling is `tests/test_release_bundle_wiki.sh` at the repo root — `test-mcp-declaration-hygiene.sh` is arguably a manifest-hygiene suite, not a wiki one. The item is satisfied on either reading because **both** sets were run and all five suites exited 0, but the contract item itself looks mis-authored, and that is a contract-authoring observation a human may want to note rather than something this PR can fix.

**Verification worth re-running rather than trusting:** the guard's teeth were shown, not asserted — deleting one `PAGE_PARITY` entry reds **only L10** (L1–L9 stay green), and `mutation-check.sh` returns `mutated_result: red` / `restored_result: green` / `verdict: guard_verified` at exit 0 against the final branch state.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1742

**Summary:** Both perspectives independently graded all 18 committed review-contract items as satisfied on the same shipped-cardinality mutation guard, with zero findings and no locus disagreement — safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- The recorded mutation triple is genuinely a run, not an assertion — reproducing the recipe verbatim in an independent clone returned byte-identical values. (plugin-dev:plugin-validator)
- The floor has teeth against every member of the list: deleting each of the 12 PAGE_PARITY entries in turn reds L10 in all 12 cases. (plugin-dev:plugin-validator)
- The author avoided the exact false-attribution trap the risk watchlist called out — arch-er-diagram.md reds both L4 and L10 while the chosen entry reds L10 alone. (plugin-dev:plugin-validator)
- The rejected binding arm is recorded in the PR body with the reason it was dropped, and the pre-existing liveness guard is left intact. (plugin-dev:plugin-validator)
- item 5 satisfied: manually counted the shipped PAGE_PARITY list at head — exactly 12 filenames — matching the floor literal 12. (cogni-service:contract-grading-reviewer)
- item 16 satisfied: reproduced the full-suite output under the mutation by hand — only L10 goes red, L1-L9 including L4's arch-er-diagram.md-keyed assertion all stay PASS. (cogni-service:contract-grading-reviewer)
- item 17 satisfied: independently re-ran mutation-check.sh with the PR body's exact recipe and got guard_verified with exit 0, matching the recorded triple rather than trusting the assertion. (cogni-service:contract-grading-reviewer)
- item 18 satisfied: grepped the full diff for issue-reference and version-token patterns with zero hits; check-breadcrumbs.sh reports clean. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1758, opened 2026-09-02
cogni-knowledge: the check/grade census guards one of the three suites that share the convention


claude-cogni-work:
## Plugin
cogni-knowledge

## Current behavior
#1753 added `klib-50`, a bidirectional check/grade census, to `cogni-knowledge/tests/test_knowledge_lib.sh`. It scans `"$0"` — its own file — so it guards exactly one suite.

Three suites in `cogni-knowledge/tests/` share the two-part convention it guards (a python-side `check("<tag>", assert_<tag>)` registration whose verdict is only ever read by a matching bash-side `grade <tag> "…"` line, each carrying its own byte-identical copy of `grade()`):

- `test_knowledge_lib.sh` — guarded as of #1753
- `test_ingest_contract.sh` — **unguarded**
- `test_pdf_extract.sh` — **unguarded**

The other suites in that directory are structurally immune and are **not** in scope here: `test_refresh_resweep_contract.sh` and `test_knowledge_wiki_probe.sh` use `assert_grep` / inline `green`+`red`, where the call *is* both the registration and the grading in one statement; `test_control_path_flip.sh` has a python-side `check(label, cond, detail)` that accumulates into `failures` and self-grades via `sys.exit(1 if failures else 0)`.

## Desired behavior
All three suites that share the convention are guarded against the silent-orphan class, without three divergent copies of the census.

## Motivation
Both suites are balanced today (4 registrations / 4 grade lines each, verified), so nothing is currently hiding — the exposure is future. A `check("…")` added to either without its `grade` line is silently uncounted in exactly the way `page_type_line` and `parse_distilled_claims_with_backlinks` were in #1753, and the suite stays green.

Copy-pasting the census into the other two is the wrong fix: it is ~20 lines carrying four `|| true` guards (required because `set -eu` is on and a zero-match grep would abort the suite before the case could report), two anchoring subtleties (the extractors must match neither their own pattern literals nor the `grade() {` definition), and an `LC_ALL=C` collation pin. That is precisely the shape that diverges across copies.

This was deliberately **not** done inside #1753: it would have shipped a shared helper whose only motivated caller was the one file, and #1753's own scope fence admitted only `cogni-knowledge/` paths.

## Proposed solution
[...]

PR #1759, opened 2026-09-02
cogni-knowledge: grade two orphaned klib cases + permanent check/grade census

### LINKED ISSUE #1753: cogni-knowledge: two orphaned check(...) cases in test_knowledge_lib.sh are never graded
## Plugin
cogni-knowledge

## Current behavior
`cogni-knowledge/tests/test_knowledge_lib.sh` grades cases through a **two-part** convention:

1. an `assert_<tag>()` function registered via `check("<tag>", assert_<tag>)` inside the python heredoc (lines 72-1384), which prints `"<tag>: OK"` or `"<tag>: FAIL <msg>"` into `$OUT`; and
2. a matching `grade <tag> "klib-NN …"` line in the bash section (lines 1401-1442), which is what actually greps `$OUT` and increments `errors`.

The `grade` helper (lines 1389-1399) is the only thing that reads a case's result. So a `check(...)` with **no** matching `grade` line runs its assertion, prints its verdict into `$OUT`, and is then never read — silently invisible, never counted, and **unable to fail the suite**.

Two such orphans exist today:

- `check("page_type_line", assert_page_type_line)` — line 1364
- `check("parse_distilled_claims_with_backlinks", assert_parse_distilled_claims_with_backlinks)` — line 1375

Neither has a `grade` line anywhere in the 1401-1442 block. A regression in either function would leave the suite green.

There is no secondary safety net: `scripts/run-plugin-tests.py` classifies purely on exit code and never parses stdout (its own docstring: *"this runner never parses stdout to decide an outcome. Exit status is the whole signal."*).

Note the asymmetry — the *inverse* orphan is safe. A `grade` with no `check` falls into `grade()`'s `*` catch-all, prints `FAIL: … (no result line …)` and increments `errors`. Only the missing-`grade` direction is silent, which is why this went unnoticed.

## Desired behavior
Both cases are counted toward the suite's `errors` total, and no third orphan can hide.

## Motivation
Two assertions currently pay their runtime cost and buy nothing. Until they are graded, `test_knowledge_lib.sh` overstates its own coverage by two cases.

## Acceptance criteria
- [ ] `grade page_type_line "klib-NN …"` exists in the bash grade block and the case reports PASS. Evidence: `bash cogni-knowledge/tests/test_knowledge_lib.sh` prints a `PASS:` line whose description begins with that case's `klib-NN` token. Falsified by the run printing no line for `page_type_line`.
- [ ] `grade parse_distilled_claims_with_backlinks "klib-NN …"` exists and the case reports PASS, by the same evidence. Falsified the same way.
- [ ] A census confirms every `check(` tag in the file has exactly one matching `grade` tag, and vice versa. Evidence: extracting the tag from every `check("<tag>"` line and from every `grade <tag>` line yields two identical sets. Falsified by any tag present in one set and absent from the other.
- [ ] The new ids continue the allocation counter rather than reusing one. Evidence: the two chosen `klib-NN` ids are absent from the file at base. Falsified by either id already appearing at `origin/main`.
- [ ] If grading either case surfaces a **pre-existing** failure, it is fixed in the same PR and the fix is described. A red case that is merely re-silenced does not satisfy this issue.
- [ ] Scope boundary: `git diff --name-only origin/main...HEAD` lists only paths beginning `cogni-knowledge/`, and `git diff origin/main...HEAD -- '*plugin.json'` is empty.

_Deferred from #1748 (citation-family-aware primitives), which touches the same test file but adds correctly-wired cases and does not fix these two orphans._

_Affected components: cogni-knowledge/tests/test_knowledge_lib.sh_

_Complexity hint: low_

claude-cogni-work:
Closes #1753.

## Summary

- Add `grade page_type_line "klib-48 …"` and `grade parse_distilled_claims_with_backlinks "klib-49 …"` to the grade-call block of `cogni-knowledge/tests/test_knowledge_lib.sh`, so the two registrations at base lines 1504 and 1515 finally drive the suite's `errors` total instead of being silently invisible.
- Add **klib-50**, a permanent in-suite census that runs on every invocation and folds into `errors`: it extracts the registration-tag set and the grade-tag set from the suite's own source (`$0`) and fails on any registration-without-grade, any grade-without-registration, any duplicate on either side, or an empty extraction on either side.
- New ids continue the allocation counter — klib-01 … klib-47 exist at base (contiguous, max 47), so 48/49/50 are the next free ids. No existing klib-02 … klib-47 grade line is modified, and no registration is removed.

One file, 48 added lines, nothing deleted.

## Why the census is permanent rather than a one-time check

The issue's Desired behavior is *"no third orphan can hide"*. That is false under a one-time verification: a census run once during development constrains nothing about the next case someone adds — which is exactly how these two arose. So the census ships as a case in the suite, not as a command in this PR body.

It is **bash-only**, placed after the grade-call block and before the final `if [ $errors -gt 0 ]` arm. The python heredoc runs as a subprocess (`OUT=$(python3 - "$SCRIPTS_DIR" "$WORK" <<'PY' … PY)`) with no handle on the shell script's own source path — and more decisively, half the census's subject, the `grade` lines, exists *only* as script text and never enters that subprocess at all. `klib-01` (base 43-64) is the file's own precedent for a bash-only case. Unlike `klib-01` this one is **non-fatal**: it increments `errors` rather than `exit 1`, so the rest of the report still prints.

## The three traps it is written against

1. **`set -eu` (base line 24).** A zero-match `grep` exits 1, which under `set -e` would abort the whole suite *before* the census could print its own FAIL line — and would also leave `mutation-check.sh` with no result line to classify, reporting `case_not_found` instead of a red verdict. Every extraction is `|| true`-guarded, following the file's own precedent at base line 1536 (`… | grep "^${tag}:" || true`).
2. **Self-matching extractors.** A bare `check\(` also matches the `def check(tag, fn):` definition and a prose comment (50 hits vs the correct 48); a bare `grade` matches the `grade() {` definition at base 1533 and four comment lines using the English word "grades"/"degrades". The anchored forms give exactly 48 and 46 at base, 48 and 48 at HEAD. They also cannot match the census's *own* pattern literals: the source bytes there are an escaped paren, never a bare `check(`, and no census line begins at column 0 with `grade` followed by whitespace.
3. **Locale collation.** `comm` compares bytewise while `sort` is locale-sensitive, and the tags contain `_`, whose collation weight differs between a UTF-8 locale and C. Both sorts and both `comm` invocations are pinned to `LC_ALL=C` so the two streams cannot desynchronise on a differently-configured runner.

The two extracted sets land as files in the suite's existing `$WORK` temp dir (base line 66, already `trap`-cleaned on EXIT) rather than in shell variables. That is what lets emptiness be `[ ! -s … ]` and lets `uniq -d` / `comm` read the files directly — a variable round-trip through `printf '%s\n'` emits a phantom blank line for an empty set, which every consumer would then have to scrub separately.

## Scope decisions

- **In:** one file, `cogni-knowledge/tests/test_knowledge_lib.sh`.
- **Out:** no `_knowledge_lib.py` change, no other test file, no `plugin.json` (the patch bump is post-merge).
- **Neither orphaned assertion is weakened to reach green.** `assert_page_type_line` (base 274-293) and `assert_parse_distilled_claims_with_backlinks` (base 596-638) are byte-identical in this diff — the diff adds 48 lines and deletes none. They were already passing under the python subprocess; grading them simply makes their result count.
- **The issue's "fix any pre-existing failure surfaced by grading" criterion is vacuous here**, and that is a finding rather than an omission: the suite is exit 0 / 47 PASS / 0 FAIL at base, and exit 0 / 50 PASS / 0 FAIL with all three new cases wired. Grading the two orphans surfaced nothing red, so nothing is being re-silenced.

## Verification

Run at the branch head:

- `bash cogni-knowledge/tests/test_knowledge_lib.sh` → exit 0, **50 PASS / 0 FAIL** (47 at base + klib-48/49/50).
- Each new case prints its id as the second whitespace-separated token, per the invariant recorded at base line 49 — confirmed for `klib-48`, `klib-49`, `klib-50`.
- `grep -c '^grade page_type_line ' cogni-knowledge/tests/test_knowledge_lib.sh` → 1 at HEAD, 0 at `origin/main`; same for `'^grade parse_distilled_claims_with_backlinks '`.
- Registration-tag count and grade-tag count are both **48** at HEAD (48 / 46 at base).
- `git diff --name-only origin/main...HEAD` lists only `cogni-knowledge/tests/test_knowledge_lib.sh`; `git diff origin/main...HEAD -- '*plugin.json'` is empty.

**All five verdict arms were driven red by hand and each reported its own named bucket** — re-run after the pre-commit `/simplify` pass rewrote the block, not only before it. The file was restored between each, and the suite returns to exit 0 / 50 PASS afterwards:

| Mutation applied | Suite rc | Verdict reported |
|---|---|---|
| delete the existing `grade identity …` line | 1 | `ORPHAN-REGISTRATION … : identity` |
| add `grade nonexistent_tag "klib-98 …"` with no registration | 1 | `ORPHAN-GRADE … : nonexistent_tag` |
| duplicate the `grade identity …` line | 1 | `DUP-GRADE … : identity` |
| duplicate the `check("identity", …)` registration | 1 | `DUP-REGISTRATION … : identity` |
| rename `grade()` and all its call sites so the grade-side extractor matches nothing | 1 | `EMPTY - registrations=48 grade lines=0` |

That last row is also the direct proof of trap 1: with the extraction returning nothing, the suite *reports* rather than aborting under `set -eu`.

## Mutation recipe

Deletes the existing `grade identity …` line (base 1545). `identity` is unique as both a registration tag and a grade tag, so removing it makes the registration side strictly larger and must flip klib-50 green → red without touching any other case's grading.

Run from the root of a checkout of this branch. `mutation-check.sh` ships with `cogni-service` in the `managed-service` repo — `insight-wave`'s own `cogni-consult/scripts/mutation-check.sh` and `cogni-portfolio/scripts/mutation-check.sh` are different, bespoke scripts and are not this harness, per `cogni-knowledge/tests/README.md:69-72`:

```
  --root . \
  --file cogni-knowledge/tests/test_knowledge_lib.sh \
  --expr 's{^grade identity[ \t][^\n]*\n}{}m' \
  --test 'bash cogni-knowledge/tests/test_knowledge_lib.sh' \
  --case klib-50
```

Expected: `guard_verified` — `PASS: klib-50 …` at HEAD, `FAIL: klib-50 …` under the mutation.

## Follow-up issues

- **#1758** — cogni-knowledge: the check/grade census guards one of the three suites that share the convention.

`test_ingest_contract.sh` and `test_pdf_extract.sh` carry the same two-part convention, each with its own byte-identical `grade()`, and neither is guarded by this census. Both are balanced 4/4 today, so nothing is hiding — the exposure is future. Deliberately not folded in here: it would ship a shared helper whose only motivated caller is this file, and the repo-tier variant (`scripts/check-check-grade-pairing.py`, alongside the existing CI-wired `check-case-id-pairing.py`) lands outside this issue's `cogni-knowledge/` scope fence.

Not in scope for that follow-up, and checked: `test_refresh_resweep_contract.sh` and `test_knowledge_wiki_probe.sh` use `assert_grep` / inline `green`+`red`, where the call *is* both registration and grading; `test_control_path_flip.sh` self-grades via `sys.exit(1 if failures else 0)`. All three are structurally immune to this defect class.

## Plan record

Resolution plan: https://github.com/cogni-work/insight-wave/issues/1753#issuecomment-5504677493

### DIFF
--- a/cogni-knowledge/tests/test_knowledge_lib.sh
+++ b/cogni-knowledge/tests/test_knowledge_lib.sh
@@ -1588,6 +1588,54 @@ grade citation_family_dispatch "klib-44 citation_family (#1748) — the first fu
 grade strip_author_date_citation_markers "klib-45 strip_author_date_citation_markers (#1748) — APA/MLA/Harvard markers removed with no residue, multiple per sentence, no-op when absent; scheme-anchored so ordinary parenthesized prose links and the [[N]] anti-pattern survive; strip_citation_markers BRANCHES (author-date marker survives under ieee, numbered marker survives under apa)"
 grade extract_author_date_citation_urls "klib-46 extract_author_date_citation_urls (#1748) — the edge set klib-11 pins, transposed: appearance order, file:// first-class, unbracketed file:// with a literal space captured whole, angle-bracketed form, mixed file+http both in order, URL-less marker and ''→[]; scheme-anchored; extract_citation_urls dispatches per family"
 grade build_author_date_reference_list "klib-47 build_author_date_reference_list (#1748) — dedup by SOURCE identity not rendered string, alphabetical by surname across 'Last, First' and 'First Last', un-numbered (no **[N]**/[[N]]), ('','')+surname-less publisher surrogate sort last without raising, slug tiebreak, empty/None→[]"
+grade page_type_line "klib-48 page_type_line (#931) — the reader-facing Type: <Display> · <stage> header: per-type display name and stage word for all seven types, U+00B7 middle dot as the separator and never an ASCII substitute, case-insensitive key, and the fail-safe arms (unknown -> title-cased + raw, None/empty -> Unknown + raw, non-str int coerced) that must never raise"
+grade parse_distilled_claims_with_backlinks "klib-49 parse_distilled_claims_with_backlinks (#885) — the dual-level retrieval join reader: claim_id + text + backlinks and nothing else, inline list parsed with or without a space after the comma, order preserved, missing/empty backlinks normalized to [] so every claim carries the key, the with_id sibling left byte-identical (additive), and inline []/no key/empty/no-frontmatter -> []"
+
+# --- klib-50: check/grade census -------------------------------------------
+# The convention this file grades through has no structural enforcement: a
+# python-side registration prints "<tag>: OK"/"<tag>: FAIL" into $OUT, and only
+# a bash-side grade line ever reads it, so a registration with no grade line is
+# never read and can never fail the suite. #1753 found two such orphans.
+#
+# Bash-only for the same reason klib-01 (line 43) is: the python heredoc runs as
+# a subprocess with no handle on this script's source, and half the subject here
+# — the grade lines — is script text that never enters it. Unlike klib-01 this
+# case is NON-FATAL, folding into $errors so the rest of the report still prints.
+#
+# Three properties are load-bearing. `set -eu` is on (line 24) and grep exits 1
+# on zero matches, so every extraction is `|| true`-guarded as grade() is at line
+# 1536 — otherwise the suite aborts before this case can report. The extractors
+# are anchored so they match neither their own pattern literals (the bytes here
+# are an escaped paren, never a bare one) nor the grade() definition. LC_ALL=C
+# keeps sort and comm byte-collated, since the tags contain `_`.
+grep -oE 'check\("[a-z0-9_]+"' "$0" | sed 's/^check("//; s/"$//' | LC_ALL=C sort > "$WORK/census_reg" || true
+awk '/^grade[ \t]/{print $2}' "$0" | LC_ALL=C sort > "$WORK/census_graded" || true
+census_dup_reg=$(LC_ALL=C uniq -d "$WORK/census_reg" | tr '\n' ' ' || true)
+census_dup_graded=$(LC_ALL=C uniq -d "$WORK/census_graded" | tr '\n' ' ' || true)
+census_only_reg=$(LC_ALL=C comm -23 "$WORK/census_reg" "$WORK/census_graded" | tr '\n' ' ' || true)
+census_only_graded=$(LC_ALL=C comm -13 "$WORK/census_reg" "$WORK/census_graded" | tr '\n' ' ' || true)
+
+census_verdict="PASS"
+if [ ! -s "$WORK/census_reg" ] || [ ! -s "$WORK/census_graded" ]; then
+  census_verdict="EMPTY - registrations=$(wc -l < "$WORK/census_reg") grade lines=$(wc -l < "$WORK/census_graded"); an extractor matched nothing, so the convention was renamed or the scan is broken"
+elif [ -n "$census_dup_reg" ]; then
+  census_verdict="DUP-REGISTRATION - tag registered more than once: $census_dup_reg"
+elif [ -n "$census_dup_graded" ]; then
+  census_verdict="DUP-GRADE - tag graded more than once: $census_dup_graded"
+elif [ -n "$census_only_reg" ]; then
+  census_verdict="ORPHAN-REGISTRATION - registered but never graded, so its result is never read: $census_only_reg"
+elif [ -n "$census_only_graded" ]; then
+  census_verdict="ORPHAN-GRADE - graded but never registered: $census_only_graded"
+fi
+
+census_desc="klib-50 check/grade census - every registered tag has exactly one grade line and every grade line exactly one registration; duplicates on either side and an empty extraction are rejected too, so a third orphan cannot hide"
+if [ "$census_verdict" = "PASS" ]; then
+  green "PASS: $census_desc"
+else
+  red "FAIL: $census_desc ($census_verdict)"
+  errors=$((errors + 1))
+fi
+
 
 if [ $errors -gt 0 ]; then
   red "$errors case(s) failed."


claude-cogni-work:
## Ship-ready

Author instance has handed this PR off for independent review. Draft, carrying `svc:needs-review`. No review verdict was formed on this side — that is the merger's call.

**Handoff preflight:** exit 0, clean. `discovered=7 resolved=7 min=7`, `gating_applicable=3 verdicts=3 passed=3 failed=0`, `skipped=4`.
- Graded: `run-tests` (suites green for cogni-knowledge) · `version-bump` (no version-line touch) · `mutation-recipe` (recorded recipe runnable as written — `klib-50` found in the file).
- Skipped, all `not_applicable` on a test-file-only diff: `frontmatter`, `breadcrumbs` (no touched `SKILL.md` or `agents/*.md`), `skill-length`, `agent-length`.

**Author self-grade against the issue-time contract** (`--target issue --origin triage`, derived 2026-09-02T04:48:25Z, before any branch was cut — 14 items): **14 satisfied, 0 violated, 0 unclear.** No fix pass was needed and there is **no residual** for the merger to inherit.

Two things worth the reviewer's attention, neither a residual:

1. **The shipped grade-side extractor is `awk '/^grade[ \t]/{print $2}'`, not the `grep -oE … | awk` form the contract's `expected_approach` anticipated.** That substitution came out of the pre-commit `/simplify` pass. The grader judged it against the anchoring *property* the item names — excluding the `grade() {` definition — rather than the anticipated spelling, and confirmed it yields 46 at base / 48 at HEAD with an empty set-difference in both directions at HEAD. Flagging it because it is a deliberate divergence from the contract's own worked example.
2. **The issue's "fix any pre-existing failure surfaced by grading" criterion is vacuous**, and that is a finding rather than a skipped item: both orphaned assertions already pass at base (verified twice, independently, before the fix was written). Nothing red is being re-silenced. If you expected this PR to contain a behavior fix, that expectation is what is wrong, not the PR.

**Evidence the merger may want to re-run rather than trust:** all five census verdict arms were driven red by hand *after* `/simplify` rewrote the block, each reporting its own named bucket, file restored between each — the table is in the PR body. The `## Mutation recipe` section is executable as written (literal flag values, repo-root-relative `--file`, `--case klib-50` matching the printed label token).

Follow-up filed from this run: **#1758** (the census guards one of the three suites sharing the convention).

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1759

**Summary:** Both perspectives independently executed all 14 committed contract items and graded every one satisfied; the only findings are two merger-added preference-level nits from plugin-validator, both demoted as advisory additions (advisory-addition-predicate.sh: demotable=true) and both minor, so the PR passes clean.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/tests/test_knowledge_lib.sh:1611` — The registration extractor's character class admits only lowercase snake_case tags. All 48 tags in the file today are lowercase snake, and a mis-cased tag that IS graded still fires loudly as ORPHAN-GRADE because the awk side prints $2 verbatim. The residual silent window is narrow and hypothetical: a future registration with an uppercase or hyphenated tag AND no grade line would be invisible to both sides at once. Contract item 12 describes the mechanism as extracting from every check("<tag>" line, which this matches, so no acceptance item is at risk - a hardening observation.
  - *Suggested fix:* Optional follow-up: widen the character class, or pair it with a count assertion against the loose form minus the known non-registration lines.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/tests/test_knowledge_lib.sh:1620` — Cosmetic: the EMPTY arm's operator message interpolates the wc -l of the census file unpadded, so on macOS/BSD it renders with wc's leading column padding - the actual line reads 'EMPTY - registrations=      48 grade lines=       0'. The PR body's verification table renders that row without the padding, so the recorded output is slightly idealized versus what the suite actually prints. The arm fires correctly and names the right bucket, so this is presentation only, not a fault.
  - *Suggested fix:* Optional follow-up: strip the wc padding in the EMPTY message, and correct the PR-body table row to the literal printed text.

### Positive observations

- The mutation recipe in the PR body executes verbatim to guard_verified (mutated red, restored green) - all flag values literal, --file repo-root-relative, --case klib-50 binding to the printed label token rather than resolving to case_not_found. (plugin-validator)
- Four further census arms beyond the recorded recipe (ORPHAN-GRADE, DUP-GRADE, DUP-REGISTRATION, EMPTY) each drove the suite to rc=1 with its own distinctly named bucket, so the multiplicity and bidirectionality claims are executed facts, not assertions. (plugin-validator)
- The EMPTY arm is the concrete answer to the risk-watchlist 'trivially green over two empty sets' hole - with the grade side renamed out of existence the census reports rather than passing, and reports rather than aborting under set -eu, proving the || true guards. (plugin-validator)
- The census avoids the self-matching trap in both directions by construction, not by luck. Extraction yields 48/48 with an empty comm difference both ways. (plugin-validator)
- The block increments errors rather than printing a bare red line, so it is blocking under run-plugin-tests.py's exit-code-only classification. (plugin-validator)
- Both new grade descriptions are substantive, naming the specific contract each case protects rather than being bare-id stubs. (plugin-validator)
- The PR body is honest about the pre-existing-failure criterion being vacuous, and the diff is 48 additions / 0 deletions - the strongest possible evidence that no assertion was softened. (plugin-validator)
- The follow-up scoping (#1758) correctly identifies which sibling suites share the convention and which are structurally immune; confirmed independently. (plugin-validator)
- item 1: changed-file set is exactly the one mandated path; no plugin.json diff. (contract-grading-reviewer)
- item 2/3: both grade lines present at HEAD (lines 1591, 1592), absent at base, inside the grade-call block before the errors arm. (contract-grading-reviewer)
- item 4: unconditional census block increments errors via red + errors+=1, a direct match to the cited precedent shape rather than a substitution needing recorded reasoning. (contract-grading-reviewer)
- item 5: all three new PASS lines carry the klib-NN id as the second whitespace-separated token. (contract-grading-reviewer)
- item 6: klib-48/49/50 each absent at base; base max 47; contiguous, no gap, no reuse. (contract-grading-reviewer)
- item 7: zero deleted content lines in the diff. (contract-grading-reviewer)
- item 8: single diff hunk starts at line 1588, so both assertion bodies are outside it and byte-identical to base. (contract-grading-reviewer)
- item 9: suite exits 0 with 50 PASS and 0 FAIL. (contract-grading-reviewer)
- item 10: comm -23 and comm -13 both checked and both drive red. (contract-grading-reviewer)
- item 11: uniq -d against non-deduplicated sorted multisets with explicit DUP arms - a duplicate is rejected independently of the equality check. (contract-grading-reviewer)
- item 12: independent re-extraction outside the script gives 48 vs 48 with zero diff; the awk extractor provably does not match the grade() definition, and the check extractor provably does not match its own source line. (contract-grading-reviewer)
- item 13: ran mutation-check.sh with the recorded literal flags - guard_verified true (mutated red, restored green). (contract-grading-reviewer)
- item 14: check-breadcrumbs clean; check-structure-all clean across all 8 plugins. (contract-grading-reviewer)

### Recommendation

---

Issue #1763, opened 2026-09-02
repo: decide whether the retirement-ledger gate covers plugins beyond cogni-workspace, and extend it if so


claude-cogni-work:
Deferred from the resolution of #1723, which adds `scripts/check-retirement-ledger.py`.

## Current behavior

The guard's subject is hardcoded to `cogni-workspace/skills/<one-level>/SKILL.md`, because the retirement ledger it reads — `cogni-workspace/references/retired-trigger-phrases.tsv` — is cogni-workspace's own. Every other plugin with a `skills/` tree (`cogni-portfolio`, `cogni-trends`, `cogni-knowledge`, `cogni-consult`, `cogni-marketing`, `cogni-sales`, `cogni-website`, `cogni-docs`) can retire a skill and drop its trigger phrases with nothing recording it.

## Desired behavior

A decision, recorded in the repo, on whether the retirement-ledger invariant should hold repo-wide — and if so, the guard extended to enforce it.

## Motivation

This is a **design call before it is a code change**, which is why it was not folded into #1723. Extending the guard requires first settling whether each plugin carries its own ledger (matching the current per-plugin `references/` convention, but multiplying the files a guard must locate) or whether one shared repo-root ledger owns every plugin's retirement events (one file, but it breaks the ledger's current co-location with the suite that reads it). Shipping either shape without that decision would bake it in silently.

Note the coupling: whichever way it goes, C10 in `cogni-workspace/tests/test-skill-trigger-phrases.sh` and the parity case C11 both read the cogni-workspace ledger by path, so a relocation is not a guard-only change.

## Acceptance criteria

- [ ] The per-plugin-ledger vs shared-ledger question is decided and the decision is recorded in the repo (CLAUDE.md or a reference doc), with its rationale.
- [ ] If the decision is to extend coverage, `scripts/check-retirement-ledger.py` enforces the invariant for every plugin in scope, with a fixture case per newly-covered tree.
- [ ] If the decision is to keep it cogni-workspace-scoped, that scoping is stated where a reader of the guard would look, so the narrowness reads as deliberate rather than as an oversight.

claude-cogni-work:
## Claimed for triage

cogni-service has claimed this issue at pick time (`svc:in-progress`, applied 2026-09-03T12:44:49Z by the author identity). Triage and, if it routes there, resolution follow in this drain.

A sibling drain reading this claim should back off rather than re-pick.

claude-cogni-work:
## Triage Report

[...]

Issue #1762, opened 2026-09-02
cogni-workspace: the retirement-ledger gate does not cover a phrase dropped from a MODIFIED surviving skill description


claude-cogni-work:
Deferred from the resolution of #1723, which adds `scripts/check-retirement-ledger.py` — a merge-base-anchored guard that fails a PR **deleting** a `cogni-workspace/skills/<name>/SKILL.md` whose trigger phrases gain neither a retirement-ledger row nor a live re-claim.

## Current behavior

That guard scopes its subject with `--diff-filter=D`: it only ever looks at deleted `SKILL.md` paths. A trigger phrase can leave the claim surface without any file being deleted — an author edits a surviving skill's `description:` and drops one of its double-quoted phrases. The phrase then vanishes from the live set with no ledger row and no sibling re-claim, and nothing notices: C10 is blind to it for the same reason it is blind to the deletion case (the phrase is in neither the `claimed` nor the `retired` set, so neither `comm` direction fires), and the new deletion guard never looks at modified files.

## Desired behavior

The same accounting rule applies to modification: a phrase present in a `SKILL.md`'s frontmatter at the merge base, absent from that same file at HEAD, and neither re-claimed by a surviving skill nor recorded in `cogni-workspace/references/retired-trigger-phrases.tsv` with a non-empty `reason`, fails deterministically.

## Motivation

Same omission class as #1723, different diff shape. It needs a per-skill base-vs-head phrase-set diff over **modified** files rather than a deletion scan, so it is a second arm (or a second guard) with its own fixture cases — not a widening of the deletion one. Keeping it separate is what let #1723 ship a bounded, reviewable change.

## Acceptance criteria

- [ ] A modified `cogni-workspace/skills/<name>/SKILL.md` that drops a double-quoted trigger phrase present at the merge base, where no surviving skill re-claims it and no ledger row records it, fails deterministically.
- [ ] The added arm is shown to have teeth via a mutation-style red-then-green trace against its own case id.
- [ ] A description edit that only rewords prose around unchanged quoted phrases still exits 0, so ordinary skill edits are not taxed.

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-09-03T08:45:22Z` (GitHub clock).

Released automatically if triage parks, decomposes, or closes this issue; carried
into resolution otherwise. A claim older than the freshness TTL reads as a strand
and becomes pickable again.

[...]

PR #1764, opened 2026-09-02
Merge-base gate: a skill retirement must write a ledger row

### LINKED ISSUE #1723: cogni-workspace: nothing forces a skill retirement to write a ledger row, so the orphan arm stays green on a shrunken claim surface
Deferred from the resolution of the orphaned-trigger-phrase issue (the `story-to-storyboard` to `story-to-web` fold follow-up).

That change adds case C10 to `cogni-workspace/tests/test-skill-trigger-phrases.sh`, which proves the retirement ledger and the live claim surface agree with each other in both directions: a phrase recorded `claimed` that no live skill yields is an orphan, and a phrase recorded `retired` that a live skill does yield is a stale record.

**What it does not do.** Nothing forces a *future* retirement to write a row at all. An author who deletes a skill directory and adds no ledger entry leaves C10 green on a silently shrunken claim surface — the phrases simply vanish from both sides of the comparison at once, so the two sides still agree. That is the same class of invisibility the parent issue reports one instance of: the guard is green because it stopped having anything to look at, not because nothing was lost.

Closing it needs a different mechanism from the parent's. C10 is a tree-only check — it reads the current `skills/` tree and the current ledger. Detecting an *omission at retirement time* requires comparing against the merge base, so it needs git-history or base-vs-head diff awareness, and is likely a CI step or a separate suite rather than another arm in this one.

## Acceptance criteria

- [ ] A deleted `skills/<name>/SKILL.md` whose frontmatter trigger phrases gain no retirement-ledger row fails a deterministic check against the merge base.
- [ ] The check is shown to have teeth via a mutation-style red-then-green trace, not merely a passing run.
- [ ] A skill deletion whose phrases ARE all accounted for — re-claimed by a surviving skill, or recorded retired with a reason — passes, so the check exits 0 on a clean base.

claude-cogni-work:
Closes #1723.

## Summary

C10 in `cogni-workspace/tests/test-skill-trigger-phrases.sh` cross-checks the retirement ledger against the live tree — but both of its directions read the two *current* sides. Delete a skill and write no ledger row, and the phrases leave the record and the tree at once: the two sides still agree, and the arm stays green on a claim surface that just shrank. Seeing that requires the base side, which a tree-only suite has no git ref to reach.

- **New guard `scripts/check-retirement-ledger.py`** (stdlib-only). Anchors on `git merge-base <base-ref> HEAD` — never the base tip — lists deletions with `--diff-filter=D`, recovers each deleted `cogni-workspace/skills/<name>/SKILL.md` from the merge base via `git show`, extracts its frontmatter trigger phrases, and fails when a phrase is accounted for by neither side at HEAD: not yielded by a surviving live skill, and not carried by a `retired-trigger-phrases.tsv` row with a non-empty `reason`.
- `{success, data, error}` envelope; exit **0** clean *or* degraded, **1** violations, **2** script error. An unresolvable base ref (shallow clone, offline) reports `status: "degraded"` and exits 0 rather than false-flagging every PR — which is exactly why the CI job re-grades that same invocation and fails the build on `degraded`.
- **New CI job `retirement-ledger-gate`** in `.github/workflows/lint.yml`, copying the `version-bump-gate` shape: `fetch-depth: 0`, `ref: ${{ github.event.pull_request.head.sha }}` (not the auto-generated merge ref, which collapses the merge base onto main's tip), and a same-step **one-invocation** `status != "degraded"` vacuity assertion.
- **New self-test `tests/test_check_retirement_ledger.sh`** — 13 cases, each building its own git fixture repo, so no case depends on the real checkout having a base ref. Cases `retirement-ledger-11-unparseable-base-blob` and `retirement-ledger-12-mixed-violation-kinds` cover the unparseable-base-blob observation channel: 11 pins that the observation is recorded AND that it does not move the exit code, 12 pins that an observation and a violation co-report without either contaminating the other's count. Case `retirement-ledger-13-move-out-of-skills-is-a-retirement` pins that a retirement performed as a `git mv` out of `skills/` is still a deletion the ledger must account for.
- **New parity case C11** in the trigger-phrase suite, diffing the guard's `--emit-phrases` output against `emit_phrases` over the real tree (200 phrase lines, byte-identical), with an anti-vacuity floor so two empty outputs cannot match trivially.
- Registration bullet in root `CLAUDE.md`; header-only paragraph in the ledger.

## Why the ledger is read from the HEAD side

Deliberate, and load-bearing. The row that satisfies this guard is the one the author adds *on the branch*. Read the ledger from the base side and that row is invisible — the documented remedy would never satisfy the check, making the guard unfixable rather than merely wrong.

## Evidence

| Check | Result |
|---|---|
| `bash tests/test_check_retirement_ledger.sh` | 13/13 PASS, exit 0 |
| `bash cogni-workspace/tests/test-skill-trigger-phrases.sh` | C10 **and** C11 PASS, exit 0 |
| `python3 scripts/run-plugin-tests.py` | **all 140 suites passed** (the new suite discovered at `[136/140]`) |
| `python3 scripts/check-retirement-ledger.py` at head | `clean: true`, `status: "ok"`, merge base `5964ad22`, exit 0 |
| `check-workflow-yaml.py` / `check-case-id-pairing.py` / `check-result-line-plainness.py` / `check-breadcrumbs.py` / `check-code-span-nesting.py` | all exit 0 |
| `check-version-bump.py` | `status: ok`, `plugins_touched: []` — no version touched; the bump is post-merge |
| Mutation recipe | **`guard_verified`** — mutated `red`, restored `green` |

**C11 was separately proven able to fail**, not just observed passing: mutating the guard's key derivation (`" ".join(phrase.split()).lower()` → drop the `.lower()`) turns C11 **red**, and restoring it turns it green. Worth recording that the first falsifier tried — disabling the `>`/`|` block-scalar unwrap — left C11 green, and that is correct rather than a gap: dropping the `>-` line removes no double-quoted span, so the extracted phrase set is identical either way. The mutation was a genuine no-op, not a blind spot.

## Mutation recipe

```bash
bash /Users/stephandehaas/.claude/plugins/cache/managed-service/cogni-service/0.0.487/scripts/mutation-check.sh --root /Users/stephandehaas/GitHub/dev/insight-wave/.git/cogni-worktrees/fix/issue-1723-retirement-ledger-merge-base-gate --file scripts/check-retirement-ledger.py --expr 's/--diff-filter=D/--diff-filter=X/g' --test tests/test_check_retirement_ledger.sh --case retirement-ledger-01-missing-row
```

Swapping the deletion filter for one that matches nothing makes the guard find zero deleted skills and report clean on a tree that omits a ledger row. Case `retirement-ledger-01-missing-row` goes RED under it and GREEN on restore — run above, verdict `guard_verified`.

The `/g` is load-bearing: `--expr` is a `perl -0pi` expression, which slurps the whole file, and the module docstring names the same `--diff-filter=D` literal. A non-global `s///` would rewrite only the docstring occurrence, leave the git argument list intact, keep the case green, and report a false `vacuous_guard`. `--root` names the worktree the branch was built and verified in; re-point it at any checkout of this branch to replay.

Both paths in the recipe are **substitution points, not literals to copy**: the harness path is whichever cogni-service plugin-cache version the replaying reviewer has installed, and `--root` is any checkout of this branch.

## Scope decisions

**In.** Exactly the three acceptance criteria — a deleted `SKILL.md` with unaccounted phrases fails against the merge base; the check is shown to have teeth by a replayable red-then-green trace; an accounted-for deletion and a clean base both exit 0.

**Deliberately not taken.** A checked-in baseline-snapshot file compared against the current tree. It needs no git history, but it reintroduces the stale-record class the ledger already carries — the snapshot becomes a second thing to keep in step, and a retirement that forgets the ledger would equally forget the snapshot. The issue's own criterion says "against the merge base".

**Fence.** Only `scripts/**`, `tests/**`, `.github/workflows/**`, root `CLAUDE.md`, `cogni-workspace/tests/**`, and the ledger **header**. No plugin `skills/**` or `agents/**` body, no `.claude-plugin/**` manifest, and **no ledger data row** — the 11 existing rows are byte-identical to base, verified by diff, so C10's `claimed > 0 && retired > 0` anti-vacuity floor reads exactly what it read before.

## Follow-up issues

- **#1762** — the gate does not cover a phrase dropped from a **modified** surviving skill description. Same omission class, different diff shape (a per-skill base-vs-head phrase-set diff over modified files, not a deletion scan).
- **#1763** — decide whether the gate covers plugins beyond cogni-workspace, and extend it if so. A per-plugin-ledger vs shared-ledger design call before it is a code change.

Both were duplicate-checked against the open board and 30 days of closed issues; both returned `unique`.

## Open questions — resolved

All three were ruled by the maintainer on this PR before merge; the rulings, with their evidence, are in the ruling comment on this PR.

**1. Should `emit_phrases` delegate to the guard instead of C11 pinning two copies?** — **Declined.** Delegation would make every case C1–C11 in a cogni-workspace plugin suite depend on a repo-root script (C1–C8 transitively through `scan_skills`), inverting the ownership direction; and decisively, under delegation C11 would diff the guard against itself and be trivially green — the pin that justifies the removal is destroyed by the removal. The two extractors stay, pinned by C11. Settled, so no follow-up issue.

**2. The issue-time review contract's item 17 names a script that does not exist in this repo.** — **Graded on the property; satisfied.** Item 17 is tagged `[fidelity: purpose — the teeth claim must be replayable]`, and it is. Verified against `origin/main`: `git ls-tree -r --name-only origin/main | grep -i mutation` returns exactly `cogni-consult/scripts/mutation-check.sh` and `cogni-portfolio/scripts/mutation-check.sh`, both bespoke and **no-argument**; `origin/main:scripts/mutation-check.sh` does not exist. Mechanism mis-authored, property met. The template defect is filed against the repo that owns the contract generator.

**3. Token-scope observation.** The runner token is over-scoped against the documented least-privilege posture (`extra: gist, read:org, workflow`; `forbidden: workflow`), and the push proceeded under the settled `--allow-overscoped` opt-out. `workflow` — the scope flagged as forbidden — is precisely what makes this PR's `lint.yml` change pushable at all, so a least-privilege token per the documented posture could not have opened this PR. **Recorded and filed upstream** as a cogni-service posture contradiction, not an insight-wave change.

## Contract item 11 — deviation stated

The guard routes a merge-base `SKILL.md` blob it can **read but not parse** to a non-failing `data.observations[]` entry at exit 0, where item 11 read literally asks for a violation or exit 2. This is a deliberate, maintainer-ruled deviation graded on the item's own `[fidelity: purpose — silence must never come from the guard failing to look]` tag: the fact is surfaced in the envelope and on stderr, so it is not a silent skip, and the exit-0 summary partitions its counts rather than claiming full accounting. A blob that cannot be **read** still exits 2. The rationale — that nothing a branch commits changes its own merge base, so a blocking demand here is unsatisfiable by construction — is argued durably in the module docstring at `scripts/check-retirement-ledger.py:79-102`, and is restated here so the deviation is visible in the merged record rather than only in the source.

## Simplify pass

Four parallel quality agents ran pre-commit. Three independently converged on one root cause — the extractor duplication. Applied: a single shared `iter_skills()` traversal, so C11's pin now covers the live-side accounting walk transitively rather than only the parity copy (this was the substantive one — the unpinned walk built the set the whole comparison rests on); `clean`/`count` derived at render time instead of hand-maintained at three exits; one consistent element schema for `deleted_skills`; `SKILL_PATH_RE` built from `SKILLS_REL` so the path is spelled once; and a C11 failure message that no longer reports the tree total as the delta. Skipped with reasons: the `check_eq` helper (a choice between two existing house styles, and the chosen one matches the direct template `tests/test_check_version_bump.sh`), the CI-job `strategy.matrix` collapse (would restructure an existing job outside this diff's fence), and open question 1 above. The efficiency pass found nothing: the live-tree walk runs once, only after the no-deletion early exit, and measures 2 ms.


claude-cogni-work:
## Ship-ready

Author instance has handed this PR off for independent merger review. Draft, `svc:needs-review` applied. Issue #1723 moved `svc:in-progress` → `svc:resolved`.

### Handoff preflight — exit 0, clean

3 passed / 4 skipped / 0 violations.

- **Graded:** `run-tests` (suites green for cogni-workspace) · `version-bump` (no version-line touch) · `mutation-recipe` (recorded recipe runnable as written — `retirement-ledger-01-missing-row` resolved in `tests/test_check_retirement_ledger.sh`)
- **Skipped:** `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` — all `not_applicable`, the diff touches no `SKILL.md` and no `agents/*.md`.

### Author self-grade against the issue-time contract — 22/22 satisfied, no residual

Graded against the contract committed on the issue at `2026-09-02T10:47:47Z`, before any branch existed (`--target issue --origin triage`). Advisory only — it carries no verdict label and does not substitute for the merger's own review.

The grader executed premises rather than reading for them: it re-derived extractor parity independently (200 lines, byte-identical), ran the CI job's `run:` block verbatim against both the real head **and** a forced-`degraded` status to prove the vacuity assertion fires, deleted `refs/remotes/origin/main` and re-ran both suites to confirm shallow-job tolerance, and replayed the recorded mutation recipe to reproduce `guard_verified`.

**Residual: none.** No item graded `violated` or `unclear`, so no fix pass ran.

### One contract item the merger should grade on property, not mechanism

**Item 17** requires the mutation recipe to target a repo-root `scripts/mutation-check.sh` in this repo with a root-relative `--file`. That script does not exist here — `git ls-tree -r --name-only origin/main | grep -i mutation` returns only `cogni-consult/scripts/mutation-check.sh` and `cogni-portfolio/scripts/mutation-check.sh`, both bespoke and no-argument, and `cogni-workspace/scripts/` has none. The flagged five-flag tool is cogni-service's own instrument.

The item's **property** — a replayable red-then-green teeth trace — is satisfied and was independently reproduced. Its **path prefix** is mis-authored, and it traces to a false premise in this run's own triage comment rather than to anything in issue #1723. Flagging it explicitly so it is adjudicated rather than silently failed.

### Open questions carried into review

Three, all in the PR body under `## Open questions`. The first is a genuine structural call the author declined to make unilaterally: whether the suite's `emit_phrases` should **delegate** to the guard's `--emit-phrases` — deleting ~35 lines of duplicated Python and case C11 outright — rather than keeping two copies pinned by a parity case. It inverts a dependency (a cogni-workspace suite would depend on a repo-root script), which is a question about which tree owns the extractor.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1764

**Summary:** 21 of 22 committed contract items pass with executed evidence, but item 11 is independently reproduced as violated by both substantive perspectives — a base-side SKILL.md whose frontmatter fails to parse is silently skipped rather than surfaced, so the guard reports a false 'every trigger phrase accounted for' on a file it never parsed; the 763-line diff-size note is a non-gating advisory and does not change this outcome.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[major]** (auto-fixable) `review-contract:item-11` — Item 11 requires that a base-side SKILL.md blob that cannot be read OR whose frontmatter does not parse be surfaced as a violation or a script error (exit 2), never a silent skip that still yields exit 0. Only the first half exists. phrases_from_skill_text() returns ([], name) when the ^---\n(.*?)\n---\n match fails, and again when the ^description: search fails - both silently. Reproduced on a fixture by BOTH independent perspectives: base commit holds cogni-workspace/skills/alpha/SKILL.md with NO frontmatter, branch deletes it, ledger carries no row. Result: rc=0, {"status":"ok","violations":[],"deleted_skills":[{"skill":"alpha","phrases":0}],"clean":true}, stderr 'every trigger phrase accounted for.' The guard asserts full accounting for a file it never parsed - silence produced by failing to look, the exact invisibility class this guard exists to close.
  - *Suggested fix:* Distinguish 'no frontmatter block' / 'no description key' from 'parsed, zero quoted spans' (e.g. return None for the former) and treat None in the deleted-skill loop as a violation (check: 'base-blob-unparseable', naming skill+path) or a script error (exit 2), matching the treatment the unreadable-blob arm already gets a few lines above. Add tests/test_check_retirement_ledger.sh case retirement-ledger-11-unparseable-base-blob asserting the non-zero exit and the new check code, so the branch has a red arm rather than only a code path.

- **[minor]** (advisory-addition) `PR diff` — Diff is 763 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

### Positive observations

- Merge-base anchoring is regression-tested, not merely asserted: case retirement-ledger-06 builds a fixture where main advances past the fork and deletes a skill of its own while the branch deletes nothing - the exact false-flag a base-TIP anchor would produce. Ran green. (plugin-dev:plugin-validator)
- Extractor duplication is closed by a real pin: C11 diffs the guard's --emit-phrases against the suite's emit_phrases over the real tree with an anti-vacuity floor, and case 10 proves --emit-phrases spawns no git call so C11 cannot redden the shallow plugin-test-suites job. (plugin-dev:plugin-validator)
- The shared iter_skills() traversal makes C11's pin cover the live-accounting walk transitively, closing the fail-open direction where an under-collecting live set would demand rows for phrases a surviving skill still claims. (plugin-dev:plugin-validator)
- The PR body's Open question 2 correctly falsifies a premise in its own review contract (repo-root scripts/mutation-check.sh does not exist here) and flags it for the merger rather than silently substituting. Verified independently. (plugin-dev:plugin-validator)
- Item 6 confirmed by independent execution rather than accepted: extracted the suite's emit_phrases standalone and diffed against --emit-phrases over the real tree - 200 lines each, byte-identical. (cogni-service:contract-grading-reviewer)
- Item 13 confirmed by replaying the CI job's exact run: block - a real run exits the step 0 and a forced-degraded run exits 1, from one invocation. (cogni-service:contract-grading-reviewer)
- Item 22 confirmed: the ledger diff is header-comment-only, so no data row moved and C10's claimed>0/retired>0 floor reads exactly what it read before. (cogni-service:contract-grading-reviewer)

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1764
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator

**Summary:** plugin-validator found one major defect (the rewritten base-blob-unparseable FIX_HINTS remedy still loops the operator back to the identical failure) plus four minors, one of which is deliberately deferred with a written waiver; plugin-validator was the sole substantive perspective routed on this PR (no SKILL.md / agents/*.md / resume / loop-bearing paths touched), so this gate's coverage is narrower than the usual multi-perspective fan-out.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[major]** (auto-fixable) `scripts/check-retirement-ledger.py:408-416` — The rewritten hint's third clause is false as written, proven by execution. Its first two claims hold: a ledger row does NOT clear the arm (fixture: deleted no-frontmatter skill + two well-formed rows -> rc=1, base-blob-unparseable), and restoring the file at HEAD does (rc=0). But 'retire it in a follow-up commit that gives it a frontmatter description: block first, so its phrases become readable and the ordinary ledger remedy applies' fails on its natural reading: a follow-up branch that commits the description block and then the deletion still resolves git show <merge-base>:<path> to the OLD unparseable blob, because nothing a branch commits can change its own merge base. Fixture C returns rc=1 base-blob-unparseable and re-prints THIS SAME HINT — the operator is looped back into the identical failure with no stated exit. The remedy only works when the description block is already on the BASE side before the deletion branch forks (fixture D: rc=0 clean). This is the 'unfixable rather than merely wrong' failure the module docstring's THE LEDGER IS READ FROM THE HEAD SIDE paragraph makes load-bearing, for the third consecutive cycle.
  - *Suggested fix:* Say where the description block has to land, not just when — or state the constraint as a fact and stop prescribing a multi-step procedure. Then re-run the fixture-C shape and confirm the printed remedy is one the operator can actually execute.

- **[minor]** (auto-fixable) `PR body, Summary bullet 4 and the Evidence table` — The PR body still claims the self-test has '10 cases' and evidences '10/10 PASS'. The suite now carries 12 cases and reports 12 PASS lines; cases 11 and 12 cover the unparseable-base-blob arm this cycle is about, so the merge record understates the delta's own coverage.
  - *Suggested fix:* Update both to 12 cases / 12 PASS and name cases 11 and 12 by id.

- **[minor]** `PR diff` — Diff is 912 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (waived) `scripts/check-retirement-ledger.py:195-199,215-216` — Re-raised for the record and verified rather than taken on report: mutating away emit_phrases' UNPARSEABLE continue, and separately live_phrase_keys' filter, each leaves BOTH suites green. No fixture has an unparseable skill LIVE at HEAD, so neither sentinel arm is exercised.
  - *Justification:* Deferred deliberately and recorded as a follow-up. The deferral is correct: the sentinel is truthy by design, so an uncovered drift raises TypeError and surfaces as success:false / exit 2, which CI treats as a build failure — the gap fails LOUD, not open, the opposite direction from the invisibility class this guard closes.

- **[minor]** (auto-fixable) `scripts/check-retirement-ledger.py:333-335` — PRE-EXISTING (unchanged from base 62483bd8): the ledger-missing detail says the ledger is 'missing or unreadable at HEAD', but an unreadable-yet-present ledger never produces this violation — is_file() is True for a chmod-000 file, so ledger_phrase_keys' open() raises and the top-level handler returns exit 2. Fails safe, no false green, but the string names a case it does not cover.
  - *Suggested fix:* Narrow the detail to 'missing at HEAD', or catch OSError around ledger_phrase_keys.

- **[minor]** (auto-fixable) `scripts/check-retirement-ledger.py:383-385` — The new comment justifying v.get("skill") or v["path"] names a case this loop can never see. The ledger-missing arm renders, writes its own FIX_HINTS and returns 1 before the summary block — verified by execution. Every violation that CAN reach the loop always carries a non-empty skill, so the fallback is unreachable by construction. The prior cycle's finding 3 rested on a false premise, and the fix documented a case the pass does not handle.
  - *Suggested fix:* State the actual invariant: every violation that reaches here carries both skill and detail; the .get is defensive should a future arm fall through instead of returning early.

### Positive observations

- All three acceptance criteria verify by execution, not by reading: AC1 case retirement-ledger-01-missing-row (exit 1, retirement-row-missing, 2 violations); AC2 the recorded mutation recipe replayed at this head returns verdict guard_verified (mutated red, restored green) with literal flag values throughout; AC3 cases 02 and 03 both exit 0 clean. (plugin-dev:plugin-validator)
- Cases 11 and 12 are genuinely RED when only scripts/check-retirement-ledger.py is reverted to 62483bd8 — case 11 collapses to 0|ok|EMPTY|0 on BOTH arms and case 12 to 1|ok|retirement-row-missing|2, confirming the sentinel is what the cases bind to. (plugin-dev:plugin-validator)
- Both suites green at head: 12/12 in tests/test_check_retirement_ledger.sh, 11/11 in the trigger-phrase suite including the C11 parity pin with its anti-vacuity floor. (plugin-dev:plugin-validator)
- The core of the prior cycle's fix is TRUE and confirmed independently: two well-formed ledger rows leave a deleted no-frontmatter skill at rc=1 base-blob-unparseable, and restoring the file at HEAD returns rc=0. The inert-remedy claim the prior cycle raised is genuinely retired. (plugin-dev:plugin-validator)
- The key-agnostic loop change removes a latent crash: at 62483bd8 a ledger-missing violation reaching the summary loop would have raised KeyError after render() had already emitted a success envelope, printing a second contradictory envelope. (plugin-dev:plugin-validator)
- The guard runs for real against this branch — status ok, clean true, merge_base 5964ad22, never degraded — and the CI job re-grades that SAME invocation's JSON to fail the build on degraded. (plugin-dev:plugin-validator)
- Plugin-structural surface clean: no skills/**, agents/** or .claude-plugin/** touched; check-frontmatter.sh on cogni-workspace returns 0 violations; the ledger diff is header-comment-only so C10's anti-vacuity floor reads what it read before; no plugin.json version touched. (plugin-dev:plugin-validator)
- Handoff preflight at this head is a real green, not a vacuous one: 7/7 instruments resolved, 3 gating checks actually ran and passed (run-tests green, version-bump no touch, mutation-recipe runnable as written), 4 skipped as genuinely not-applicable. (plugin-dev:plugin-validator)

### Recommendation

---

claude-cogni-work:
## Revision posted

Auto-revise cycle off the `needs_revision` verdict at
https://github.com/cogni-work/insight-wave/pull/1764#issuecomment-5509055749.
No human comments, no reviews, no unresolved review threads — the verdict was the
sole driver.

Two commits landed this cycle: `9511668e` (the item-11 fix) and `ffb4c02d` (a
correction to that fix's own operator text, raised by the re-review).

### Findings addressed

**[major] `review-contract:item-11` — resolved in `9511668e`.**
`phrases_from_skill_text()` returned `([], name)` from **both** silent arms — no
frontmatter block, and no `description:` key — so a deleted base-side `SKILL.md`
that could not be parsed produced `rc=0`, `violations: []`, `clean: true` and the
stderr claim *"every trigger phrase accounted for"*.

It now returns an `UNPARSEABLE` sentinel from both arms, and each of the three
consumers states its own decision:

- `emit_phrases()` skips it after counting the file as found — byte-for-byte the
  suite copy's behaviour, so C11's parity pin stays valid. Confirmed:
  `--emit-phrases` over the real tree is **byte-identical** to base (200 lines).
- `live_phrase_keys()` skips it. That is the safe direction — this set is what
  *excuses* a retired phrase, so omitting a file can only make the guard demand
  more, never less.
- the deleted-skill loop raises `base-blob-unparseable` and exits 1 — the
  analogue of the exit-2 the unreadable-blob arm already got.

The sentinel is a **truthy** `object()` rather than `None` on purpose: the one
cheap way back into this defect class is a future consumer writing `if keys:`
before iterating. Against a falsy sentinel that reads as "nothing to check" and
silently skips; against this one it falls through to iteration and raises
`TypeError` at once. Verified directly, not just asserted.

**[minor] PR diff size — not actioned, and not actionable here.** Splitting an
already-open PR means closing it and re-opening two, which discards the committed
review contract, the verdict history and the item-11 grading. Reviewing the branch
as it stands is the better position.

### Test evidence

`tests/test_check_retirement_ledger.sh` gains **two** cases (10 → 12):

- `retirement-ledger-11-unparseable-base-blob` — both silent arms under one id
  (no frontmatter block; a block with no `description:` key), so neither can
  report green alone.
- `retirement-ledger-12-mixed-violation-kinds` — one deletion of each kind in one
  branch, pinning that the two violation classes co-report.

**Both are proven RED, not merely observed passing.** Reverting *only*
`scripts/check-retirement-ledger.py` to `62483bd8` in an otherwise identical tree:

| Case | at `62483bd8` | at head |
|---|---|---|
| 11 arm *no frontmatter* | `0\|ok\|EMPTY\|0` | `1\|ok\|base-blob-unparseable\|1` |
| 11 arm *no description* | `0\|ok\|EMPTY\|0` | `1\|ok\|base-blob-unparseable\|1` |
| 12 mixed kinds | `1\|ok\|retirement-row-missing\|2` | `1\|ok\|base-blob-unparseable,retirement-row-missing\|3` |

Case 12 is not entailed by case 11 — it fails distinctly under the reverted guard.

Green at head: ledger suite **12/12**, trigger-phrase suite **11/11** (C11
included), guard `clean: true` / `status: ok`, all five repo lint guards exit 0,
handoff preflight `clean` with 3 gating checks actually run (7/7 instruments
resolved).

### What the re-review then caught in my own fix

`ffb4c02d` corrects the `base-blob-unparseable` `FIX_HINTS` I wrote in
`9511668e`. The original text prescribed recording ledger rows — which **cannot**
clear that arm, because the violation is appended before `accounted` is ever
consulted. Verified on a fixture: a deleted no-frontmatter skill with two
well-formed `retired` rows still returns `rc=1 base-blob-unparseable`; restoring
the file at HEAD returns `rc=0 clean`.

**This is still not fully closed** — see the fresh verdict.

### Follow-up issues

- **#1765** — two of the three `UNPARSEABLE` consumer arms (`emit_phrases`,
  `live_phrase_keys`) are unpinned: deleting either leaves both suites green,
  because no fixture leaves an unparseable skill *live at HEAD*. Waived here with
  justification (the gap fails **loud** — a truthy sentinel raises `TypeError` →
  exit 2 → CI build failure — rather than open), and filed rather than left as
  prose.

### Still open

The fresh verdict is **`needs_revision`**, on a major in the corrected hint
itself: its remaining clause ("retire it in a follow-up commit that gives it a
frontmatter `description:` block first") is *also* false on its natural reading —
a branch cannot change its own merge base, so that follow-up re-prints the
identical failure. The remedy only works once the description block has **merged
to the base** before the deletion branch forks.

That is the third consecutive cycle on this one passage, which is itself the
signal: the fix is to state the constraint as a fact and stop prescribing a
multi-step procedure. **Not handed off to the merger** — the author does not pass
a known-negative PR across the identity boundary. This PR stays re-pickable by
the next author drain.

sdh07:
## Maintainer ruling — `base-blob-unparseable` becomes a non-failing observation

This PR has returned `NEEDS_REVISION` on three consecutive cycles, each one re-wording the
`FIX_HINTS` remedy for the `base-blob-unparseable` arm. Ruling: **stop re-wording the remedy and
remove the obligation.** The arm is downgraded from a blocking violation to a non-failing
observation.

### Why

Two measurements decide it:

| Evidence | Result |
|---|---|
| `cogni-workspace` skill deletions across this repo's whole history | 8 — every base-side blob parseable |
| Live `cogni-workspace` skills with no parseable `description:` | 0 of 22 |

The class is not merely rare, it is empty by construction. A `SKILL.md` with no frontmatter block,
or a block with no `description:` key, is not a loadable skill — the plugin loader requires both. It
therefore never advertised a trigger phrase, so deleting it loses no claim surface. The arm's
premise, "unknown is not zero, so demand accounting", does not hold here: unknown *is* knowable, and
it is zero.

That is also why no executable remedy could be written. Three cycles have now demonstrated that
every candidate loops the operator back into the same failure, because nothing a branch commits can
change its own merge base. A gate that cannot be satisfied is worse than no gate — it converts a
correct PR into an indefinite bounce.

### What the original finding did catch, and what survives

The first cycle's complaint was real but narrower than the fix applied to it: the gate printed
`every trigger phrase accounted for` about a file it never parsed. That is an honesty defect in the
**message**, not a missing obligation on the author. The `UNPARSEABLE` sentinel already on this
branch is the correct distinction and stays — it is what makes the fact expressible at all. Only its
consequence was miscalibrated.

### The change

1. Keep the `UNPARSEABLE` sentinel and the `phrases: null` envelope field. Nothing already written
   is discarded.
2. Move the finding from `violations[]` to an `observations[]` entry. Exit stays **0** when nothing
   else is wrong.
3. Rewrite the exit-0 stderr line so it cannot overclaim — report deleted-skill count, how many were
   fully accounted for, and how many were unparseable at the merge base with no derivable phrase set.
4. **Delete the `base-blob-unparseable` `FIX_HINTS` block.** Nothing to fix means no remedy to name.
   This is what dissolves the loop rather than re-wording it a fourth time.
5. Keep cases `retirement-ledger-11-unparseable-base-blob` and
   `retirement-ledger-12-mixed-violation-kinds`, re-pointed to assert a clean exit **plus** a
   populated `observations[]`. They still discriminate: against a gate that silently skips, the
   observation is absent, so the arms stay red against the pre-fix behaviour.

### Precedent

This is the repo's established pattern rather than a new one. `CLAUDE.md` records it for
`check-mcp-tool-grant.py`: registry-accuracy gaps take a non-failing channel and never move the exit
code, surfaced as an observation.

### The alternative, and why it was not taken

Keeping the arm blocking would require the hint to name a two-**PR** procedure, not a two-commit one:
the merge base moves only when `main` moves, so a `description:` block must land on `main` first and
the deletion must happen on a branch forked after that merge. That is executable, but it buys a
blocking gate whose expected firing rate on real work is approximately zero, at the cost of a full
extra merge cycle whenever it does fire. Not worth it.

### Scope note

The advisory diff-size finding stays waived. The PR-body count drift (the body still says 10 cases /
10 PASS; the suite now carries 12) should be corrected in the same cycle.

---

*Ruling made by the maintainer on explicit in-session instruction. The measurements above were
produced by this session against the branch head and are machine-verifiable; the decision to
downgrade the arm is the maintainer's.*

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1764
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator

**Summary:** Both perspectives passed with only minor, non-gating findings (one waived with justification, one an un-waived advisory minor on diff size); coverage note: no review_contract was passed on this dispatch and the PR touches no SKILL.md, agents/*.md, *-resume/SKILL.md or loop-bearing path, so plugin-dev:plugin-validator was the sole substantive perspective routed, with the two deterministic pre-gates (check-frontmatter.sh: 0 offenders; check-version-bump.sh: status ok, 0 violations) running clean alongside it.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (2)

- **[minor]** (waived) `.github/workflows/lint.yml:63 (version-bump-gate step)` — Follow-up only, explicitly out of scope for this PR. The sibling version-bump gate carries the identical inherited-`set -e` flaw this PR just fixed in the retirement-ledger step: its run: block also relies on `set -uo pipefail` without a preceding `set +e`, so under GitHub's `bash -e {0}` it aborts at its capture line and suppresses its own diagnostics on the non-zero path.
  - *Justification:* Scoped out as a separate already-merged workflow's defect; permitted only as a follow-up note. Must not gate this PR.

- **[minor]** `PR diff` — Diff is 979 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

### Positive observations

- Fix 1 is demonstrably load-bearing: removing --no-renames reproduces the exact fail-open the case documents, so case 13 is a real discriminator rather than a restatement of case 1. (plugin-dev:plugin-validator)
- Fix 1 introduces no false positive on the adjacent shape - a skill directory renamed within skills/ still exits 0. (plugin-dev:plugin-validator)
- Fix 2 verified on all four exit paths under bash -e, including the negative control: without set +e the violation and error paths print nothing at all, proving the remedial stderr surface was genuinely unreachable before. (plugin-dev:plugin-validator)
- The vacuity re-grade reads the SAME invocation's JSON rather than re-running the script, so a degraded run fails the build instead of passing vacuously. (plugin-dev:plugin-validator)
- Extractor parity is pinned transitively: iter_skills is a single traversal shared by both emit_phrases and live_phrase_keys, so C11's pin covers the live side too. (plugin-dev:plugin-validator)
- The UNPARSEABLE sentinel is deliberately truthy, so a future `if keys:` consumer raises TypeError immediately rather than silently reading a failed parse as 'nothing to check'. (plugin-dev:plugin-validator)
- The maintainer's base-blob-unparseable ruling is implemented and pinned in BOTH failure directions - case 11 reddens when the sentinel is reverted (under-reporting) and again when observations.append is flipped to violations.append (over-blocking). (plugin-dev:plugin-validator)
- Rationale is stated at each call site rather than only in the commit message, so both fixes survive a future reader who never sees this PR. (plugin-dev:plugin-validator)

### Recommendation

---

claude-cogni-work:
## Revision posted

Two commits on this cycle: `d09c11b0` implements the maintainer ruling, `8190093c` closes two fail-opens found by review of that ruled head. Fresh verdict: **PASS**.

### Findings resolved

| Finding | Resolution |
|---|---|
| `review-contract:item-11` — base-blob-unparseable silently skipped (major, raised on 3 consecutive cycles) | **Settled by maintainer ruling**, not by re-wording. The arm is now a non-failing observation in `data.observations[]`; an unreadable blob still exits 2. The `FIX_HINTS` arm is deleted — no remedy existed, since nothing a branch commits changes its own merge base. Exit-0 summary now partitions its counts instead of claiming full accounting. `d09c11b0` |
| `.github/workflows/lint.yml` — step suppressed its own diagnostics (major) | `set +e` first in the `run:` block. GitHub runs it as `bash -e {0}` and `set -uo pipefail` does not clear an inherited `-e`, so the step aborted at the capture line and envelope + `FIX_HINTS` never reached the log. Failing exit preserved by the block's explicit `exit "$rc"`. `8190093c` |
| `scripts/check-retirement-ledger.py` — `git mv` retirement invisible (major) | `--no-renames` added. Rename detection is on by default since git 2.9, so archiving a SKILL.md out of `skills/` was paired as R, the D filter saw nothing, and the guard exited 0 announcing "no skill deletion" while the phrases had left both the live set and the record. `8190093c` |
| `PR diff` — 979 lines vs the 400 advisory bound (minor) | **Waived.** Advisory and non-gating. Splitting is unavailable in a revise cycle — the PR is open with verdicts against it — and the unit is not separable: the guard, its CI job, its 13-case self-test and the ledger it reads are not independently shippable. |
| `.github/workflows/lint.yml:63` — sibling version-bump gate has the same inherited flaw (minor) | **Waived here, filed as #1766.** A separate already-merged workflow's defect; widening this PR to cover it was explicitly declined. |

### Verification

Every claim below was reproduced on a fixture, not taken on report.

| Check | Result |
|---|---|
| `bash tests/test_check_retirement_ledger.sh` | 13/13 PASS, exit 0 |
| `bash cogni-workspace/tests/test-skill-trigger-phrases.sh` | C1–C11 PASS (C11 extractor-parity pin intact) |
| `python3 scripts/run-plugin-tests.py` | 140 suites, 0 failures |
| Repo guards | breadcrumbs, result-line-plainness, case-id-pairing, version-bump, code-span-nesting, workflow-yaml — all exit 0 |
| Case 11 mutation, **both directions** | reverting the `UNPARSEABLE` sentinel reddens it (under-reporting); flipping `observations.append` → `violations.append` reddens it (over-blocking) |
| Case 13 mutation | removing `--no-renames` reddens it, reproducing `0|ok|EMPTY|0|EMPTY` — the guard green while the phrases are gone |
| CI step under `bash -e` | without `set +e`, violation and script-error paths emit empty stdout **and** stderr; with it, envelope + `FIX_HINTS` print and the step still exits 1 |

### Coverage caveat

No `review_contract` was passed on this dispatch, and the PR touches no `SKILL.md`, `agents/*.md`, `*-resume/SKILL.md` or loop-bearing path — so `plugin-dev:plugin-validator` was the **sole** substantive perspective routed. Both deterministic pre-gates ran clean (frontmatter: 0 offenders; version-bump: `status: ok`). This gate's coverage is narrower than a full fan-out; the merger's independent contract review is the backstop.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1764

**Summary:** 21 of the 22 committed review-contract items are satisfied by every perspective that graded them. Item 11 is a genuine split judgment rather than shared uncertainty - plugin-dev:plugin-validator flagged it INTENT-UNCLEAR (major, un-demoted since it sits on the contract) while cogni-service:contract-grading-reviewer independently graded the same locus satisfied - so the merged verdict is needs_revision on that one un-marked major finding, routed for a human call rather than an author bounce. Two merger-added preference findings (mutation-recipe path portability, single-plugin ledger scope) are advisory additions and do not gate. The oversized-diff observation (1022 lines vs the 400-line bound) is an ordinary un-waived minor. Deterministic evidence at head: 140/140 discovered suites pass, the new 13-case suite passes, C10 and C11 pass, frontmatter and version-touch pre-gates clean, and the recorded mutation trace replays as guard_verified (red under mutation, green on restore).

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (4)

- **[major]** `review-contract:item-11` — INTENT-UNCLEAR: Item 11 requires that a base-side SKILL.md blob which cannot be read OR whose frontmatter does not parse be surfaced as a violation or a script error (exit 2), never a silent skip at exit 0. The diff splits the two halves. Cannot-be-READ matches the item exactly (scripts/check-retirement-ledger.py:384-391 renders an error envelope and returns 2). Cannot-be-PARSED deliberately does not: the UNPARSEABLE sentinel routes to a non-gating observations[] entry at :399-419 with a stderr note and exit 0, pinned by cases retirement-ledger-11-unparseable-base-blob and -12-mixed-violation-kinds which assert exit 0 on that arm. This is not the fail-open the item guards against - the finding IS surfaced in the envelope and on stderr, so it is not a silent skip - but it is also not a violation and not exit 2, so the item is neither cleanly satisfied nor cleanly violated. The divergence is argued at length in the module docstring :79-102 (the merge-base blob is history, so no commit on the branch can make it parse; a demand there would be unsatisfiable) and was introduced by an explicit prior review round (commit d09c11b0). Mitigating and independently verified: an UNPARSEABLE result requires either no frontmatter block or one with no description: key, and case C9 in cogni-workspace/tests/test-skill-trigger-phrases.sh already fails any real cogni-workspace skill yielding zero trigger phrases - so on any merge base whose CI was green, no genuine skill file can reach this arm. SPLIT JUDGMENT: cogni-service:contract-grading-reviewer independently examined this same locus and graded it satisfied on purpose-fidelity grounds, so the divergence is between two confident perspectives, not reviewer uncertainty. The finding is not demoted; it is routed to human judgment. A human should decide whether item 11 is graded on the property (no silent skip - satisfied) or the mechanism (violation/exit 2 - unsatisfied).
  - *Suggested fix:* No code change recommended by either perspective. If item 11 is read literally, the minimal reconciliation is to gate the observation channel under an opt-in flag, or to state the deviation and its rationale in the PR body's scope section - it is currently argued only in the script docstring and the suite header, so the divergence is invisible at review time.

- **[minor]** (advisory-addition) `PR body -> Mutation recipe section` — The recorded recipe satisfies item 17's flag requirements, but both path arguments are machine-local: the harness is invoked from a cogni-service 0.0.487 plugin-cache path and --root points at the author's own worktree. The next reviewer must substitute both before the trace replays. The teeth claim itself is true - it was reproduced here both by re-pointing --root at the PR-head worktree (verdict guard_verified) and by applying the expression manually.
  - *Suggested fix:* Note in the recipe block that the harness path is the reviewer's own cogni-service plugin-cache path and --root is any checkout of this branch, so both are substitution points rather than literals to copy.

- **[minor]** `PR diff` — Diff is 1022 changed lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

- **[minor]** (advisory-addition) `scripts/check-retirement-ledger.py:125-131` — SKILLS_REL is the single hardcoded literal 'cogni-workspace/skills', so SKILL_PATH_RE matches only that one plugin; a retirement deleting a skill in another plugin produces zero matches and the guard exits 0. This is the documented and correct scope for this PR - cogni-workspace is the only plugin carrying a retirement ledger - and #1763 records the design call on whether to extend it. Raised only so the boundary is on record: the CI job name and the CLAUDE.md registration paragraph read more broadly than the guard's actual subject.
  - *Suggested fix:* Optionally narrow the CI job name or the CLAUDE.md sentence to name cogni-workspace explicitly, so the advertised subject and the implemented subject read the same. No code change needed.

### Positive observations

- Both perspectives independently replayed the teeth. The PR's own recipe expression (--diff-filter=D -> X) turns case retirement-ledger-01-missing-row RED and green on restore (verdict guard_verified), and mutating the guard's key derivation (dropping .lower()) turns C11 RED.
- Deterministic evidence at head: 140/140 discovered suites pass; the new tests/test_check_retirement_ledger.sh passes 13/13 with 13 unique first-token case ids; C10 and the new C11 both PASS; check-result-line-plainness, check-case-id-pairing, check-workflow-yaml, check-breadcrumbs and check-code-span-nesting all exit 0; no version touched.
- --no-renames (:346-348) closes a second independently-reachable route into the same fail-open that nothing in the checklist asked for: a retirement carried out as a git mv out of skills/ reports as D rather than R. Case retirement-ledger-13 pins it.
- iter_skills() (:192-213) is a single traversal shared by the C11-pinned parity mode and by live_phrase_keys, so C11's pin covers the live-side accounting walk transitively rather than only the parity copy.
- The UNPARSEABLE sentinel is deliberately truthy rather than None (:134-139): a falsy sentinel would let a future 'if keys:' read as nothing-to-check and silently reopen the exact class the guard exists to close.
- The CI step's set +e ordering is documented as load-bearing - GitHub runs run: as bash -e {0}, so without it the step would abort on the capture line and the FIX_HINTS remedial surface would never reach the log, while the failing exit is preserved by the explicit exit $rc.
- The suite unsets RETIREMENT_LEDGER_BASE_REF at the top so an ambient CI value cannot silently grade every case against the wrong baseline while still reporting green; run_gate also asserts the observation field on EVERY case, not only the two expecting one.
- Case retirement-ledger-06-merge-base-anchoring is a genuine regression test for the base-tip-versus-merge-base decision: main advances past the fork and deletes a skill of its own while the branch deletes nothing, and the guard reports clean. A base-tip anchor would blame the branch.
- item 13: fetch-depth: 0 at lint.yml:22, ref head.sha at :28, and a same-step one-invocation status != degraded vacuity assertion at :42 and :59-69, mirroring the version-bump-gate shape.
- item 22: the changed-file set matches the contract's boundary predicate exactly; the ledger diff is header-comment-only, with no data row added, removed or altered, so C10's anti-vacuity floor reads exactly what it read before.

### Recommendation

---

sdh07:
## Maintainer ruling — item 11 graded on the property; the three open questions closed

This PR is not held on a defect. It is held on a **stale contract record**. Ruling on every
outstanding item below; the change is mergeable as it stands.

### Item 11 — graded on the property. Satisfied. No code change.

The item is tagged `[fidelity: purpose — silence must never come from the guard failing to look]`.
The contract's own grading grammar therefore settles which reading applies: purpose, not mechanism.
Items 3 and 6 are the ones tagged `[fidelity: literal]`; item 11 is not one of them.

The purpose is met. The fact is surfaced in `data.observations[]` and on stderr
(`scripts/check-retirement-ledger.py:412-418`, `:435-436`), and the exit-0 summary partitions its
counts (`:470-479`) instead of claiming full accounting over a file it never parsed. That message
dishonesty was the original finding, and it is fixed.

Three facts close the residual doubt:

- The **unreadable** arm still exits 2 (`:384-391`). Only the *determinate* unparseable arm is
  downgraded. A `SKILL.md` with no frontmatter `description:` is not loadable as a skill, so it
  advertised no trigger phrase and its deletion removes no claim surface.
- Nothing a branch commits changes its own merge base, so a blocking demand here is unsatisfiable
  by construction — argued at `:79-102`. A gate that cannot be cleared converts a correct PR into
  an indefinite bounce.
- Case C9 in `cogni-workspace/tests/test-skill-trigger-phrases.sh:411-432` already fails any live
  cogni-workspace skill yielding zero phrases, so on any merge base whose CI was green, no genuine
  skill file can reach this arm.

Precedent is this repo's own: root `CLAUDE.md` records the same non-failing observation channel for
`scripts/check-mcp-tool-grant.py`'s registry-accuracy gaps.

### Why the split judgment arose at all

Worth recording, because the code was never the thing out of date. This locus was ruled on
2026-09-02 at 13:29, the ruling was implemented in `d09c11b0`, and it is pinned in **both** failure
directions. The gatekeeper cycle that followed grades against the **committed contract text**,
which was never amended to record the ruling — so it re-raised a settled question as new. The gate
behaved correctly in routing a two-perspective split to a human rather than bouncing the author;
what it lacked was any way to see that the item had already been adjudicated.

The general lesson, noted rather than filed: a contract item settled by a maintainer ruling should
be marked as settled **in the contract record**, so the next cycle grades against the ruling
instead of re-litigating it. If this recurs it earns its own issue.

### Open question 1 — extractor delegation: declined

Keep the guard's extractor and the suite's `emit_phrases` as two copies pinned by C11. Two reasons,
the second decisive:

- Delegation would make every case C1–C11 in a **cogni-workspace plugin suite** depend on a
  repo-root script — C1–C8 transitively through `scan_skills` — inverting the ownership direction
  for the sake of a cleanup.
- Under delegation, **C11 would diff the guard against itself** and be trivially green. The pin
  that justifies the removal is destroyed by the removal.

The contract's own risk watchlist records this as judgment-only, either mechanism acceptable. The
answer is settled; no follow-up issue.

### Open question 2 — contract item 17: graded on the property. Satisfied.

Item 17 is likewise tagged `[fidelity: purpose — the teeth claim must be replayable by the next
reviewer rather than re-derived]`. It is replayable, and was independently replayed.

The `scripts/mutation-check.sh` path prefix is a false premise inherited from the triage comment,
and the PR was right to flag it rather than silently substitute. Confirmed against `origin/main`:

```
$ git ls-tree -r --name-only origin/main | grep -i mutation
cogni-consult/scripts/mutation-check.sh
cogni-portfolio/scripts/mutation-check.sh
```

Both are bespoke and no-argument; no repo-root copy exists. The five-flag harness is cogni-service's
own instrument. Mechanism mis-authored, property met. Filed against the repo that owns the
contract template.

### Open question 3 — recorded, filed upstream

The token-posture contradiction is real — `workflow` is listed forbidden, yet it is exactly what
makes a `lint.yml` change pushable, so a least-privilege token per the documented posture could not
have opened this PR. That is a cogni-service posture question, not an insight-wave one. Filed there.

### Residual advisory — deliberately deferred

The CI job name `Retirement-ledger gate (a skill retirement accounts for its trigger phrases)` reads
repo-wide while the guard is scoped to `cogni-workspace/skills` (`:125-131`). The root `CLAUDE.md`
bullet is already correctly scoped and must quote the job name verbatim, so a rename is a
two-surface change — and #1763 owns whether the scope extends at all, which would churn it straight
back. Boundary recorded; nothing changed.

### Disposition

Proceeding to merge. CI is green on all 14 checks at `8190093c`, 140/140 discovered suites pass, the
recorded teeth trace replays as `guard_verified`, and no version is touched.

Issue #1767, opened 2026-09-02
repo: two more workflow run: blocks suppress their diagnostics on the inherited -e


claude-cogni-work:
## Problem

Parent: #1766, which fixed this defect in `.github/workflows/lint.yml`'s `version-bump-gate` step and — per its own Scope clause ("name them here rather than fixing them silently") — **named rather than fixed** the remaining two instances. This is that follow-up.

GitHub Actions runs a `run:` block as `bash -e {0}`. `set -uo pipefail` does not clear an inherited `-e`; only `set +e` does. A block that captures a command's output and then reads `$?` therefore aborts at the capture line on any non-zero exit, so everything after it — the `cat`s, the annotations, the exit normalisation — is unreachable.

Two blocks still carry the pattern:

1. **`.github/workflows/lint.yml`, "Discover and run every plugin and repo-root test suite" (116–134).** `python3 scripts/run-plugin-tests.py > plugin-tests.json` at 121 is bare, so a failing suite aborts before `rc=$?` (122) and before the per-suite `::error file=` annotation loop (127–132). Its own comment at 118–119 — *"Not `set -e`: the runner's exit code is the result we report on, so it must be captured rather than abort the step"* — asserts precisely the property the code does not have. Not writing `set -e` is not the same as clearing an inherited one.

2. **`.github/workflows/cogni-version-bump.yml`, "Apply the patch bump" (141–183).** The highest-stakes of the three. `python3 scripts/apply-version-bump.py $ARGS > bump.json 2> bump.err` at 151 is bare, so the abort precedes `count` (175) and `deferred_failure` (179) reaching `$GITHUB_OUTPUT`. The two downstream steps gated on those outputs — "Commit, verify, and push" (186) and "Fail the run on a hard skip" (287) — then evaluate their `if:` against never-set values and are **silently skipped**, so the deferred-failure design the block's own comment at 176–179 describes can never execute.

Explicitly **not** affected: `cogni-version-bump.yml` "Commit, verify, and push" (185–284). It uses the same capture idiom inside `verify()` (233–249), but `verify()` is only ever invoked as `if ! verify` (251, 277), and bash exempts an entire function body from `-e` while it is being used as a tested condition. Verified empirically. Do not "fix" it — adding `set +e` there would strip intentional early-abort protection from the step's other commands.

## Acceptance criteria

[...]

PR #1768, opened 2026-09-02
lint.yml: unmask the version-bump gate's diagnostics with set +e

### LINKED ISSUE #1766: repo: the version-bump gate's CI step suppresses its own diagnostics on every non-zero path
## Problem

The version-bump gate's CI step suppresses its own diagnostics on every non-zero path, so an author who trips it sees a bare red with no explanation.

GitHub Actions runs a `run:` block as `bash -e {0}`. `set -uo pipefail` does **not** clear an inherited `-e` — only `set +e` does. The step's first statement after that line captures the guard's output:

```
set -uo pipefail
python3 scripts/check-version-bump.py > version-gate.json 2> version-gate.err
rc=$?
cat version-gate.json
cat version-gate.err >&2
```

The moment the guard exits non-zero, `-e` aborts the step at the capture line. `rc=$?`, both `cat` calls, and everything after them never execute. The build still fails, so the gate's verdict is correct — but the envelope, the violation list and the remediation hint are all unreachable.

## Evidence

Found while reviewing the sibling retirement-ledger gate, which had been copied from this step and inherited the same flaw. Verified there by extracting the real `run:` block and executing it under `bash -e` against stub guards on all four exit paths: the violation and script-error paths produced **completely empty** stdout and stderr, and the script-error path exited 2 rather than the intended normalized 1 — making its `::error::` annotation and the 2-to-1 normalization dead code.

The retirement-ledger step was fixed by inserting `set +e` as the first line of its `run:` block, with the rationale stated in a comment. This issue tracks applying the same fix to the version-bump gate.

## Acceptance criteria

- [ ] The version-bump gate's `run:` block begins with `set +e`, ahead of `set -uo pipefail`, with a comment stating why it is load-bearing.
- [ ] On the violation path the step prints the envelope and the guard's stderr, and still exits non-zero — the explicit `exit` at the end of the block preserves the failing status, so the fix must not turn a failing gate green.
- [ ] The script-error path reaches its `::error::` annotation and exits with the intended normalized status.
- [ ] Verified the way the sibling was: run the extracted block under `bash -e` against stub guards covering clean, degraded, violation and script-error, and record the observed output and exit per path.

## Scope

Workflow-file change only. Do not modify `scripts/check-version-bump.py` — the guard's logic is correct; only its step's diagnostics are unreachable. Check whether any other `run:` block in `.github/workflows/` follows the same `set -uo pipefail`-without-`set +e` capture pattern and name them here rather than fixing them silently.

claude-cogni-work:
Closes #1766.

## Summary

The version-bump gate's `run:` block suppressed its own diagnostics on every non-zero path. GitHub Actions runs a `run:` block as `bash -e {0}`, and `set -uo pipefail` does **not** clear an inherited `-e` — only `set +e` does. So the moment `check-version-bump.py` exited non-zero, `-e` aborted the step at the capture line: `rc=$?`, both `cat`s, the `rc -eq 2` normalization and the terminal `exit "$rc"` never ran. The build went red with an empty log — no envelope, no violation list, no fix hint.

This inserts `set +e` as the block's first executable statement, immediately above the existing `set -uo pipefail`, with the rationale in a comment directly below the pair. The failing status is preserved by the block's existing terminal `exit "$rc"`, not by `-e`, so a failing gate stays red.

Nothing else in the block changes. Both `cat`s, the `rc -eq 2` → `::error::version-bump gate errored` → `exit 1` arm, the degraded-vacuity heredoc with `vrc=$?` and `[ "$vrc" -eq 0 ] || exit 1`, and the terminal `exit "$rc"` are byte-identical; the checkout's `fetch-depth: 0` and `ref:` lines are untouched. `scripts/check-version-bump.py` is not modified. No `shell:` key was added — an explicit `shell: bash` supplies `--noprofile --norc -eo pipefail` and would **re-arm** `-e` — and there is no `continue-on-error:`, no `|| true`, no unconditional `exit 0`.

Changed paths: `.github/workflows/lint.yml` only (7 insertions).

**Ordering note.** The comment sits *below* the two `set` lines deliberately, so `set +e` and `set -uo pipefail` read as one adjacent pair and both stay inside a short `grep -A4` window on the step name. A comment wedged above either one pushes `set -uo pipefail` out of that window.

## Scope decisions

The issue's Scope clause says to check for other `run:` blocks sharing the pattern and "name them here rather than fixing them silently". Two readings exist — attach the negation to *silently* (fix and name), or to the whole clause (name, do not fix). **This PR takes the narrower reading**: the title, the Problem section and all four acceptance criteria are scoped to this one gate, and `cogni-version-bump.yml` "Apply the patch bump" is post-merge *write* behaviour whose fix changes which downstream steps run — that deserves its own verification, not a ride-along in a diagnostics PR.

The other affected blocks are named below **and filed as #1767**, so naming does not mean losing them.

## Sweep — every other `run:` block in `.github/workflows/`

All four workflow files re-enumerated (`cla.yml`, `cogni-version-bump.yml`, `lint.yml`, `workflow-guard.yml`).

| Workflow | Job / step | Lines | Verdict | Action here |
|---|---|---|---|---|
| `lint.yml` | `version-bump-gate` / "Assert no plugin version was touched on this branch" | 63–93 | **Affected** — the reported instance | **Fixed** |
| `lint.yml` | `plugin-test-suites` / "Discover and run every plugin and repo-root test suite" | 116–134 | **Affected** | Named, tracked in #1767 |
| `cogni-version-bump.yml` | `bump` / "Apply the patch bump" | 141–183 | **Affected** | Named, tracked in #1767 |
| `cogni-version-bump.yml` | `bump` / "Commit, verify, and push" | 185–284 | **NOT affected** | Untouched |
| `cogni-version-bump.yml` | `bump` / "Compute the touched-file set" | 88–139 | Not affected — its risky call is guarded by `if ! git diff` | Untouched |
| `lint.yml` (×10), `cla.yml`, `workflow-guard.yml` | single-command guard steps | — | Not affected — `set -euo pipefail` by design, one bare unredirected command as the last statement, so an early abort loses no diagnostics | Untouched |

**`plugin-test-suites` (116–134)** additionally carries a *false comment* at 118–119: "Not `set -e`: the runner's exit code is the result we report on, so it must be captured rather than abort the step." That is precisely the property the code does not have — not writing `set -e` does not clear an inherited one. The correction belongs with the fix, in #1767.

**`Apply the patch bump` (141–183)** is the highest-stakes of the three: the abort precedes `count` (175) and `deferred_failure` (179) reaching `$GITHUB_OUTPUT`, so the `if:`-gated steps at 186 and 287 evaluate against never-set values and are **silently skipped**.

**Why "Commit, verify, and push" is explicitly not affected.** It uses the same capture idiom, but only inside `verify()` (233–249), which is invoked solely as `if ! verify` (251, 277). Bash exempts an entire function body from `-e` while the call sits in a tested context. Verified empirically. Adding `set +e` there would strip intentional early-abort protection from the step's other commands.

## Verification (AC4)

The block was extracted **programmatically from the shipped `lint.yml`** — never hand-transcribed — and executed under `bash -e` against a stubbed guard on all four exit paths, plus a falsifier.

### Extraction (verbatim, replayable from the repo root)

```bash
mkdir -p /tmp/vg && cat > /tmp/vg/extract.py <<'PY'
import sys
STEP = "- name: Assert no plugin version was touched on this branch"
lines = open(sys.argv[1]).read().split("\n")
i = next(n for n, l in enumerate(lines) if l.strip() == STEP)
j = next(n for n in range(i + 1, len(lines)) if lines[n].strip() == "run: |")
body, ind = [], None
for l in lines[j + 1:]:
    if l.strip() == "":
        body.append(""); continue
    cur = len(l) - len(l.lstrip(" "))
    if ind is None: ind = cur
    elif cur < ind: break
    body.append(l[ind:])
while body and body[-1] == "": body.pop()
sys.stdout.write("\n".join(body) + "\n")
PY
python3 /tmp/vg/extract.py .github/workflows/lint.yml > /tmp/vg/block.sh
```

Integrity assertions (both load-bearing — `<<'PY'` requires its terminator at column 0, so a dedent that left `PY` indented would make every record below a false result):

```bash
grep -qx "python3 - <<'PY'" /tmp/vg/block.sh && grep -qx 'PY' /tmp/vg/block.sh   # -> OK
wc -l < /tmp/vg/block.sh                                                          # -> 36
```

### Falsifier derivation, with non-vacuity assertions

```bash
grep -vxF 'set +e' /tmp/vg/block.sh > /tmp/vg/falsifier.sh
[ "$(( $(wc -l < /tmp/vg/block.sh) - $(wc -l < /tmp/vg/falsifier.sh) ))" -eq 1 ] || exit 1
grep -qxF 'set +e' /tmp/vg/block.sh    || exit 1
grep -qxF 'set +e' /tmp/vg/falsifier.sh && exit 1
```

`grep -v` happily emits an *unchanged* file when nothing matched, so a falsifier identical to the block would "pass" every path and evidence nothing. These three assertions are what rule that out. `-x -F` is whole-line fixed-string, so it removes the bare statement and leaves the comment lines mentioning `set +e` intact.

### Stub harness

The stub is a `python3` **shim on PATH that dispatches on argv**, not a bare replacement interpreter. This matters: the block invokes `python3` twice — once for the guard and once for its own `python3 - <<'PY'` vacuity check — so a stub answering every `python3` call would swallow the vacuity check and make the degraded path unexercisable.

```bash
cat > /tmp/vg/bin/python3 <<'SH'
#!/usr/bin/env bash
if [ "${1:-}" = "scripts/check-version-bump.py" ]; then
  printf '%s' "$STUB_STDOUT"; printf '%s' "$STUB_STDERR" >&2; exit "$STUB_RC"
fi
exec "$REAL_PY" "$@"
SH
```

Payloads are passed through the environment, so no shell quoting or heredoc interpolation can mangle the JSON.

### Observed results

```
SHIPPED BLOCK (36 lines, extracted from .github/workflows/lint.yml)

Path 1 clean — stub exit 0
  stdout (124 bytes):
    {"success": true, "data": {"status": "clean", "violations": []}, "error": ""}version-bump gate ran for real (status: clean)
  stderr (25 bytes):
    version-bump gate: clean
  block exit: 0

Path 2 violation — stub exit 1
  stdout (175 bytes):
    {"success": false, "data": {"status": "clean", "violations": [{"plugin":"cogni-trends","kind":"version-touched"}]}, "error": ""}version-bump gate ran for real (status: clean)
  stderr (180 bytes):
    FAIL: 1 version violation(s):
      - cogni-trends/.claude-plugin/plugin.json: version touched on a feature branch
    Fix: revert the version line; the post-merge workflow owns the bump.
  block exit: 1

Path 3 script-error — stub exit 2
  stdout (106 bytes):
    {"success": false, "data": null, "error": "marketplace.json not found"}::error::version-bump gate errored
  stderr (0 bytes):
    (empty)
  block exit: 1

Path 4 degraded — stub exit 0
  stdout (136 bytes):
    {"success": true, "data": {"status": "degraded", "degraded_reason": "base ref origin/main unresolvable", "violations": []}, "error": ""}  stderr (218 bytes):
    version-bump gate: base ref unresolvable
    version-bump gate is VACUOUS (status: degraded) — base ref origin/main unresolvable
    It compared nothing. This usually means fetch-depth: 0 was removed from the checkout step.
  block exit: 1

FALSIFIER (identical, set +e removed)

Path 1 clean — stub exit 0
  stdout (124 bytes):
    {"success": true, "data": {"status": "clean", "violations": []}, "error": ""}version-bump gate ran for real (status: clean)
  stderr (25 bytes):
    version-bump gate: clean
  block exit: 0

Path 2 violation — stub exit 1
  stdout (0 bytes):
    (empty)
  stderr (0 bytes):
    (empty)
  block exit: 1

Path 3 script-error — stub exit 2
  stdout (0 bytes):
    (empty)
  stderr (0 bytes):
    (empty)
  block exit: 2

Path 4 degraded — stub exit 0
  stdout (136 bytes):
    {"success": true, "data": {"status": "degraded", "degraded_reason": "base ref origin/main unresolvable", "violations": []}, "error": ""}  stderr (218 bytes):
    version-bump gate: base ref unresolvable
    version-bump gate is VACUOUS (status: degraded) — base ref origin/main unresolvable
    It compared nothing. This usually means fetch-depth: 0 was removed from the checkout step.
  block exit: 1

```

### Reading the falsifier

| path | shipped block | `set +e` removed |
|---|---|---|
| 1 clean (guard rc 0) | envelope + summary, **exit 0** | unchanged, exit 0 |
| 2 violation (guard rc 1) | envelope + FAIL list + Fix hint, **exit 1** | **0 bytes stdout, 0 bytes stderr**, exit 1 |
| 3 script error (guard rc 2) | envelope + `::error::` annotation, **exit 1** | **0 bytes stdout, 0 bytes stderr**, exit **2** |
| 4 degraded (guard rc 0) | envelope + VACUOUS message, **exit 1** | unchanged, exit 1 |

Exactly the two non-zero guard paths flip, which is the reported defect reproduced; path 3 additionally shows the lost 2-to-1 exit normalization. Paths 1 and 4 are expected to be identical — the stub exits 0 on both, so nothing aborts — and recording them is what shows the falsifier is not simply breaking everything.

## Test plan

- [x] `python3 scripts/check-workflow-yaml.py` — exit 0 (this is the repo's own workflow-YAML guard; the repo ships no PyYAML dependency, so this, not `yaml.safe_load`, is the parse check).
- [x] `python3 scripts/check-version-bump.py` — exit 0.
- [x] `python3 scripts/run-plugin-tests.py` — exit 0, 139 suites, `failed_suites: []`.
- [x] `python3 scripts/check-result-line-plainness.py` — exit 0.
- [x] `python3 scripts/check-case-id-pairing.py` — exit 0.
- [x] `git diff --name-only origin/main...HEAD` → `.github/workflows/lint.yml` only; nothing under `scripts/**`, no `plugin.json` / `marketplace.json` version line.
- [x] `grep -n -A4 'Assert no plugin version was touched' .github/workflows/lint.yml` shows `set +e` then `set -uo pipefail` on consecutive lines.

## Mutation recipe

Not applicable, stated rather than omitted: the mutation-recipe check fires when the touched set adds a guard under a `scripts/` directory. This diff touches only `.github/workflows/lint.yml` and adds no guard and no test case, so there is no committed case id to drive RED→GREEN. The attribution role a mutation recipe would play is carried here by the falsifier transcript above, which is derived from the shipped file and asserted to be a one-line delta.

## Note for the reviewer

This diff touches only a repo-root file, so `check-preflight.sh` resolves to **no plugin** and its plugin-scoped instruments grade nothing. A green preflight on this PR is not evidence about this change — the path records and the falsifier are. Please weigh those, not the preflight's exit code.

## Follow-up issues

- **#1767** — repo: two more workflow `run:` blocks suppress their diagnostics on the inherited `-e` (the two AFFECTED rows named in the sweep table above, plus the false comment at `lint.yml` 118–119).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -62,7 +62,14 @@ jobs:
 
       - name: Assert no plugin version was touched on this branch
         run: |
+          set +e
           set -uo pipefail
+          # `set +e` above is load-bearing and must stay first. Actions runs this
+          # block as `bash -e {0}`, so -e is inherited before line 1 executes, and
+          # `set -uo pipefail` does NOT clear an inherited -e — only `set +e` does.
+          # Without it the guard's non-zero exit aborts the step at the capture
+          # below: a bare red with no envelope, no violation list, no fix hint.
+          # The failing status is preserved by the terminal `exit "$rc"`, not by -e.
           python3 scripts/check-version-bump.py > version-gate.json 2> version-gate.err
           rc=$?
           cat version-gate.json


claude-cogni-work:
## Ship-ready

Author instance has handed this PR off for independent review by the merger instance. Draft, carrying `svc:needs-review`. No review verdict was posted by the author — that is the merger's to form.

Graded against the `origin: triage` contract on #1766 (`derived_at 2026-09-02T14:52:48Z`) — written before this branch existed, so it is diff-blind by chronology.

- **20 items graded · 20 satisfied · 0 violated · 0 unclear**
- **Residual carried to the merger: none.** No fix pass was needed or run.
- Items 15, 16 and 19 are **vacuously** satisfied — this PR deliberately leaves the two sibling blocks unmodified and adds no test file, so their "if X is fixed…" conditionals do not trigger.

The self-grade re-ran every executable claim against the head ref (`89d3a331`) rather than reasoning from the diff: `check-workflow-yaml.py`, `check-version-bump.py`, `run-plugin-tests.py` (139/139), `check-result-line-plainness.py`, `check-case-id-pairing.py`, plus an independent re-derivation of both numbers the PR body asserts — the 36-line extracted block and the one-line falsifier delta — and a re-run of the four-path harness and the falsifier from the body's own recorded commands. All matched.

This is advisory. The merger forms its own contract and grades independently.

### Handoff preflight

`check-preflight.sh` exited **0**, status `no_applicable_checks` — and it is worth stating plainly rather than reporting as a clean bill of health:

```
discovered=7  passed=0  failed=0  skipped=7  gating_applicable=0
```

**Every instrument skipped; the preflight graded nothing.** This is a repo-root diff (`.github/workflows/lint.yml` only), which resolves to no plugin, so `version-bump` skipped on `touched set resolves to no plugin` and the other six on `not_applicable` (no touched `SKILL.md`, `agents/*.md`, `tests/` or `scripts/` guard path). The accounting identity `7 = 0 + 0 + 7` holds, so nothing was lost silently — but a green preflight here is not evidence about this change.

**What to weigh instead:** the four per-path transcripts and the falsifier in the PR body. Those are derived from the shipped file, and the falsifier is asserted to be a one-line delta, so they are attributable in the way the preflight is not.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1768

**Summary:** Both perspectives passed with no findings; all 20 committed review-contract items graded satisfied by each, and independent re-derivation of the PR's literal claims (36-line extraction, 1-line falsifier delta, 139/139 suites) confirmed true at head.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Positive observations

- Minimal surgical diff: 0 deletions is itself the proof for contract item 18. (plugin-validator)
- Rationale comment defuses the 'delete this redundant no-op' regression. (plugin-validator)
- Falsifier guarded by three non-vacuity assertions, closing the grep -v silent-no-op class. (plugin-validator)
- Stub is an argv-dispatching PATH shim, so the block's own vacuity heredoc still runs. (plugin-validator)
- Sweep names the NOT-affected block explicitly rather than omitting it. (plugin-validator)
- PR body proactively discloses that check-preflight resolves to no plugin here. (plugin-validator)
- Every literal item verified by running its own stated command at head. (contract-grading-reviewer)
- Both asserted numbers independently re-derived and matched. (contract-grading-reviewer)

### Recommendation

---

PR #1782, opened 2026-09-02
cogni-knowledge: population-derived per-fence probe_plugin guard

### LINKED ISSUE #1720: cogni-knowledge: consolidate the 15 duplicated inline probe_plugin() definitions across the live knowledge-* skills

claude-cogni-work:
Closes #1720.

## Census (re-derived at the time of the work, `origin/main` `fb89b933`)

`probe_plugin() {` occurs in **3 files under `cogni-knowledge/`, exactly 1 of them live**:

| File | Line | Status |
|---|---|---|
| `cogni-knowledge/skills/knowledge-refresh/SKILL.md` | 77 | **live** — the sole consumer; its sole call is at :85, inside the same fence (opened :73, closed :86) |
| `cogni-knowledge/_archive/skills/knowledge-report/SKILL.md` | 50 | archived — outside the live test glob |
| `cogni-knowledge/_archive/skills/knowledge-research/SKILL.md` | 50 | archived — outside the live test glob |

No other live `cogni-knowledge/skills/*/SKILL.md` carries a `probe_plugin` token at all.
The issue claimed 15 live definitions.

## Why the proposed consolidation is declined

The other named siblings were **removed**, not left duplicated, when the wiki engine was
vendored and the research chain clean-broken. Their absence is already asserted by roughly
fifteen `assert_not_grep 'probe_plugin'` arms across the contract suites
(`knowledge-wiki-probe-01..04`, `compose-contract-16`, `finalize-11b`, `khealth-06`,
`ingest-contract-10b`, `ingest-source-05b`, `lint-06`, `prefill-05`, `verify-contract-26`,
`skill-contracts-14/-16/-89/-90b/-102/-103`). With one live definition there is nothing to
de-duplicate.

Worse, extracting it would be a regression. A skill's fenced Bash blocks do not share shell
state, so moving the definition into a sourced snippet and dropping the inline copy makes
the calling fence at :85 stop defining what it calls — the standalone exit-127 defect
`refresh-resweep-28` exists to catch, which would report `FAIL-NODEF`.

This PR therefore modifies **no** file under `cogni-knowledge/skills/`, creates **no** shared
probe snippet, and leaves `cogni-knowledge/scripts/resolve-wiki-scripts.sh` untouched.

## What it adds

One new suite, `cogni-knowledge/tests/test_probe_plugin_fence.sh` (slug `probe-plugin-fence`),
delivering the issue's one surviving acceptance criterion: a **population-derived, set-level,
per-fence** guard. For every live `skills/*/SKILL.md` it derives the consuming set by scanning,
and requires each fence that *calls* `probe_plugin` to define it — or to `source`/`.` a file
that defines it — inside that **same** fence. One-element population today, auto-covering any
future consumer with no edit here (the same drift the `resolve-wiki` sibling's derived
population fixed after its hand-maintained list fell to 8 names against 13 actual callers).

| Case | Asserts |
|---|---|
| `probe-plugin-fence-01` | vacuity floor — red when no `SKILL.md` was scanned at all |
| `probe-plugin-fence-02` | **the deliverable** — every derived calling fence defines `probe_plugin`, or sources a file that does, in that same fence |
| `probe-plugin-fence-03` | no orphan definition — a fence defining `probe_plugin` also calls it there |
| `probe-plugin-fence-04` | silent shrink — a skill naming `probe_plugin` but yielding no derived call site is graded by nothing |
| `probe-plugin-fence-05`..`-09` | self-tests driving the **same** `scan_file()` against `mktemp -d` fixtures, so a self-test cannot diverge from what grades the tree |

Three scanning properties are load-bearing and stated in the suite header: a
whitespace-tolerant fence opener (an indented fence under a numbered step would otherwise
stop the toggle and flatten the file); a call detector matching any in-fence line naming
`probe_plugin`, wider than the per-file predicate's `/probe_plugin[ \t]+[A-Za-z]/`, which
misses `probe_plugin "$plugin" claims`; and an inline definition that short-circuits the
source path.

## Two defects closed before the code was written

1. **A dead FAIL arm.** The widened call detector also matches the definition line
   `probe_plugin() {`. Without a `next` on the `hasdef` rule, `hasdef` would always imply
   `hascall`, making the orphan branch unreachable **for every possible input** — case `-03`'s
   FAIL arm dead while both arms stayed statically present, which `check-case-id-pairing.py`
   cannot see. Verified empirically: the naive spelling emits `CALL` for a def-only fence
   where the shipped spelling emits `ORPHAN`.
2. **A day-one red.** The live fence sources `${CLAUDE_PLUGIN_ROOT}/scripts/resolve-wiki-scripts.sh`,
   which does **not** define `probe_plugin`. Both the no-source and sourced verdicts are gated
   on `hascall && !hasdef`, so an inline definition short-circuits before the source path is
   consulted. Adjudicating on that source would have redded this suite against a correct tree.

## One deliberate divergence from the sibling it is modelled on

`resolve-wiki-14` reds on a zero **call** count as well as a zero scan count. Case
`probe-plugin-fence-01` deliberately does not: that sibling grades a thirteen-skill
population where zero callers can only mean a broken scan, whereas this population is one, and
a legitimate future removal of the sole consumer must not force an edit to this file. The
scan-is-broken direction is case `-04`'s job instead, and it keys off the tree rather than a
literal. Stated in the suite header so it reads as a choice, not a transcription slip.

## Verification

| Check | Result |
|---|---|
| `bash cogni-knowledge/tests/test_probe_plugin_fence.sh` | 9/9 green, exit 0 — baseline `scanned=21, calling=1` |
| Forced-red vacuity: `PROBE_PLUGIN_FENCE_SKILLS_DIR=$(mktemp -d) bash …` | `FAIL: probe-plugin-fence-01`, exit 1 |
| Case-id uniqueness census (run the suite, group emitted first tokens) | empty — no collision |
| `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` | **80/80 pass**, 0 fail; the new suite is discovered at `[55/80]` |
| `refresh-resweep-28` / `-29` | still green, unmodified |
| `scripts/check-case-id-pairing.py` | 0 findings |
| `scripts/check-result-line-plainness.py` | 0 findings |
| `scripts/check-breadcrumbs.py` | exit 0 |
| `scripts/check-code-span-nesting.py` | exit 0 |
| `scripts/check-version-bump.py` | `clean: true`, 0 plugins touched |
| `git diff --stat origin/main` | 1 file changed, 325 insertions — no `version` line in either manifest |

## Mutation recipe

The harness is not in this repo — `cogni-knowledge/tests/README.md` records that it ships with
the cogni-service managed-service tooling, and the same-named scripts under `cogni-portfolio/`
and `cogni-consult/` are different, bespoke scripts. Run it from the managed-service checkout:

```sh
bash <managed-service>/cogni-service/scripts/mutation-check.sh \
  --root <insight-wave> \
  --file cogni-knowledge/skills/knowledge-refresh/SKILL.md \
  --expr 's/^probe_plugin\(\) \{\n(?:.*\n)*?^\}\n//m' \
  --test 'bash cogni-knowledge/tests/test_probe_plugin_fence.sh' \
  --case 'probe-plugin-fence-02'
```

The expression deletes the whole `probe_plugin() { ... }` block from the knowledge-refresh
fence while leaving the call at line 85. **Run directly and confirmed:** the suite then prints

```
FAIL: probe-plugin-fence-02 fence(s) call probe_plugin without defining it, or sourcing a file that defines it, in that same fence (skill:fence-open-line):  knowledge-refresh:73
```

and exits 1; restoring the file returns it to exit 0. The offender names
`knowledge-refresh:73`, which is what proves the population sweep reaches the sole live
consumer — a `vacuous_guard` verdict would mean it does not.

## Plan record

https://github.com/cogni-work/insight-wave/issues/1720#issuecomment-5515841222

### DIFF
--- a/cogni-knowledge/tests/test_probe_plugin_fence.sh
+++ b/cogni-knowledge/tests/test_probe_plugin_fence.sh
@@ -1,0 +1,325 @@
+#!/usr/bin/env bash
+# test_probe_plugin_fence.sh - population-derived, per-fence probe_plugin guard.
+#
+# A skill's fenced Bash blocks do not share shell state. A fence that CALLS
+# probe_plugin must therefore define it -- or source a file that does -- inside
+# that same fence, or the call is a standalone exit-127 at run time. One suite
+# already asserts that property, but only for the single file it names by hand.
+# This one derives the graded population from the tree instead: every
+# skills/*/SKILL.md is scanned, and a skill enters the set by calling
+# probe_plugin inside a fence. A newly-calling skill is graded on its first run
+# with no edit here, which is the drift a hand-named subject cannot survive.
+#
+# Three scanning properties are load-bearing, each for a stated reason:
+#
+#   - The fence opener tolerates leading whitespace. A fence nested under a
+#     numbered step is indented, and an anchored opener would stop toggling
+#     there and grade the rest of the file as one flat blob.
+#
+#   - The call detector matches ANY in-fence line naming probe_plugin, not just
+#     a bare-literal first argument. A detector keyed on an alphabetic argument
+#     misses the quoted-variable form, which is precisely the next consumer a
+#     derived population is supposed to catch for free.
+#
+#     That widening removes the property that would otherwise let the definition
+#     rule fall through: `probe_plugin() {` also names probe_plugin. So the
+#     definition rule ends in `next`, keeping a def-only fence out of the call
+#     set. Without it, hasdef would always imply hascall, the orphan branch
+#     would be unreachable for every possible input, and case 03's FAIL arm
+#     would be dead while both arms stayed statically present -- which the
+#     case-id pairing guard cannot see.
+#
+#   - An inline definition SHORT-CIRCUITS the source path. Both the no-source
+#     and the sourced verdicts are gated on `hascall && !hasdef`, so a fence
+#     that defines probe_plugin inline is clean and its source lines are never
+#     resolved. The sole live consumer sources a wiki-engine resolver that does
+#     not define probe_plugin; adjudicating on that source would red this suite
+#     against a correct tree.
+#
+# Case 01 is a vacuity floor on the SCANNED count only. It deliberately does
+# NOT red on a zero call count, unlike the sibling suite it is modelled on: that
+# sibling grades a thirteen-skill population, where zero callers can only mean a
+# broken scan, whereas the population here is one, and a legitimate removal of
+# the sole consumer must not force an edit to this file. The scan-is-broken
+# direction is case 04's job instead, and it keys off the tree: a skill whose
+# text names probe_plugin while yielding no derived call site is graded by
+# nothing, which is the failure a shrinking population actually produces.
+#
+# Cases 05-09 drive the SAME scan_file() against synthetic fixtures, so a
+# self-test can never diverge from what grades the tree. They are what proves
+# the detector discriminates rather than always agreeing.
+#
+# bash 3.2 + stdlib only.
+
+set -eu
+
+PLUGIN_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+# The scan root is overridable so case 01's liveness arm is reachable: point it
+# at an empty dir and the derived set is empty, which must go RED.
+SKILLS_DIR="${PROBE_PLUGIN_FENCE_SKILLS_DIR:-$PLUGIN_ROOT/skills}"
+
+. "$(dirname "$0")/fixtures/test_helpers.sh"
+
+errors=0
+
+# The fence-scoped pass. Emits only BARE records -- never a PASS:/FAIL: label --
+# so every result line in this suite still originates at a red/green call and
+# stays visible to scripts/check-case-id-pairing.py:
+#   CALL   <skill>                     once per file with any fenced call
+#   NODEF  <skill>:<fence-open-line>   calling fence, no def, no source at all
+#   SRC    <skill>:<fence-open-line> <token>
+#                                      calling fence, no def, one record per
+#                                      buffered source token for the shell to
+#                                      adjudicate
+#   ORPHAN <skill>:<fence-open-line>   fence defines probe_plugin, never calls it
+#
+# Rule ORDER is load-bearing and the two `next` statements are not optional --
+# see the header. Rule 4 is the catch-all, reached only by lines rules 2 and 3
+# did not consume.
+scan_file() {
+  awk -v skill="$2" '
+    /^[ \t]*```/ {
+      if (inf) {
+        if (hascall) {
+          calls = 1
+          if (!hasdef) {
+            if (nsrc == 0) print "NODEF " skill ":" fl
+            else for (i = 1; i <= nsrc; i++) print "SRC " skill ":" fl " " src[i]
+          }
+        } else if (hasdef) {
+          print "ORPHAN " skill ":" fl
+        }
+      }
+      inf = !inf
+      if (inf) { hasdef = 0; hascall = 0; nsrc = 0; fl = NR }
+      next
+    }
+    inf && /probe_plugin\(\)[ \t]*\{/ { hasdef = 1; next }
+    inf && /^[ \t]*(\.|source)[ \t]+/ {
+      for (i = 1; i <= NF; i++) {
+        if ($i == "." || $i == "source") { nsrc = nsrc + 1; src[nsrc] = $(i + 1); break }
+      }
+      next
+    }
+    inf && /probe_plugin/ { hascall = 1 }
+    END { if (calls) print "CALL " skill }
+  ' "$1"
+}
+
+# Does the file a buffered source token names define probe_plugin? The token
+# arrives as awk read it, so surrounding quotes and a CLAUDE_PLUGIN_ROOT prefix
+# are still on it. An unresolvable path is deliberately NOT satisfying: a fence
+# whose only vouching source cannot be read is vouched for by nothing.
+src_defines_probe() {
+  sp_path=$(printf '%s' "$1" | tr -d "\"'" | sed -e "s|[\$]{CLAUDE_PLUGIN_ROOT}|$2|g" -e "s|[\$]CLAUDE_PLUGIN_ROOT|$2|g")
+  [ -f "$sp_path" ] || return 1
+  grep -qE 'probe_plugin\(\)[ \t]*\{' "$sp_path"
+}
+
+# Turn one file's scan into space-joined offender tokens for case 02: every
+# NODEF fence, plus every sourced fence where no buffered source resolves to a
+# file defining probe_plugin.
+adjudicate_nodef() {
+  an_scan=$1
+  an_root=$2
+  printf '%s\n' "$an_scan" | sed -n 's/^NODEF /  /p' | tr '\n' ' '
+  an_keys=$(printf '%s\n' "$an_scan" | sed -n 's/^SRC \([^ ]*\) .*/\1/p' | LC_ALL=C sort -u)
+  for an_key in $an_keys; do
+    an_ok=no
+    for an_tok in $(printf '%s\n' "$an_scan" | awk -v k="$an_key" '$1 == "SRC" && $2 == k { print $3 }'); do
+      if src_defines_probe "$an_tok" "$an_root"; then
+        an_ok=yes
+      fi
+    done
+    if [ "$an_ok" = no ]; then
+      printf '  %s ' "$an_key"
+    fi
+  done
+}
+
+# -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


claude-cogni-work:
## Ship-ready

Handed off by the author instance for independent merger review. Draft, carrying
`svc:needs-review`. No review verdict was formed on this side and no `svc:review-*`
label was applied — the independent review belongs to the merger.

**Handoff preflight:** exit 0, clean. 3 passed — `run-tests` (cogni-knowledge suites
green), `version-bump` (no version-line touch), `mutation-recipe` (the recorded recipe
is runnable as written and `probe-plugin-fence-02` resolves in the suite). 4 skipped,
all `not_applicable`: `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` —
the diff touches no `SKILL.md` and no `agents/*.md`.

**Author self-grade against the issue-time contract** (`--target issue --origin triage`,
derived `2026-09-02T20:04:16Z`, before any branch was cut): **18/18 satisfied, 0 violated,
0 unclear.** No fix pass was needed and there is **no residual** for you to inherit.

Two contract items the grader was asked to scrutinise, and how they landed:

- **Item 13 (mutation recipe).** The item cites `scripts/mutation-check.sh`, which does
  not exist in this repo — the harness ships from the managed-service tooling, as
  `cogni-knowledge/tests/README.md` records. Graded *satisfied*: a defect in the
  contract's own premise, not a PR gap. The PR states the full
  `--root/--file/--expr/--test/--case` invocation and additionally reports the
  directly-run equivalent producing `FAIL: probe-plugin-fence-02 … knowledge-refresh:73`,
  exit 1, restored to exit 0.
- **Item 10 (vacuity floor).** The PR deliberately diverges from `resolve-wiki-14` by
  *not* reddening on a zero call count, because this population is one and a legitimate
  future removal of the sole consumer must not force an edit to the suite. Graded
  *satisfied*: the contract's own text offers two acceptable shapes and the PR implements
  both (case `-01` zero-scanned, case `-04` uncovered-name). The divergence is stated in
  the suite header so it reads as a choice.

**The one thing worth your independent eye:** this PR *declines* the issue's stated
proposal. The issue asked to consolidate 15 duplicated `probe_plugin()` definitions; the
census re-derived at `origin/main` `fb89b933` finds **1** live definition, not 15. The
judgment that the remaining deliverable is the guard alone — and that building the
proposed shared snippet would be a regression — is the load-bearing call here, not the
code.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1782

**Summary:** Both perspectives passed with all 18 committed contract items graded satisfied; the sole finding — a minor, merger-added preference on the SOURCE detector's fails-closed recall floor — is a preference addition on the gatekeeper path, so it is excluded from gating per the advisory-addition pre-step and the PR merges clean.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (advisory-addition) `cogni-knowledge/tests/test_probe_plugin_fence.sh:98` — The source-line rule anchors on a line-initial `.`/`source`, so a source reached through a compound command (e.g. `cd "$dir" && source snippet.sh`) buffers no SRC token and the fence falls through to the NODEF branch — reported as an offender though it legitimately sources a defining file. The direction fails CLOSED (a false red, never a false green). The suite header states three load-bearing scanning properties but this recall limit of the SOURCE detector is not among them; the header's stated recall reasoning covers the CALL detector only. No live consumer is affected today: the sole calling fence defines probe_plugin inline and short-circuits the source path.
  - *Suggested fix:* Either widen the rule to match a `.`/`source` token anywhere on an in-fence line (mirroring the deliberate widening of the call detector), or add a fourth header bullet stating the line-initial recall floor and that it fails closed. No behaviour change is required for the current tree.

### Positive observations

- Contract items 1-18 all graded satisfied with empirical evidence: the suite was run (9/9 green, exit 0, scanned 21 / calling 1), its vacuity arm forced red via the scan-root override, and the PR body's mutation recipe independently reproduced (FAIL: probe-plugin-fence-02, offender knowledge-refresh:73, exit 1). (contract-grading-reviewer)
- Scope fence holds by construction: git diff shows exactly one added file; knowledge-refresh/SKILL.md is byte-identical to base with def :77, call :85, fence :73-86. (contract-grading-reviewer)
- Per-fence (not per-file) granularity proven behaviourally: a split-fence fixture reported both a NODEF and an ORPHAN offender. (contract-grading-reviewer)
- The guard discriminates rather than merely passing: four distinct red configurations each produced the expected named offender and exit 1, with the unmutated control green. (plugin-validator)
- The awk rule-order defect the header calls out is real and correctly handled — the `next` on the definition rule is what keeps case 03's FAIL arm reachable, confirmed empirically rather than statically. (plugin-validator)
- Cases 05-09 drive the SAME scan_file()/adjudicate_nodef() that grades the tree against mktemp fixtures, so a detector that stopped discriminating cannot pass the self-tests while still reporting the tree clean. (plugin-validator)
- The suite complements rather than duplicates test_knowledge_wiki_probe.sh: those cases assert probe ABSENCE, this one asserts fence CO-LOCATION for callers. No overlapping-sweep drift risk. (plugin-validator)
- The ORPHANDEF direction of refresh-resweep-28 is preserved (case 03) rather than silently narrowed, and the choice is stated in the header. (plugin-validator)
- The zero-call-count divergence from resolve-wiki-14 is a stated, reasoned choice with the scan-is-broken direction assigned to case 04. (plugin-validator)
- bash 3.2 + stdlib only holds in practice: green under /bin/bash 3.2.57 as well as bash 5.3. (plugin-validator)
- The scan root is exposed via PROBE_PLUGIN_FENCE_SKILLS_DIR purely so the vacuity arm stays reachable, defaulting to live skills/ so the two _archive definitions never enter the graded set. (plugin-validator)

### Recommendation

---

PR #1800, opened 2026-09-02
repo: set +e in the two remaining -e-suppressed workflow blocks

### LINKED ISSUE #1767: repo: two more workflow run: blocks suppress their diagnostics on the inherited -e

claude-cogni-work:
Closes #1767. Follow-up to #1766, which fixed this defect in `version-bump-gate` and — per its own Scope clause — named rather than fixed the remaining two instances.

## Summary

Actions runs a `run:` block as `bash -e {0}`, so `-e` is in force before the block's first line. `set -uo pipefail` does not clear it; only `set +e` does. Both blocks captured a guard's output and then read `$?`, so a non-zero guard exit aborted the step **at the capturing command itself** and every diagnostic below it was unreachable.

- **`.github/workflows/lint.yml`, "Discover and run every plugin and repo-root test suite"** — opens with `set +e` ahead of `set -uo pipefail`, so `rc=$?`, the per-suite `::error file=` annotation loop and the terminal `exit "$rc"` are reachable on a failing sweep. The block's comment asserted precisely the property the code lacked; replaced with one that states the real mechanism, matching the #1766 house style (both `set` lines adjacent, rationale below the pair).
- **`.github/workflows/cogni-version-bump.yml`, "Apply the patch bump"** — opens with `set +e`, so both `cat`s, the skip annotations and the `count` / `deferred_failure` writes reach `$GITHUB_OUTPUT` on every exit path. The two `if:`-gated downstream steps stop evaluating against never-set values.
- **New `tests/test_workflow_run_block_e_inheritance.sh`** — 17 cases.

## Scope decisions

**"Commit, verify, and push" is deliberately NOT changed.** Its only capture sits inside `verify()`, which is invoked solely as `if ! verify` (two call sites), and bash exempts an entire function body from `-e` while it serves as a tested condition. The pattern is therefore not present there. Verified at base rather than assumed — both call sites re-checked at head, and the step is byte-identical to base (the diff's two hunks land ~35 lines above it).

**No terminal `exit "$RC"` was added to "Apply the patch bump".** That block deliberately has no terminal exit and falls through to an implicit 0: on RC=1 (a hard skip) the step is *meant* to succeed so the healthy plugins still commit, with the run failed later by "Fail the run on a hard skip". Adding one "for symmetry" with the lint block would re-skip "Commit, verify, and push" through a new door while every `set +e` check still passed. Cases `wfe-09`, `wfe-10` and `wfe-14` lock it — and injecting that exit reddens all three.

**No standing class-wide CI guard.** #1766 and #1767 both explicitly excluded it as a separately reviewable unit. This suite asserts the behaviour of two named blocks and never scans other workflows for the pattern.

**No `plugin.json` version bump** — the post-merge workflow owns it.

## Verification

Both blocks are extracted **programmatically from the shipped YAML** by a stdlib line scanner (PyYAML is unavailable in this repo and nothing imports it), then replayed under `bash -e` against stub guards driving every documented exit path, each case in its own `mktemp -d` with a fresh `$GITHUB_OUTPUT`.

| Case | Path | Observed |
|---|---|---|
| `wfe-03` | lint rc=0 | exit 0, no annotation |
| `wfe-04` | lint rc=1 | `::error file=…::test suite failed: …` emitted, exit 1 |
| `wfe-05` | lint rc=2 | exit 2, annotation loop reached, no traceback |
| `wfe-06` | lint **falsifier** | `set +e` removed → annotation **absent** |
| `wfe-08` | bump RC=0 | `count=1`, `deferred_failure=0`, exit 0 |
| `wfe-09` | bump RC=1 | skip annotation emitted, **both** outputs written, exit 0 |
| `wfe-10` | bump RC=1, none bumped | `count=0`, `deferred_failure=1`, "nothing to do", exit 0 |
| `wfe-11` | bump RC=2 | `::error::apply-version-bump.py errored`, exit 1, no `count` |
| `wfe-13` | bump **falsifier** | `set +e` removed → `deferred_failure` **absent**, non-zero exit |

`wfe-04` keys on the **annotation**, never the exit code: the block exits 1 on a failing sweep both with and without the fix, so an exit-code assertion there could never redden. The annotation is what the fix restores.

**Two things the mutation pass changed, worth flagging to a reviewer:**

1. The suite originally located `set +e` with a plain substring match. The retirement-ledger gate's *comment* contains the text `` `set +e` FIRST ``, so the locator matched the comment at line 1 instead of the statement at line 12 — and reported that block correctly ordered while `set +e` sat two lines **below** its capture. Whether that was harmless depended only on whether a block's comment happened to sit above or below its code, which is exactly what an ordering assertion must not depend on. The locator now skips comment lines, and the mutation that exposed this now reddens `wfe-17`.
2. `wfe-15` originally asserted an **exact** count of three `set +e` lines. A future block in this class is fixed by adding a fourth, so the equality would have reddened on correct work. It is now a floor, and the reordering coverage it could never provide moved to `wfe-16`/`wfe-17`, which assert ordering per block for the two lint.yml blocks this PR does not otherwise touch.

## Mutation recipe

```
bash cogni-service/scripts/mutation-check.sh --root . --file .github/workflows/cogni-version-bump.yml --expr '/^          set +e$/d' --test tests/test_workflow_run_block_e_inheritance.sh --case wfe-09
```

Note that `cogni-service/scripts/mutation-check.sh` is not vendored in this repo, so the recipe above is recorded in its canonical portable form. The in-repo replayable equivalent, run and confirmed RED before this PR was opened:

```
perl -i -ne 'print unless /^          set \+e$/' .github/workflows/cogni-version-bump.yml
bash tests/test_workflow_run_block_e_inheritance.sh
git checkout -- .github/workflows/cogni-version-bump.yml
```

Every case was mutation-proved rather than merely observed green: removing `set +e` from either fixed block, reordering it below the capture in any of the four at-risk blocks, injecting the forbidden terminal exit, restoring the stale comment, and neutering the falsifier helper each redden exactly the expected cases.

## Test plan

- [x] `bash tests/test_workflow_run_block_e_inheritance.sh` — 17/17 PASS, exit 0 (~1.9s, well inside the 300s per-suite budget)
- [x] `python3 scripts/check-workflow-yaml.py` — clean; both edited files still parse and declare jobs
- [x] `python3 scripts/run-plugin-tests.py` — 142 suites, `failed_suites: []`; the new suite is discovered by the `tests/*.sh` glob
- [x] `python3 scripts/check-case-id-pairing.py` and `python3 scripts/check-result-line-plainness.py` — both clean
- [x] `grep -n 'Not \`set -e\`' .github/workflows/lint.yml` — no match
- [x] "Commit, verify, and push" byte-identical to base; both `if ! verify` sites intact
- [x] Changed paths are exactly the three intended; both guard scripts unmodified; no version file touched

## Open questions

- **The least-privilege token posture forbids the one scope this PR needs.** `check-token-scope.sh --strict` exits 1 here with `forbidden_scopes: ["workflow"]` — but a PR that edits `.github/workflows/*.yml` cannot be pushed without `workflow`. This run took the settled `--allow-overscoped` opt-out to proceed. Worth deciding whether the documented posture should carve out workflow-touching work, since today every such PR must opt out of its own gate.
- **The mechanism paragraph now appears five times** across `lint.yml` (three sites), `cogni-version-bump.yml` and the suite header, and two placements already coexist in `lint.yml` alone (`:125` puts it before `set +e`, the others after). Deliberately **not** filed as a follow-up and **not** harmonized here: harmonizing means churning a block this issue did not name, and filing would re-open a scope line both #1766 and #1767 drew. Recorded here instead so the next person to touch these blocks sees it.
- **A class-wide guard remains unfiled by design**, but the exclusion's stated cost is now partly stale: `scripts/check-workflow-yaml.py` already discovers every workflow and already tracks `run: |` block bodies by indent column — purely to *skip* them. A third violation kind would reuse its discovery, envelope, CI job and suite. Its docstring does disclaim being a parser, so extending into block bodies is a real design change, not a free rider. Noted so a future scoping decision is made against current facts.

### DIFF
--- a/.github/workflows/cogni-version-bump.yml
+++ b/.github/workflows/cogni-version-bump.yml
@@ -144,7 +144,19 @@ jobs:
         env:
           DRY_RUN: ${{ inputs.dry_run }}
         run: |
+          set +e
           set -uo pipefail
+          # `set +e` above is load-bearing and must stay first. Actions runs this
+          # block as `bash -e {0}`, so -e is inherited before line 1 executes, and
+          # `set -uo pipefail` does NOT clear an inherited -e — only `set +e` does.
+          # Without it RC=1 (a hard skip, where the healthy plugins WERE written)
+          # aborts the step at the apply-version-bump command itself, so both
+          # `cat`s, the skip annotations and the `count` / `deferred_failure`
+          # outputs never happen — silently skipping BOTH downstream `if:`-gated
+          # steps.
+          # Deliberately NO terminal `exit "$RC"` here, unlike the gate blocks in
+          # lint.yml — this step must SUCCEED on RC=1 so the healthy plugins still
+          # commit; "Fail the run on a hard skip" fails the run at the very end.
           ARGS="--changed-file changed-files.txt"
           if [ "${DRY_RUN:-}" = "true" ]; then ARGS="$ARGS --dry-run"; fi
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -185,9 +185,16 @@ jobs:
 
       - name: Discover and run every plugin and repo-root test suite
         run: |
-          # Not `set -e`: the runner's exit code is the result we report on,
-          # so it must be captured rather than abort the step.
+          set +e
           set -uo pipefail
+          # `set +e` above is load-bearing and must stay first. Actions runs this
+          # block as `bash -e {0}`, so -e is inherited before line 1 executes, and
+          # `set -uo pipefail` does NOT clear an inherited -e — only `set +e` does.
+          # Without it a failing sweep aborts the step at the runner command itself,
+          # so `rc=$?`, the per-suite annotation loop and the terminal exit are all
+          # unreachable: a bare red naming no failing suite, on the one job whose
+          # whole purpose is to name them.
+          # The failing status is preserved by the terminal `exit "$rc"`, not by -e.
           python3 scripts/run-plugin-tests.py > plugin-tests.json
           rc=$?
 
--- a/tests/test_workflow_run_block_e_inheritance.sh
+++ b/tests/test_workflow_run_block_e_inheritance.sh
@@ -1,0 +1,337 @@
+#!/usr/bin/env bash
+#
+# Self-test for the two GitHub Actions `run:` blocks that must clear the
+# inherited `-e` before capturing a guard's exit code (issue #1767, follow-up
+# to #1766).
+#
+# Actions runs a `run:` block as `bash -e {0}`, so `-e` is in force before the
+# block's first line. `set -uo pipefail` does NOT clear it — only `set +e` does.
+# A block that captures a command's output and then reads `$?` therefore aborts
+# at the capturing command itself on any non-zero exit, and every diagnostic
+# below it is unreachable.
+#
+# Blocks under test, both extracted PROGRAMMATICALLY from the shipped YAML so
+# this suite cannot go green against a workflow that has since changed:
+#   .github/workflows/lint.yml            "Discover and run every plugin and repo-root test suite"
+#   .github/workflows/cogni-version-bump.yml  "Apply the patch bump"
+#
+# NOT the standing class-wide guard. This asserts the behaviour of exactly two
+# named blocks and never scans other workflows for the pattern — a repo-wide
+# guard for the class was considered during #1766's planning and deliberately
+# excluded from both #1766 and #1767.
+#
+# Cases:
+#   wfe-01  lint block extracts and its first executable line is `set +e`
+#   wfe-02  in the lint block `set +e` precedes both `set -uo pipefail` and the capture
+#   wfe-03  lint rc=0 exits 0 and emits no per-suite annotation
+#   wfe-04  lint rc=1 EMITS the per-suite `::error file=` annotation  <- the defect path
+#   wfe-05  lint rc=2 exits 2 with the annotation loop still reached
+#   wfe-06  lint FALSIFIER: `set +e` removed, rc=1, annotation ABSENT
+#   wfe-07  bump block extracts and its first executable line is `set +e`
+#   wfe-08  bump RC=0 writes count and deferred_failure=0 to $GITHUB_OUTPUT, exits 0
+#   wfe-09  bump RC=1 hard skip emits its skip annotation and writes BOTH outputs
+#   wfe-10  bump RC=1 with nothing bumped writes count=0 and says nothing to do
+#   wfe-11  bump RC=2 errors loudly, exits 1, and writes no count
+#   wfe-12  bump passes --dry-run through when DRY_RUN=true
+#   wfe-13  bump FALSIFIER: `set +e` removed, RC=1, deferred_failure ABSENT
+#   wfe-14  the bump block adds no terminal `exit "$RC"` (it must SUCCEED on RC=1)
+#   wfe-15  lint.yml carries at least three exact `set +e` lines and no stale comment
+#   wfe-16  the version-touch gate still keeps `set +e` ahead of its capture
+#   wfe-17  the retirement-ledger gate still keeps `set +e` ahead of its capture
+#
+# wfe-04 keys on the ANNOTATION, never on the exit code: the block exits 1 on a
+# failing sweep both with and without the fix (with it, via the terminal
+# `exit "$rc"`; without it, via the aborting command), so an exit-code assertion
+# there could never redden. The annotation is what the fix actually restores.
+#
+# Every case runs the extracted block under `bash -e` in its OWN mktemp -d, with
+# $GITHUB_OUTPUT pointed at a fresh per-case file. Both are load-bearing: the
+# blocks write scratch files into the cwd, and the bump block appends to
+# "$GITHUB_OUTPUT" under `set -u`, which aborts the shell on an unset variable
+# EVEN WITH `set +e` in force — so a shared or unset path would make every bump
+# case pass or fail for the wrong reason.
+#
+# Satisfies the runner's three-property contract: exits non-zero on failure and
+# zero otherwise, runs as `bash <path>` with no arguments from any working
+# directory, and needs no network or credentials, writing only inside its own
+# mktemp -d.
+#
+# Result-line ids: every emitted PASS:/FAIL: line carries a first-token id
+# (wfe-NN), unique per emitted line, followed by a SPACE and never a colon
+# abutting it, so `mutation-check.sh --case <id>` addresses exactly one
+# assertion.
+
+set -eu
+
+TESTS_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$TESTS_DIR/.." && pwd)"
+LINT_YML="$REPO_ROOT/.github/workflows/lint.yml"
+BUMP_YML="$REPO_ROOT/.github/workflows/cogni-version-bump.yml"
+
+REAL_PYTHON3="$(command -v python3)"
+
+WORK="$(mktemp -d)"
+trap 'rm -rf "$WORK"' EXIT
+
+# Plain, uncoloured result lines — unconditionally. A result line must start
+# with the literal PASS:/FAIL: so a mutation harness can classify it with
+# `^[[:space:]]*FAIL:[[:space:]]+`; an escape sequence before the label defeats
+# that match. Deciding by `[ -t 1 ]` would make parsability depend on whether
+# the suite ran under a pty.
+red()   { printf '%s\n' "$1"; }
+green() { printf '%s\n' "$1"; }
+
+FAILED=0
+check() {  # check <label> <condition-exit-code>
+  if [ "$2" -eq 0 ]; then
+    green "PASS: $1"
+  else
+    red "FAIL: $1"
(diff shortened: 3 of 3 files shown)


claude-cogni-work:
## Ship-ready

Handed off for independent review. Authored under `--role author`; this instance did not review the diff or flip it ready.

Graded against the acceptance contract committed on #1767 at `2026-09-02T22:51:56Z` — before this branch was cut, so the bar is diff-blind by chronology.

**16 items graded · 16 satisfied · 0 violated · 0 unclear.** No fix pass was needed, and there is **no residual**.

The grader independently confirmed one weakness the author flagged for it rather than asserting: **contract item 9 is vacuous as written.** It asks that `grep -c 'set +e' .github/workflows/lint.yml` be `>= 3` at head, but that loose count is already **5 at base** — three of the five are comment lines that merely mention the token — so the item could never fail regardless of whether the fix landed. It is recorded satisfied because it is satisfied, not because it discriminates. The property it was reaching for is carried instead by `wfe-15` (an exact-statement floor) and by `wfe-16`/`wfe-17`, which assert ordering per block.

### Preflight

`check-preflight.sh` exited **0, clean, zero violations — having graded nothing**:

```
check-preflight: no applicable checks for this touched set (3 file(s), no plugin) — clean
```

Every instrument is plugin-scoped and this is a repo-root diff, so none applied. Worth the reviewer's attention: the `mutation-recipe` check did not fire either, even though the diff adds a test suite — its subject pattern is `*/tests/test[-_]*.sh`, which needs a directory component before `tests/`, and this suite lives at repo-root `tests/`. A `## Mutation recipe` section is present and was replayed by hand (confirmed RED before the PR opened), but nothing automated verified it. **Treat this green as absence of evidence, not evidence of absence.**

### Two things worth a reviewer's eye

1. **The suite found a defect in itself during the mutation pass.** Its `set +e` locator was a plain substring match, and the retirement-ledger gate's comment contains the text `` `set +e` FIRST `` — so it matched the comment at line 1 rather than the statement at line 12, and reported that block correctly ordered while `set +e` sat two lines *below* its capture. It only ever worked for the other blocks because their comments happen to sit *below* the code. The locator now skips comment lines; the mutation that exposed this reddens `wfe-17`.

2. **`wfe-14` is entailed by `wfe-09`/`wfe-10` and strictly weaker.** Injecting `exit "$RC"` reddens all three, but injecting `[ "$RC" -eq 0 ] || exit 1` — which breaks the same invariant — reddens only the behavioural pair while `wfe-14` stays green. It was kept because contract item 12 explicitly asks for a grep of the step body's tail alongside the RC=1 transcript. If a reviewer would rather drop it, the behavioural cases already cover the intent and more.

### Open questions

Three are recorded in the PR body under `## Open questions` — a token-scope posture that forbids the one scope a workflow-editing PR needs, a mechanism comment now duplicated across five sites, and the deliberately-unfiled class-wide guard. None blocks review. **Nothing was filed as a follow-up issue**, deliberately: two of the three would re-open scope lines that #1766 and #1767 both drew explicitly.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1800

**Summary:** Both perspectives passed independently and all 16 committed review-contract items are graded satisfied by both reviewers with executed evidence; the single un-waived finding is a merger-added minor preference (test-breadth gap on wfe-11, not a regression) demoted by the gatekeeper-path advisory-addition pre-step, so it renders for visibility but does not gate — safe to merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (advisory-addition) `tests/test_workflow_run_block_e_inheritance.sh (case wfe-11)` — wfe-11 exercises the RC=2 path but binds only three properties — the ::error::apply-version-bump.py errored annotation, exit 1, and the absence of count= in $GITHUB_OUTPUT. It never asserts the output of `cat bump.json` or `cat bump.err`, which contract item 11 names by hand. Probed on a scratch copy: deleting `cat bump.err >&2` and deleting `cat bump.json` each leave the suite at 17/17 PASS, exit 0. The property itself is TRUE at head, so this is a guard-breadth gap on a brand-new test, not a fault in the shipped workflow; nothing at head regresses. wfe-11 is also not vacuous: removing set +e from the bump block reddens it. NON-BLOCKING; recorded so a future edit to the RC=2 diagnostics is not silently uncovered.
  - *Suggested fix:* Add `grep -q 'verification error' "$D/stdout" && grep -q 'apply-version-bump: verification error' "$D/stderr"` to the wfe-11 condition.

### Positive observations

- All 16 committed contract items graded satisfied by BOTH perspectives with executed evidence (greps at head, check-workflow-yaml.py, a full green harness run, and manual falsifier mutations).
- item 3: resolved the conflicting citations against the actual base file — the false comment is at base 188-189 as the contract cites; the issue's own AC bullet citing 118-119 was stale, not the contract.
- item 5: the true PR diff is merge-base-anchored (eeb1ce26), exactly the 3 declared files; a naive origin/main diff would have shown 6 and produced a spurious scope violation.
- item 15: set +e stripped from both blocks reddens the matching cases for the diagnostic-loss reason — the falsifier has teeth.
- The risk_watchlist's highest-value wrong fix is explicitly guarded, not merely avoided: the bump block has no terminal exit "$RC"; wfe-14 asserts its absence and injecting it reddens three cases.
- Both blocks are extracted programmatically from the shipped YAML, so the suite cannot go green against a workflow that has since changed.
- No plugin surface touched: all 8 manifests parse and mirror versions exactly; no plugin.json version bumped in the branch, which is correct — the post-merge workflow owns it.
- Full repo sweep at head: 142 suites, 142 passed, 0 failed; the new harness is discovered by the runner and runs 17/17 in 1.94s.

### Recommendation

---

Issue #1803, opened 2026-09-03
cogni-knowledge: adopting the check/grade census is itself an unread manual registration


claude-cogni-work:
## Plugin
cogni-knowledge

## Current behavior
#1758 lifted the check/grade census into `cogni-knowledge/tests/fixtures/test_helpers.sh` as `check_grade_census` and wired it into the three suites that share the convention — `test_knowledge_lib.sh`, `test_ingest_contract.sh`, `test_pdf_extract.sh`. That closes the inner loop: a `check("<tag>", …)` registration with no matching `grade <tag>` line now reddens its suite.

The **outer** loop is still open, and it is the same shape. Adopting the census is itself a manual registration that nothing reads. A fourth suite that starts using the convention inherits the pre-#1753 state silently: it carries `check("…")` registrations and `^grade` lines, calls `check_grade_census` nowhere, and stays green while an orphan hides in it. Nothing reports that.

Verified at `origin/main` after #1758's PR: the population is exactly three today — `grep -rlE 'check\("[a-z0-9_]+"'` and `grep -rlE '^grade[ \t]'` over `cogni-knowledge/tests/` return the same three files, and all three call the helper. So coverage is complete *now*; the exposure is the next adopter.

This is the same class the census itself exists to close, one level up — which is why it is worth closing the same way rather than by convention.

## Desired behavior
A suite that adopts the check/grade convention without calling `check_grade_census` is reported, not silently uncovered.

## Motivation
#1758's own motivation applies verbatim, one level out: "Both suites are balanced today, so nothing is currently hiding — the exposure is future." A guard whose population is a hand-maintained list has a recall floor nobody measures. The three-file roster is also currently restated in prose in two places (`fixtures/test_helpers.sh`'s header and `tests/README.md`), which is the dual-site drift shape the rest of this tree is careful about — a derived population would retire both restatements.

## Proposed solution
Add one case to a single suite that censuses **enrolment**: scan sibling `*/tests/*.sh` for files carrying both a `check("<tag>"` registration and a `^grade` line, and fail if any of them does not also call `check_grade_census`. That keeps the check inside a suite's own output (so it satisfies the same bar #1758's AC1 set, which a `scripts/`-tier CI-only guard cannot), and makes the guard self-enrolling instead of self-listing.

Worth deciding when picked up: which suite hosts it. `test_knowledge_lib.sh` is the incident's origin but is already the largest suite; a small dedicated suite is the alternative.

## Acceptance criteria
[...]

PR #1804, opened 2026-09-03
cogni-knowledge: share the check/grade census across three suites

### LINKED ISSUE #1758: cogni-knowledge: the check/grade census guards one of the three suites that share the convention

claude-cogni-work:
Closes #1758.

## Summary
- Lifts the #1753 check/grade census out of `cogni-knowledge/tests/test_knowledge_lib.sh` (L1594-1637 at base) into `cogni-knowledge/tests/fixtures/test_helpers.sh` as `check_grade_census FILE`, which takes the file to census as an argument and returns the verdict string on stdout.
- All three suites that share the two-part convention now run it: `test_knowledge_lib.sh` (`klib-50`, id spelling unchanged), `test_ingest_contract.sh` (`ingest-contract-125-check-grade-census`), `test_pdf_extract.sh` (`pdf-extract-06`). Each keeps its own `PASS:`/`FAIL:` arms in its own file.
- The verdict computation now exists in exactly one place under `cogni-knowledge/`. All five verdict arms (EMPTY / DUP-REGISTRATION / DUP-GRADE / ORPHAN-REGISTRATION / ORPHAN-GRADE), the `|| true` guards, the self-non-matching anchors and the `LC_ALL=C` collation pin travel with it, along with the substance of the #1753 rationale — including the bash-only paragraph, which the first draft dropped.

### Design forks closed here, deliberately not re-opened
- **Shared helper, not a repo-tier `scripts/check-check-grade-pairing.py`.** The issue floats that alternative as "worth an explicit decision". It is already foreclosed by the issue's own **AC1**, which requires the *suite* to exit non-zero and name the tag: a scanner wired into `.github/workflows/lint.yml` fails CI, never the suite. Every changed path here is under `cogni-knowledge/**` — no `scripts/**`, no `.github/**`.
- **Emission arms stay in the suite files; only the computation moves.** `scripts/check-case-id-pairing.py` `discover()` (L144-155) uses two non-recursive globs, `tests/*.sh` and `*/tests/*.sh`, and its docstring (L71-75) names `fixtures/test_helpers.sh` as "one path segment too deep".
- **Named `check_grade_census`, not `assert_check_grade_census`** as the issue proposed. It increments no `errors` and prints no label — it returns a value on stdout, which is `case_slug`'s shape (`test_helpers.sh` L27-29), not `assert_grep`'s. An `assert_` prefix would misdescribe it.
- **No caller `$WORK`, no `EXIT` handler, returns 0 on every path.** `test_ingest_contract.sh` defines no `$WORK` at all while `test_knowledge_lib.sh:66-67` and `test_pdf_extract.sh:28-29` each own a `WORK` + `EXIT`-handler pair, so the helper mints its own `mktemp -d` and cleans up inline. It installs no `EXIT` handler of its own: a second one silently *replaces* the caller's and leaks that caller's `$WORK` with every suite still green. Under `set -e`, `v=$(check_grade_census "$0")` inherits the function's status, so the body ends in an explicit `return 0`.
- **Next-unused ids, no renumbering** (`tests/README.md` rule 3). In `test_ingest_contract.sh` ids 121-124 sit **mid-file at L216-219** while the trailing grade lines at L436-439 carry 117-120 — reading only the trailing lines yields a false max of 120 and would mint a colliding 121. The true next unused id is `ingest-contract-125`; for `test_pdf_extract.sh` it is `pdf-extract-06`. Verified by a full-file id census, not by grepping grade lines.

### Changed during review: the case ids are now spelled literally in both arms

The first draft reproduced the base shape — `census_desc="klib-50 …"` plus `green "PASS: $census_desc"`. The pre-commit `/simplify` review flagged that this makes the case **invisible to `scripts/check-case-id-pairing.py`**, because an id token that is entirely an expansion is skipped. Probed both directions:

| shape | delete the green arm → | 
|---|---|
| `green "PASS: $census_desc"` (base shape) | **0 violations** — the guard cannot see the case |
| `green "PASS: pdf-extract-06 $census_desc"` | `unpaired_fail_id pdf-extract-06` |

So the stated reason for keeping the arms in the suite file was not actually being obtained. Fixed by moving the id out of `census_desc` and spelling it literally in both arms of all three cases. **The emitted lines are byte-identical to before** (verified by diffing captured output), so `mutation-check --case` still resolves and `klib-50`'s output is unchanged from #1753. This also makes the issue's AC4 evidence — `grep -n 'klib-50' test_knowledge_lib.sh` showing a green `PASS: klib-50 …` and a red `FAIL: klib-50 …` arm — pass, which it did **not** at base.

### Two claims corrected rather than relocated

The lift would otherwise have carried two false statements from #1753 into shared infrastructure that 80 suites source:

1. **The `|| true` guards are not what keeps the suite alive.** #1753's comment says an unguarded extraction "aborts the suite before this case can report". Measured: every guard sits on the **last element** of its pipeline, where `sort` and `tr` succeed on empty input, so a zero-match extraction survives `set -eu` without them — and under `set -o pipefail` too. The guards are **kept** (they are the guard a future edit needs if the grep or awk ever moves to a pipeline tail), but the header now says what they actually do.
2. **Pointing the helper at itself is not a trap.** #1758's plan warned it "would census the patterns instead of the convention". Measured: `check_grade_census fixtures/test_helpers.sh` returns `EMPTY - registrations=0 grade lines=0` — neither extractor matches its own literal, and the EMPTY arm gives the honest answer. The header now says so instead.

### Deliberately not touched
- **`grade()` is not byte-identical across the three suites**, contrary to the issue body: `test_knowledge_lib.sh:1533-1542` and `test_pdf_extract.sh:178-188` are, but `test_ingest_contract.sh:424-434` differs in its crash-message text (`(no result line — python crashed?)` vs `(no result line for '$tag' — python subprocess crashed?)`). The census never reads `grade()`, so nothing is changed there and no "byte-identical" claim is relied on. Hoisting `grade()` into the helper was raised by the review and skipped: it is a separate deduplication the issue does not ask for.
- The dead `errors_before=$errors` at `test_ingest_contract.sh:422`, immediately adjacent to the insertion point — a drive-by fix widens the diff for no acceptance-relevant reason.
- Rewriting the pipeline as a single `awk` pass was raised by the review and skipped: it would delete the `|| true` and `LC_ALL=C` surfaces the acceptance bar names, and the measured cost is ~45 ms across a full three-suite run.
- No `plugin.json` version bump — the post-merge workflow owns it.

### Grading note for AC3
`grep -rl 'check\("[a-z0-9_]+"' cogni-knowledge/` is ambiguous: read as a regex it matches every suite carrying a real registration (all three), not just the file carrying the census source. Use the **fixed-string** form — `grep -rlF 'check\("[a-z0-9_]+"' cogni-knowledge/` — which returns exactly one path before this PR (`test_knowledge_lib.sh`) and exactly one after (`fixtures/test_helpers.sh`).

### Balance at base
All three suites are balanced today and would report PASS: `test_knowledge_lib.sh` **48** registrations / 48 grade tags, `test_ingest_contract.sh` 4/4, `test_pdf_extract.sh` 4/4 — measured by running the census pipeline against the origin blobs. Nothing is currently hiding; the exposure this closes is future.

## Test plan

All run at head, all green:

- [x] `bash cogni-knowledge/tests/test_knowledge_lib.sh` — exit 0, prints `PASS: klib-50 …`, both arms in the suite file.
- [x] `bash cogni-knowledge/tests/test_ingest_contract.sh` — exit 0, prints `PASS: ingest-contract-125-check-grade-census …`.
- [x] `bash cogni-knowledge/tests/test_pdf_extract.sh` — exit 0, prints `PASS: pdf-extract-06 …`.
- [x] `python3 scripts/run-plugin-tests.py` — **142/142 suites pass, 0 failed** (guards the other 139 suites, 80 of which source `fixtures/test_helpers.sh`).
- [x] `python3 scripts/check-case-id-pairing.py` — 0 violations. Also probed positively: deleting each census green arm in turn now yields `unpaired_fail_id` naming that id.
- [x] `python3 scripts/check-result-line-plainness.py` — 0 violations.
- [x] `check-breadcrumbs`, `check-external-dispatch`, `check-code-span-nesting`, `check-readme-inventory-sync`, `check-attribution-path-integrity`, `check-mcp-tool-grant`, `check-marketplace-manifest-sync` — all exit 0.
- [x] `grep -rlF 'check\("[a-z0-9_]+"' cogni-knowledge/` returns exactly one path: `cogni-knowledge/tests/fixtures/test_helpers.sh`.
- [x] `grep -n 'trap' cogni-knowledge/tests/fixtures/test_helpers.sh` returns nothing; `grep -c 'klib-5' …/test_helpers.sh` is 0.

**Mutation matrix — 4 directions × 3 suites, all 12 confirmed.** Each mutation applied, the suite run, then restored from a file snapshot; every one exits non-zero with the census case named and the offending tag reported:

| direction | `test_ingest_contract.sh` | `test_pdf_extract.sh` | `test_knowledge_lib.sh` (non-regression) |
|---|---|---|---|
| registration with no grade line | ORPHAN-REGISTRATION `probe_tag` | ORPHAN-REGISTRATION `probe_tag` | ORPHAN-REGISTRATION `probe_tag` |
| grade line with no registration | ORPHAN-GRADE `probe_tag` | ORPHAN-GRADE `probe_tag` | ORPHAN-GRADE `probe_tag` |
| duplicate registration | DUP-REGISTRATION `is_pdf_response` | DUP-REGISTRATION `not_found_subprocess` | DUP-REGISTRATION `parse_synthesis_sources` |
| duplicate grade line | DUP-GRADE `is_pdf_response` | DUP-GRADE `not_found_subprocess` | DUP-GRADE `identity` |

## Mutation recipe

Proves the newly-covered census in `test_pdf_extract.sh` has teeth: deleting the `normalize_pdf_body` grade line orphans its registration, so `pdf-extract-06` must go RED and return GREEN on restore.

```
bash cogni-service/scripts/mutation-check.sh --root . --file cogni-knowledge/tests/test_pdf_extract.sh --expr 's/^grade normalize_pdf_body.*\n//m' --test 'bash cogni-knowledge/tests/test_pdf_extract.sh' --case pdf-extract-06
```

`mutation-check.sh` is **not vendored in insight-wave** — it ships with the managed-service `cogni-service` plugin, so substitute that plugin's `scripts/` directory for the leading `cogni-service/scripts/` above and run from the repo root. The in-repo replayable equivalent, actually run against this commit:

```
perl -0pi -e 's/^grade normalize_pdf_body.*\n//m' cogni-knowledge/tests/test_pdf_extract.sh
bash cogni-knowledge/tests/test_pdf_extract.sh   # rc=1, FAIL: pdf-extract-06 … (ORPHAN-REGISTRATION … normalize_pdf_body)
git checkout -- cogni-knowledge/tests/test_pdf_extract.sh
bash cogni-knowledge/tests/test_pdf_extract.sh   # rc=0, PASS: pdf-extract-06 …
```

Observed exactly that: mutated `rc=1` with `FAIL: pdf-extract-06 … (ORPHAN-REGISTRATION - registered but never graded, so its result is never read: normalize_pdf_body)`, restored `rc=0` with `PASS: pdf-extract-06 …`, working tree clean afterwards.

## Follow-up issues

- **#1803** — cogni-knowledge: adopting the check/grade census is itself an unread manual registration. This PR closes the inner loop (a registration nobody grades); enrolment in the census is still a hand-maintained three-file roster that nothing reads, so a fourth adopter would inherit the pre-#1753 state silently. Coverage is complete today — the population is exactly these three suites and all three call the helper — so it is future exposure, and genuinely separable from the lift.

### DIFF
--- a/cogni-knowledge/tests/README.md
+++ b/cogni-knowledge/tests/README.md
@@ -52,6 +52,18 @@ via `trap rm -rf "$WORK" EXIT`.
   `[`, `]`, `.` or `*` — a wikilink, a glob, a path. `assert_grep` reads such
   a pattern as a regex, so it can report green against a file that does not
   contain the string at all.
+- `check_grade_census <file>` censuses the two-part check/grade convention in
+  `<file>` — every `check("<tag>", ...)` registration paired with exactly one
+  `grade <tag>` line — and returns a verdict string on stdout: `PASS`, or a
+  diagnostic naming the offending tag(s). Pass the caller's own `"$0"`. It
+  emits no label and increments no `errors`, so the calling suite spells its
+  own case id **literally in both arms** and folds the verdict into `errors`
+  itself. That is what keeps the case addressable: `fixtures/` is one path
+  segment out of reach of `scripts/check-case-id-pairing.py`, and an id that
+  is entirely an expansion is skipped by it, so an arm emitted from the helper
+  — or one whose id arrives only via `$census_desc` — is invisible to the
+  guard. `test_knowledge_lib.sh`, `test_ingest_contract.sh` and
+  `test_pdf_extract.sh` are its three callers.
 - Real Python harness (inline `python3 - <<PY ... PY` heredoc) for
   script-level assertions.
 - Fixtures are minimal — only the files the test actually exercises.
--- a/cogni-knowledge/tests/fixtures/test_helpers.sh
+++ b/cogni-knowledge/tests/fixtures/test_helpers.sh
@@ -119,13 +119,15 @@ assert_not_grep_f() {
 #
 #   Five properties are load-bearing.
 #
-#   1. Every extraction keeps the `|| true` guard #1753 wrote. Measured, it is
-#      belt-and-braces rather than the thing holding the suite up: each guard
-#      sits on the LAST element of its pipeline, where `sort` and `tr` succeed
-#      on empty input, so a zero-match extraction survives `set -eu` without
-#      them — and under `set -o pipefail` too. Keep them anyway: they are the
-#      guard a future edit needs if the grep or awk ever moves to the tail of
-#      its pipeline, which is where the abort #1753 described would be real.
+#   1. Every extraction keeps the `|| true` guard #1753 wrote. Its comment said
+#      an unguarded extraction aborts the suite under `set -eu`; measured, that
+#      is not what happens here, because each guard sits on the LAST element of
+#      its pipeline, where `sort` and `tr` succeed on empty input — so grep's
+#      exit 1 on a zero-match is already discarded by the pipe. The guards earn
+#      their keep the moment `set -o pipefail` is added: measured again, the
+#      same pipeline without them DOES abort under pipefail. None of the three
+#      callers sets pipefail today. Keep the guards — they are what makes
+#      turning it on later a one-line change rather than a debugging session.
 #   2. It returns 0 on every path. Every call site is
 #      `census_verdict=$(check_grade_census "$0")`, and under `set -e` an
 #      assignment takes its status from the substitution — a body ending in a
--- a/cogni-knowledge/tests/fixtures/test_helpers.sh
+++ b/cogni-knowledge/tests/fixtures/test_helpers.sh
@@ -97,3 +97,83 @@ assert_not_grep_f() {
     green "PASS: $description"
   fi
 }
+
+# check_grade_census FILE
+#   Census the two-part check/grade convention in FILE and print a verdict
+#   string on stdout: `PASS`, or a diagnostic naming the offending tag(s). The
+#   caller owns its own case id and both result arms — this function emits no
+#   label and increments no `errors`, so it follows case_slug's return-a-value
+#   shape rather than the assert_* family's emit-and-increment one.
+#
+#   The convention has no structural enforcement of its own: a python-side
+#   registration prints "<tag>: OK"/"<tag>: FAIL" into the suite's $OUT, and
+#   only a bash-side `grade <tag> "..."` line ever reads it, so a registration
+#   with no grade line is never read and can never fail the suite. #1753 found
+#   two such orphans.
+#
+#   The census is bash-side rather than part of the python harness because the
+#   heredoc runs as a subprocess with no handle on the suite's own source, and
+#   half the subject here — the grade lines — is script text that never enters
+#   it. Callers fold the verdict into their own $errors so the rest of the
+#   report still prints; it is not fatal.
+#
+#   Five properties are load-bearing.
+#
+#   1. Every extraction keeps the `|| true` guard #1753 wrote. Measured, it is
+#      belt-and-braces rather than the thing holding the suite up: each guard
+#      sits on the LAST element of its pipeline, where `sort` and `tr` succeed
+#      on empty input, so a zero-match extraction survives `set -eu` without
+#      them — and under `set -o pipefail` too. Keep them anyway: they are the
+#      guard a future edit needs if the grep or awk ever moves to the tail of
+#      its pipeline, which is where the abort #1753 described would be real.
+#   2. It returns 0 on every path. Every call site is
+#      `census_verdict=$(check_grade_census "$0")`, and under `set -e` an
+#      assignment takes its status from the substitution — a body ending in a
+#      failing test would kill the caller instead of reporting a verdict.
+#   3. It mints its own scratch dir and installs NO `EXIT` handler of its own.
+#      test_ingest_contract.sh defines no $WORK at all, so a caller-supplied
+#      scratch dir cannot be assumed; and test_knowledge_lib.sh and
+#      test_pdf_extract.sh each already install an `EXIT` handler that removes
+#      their $WORK. A second one here would silently REPLACE the caller's,
+#      leaking that caller's $WORK with every suite still green, so nothing
+#      would report it. Clean up inline instead, as the body below does.
+#   4. Pass FILE as the caller's own `"$0"`. Pointing it at this helper is
+#      harmless — neither extractor matches its own literal, so the census
+#      comes back `EMPTY - registrations=0 grade lines=0`, which is the honest
+#      answer: this file carries no registrations to pair.
+#   5. Neither anchor may be "simplified". The registration literal spells an
+#      escaped paren, never a bare one, so it cannot match itself; and
+#      `/^grade[ \t]/` requires whitespace after `grade`, so a `grade() {`
+#      definition line is not counted as a grade line. Relaxing either turns a
+#      clean census into a phantom finding.
+#
+#   LC_ALL=C keeps sort and comm byte-collated, since the tags contain `_`.
+check_grade_census() {
+  local file scratch dup_reg dup_graded only_reg only_graded verdict
+  file="$1"
+  scratch=$(mktemp -d)
+
+  grep -oE 'check\("[a-z0-9_]+"' "$file" | sed 's/^check("//; s/"$//' | LC_ALL=C sort > "$scratch/reg" || true
+  awk '/^grade[ \t]/{print $2}' "$file" | LC_ALL=C sort > "$scratch/graded" || true
+  dup_reg=$(LC_ALL=C uniq -d "$scratch/reg" | tr '\n' ' ' || true)
+  dup_graded=$(LC_ALL=C uniq -d "$scratch/graded" | tr '\n' ' ' || true)
+  only_reg=$(LC_ALL=C comm -23 "$scratch/reg" "$scratch/graded" | tr '\n' ' ' || true)
+  only_graded=$(LC_ALL=C comm -13 "$scratch/reg" "$scratch/graded" | tr '\n' ' ' || true)
+
+  verdict="PASS"
+  if [ ! -s "$scratch/reg" ] || [ ! -s "$scratch/graded" ]; then
+    verdict="EMPTY - registrations=$(wc -l < "$scratch/reg") grade lines=$(wc -l < "$scratch/graded"); an extractor matched nothing, so the convention was renamed or the scan is broken"
+  elif [ -n "$dup_reg" ]; then
+    verdict="DUP-REGISTRATION - tag registered more than once: $dup_reg"
+  elif [ -n "$dup_graded" ]; then
+    verdict="DUP-GRADE - tag graded more than once: $dup_graded"
+  elif [ -n "$only_reg" ]; then
+    verdict="ORPHAN-REGISTRATION - registered but never graded, so its result is never read: $only_reg"
+  elif [ -n "$only_graded" ]; then
+    verdict="ORPHAN-GRADE - graded but never registered: $only_graded"
+  fi
+
+  rm -rf "$scratch"
+  printf '%s' "$verdict"
+  return 0
+}
--- a/cogni-knowledge/tests/test_ingest_contract.sh
+++ b/cogni-knowledge/tests/test_ingest_contract.sh
@@ -438,6 +438,20 @@ grade atomic_write_text "ingest-contract-118-atomic-write-text-round atomic_writ
 grade slugify           "ingest-contract-119-slugify-lower-kebab-dash slugify — lower-kebab, dash-collapse, length cap, empty-on-non-alnum"
 grade sanitize_summary  "ingest-contract-120-sanitize-summary-u-2020 sanitize_summary — U+2020 dagger / NBSP -> regular space, accents preserved (#387)"
 
+# --- ingest-contract-125: check/grade census --------------------------------
+# Shared computation in fixtures/test_helpers.sh (check_grade_census); its
+# header carries the rationale. Guards this suite against the silent-orphan
+# class #1753 found in test_knowledge_lib.sh: a python-side registration with
+# no matching bash-side grade line is never read, so it can never fail here.
+census_verdict=$(check_grade_census "$0")
+census_desc="check/grade census - every registered tag has exactly one grade line and every grade line exactly one registration; duplicates on either side and an empty extraction are rejected too, so an orphan cannot hide"
+if [ "$census_verdict" = "PASS" ]; then
+  green "PASS: ingest-contract-125-check-grade-census $census_desc"
+else
+  red "FAIL: ingest-contract-125-check-grade-census $census_desc ($census_verdict)"
+  errors=$((errors + 1))
+fi
+
 if [ $errors -eq 0 ]; then
   green ""
   green "ALL PASS"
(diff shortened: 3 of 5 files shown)


claude-cogni-work:
## Ship-ready

Author instance has handed this off for independent review. Draft, `svc:needs-review` applied. No review verdict has been formed on this branch — the merger's review is the first.

### Handoff preflight — clean

`check-preflight.sh --head-ref origin/fix/issue-1758-shared-check-grade-census` under `COGNI_PREFLIGHT_BASE_REF=origin/main`: **exit 0, `clean: true`**, metric version 4. 7/7 instruments resolved against a `min_instruments` of 7; 3 gating verdicts, 3 passed, 0 failed, **4 skipped**:

| instrument | status | detail |
|---|---|---|
| `run-tests` | pass | suites green for cogni-knowledge |
| `mutation-recipe` | pass | recorded recipe is runnable as written — `pdf-extract-06` found in `test_pdf_extract.sh` |
| `frontmatter` | skip | not_applicable — no touched SKILL.md or agents/*.md |
| `breadcrumbs` | skip | not_applicable — no touched SKILL.md or agents/*.md |
| `skill-length` | skip | not_applicable — no touched SKILL.md |
| `agent-length` | skip | not_applicable — no touched agents/*.md |

Touched set: the 5 files listed in the PR, all under `cogni-knowledge/`. The four skips are all the "prose surfaces this diff does not touch" class — worth knowing they examined nothing rather than passing.

### Author self-grade against the issue-time contract — 20/20 satisfied

Graded by `contract-grading-reviewer` against the contract `service-triage` Step 6.5 committed to issue #1758 at `2026-09-03T02:51:32Z` (20 items), before any branch existed. Verdict `pass`, **0 findings, 0 violated, 0 unclear**. Advisory only — no verdict label, no `reviewed-head-sha` marker; the merger's review is still the first independent one.

The grader executed each item's own evidence rather than reading this PR body, including all 12 mutation cells, the recorded recipe through `mutation-check.sh` itself (`verdict: guard_verified`, red→green), both repo-tier guards, and the full 142-suite run.

Three things it checked that are worth the reviewer's attention:

1. **Item 2 (census source in exactly one file).** The item's literal evidence command is malformed — it errors with `grep: parentheses not balanced` at head — and the naive regex reading matches all three suites vacuously. The grader graded the fixed-string reading (`grep -rlF`), which returns exactly one path, `fixtures/test_helpers.sh`, and confirmed the pattern is gone from `test_knowledge_lib.sh`. Flagging the ambiguity rather than quietly picking the passing reading.
2. **Item 4 (klib-50 arms).** Its evidence did **not** hold at base — the grader ran the same grep against `origin/main` and confirmed base shows only the header comment and the `census_desc` assignment, no literal arms. The diff makes it hold, and the emitted bytes are unchanged.
3. **Item 19 (mutation recipe).** The recorded `--test` value is quoted where merged PR #1800's was not. The grader confirmed the quoting is **required**, not stylistic: the unquoted form fails at `mutation-check.sh`'s parser (`unknown argument: cogni-knowledge/tests/test_pdf_extract.sh`) because the value is multi-word. Every token is still literal.

### Residual

**None.** No item was graded `violated` or `unclear`, so no fix pass was needed against the contract.

One item the grader raised as a wording imprecision inside an otherwise-satisfied item (18) **was** fixed, in commit `64049e0f`: a clause this author added during the pre-commit review claimed the `|| true` guards are redundant "and under `set -o pipefail` too". That is false for a standalone pipeline — re-measured, the same pipeline without the guards **does** abort under pipefail; the earlier measurement survived only because the call sat inside a `v=$(...)` assignment, which masks it. The helper's property 1 now states both facts and says no caller sets pipefail today. Tests, both repo-tier guards and the full 142-suite sweep were re-run green before that commit.

### Notes carried forward for the reviewer

- The issue body asserts `grade()` is "byte-identical" across the three suites. It is not: `test_ingest_contract.sh:424-434` differs from the other two in its crash-message text. Nothing in the diff depends on the claim, and `grade()` is untouched.
- The issue floats a repo-tier `scripts/check-check-grade-pairing.py` as an open design fork. It is foreclosed by the issue's own AC1 (which requires the *suite* to exit non-zero), so it was deliberately not planned.
- `#1803` was filed for the one genuinely separable finding the `/simplify` review surfaced: enrolment in the census is itself a hand-maintained roster nothing reads.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1804

**Summary:** Both plugin-dev:plugin-validator and cogni-service:contract-grading-reviewer passed, and all 20 committed review-contract items are graded satisfied by both perspectives independently (no violated/unclear grading, no INTENT-UNCLEAR finding, no split-judgment pair). The one finding raised — a minor, non-blocking prose-imprecision preference about BSD wc padding at a non-contract locus — is a merger-added preference addition; under the gatekeeper advisory-addition pre-step it is marked advisory_addition:true and excluded from Rule 5b's all-pass severity test, so with zero un-excluded un-waived findings remaining, Rule 5b resolves to pass. It does not gate the merged verdict. (Demotion independently confirmed against scripts/advisory-addition-predicate.sh: demotable=true, with a category=defect control and a contract-locus control both returning demotable=false.)

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (1)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-knowledge/tests/fixtures/test_helpers.sh:145-147` — Header claim 4 states the self-census 'comes back `EMPTY - registrations=0 grade lines=0`'. The verdict is built from `$(wc -l < "$scratch/reg")`, whose BSD/macOS form pads to a fixed width, so the observed string is `EMPTY - registrations=       0 grade lines=       0`. Accurate under GNU wc (the CI runner); platform-dependent prose, not a fault. Affects no contract item, and the padding is inherited unchanged from the base klib-50 body.
  - *Suggested fix:* Optional, not required to merge: pipe the counts through `tr -d ' '`, or soften the header to 'registrations=0' without presenting it as the literal byte sequence.

### Positive observations

- Zero plugin-load surface touched; plugin.json and its marketplace.json mirror both stay at 1.0.112/released, correctly leaving the patch bump to the post-merge workflow. (plugin-dev:plugin-validator)
- No symbol collision: check_grade_census has exactly one definition repo-wide, and all 142 discovered suites still pass. (plugin-dev:plugin-validator)
- No scratch leak: temp-dir count identical before and after a full suite run (311/311), and the helper installs no EXIT trap, so neither caller's own trap is replaced. (plugin-dev:plugin-validator)
- item 2: ran the contract's own evidence command verbatim; it errors as written, and only the fixed-string reading matches the item's own base note (exactly test_knowledge_lib.sh at base, exactly fixtures/test_helpers.sh at head). (cogni-service:contract-grading-reviewer)
- items 12/13/14: all 13 mutations applied and run independently, each rc=1 naming the correct verdict arm and tag, every file restored (git status clean). (cogni-service:contract-grading-reviewer)
- item 19: the recorded literal-flag mutation recipe run verbatim through mutation-check.sh returns guard_verified (mutated red, restored green). (cogni-service:contract-grading-reviewer)

### Recommendation

---

PR #1812, opened 2026-09-03
Retirement-ledger gate: cover phrases dropped from modified skills

### LINKED ISSUE #1762: cogni-workspace: the retirement-ledger gate does not cover a phrase dropped from a MODIFIED surviving skill description

claude-cogni-work:
Closes #1762.

## Summary

`scripts/check-retirement-ledger.py` gains a second arm over **modified** `cogni-workspace/skills/<name>/SKILL.md` files. A phrase that file advertised at the merge base, claimed by no surviving skill and carried by no ledger row with a non-empty `reason`, is now a `description-phrase-dropped` violation (exit 1). Until now the guard scoped its subject with `--diff-filter=D`, so a phrase could leave the claim surface with no file deleted — an author edits a surviving skill's `description:` and drops a quoted phrase — and nothing noticed: C10 is blind to it for the same reason it is blind to the deletion case, and the deletion scan never looked at modified files.

**One guard, not a sibling.** Everything the second subject needs the first already establishes: the base-ref resolvability check, the merge base, the `degrade()` arm that makes an unresolvable baseline exit 0, and an `accounted` set that was always subject-agnostic. A sibling guard would have cost a second CI job replicating the `status: degraded` re-grade, a second full-history checkout, a second `CLAUDE.md` registry bullet, and a **third** copy of the phrase extractor — and C11 pins exactly one, so an unpinned third would drift and fail **open**, in precisely the direction this guard exists to close.

**What changed in the guard**

- One `git diff --name-status --no-renames --diff-filter=DM <base>...HEAD` query feeds both subjects, so `--no-renames` — already load-bearing for the archive-move case pinned by case 13 — cannot drift between the two arms. Lines partition on the first tab; only `R` and `C` carry a similarity-score suffix and the filter plus `--no-renames` exclude both, so the split is total. Every path still goes through the **unchanged** `SKILL_PATH_RE`.
- The `if not deleted: return 0` early return now fires only when **both** lists are empty. This was the structural blocker: a modification-only branch exited there, so no arm could simply be appended after the deletion loop.
- The ledger-missing arm is gated on `deleted`. A deletion has exactly one remedy and it is a ledger row; a modification has two that need no ledger at all, so a modification-only branch falls through to the per-phrase finding rather than firing a violation whose own detail text ("branch deletes a skill…") would be false. `accounted` correspondingly tolerates an absent ledger — reading it unconditionally raised into the top-level handler and exited 2.
- **No per-file head-side read.** The modified file is itself a *surviving* skill, so `live_phrase_keys` over the working tree already carries whatever its description still advertises; `key not in accounted` is the whole test. A `git show HEAD:<path>` read would have re-derived a subset of the live set at the cost of a second notion of HEAD beside the working-tree reads `live_phrase_keys` and `ledger_phrase_keys` both take.
- A file **unparseable at HEAD** therefore needs no arm of its own: it contributes nothing to the live set, so everything it advertised at the merge base is demanded — the safe direction. Whether a live file parses stays `check-frontmatter`'s subject.
- A file **unparseable at the merge base** is a `modified-base-unparseable` observation, the mirror of `base-blob-unparseable` and for the same two reasons: it advertised nothing to lose, and no edit on this branch can change its own merge base. Like its sibling it gets no `FIX_HINTS` arm — there is no executable remedy to name.
- The envelope gains `modified_skills[]` beside the untouched `deleted_skills[]`, following the same null-means-never-computed rule. `count` and `clean` stay derived inside `render` from `violations` alone.
- A latent defect is corrected: `checked = len(deleted) - len(observations)` was right only while one observation class existed and under-reports the moment a second contributes. Each subject now subtracts only its own class. Case 22 is the case that observes it.

## Scope decisions

**Out of scope, deliberately.** `.github/workflows/lint.yml` is untouched — its step selects no arm, greps no check name, and its degraded re-grade is generic, so an arm folded into the same script and the same `violations[]` / `observations[]` shape needs no workflow edit. `cogni-workspace/tests/test-skill-trigger-phrases.sh` is untouched — `phrases_from_skill_text`, `iter_skills` and `emit_phrases` are all unchanged, so C10 and C11 hold. `METRIC_VERSION` stays `"1"`: the per-phrase accounting rule did not change. No `plugin.json` / `marketplace.json` version touch — the bump is post-merge.

**Prose corrected on the same commit**, because each became a false statement the moment the arm landed: the module docstring's opener and its `SCOPED TO DELETIONS` section (replaced by a two-subjects-one-guard rationale naming the rejected sibling), the suite header invariant, the `CLAUDE.md:198` registry bullet, and the ledger's own header comment. The `CLAUDE.md` edit is minimal by that bullet's own instruction not to re-derive the guard's internal rules; in particular the quoted CI job name is left byte-identical so its pairing with `lint.yml` stays honest.

## Test plan

- [x] `bash tests/test_check_retirement_ledger.sh` → `all cases passed`, 23 PASS lines, 0 FAIL
- [x] `bash cogni-workspace/tests/test-skill-trigger-phrases.sh` → all pass, including `C10` and `C11`
- [x] `python3 scripts/check-retirement-ledger.py` on this branch → exit 0, and the stderr summary names both subjects
- [x] The recorded mutation recipe below replays `mutated_result: red` / `restored_result: green` / `verdict: guard_verified`
- [x] All 15 pre-existing cases green with their expected strings unchanged

Every new expected tuple was **measured** against a patched copy of the guard during planning, not derived on paper — and two paper-derived claims in the first plan were falsified that way: case 21's count was 3 (it is 2 — beta advertises one phrase and alpha drops one), and case 4 was cited as pinning the modified side's path scoping (its `README.md` is an **add**, invisible to the filter; cases 2 / 7 / 8 are the real transitive pin, since each overwrites the ledger and so contributes a genuine `M` row that `SKILL_PATH_RE` must reject).

| case | asserted `rc\|status\|violations\|count\|observations` |
|---|---|
| 16-description-drop-unaccounted | `1\|ok\|description-phrase-dropped\|1\|EMPTY` |
| 17-description-reword-clean | `0\|ok\|EMPTY\|0\|EMPTY` |
| 18-description-drop-accounted (both arms) | `0\|ok\|EMPTY\|0\|EMPTY` |
| 19-modified-head-unparseable | `1\|ok\|description-phrase-dropped\|2\|EMPTY` |
| 20-modified-merge-base-anchoring | `0\|ok\|EMPTY\|0\|EMPTY` |
| 21-mixed-subjects | `1\|ok\|description-phrase-dropped,retirement-row-missing\|2\|EMPTY` |
| 22-modified-base-unparseable | `0\|ok\|EMPTY\|0\|modified-base-unparseable` |
| 23-modified-no-ledger | `1\|ok\|description-phrase-dropped\|1\|EMPTY` |

Cases 22 and 23 cover behaviour this PR itself introduces and would otherwise ship untested: the new observation class (and the corrected `checked` arithmetic), and the modification-only-with-no-ledger path that dies as an unhandled-error exit 2 without the two gates above.

## Mutation recipe

```
scripts/mutation-check.sh --root . \
  --file scripts/check-retirement-ledger.py \
  --expr 's/"--diff-filter=DM"/"--diff-filter=D"/' \
  --test 'bash tests/test_check_retirement_ledger.sh' \
  --case retirement-ledger-16-description-drop-unaccounted
```

The mutation narrows the one diff query back to deletions, reintroducing exactly the regression this issue reports: case 16's branch deletes nothing, so the status split yields two empty lists, the both-empty early return fires, and the tuple moves from `1|ok|description-phrase-dropped|1|EMPTY` to `0|ok|EMPTY|0|EMPTY`. The anchor is the **quoted argument-list form** and occurs exactly once in the guard — every prose mention of the filter, here and in both files, is backticked rather than wrapped in ASCII double quotes, which is what keeps that uniqueness true. The mutant is valid Python, so only the subject moves rather than reddening every case and grading nothing. Replayed on this branch: `guard_verified`.

## Open questions

None blocking. One process note for the reviewer, recorded rather than hidden: the Step 6.4 `/simplify` quality pass was **not** run on this diff — the run was managing a constrained context budget and prioritised committing and pushing durable state over the pass. The correctness gates (both suites, the real-tree run, the mutation replay) all ran and are green; what was skipped is a quality-only pass, so treat the diff's factoring as un-reviewed by that instrument. The one place it would most plausibly have something to say is the deliberate parallelism between the deletion and modification loops, which share a blob-read shape that was left duplicated on purpose, with cross-reference comments, rather than factored into a helper.


claude-cogni-work:
## Ship-ready

Author instance is done with #1762. This PR is handed off for independent review under the two-identity topology — it is deliberately still draft, carries no review verdict, and no `svc:review-*` label. The merger forms its own contract.

### Author self-grade (advisory, against the issue-time contract)

Graded **23 items: 22 satisfied, 1 violated, 0 unclear**.

The one violation was **item 13** — its purpose clause demands coverage for a *normalization-only* edit, and no fixture changed a retained phrase's case or internal whitespace. An arm comparing raw phrase text instead of the extractor's normal form would have shipped untested. Fixed in the bounded pass as `retirement-ledger-24-description-normalize-clean` (`Alpha  ONE` → same key `alpha one`, asserts `0|ok|EMPTY|0|EMPTY`), inventory entry added, suite re-run to 24 PASS / 0 FAIL, committed as `887e118d` and pushed before this handoff.

**Residual: none.** No unclear items, and the single violation landed.

### Preflight

`check-preflight.sh` exit 0, clean — but **2 passed, 5 skipped**, which the merger should see rather than read as full coverage:

- `version-bump` (v2) — pass, no version-line touch
- `mutation-recipe` — pass, recipe runnable as written, case id found in the suite
- `run-tests` — **skipped, `not_applicable`**: this diff's two code files are repo-root `scripts/` and `tests/` paths, which resolve to no plugin, so the plugin-scoped instrument never saw them. Both suites were run directly instead — `tests/test_check_retirement_ledger.sh` → 24 PASS / 0 FAIL, `cogni-workspace/tests/test-skill-trigger-phrases.sh` → all pass including C10 and C11 — and the mutation recipe replayed `guard_verified` (mutated red, restored green).
- `frontmatter`, `breadcrumbs`, `skill-length`, `agent-length` — skipped, `not_applicable`: no `SKILL.md` or `agents/*.md` in the touched set.

### Known deviation

The Step 6.4 `/simplify` quality pass was **not** run on this diff — the run was managing a constrained context budget and prioritised landing durable state. Every correctness gate ran and is green; what is un-run is a quality-only instrument. Also recorded in the PR body's `## Open questions`. The most likely thing it would have flagged is the deliberate duplication between the deletion and modification loops, which share a blob-read shape left parallel on purpose with cross-reference comments rather than factored into a helper.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1812

**Summary:** All three perspectives passed and 23/23 committed contract items graded satisfied by both contract-grading perspectives; the only un-demoted, un-waived finding is a minor PR-size advisory, so Rule 5b (all-pass, only minor un-waived findings remain) applies.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (3)

- **[minor]** (auto-fixable) (advisory-addition) `PR body — ## Test plan` — The test plan asserts '23 PASS lines' and its per-case table stops at case 23, but the suite emits 24: commit 887e118d added retirement-ledger-24-description-normalize-clean after the body was written. The SHIPPED documentation is correct (the suite header inventory does carry case 24), so this is a stale PR-body claim only, with no artifact defect.
  - *Suggested fix:* Refresh the PR body's count to 24 and add case 24 to the expected-tuple table.

- **[minor]** `PR diff` — Diff is 409 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (advisory-addition) `scripts/check-retirement-ledger.py:412` — On a branch that BOTH deletes a skill AND drops a phrase from a surviving one while the ledger is missing, the `if deleted and not ledger_path.is_file():` arm returns 1 before the modification loop runs, so the envelope reports modified_skills: [] for a run whose diff contained a modified SKILL.md. This FAILS CLOSED (exit 1, no fail-open) and the gating is deliberate and documented inline, so item 4's purpose holds; only its sub-clause wording is literally deviated from on this one path.
  - *Suggested fix:* Optional: append the modified paths to modified_skills[] before that return so the envelope names the whole subject the run saw.

### Positive observations

- The latent `checked = len(deleted) - len(observations)` defect is fixed on the same commit rather than reproduced — a bug the diff would have INTRODUCED, found by the author. (plugin-dev:plugin-validator)
- Both pre-existing mutation recipes still return guard_verified, so widening did not blunt the deletion arm. (plugin-dev:plugin-validator)
- The no-head-side-read decision is argued at BOTH sites, closing the two-notions-of-HEAD risk the contract watchlist named. (plugin-dev:plugin-validator)
- The suite header explicitly RETRACTS a falsified paper claim (case 4's README.md is an add, invisible to the filter) rather than hiding it. (plugin-dev:plugin-validator)
- 23/23 contract items satisfied; executable items (16, 23, 10, 2) were replayed, not reasoned about. (cogni-service:contract-grading-reviewer)

### Recommendation

---

PR #1819, opened 2026-09-03
cogni-workspace: promote the user-facing output register to canonical

sdh07:
Follows #1816. Rebased onto main after that merged — not stacked.

## Why

`cogni-consult/references/user-facing-output.md` carried **7 of 9 sections of general ecosystem doctrine** from inside a vertical plugin: language and orthography (a), governed surfaces (b), the table contract (d), step announcements and brevity budgets (e), tool-call description copy (f), the anglicism test (g), and the executive register (h). Only the state-lexicon rows, the coinage table and the `{{asm:}}` rule are actually cogni-consult's.

Five other plugins have no output doctrine at all. The best register in the ecosystem was reachable by exactly one of them.

## What changed

Split on that seam. `cogni-workspace/references/user-facing-output.md` is canonical; cogni-consult's drops to an overlay carrying its own lexicon and coinages. Section letters match across both so a rule and its overlay line up.

The design mirrors **markets** — the shared-vs-overlay pattern cogni-workspace already uses: canonical in the horizontal layer, overlays carrying only plugin-specific rows.

Section (c) splits on a real seam rather than arbitrarily: the *rule* ("an engine value is never displayed raw; a missing value is a gap in the table") is general and goes canonical, while the lookup rows stay with the plugin whose engine writes them. Canonical (c) also names the open-vs-closed lexicon distinction, because conflating them is what licenses invented terms — cogni-consult's disposition triple is the worked closed case.

## Two duplications preserved deliberately

Both are load-bearing and documented as such. Neither should be "deduplicated" by a later cleanup:

- **Section (h) stays inline in the canonical file** rather than delegated, because it must reach the main loop. The agent-side contract keeps its own copy because it is *injected*, never loaded — a pointer would leave the main loop with no reachable copy.
- **cogni-consult keeps section (f) inline with its worked pair.** A description can be written wrongly without ever opening a reference, so the constraints have to sit where the model always reads them. Its own `test_step0_register_block.sh` asserts exactly that, and the three needles it pins are intact.

## The cross-plugin read — the riskiest decision here

This is the repo's **first cross-plugin file read** and is scoped as the only sanctioned one.

`$CLAUDE_PLUGIN_ROOT` resolves to the *owning* plugin, so it cannot reach the canonical file from a vertical plugin. `scripts/generate-settings.sh` emits `COGNI_WORKSPACE_PLUGIN` for every discovered plugin, and `docs/ecosystem-overview.md` already requires `manage-workspace` to have run before any plugin is used.

**The read is fail-soft by contract.** An overlay that cannot resolve the variable says so once, applies itself alone, and carries on — matching the convention already used for optional Python deps and theme fallback. A missing canonical register degrades the doctrine; it never blocks the work.

The pattern is recorded in `cogni-workspace/CLAUDE.md` so the next plugin does not reinvent it.

## Styles do not centralize — only the register does

`cogni-workspace` gains the general **Workspace Advisor** stance, generalized from `strategy-advisor.md`'s architecture (stance in the style, register in the reference) minus the two consulting-bound elements: the "senior strategy consultant" identity and the `{{asm:}}` bullet.

**Strategy Advisor stays** and now points at both halves. Migrating it into cogni-workspace was considered and rejected: `$CLAUDE_PLUGIN_ROOT` would orphan the 16 KB register it points at, and styles are session-global and cannot compose or inherit — so two picks serve two audiences. A consultant selects Strategy Advisor; a multi-plugin user selects Workspace Advisor.

## Verification

- 144/144 suites green (`scripts/run-plugin-tests.py`) and all 14 root guards clean, re-verified on the post-repair head `d8fb192a` in a clean detached worktree, plus 15/15 checks green in CI. Two suites — `cogni-consult/tests/test_resolve_assumptions.sh` and `tests/test_release_bundle_wiki.sh` — are **environment-sensitive**: a review pass saw them fail on its machine, and they failed identically on a fresh `main` worktree there, so they are not bound to this change either way.
- cogni-consult's three register-bound suites pass specifically: `test_step0_register_block.sh`, `test_generate_dashboard.sh`, `test_generate_engagement_readme.sh`.
- `orthography-drift-scan.py` still runs clean against cogni-consult, though the rule it enforces now lives canonical.
- **Dangling section citations.** The first commit repaired only `consult-personas` and `consult-design-thinking`; a review of that head found roughly eleven more still pointing at sections the split had moved, and no guard in the repo resolved a cited section *letter*, so everything stayed green over the breakage. The second commit (folding #1820) repairs the rest — the nine Step 0 blocks, `engagement-list-rendering.md`, `subagent-output-contract.md`, `interaction-language.md`, `orthography-drift-scan.py`, `README.md` and `CLAUDE.md` — and closes the gap itself: `test_reference_pointers.sh` gains case `sections-resolve` (16 citations resolved) with a red-proving arm, and `test_step0_register_block.sh` now reads the canonical file rather than the overlay alone.

## Follow-up, not in this PR

The canonical register currently has one consumer. Wiring the rest is separate work, and the survey says it is **not uniform**: cogni-knowledge, cogni-portfolio and cogni-trends have genuine lexicons worth an overlay; cogni-marketing, cogni-sales and cogni-website surface almost no stored state and would get near-empty files that drift without earning their keep. Those three want the load line and no overlay.

One open question blocks the cogni-portfolio overlay specifically: its review-verdict enum is not pinned. `pass`/`warn`/`fail` dominate the assessor agents, `revise` appears in their JSON examples, and the repo `CLAUDE.md` documents the gate as `accept`/`warn`/`fail`. Resolve it before writing that lexicon — a row for a value the engine never emits is the invented-term failure the register itself forbids.

No version bumps — the post-merge workflow owns them.


cogni-consult/references/user-facing-output.md
```diff
@@ -1,8 +1,27 @@
...
+consultant. This file is an **overlay**: the ecosystem register is canonical and
+carries the shared doctrine — language and orthography (a), governed surfaces
+(b), the never-display-a-raw-value rule (c), the table contract (d), step
+announcements and brevity budgets (e), tool-call descriptions (f), the anglicism
+test (g), and the executive register (h).
+
+Read the canonical file first:
+
+```bash
+cat "${COGNI_WORKSPACE_PLUGIN}/references/user-facing-output.md"
```
sdh07: **This bare read has no fallback, and the repo already ships the hardened form.**

`cogni-consult/scripts/discover-projects.sh:10` reaches cogni-workspace with a three-layer ladder — `${WORKSPACE_PLUGIN_ROOT:-...}` → newest dir in the plugin cache → monorepo sibling → existence test on the target file → loud actionable error. `scripts/get-market-config.py::_resolve_sibling_plugin` does the same, with the comment "a candidate wins only if the overlay file exists under it — that way half-installed or empty dirs don't win". Ten files across four plugins use that variable.

This introduces a second, fallback-free convention for the same target, under a different variable name. Two concrete failure modes it cannot survive:

- **Empty expansion reads an absolute path.** Bash without `set -u` does not error on an unset variable inside a quoted expansion, so this becomes `cat /references/user-facing-output.md`. On a host that happens to carry that path, the model silently ingests unrelated content *as the ecosystem register*.
- **A stale value still resolves.** `COGNI_WORKSPACE_PLUGIN` is an absolute path into the version-scoped plugin cache captured at `manage-workspace` time. Because this repo patch-bumps every touched plugin post-merge, the next cogni-workspace update creates a new version dir — and the variable still *resolves*, so the fail-soft branch keyed on "does not resolve" never fires.

Fix: resolve through the same ladder, and test for the **file** rather than the variable.

---

cogni-consult/references/user-facing-output.md
```diff
@@ -1,8 +1,27 @@
...
+```
+
+**If `$COGNI_WORKSPACE_PLUGIN` does not resolve**, cogni-workspace is not
```
sdh07: **This diagnosis is unsound, and the remedy reproduces the state it diagnoses.**

"cogni-workspace is not registered in this workspace" is inferred from a read failure that cannot distinguish its causes. Verified by running the command: unset and empty both produce `cat: /references/user-facing-output.md: No such file or directory`, exit 1 — byte-identical to each other and shape-identical to a stale path. At least four states produce this failure while cogni-workspace *is* registered:

1. A repo checkout during development — `scripts/check-output-style-placement.py`'s own docstring notes this variable "resolves from generated workspace settings that do not exist in a repo checkout".
2. `generate-settings.sh --update` with cogni-workspace deselected at the confirm step **prunes** `COGNI_WORKSPACE_PLUGIN` while `COGNI_WORKSPACE_ROOT` survives beside it (arm 1 spares ROOT because it is assigned unconditionally; PLUGIN has no such protection).
3. A plugin list supplied as bare name strings — `generate-settings.sh` writes `env[plugin_var]` only `if path:`, so zero `_PLUGIN` keys are written. `manage-workspace/SKILL.md:88` explicitly permits a hand-edited JSON list after the confirm step.
4. An installed cogni-workspace older than this overlay: the variable resolves to a real directory with no `references/user-facing-output.md` in it.

In 1, 3 and 4 the model states a falsehood to the consultant and sends them to `/cogni-workspace:manage-workspace`, which changes nothing. In 2 it sends them to the command that caused it.

Fix: report *what could not be read*, not why — and repair the prune in `generate-settings.sh` separately, since the register read now depends on that key.

---

cogni-consult/references/user-facing-output.md
```diff
@@ -88,90 +88,15 @@ Four notes travel with the table:
...
-Main-loop-only, deliberately: a dispatched agent's user-facing surface is the
-envelope it returns, not its own tool calls, so
-`references/subagent-output-contract.md` carries no counterpart to this section
-and none is owed.
-
-Five constraints, all binding:
+The canonical file owns the five constraints. They are restated here, and inline
+in every `consult-*` skill's Step 0, because a description can be written —
+wrongly — without ever opening any reference, so the constraints have to sit
+where the model always reads them. Edit them in the canonical file first, then
```
sdh07, line 139: **The guard named here cannot see the tier this sentence makes edit-first.**

`tests/test_step0_register_block.sh:64` sets `OWNER="$PLUGIN_DIR/references/user-facing-output.md"` — *this* file. Its byte-identity arms compare the nine Step 0 blocks to each other; its `owner-present-*` arms grep three needles against this file. Nothing in the suite opens cogni-workspace. I ran it at this head: 19/19 pass.

So the arrangement is now three tiers — canonical, this overlay, nine Step 0 blocks — and the one tier this paragraph names as edit-first ("Edit them in the canonical file first") is the only tier no assertion reads. A maintainer who follows the stated remedy gets a green CI run while eleven downstream copies still carry the old wording: exactly the silent drift the guard was written to make impossible.

Worth noting the two copies already differ — constraints 2 and 5 read "the consultant" here and "the user" in canonical — so a naive byte-equality arm cannot simply be bolted on; it would redden a tree that is behaving correctly. A comparison that folds the audience noun works, and `#1820` adds it as `canonical-basis` / `canonical-constraints` with a `goes-red-canonical` arm proving it discriminates.

Note also that all nine Step 0 blocks say the opposite of this paragraph: "Section (f) of `$CLAUDE_PLUGIN_ROOT/references/user-facing-output.md` owns these five constraints — **edit them there**". Two maintainers following the two instructions edit two different files.

---

cogni-consult/references/user-facing-output.md
```diff
@@ -235,27 +145,20 @@ worked exclusion from the list just above. The English table templates keep
...
-(c), or describe what happened in plain language.
+plugin has never written it.
 
-This table is owned here rather than borrowed. the `copywriter` skill's German-style
-table addresses sales and go-to-market vocabulary and shares no term with the
-list above; it transliterates its umlauts to ASCII, which contradicts the
-orthography this plugin mandates; and no load path reaches it from here. The
+This table is owned here rather than borrowed. The `copywriter` skill's
+German-style table addresses sales and go-to-market vocabulary and shares no term
+with the list above; it transliterates its umlauts to ASCII, which contradicts
```
sdh07: **This justification is factually wrong, and it is recorded where a future maintainer will rely on it.**

"it transliterates its umlauts to ASCII, which contradicts the orthography this plugin mandates" — that file uses correct umlauts throughout (`überarbeitet`, `Wörter`, `makroökonomischen`, `Maßnahmen`), and the `copywriter` skill forbids ASCII substitution outright:

- `template-email.md:186` and `template-brief.md:236` both check "German characters preserved exactly (ae/oe/ue/ss never substituted)"
- `translation-en-to-de.md:45` rejects output containing ASCII substitutes

So orthography is not a point of difference between the two files. The other two reasons given (different vocabulary domain, no load path) are true and sufficient on their own — this one should come out, because someone later weighing whether to consolidate will be deciding on a false premise.

Two related problems in the same section:

- **"The method above is worth citing"** — the three-step method it points at moved to canonical, so the deictic resolves to nothing in this file, and the paragraph now argues against importing a method the file no longer states.
- **The attribution is gone repo-wide.** The pre-PR text read "the method comes from the `copywriter` skill's German-style principles, cited here rather than loaded, because nothing in this plugin loads that file" — a cross-plugin argument. Canonical (g) restates the test with no attribution, and `cogni-workspace/skills/copywriter/references/01-core-principles/german-style-principles.md:299-303` now sits in the **same plugin** as an unlinked near-verbatim second copy, with nothing binding them.

---

cogni-workspace/references/user-facing-output.md
```diff
@@ -0,0 +1,214 @@
...
+
+This is the horizontal layer's copy of the doctrine, so a vertical plugin does
+not have to invent one. A plugin **overlays** it rather than restating it —
+the overlay carries only what is genuinely its own: its state lexicon rows, its
+coinage vocabulary, and any surface its own guards pin. Read this file first,
+then the overlay.
+
+**This file governs the main loop only.** A dispatched agent never sees it — a
+subagent's prompt carries no `CLAUDE.md`, no settings language, no output style
+and no skill body. A plugin whose agents produce user-facing prose owes them the
```
sdh07, line 15: **The plugin promulgating this obligation is the largest violator of it.**

This file declares that "a plugin whose agents produce user-facing prose owes them the same doctrine through its own subagent contract, injected by a `SubagentStart` hook".

`ls cogni-workspace/agents/` lists 26 agents that write user-facing prose — `copywriter`, `narrative-writer`, `report-html-writer`, `enrich-report`, `reader`, `claim-verifier`, the render-infographic agents. `cogni-workspace/hooks/hooks.json` registers only `SessionStart` and `PreToolUse`. There is no `SubagentStart` hook and no subagent contract file anywhere in `cogni-workspace/references/`.

So every cogni-workspace fan-out emits agent prose with no language, no register and no engine-vocabulary rule, while the file that states the rule sits one directory away, unreachable from the agent path. `cogni-workspace/CLAUDE.md:66` and `:81` repeat the obligation and point at cogni-consult as the model, which makes the horizontal plugin the one advertising a contract it does not implement.

Not a reason to weaken the doctrine — it is correct. But the gap should be *recorded* rather than left as an implicit claim of coverage, so nobody reads owning the register as having closed it. `#1820` states it in `cogni-workspace/CLAUDE.md` as a known gap.

---

cogni-workspace/references/user-facing-output.md
```diff
@@ -0,0 +1,214 @@
...
+   either of the first two. This does not govern a boolean marker column (a
+   glyph when true, blank when false), which is its own established idiom. And
+   rule 5 never licenses dropping a column to hide a state the reader needs
+   to see.
+
+## (e) Step announcements and brevity budgets
+
+Two tiers, split by irreversibility rather than by skill:
+
+- **Read-only step**: one sentence, at most 25 words.
```
sdh07, line 105: **Generalizing (e) deleted German vocabulary from the repo rather than overlaying it.**

`Orientierungsschritt` and `Arbeitsschritt` named these two tiers on `main` (`cogni-consult/references/user-facing-output.md:120-121`). At this head, `git grep -n 'Orientierungsschritt\|Arbeitsschritt'` over the tracked tree returns **nothing** — canonical replaced them with "Read-only step" / "Work step", and the overlay carries no (e) at all to re-claim them.

This PR's own rule (`cogni-workspace/CLAUDE.md:83`) is that an overlay carries "its coinage vocabulary". These are exactly that, and they were dropped rather than moved. A German consult session announcing a step now has no repo-sanctioned German name for either tier, so the model coins one per session — and (g)'s anglicism table cannot help, because neither term is in it.

The same happened to the worked examples in (e) and (h), each verified absent repo-wide:

| Lost | Became |
|---|---|
| "Partner, Projektleitung, CFO und Betriebsrat, jeder in seiner eigenen Logik" / "vier Agenten" | "Name the stakeholders… not 'four agents'" |
| "Teillösung 2 (die freie Content-Schicht)" / "bei 2" | "option 2 (the open content layer)" / "at 2" |
| "drei Deliverables ruhen jetzt auf einer überholten Zahl" / "das Cascade hat drei Knoten geflaggt" | English equivalent |
| "im Protokoll festgehalten" / "d-084 geschrieben" | English equivalent |

English exemplars are right for a language-neutral canonical file. But cogni-consult is German-first — its whole (g) table and the German column of (c) are German — and its overlay has no (e) or (h) to hold replacements, so the surface these examples were written to calibrate now has only English ones. This is content loss rather than a dangling pointer, and no guard covers it.

---

cogni-workspace/references/user-facing-output.md
```diff
@@ -0,0 +1,214 @@
...
+
+Resolve the **interaction language** the way the calling plugin resolves it. In
+the absence of a plugin-specific ladder, the workspace `language` key in
+`.claude/settings.local.json` is the default and Claude Code turns it into a
+`# Language` system-prompt section. A plugin that distinguishes the interaction
+language from a stored output language owns that separation in its own
+reference — do not restate a ladder here. Everything below applies in whatever
+language resolves.
+
+**German orthography.** Write ß after a long vowel or a diphthong (Maßnahme,
```
sdh07, line 31: **Third unbound copy of this rule, and it contradicts `cogni-workspace/CLAUDE.md:61` in the same commit.**

Two copies already live in this plugin:

- `hooks/on-session-start-language.sh` `RULES["de"]` injects "ß nach langem Vokal und Diphthong (Maßnahme, außerhalb, Größe), ss nach kurzem Vokal (dass, muss, Prozess). Kein schweizerisches ss an ß-Stellen." into **every** German session as `additionalContext`, with no file to load. That is the only path reaching a session carrying no skill and no overlay — i.e. the one that actually always arrives.
- `skills/copywriter/references/01-core-principles/translation-en-to-de.md:32-45` owns the fuller treatment with a `Massnahme→Maßnahme` / `Strasse→Straße` substitution table.

Lines 31-35 restate it a third time, using the same three example triples as the hook. And `cogni-workspace/CLAUDE.md:61` — unchanged by this PR — still reads "What the built-in section does **not** carry travels in `hooks/on-session-start-language.sh`: German orthography (umlauts, ß/ss)", while the new line 81 claims this file is canonical for "language and orthography". Two owners named 20 lines apart in one file.

The hook's own header comment carefully documents which rules belong where; this paragraph was written without reference to it, so the three can drift with nothing failing. Fix by naming the other two and the rule that binds them, so a change to any one is checked against the others — the hook being the copy that must never be thinner, since it is the one that always reaches the session.

---

cogni-workspace/references/user-facing-output.md
```diff
@@ -0,0 +1,214 @@
...
+2. Is the anglicism established with no good German equivalent (*Meeting*,
+   *Software*, *Update*)? If yes, keep it.
+3. Is it a hyphenated English compound? Those almost always have a German
+   rendering — replace them.
+
+Step 3 is what a plugin's own coinage table resolves to. That table belongs in
+the overlay, not here: a coinage is by definition the plugin's own, and a shared
+table would collect terms most plugins never write.
+
+**Spell an acronym out once at first use** — ICP, OMTM, UVP — then use it freely
```
sdh07, line 199: **This blanket rule contradicts the audience-tuned discipline the same plugin already owns.**

`cogni-workspace/skills/copywriter/references/01-core-principles/acronym-handling-principles.md` exists, in this plugin, and its entire subject is that expansion depth resolves from `AUDIENCE` (`expert` / `mixed` / `lay`) and that a named set is **never** expanded at all — `IT`, `EU`, `USD`, `PDF`, regulation proper nouns (`NIS2`, `DSGVO`, `BSI`), brand names, and the `IS`/`DOES`/`MEANS` arc markers.

A model reading this line will expand tokens that reference explicitly forbids expanding, and the two files give opposite answers with no cross-reference between them. Fix by making this the stated default where no audience is resolved, and deferring to that reference where one is.

**Two more in this section and the next:**

- **Line 165, the three-step test, is unqualified German in a file declared canonical for the whole ecosystem.** Line 28 says "Everything below applies in whatever language resolves", then (g) instructs "Does an equally precise German word exist?". Unlike (a), which scopes itself with a bold "**German orthography.**" heading, (g) carries no caveat. `generate-settings.sh` maps seven languages and the repo supports ten markets, so a French, Polish or Spanish workspace is handed an inapplicable rule as binding doctrine.
- **Line 205 leaks vertical vocabulary into the horizontal layer.** `"d-084 written"` is a cogni-consult decision-log id whose only other occurrence in the repo is `cogni-consult/references/subagent-output-contract.md:27`. A cogni-trends or cogni-portfolio author reading canonical (h) hits an id format their plugin never writes, alongside a `deliverables` noun their domain does not use, with no way to tell which parts are ecosystem doctrine and which are consult residue.

---

cogni-workspace/CLAUDE.md
```diff
@@ -76,6 +76,21 @@ One register, `output-styles/workspace-advisor.md`, at the **plugin root**. Four
...
+
+A vertical plugin **overlays** rather than restates it. The overlay carries only what is genuinely its own — its state lexicon rows, its coinage vocabulary, and any surface its own guards pin — and section letters match the canonical file so a rule and its overlay line up. `cogni-consult/references/user-facing-output.md` is the worked example.
+
+**Cross-plugin reads go through `$COGNI_WORKSPACE_PLUGIN`, and this is the only sanctioned use.** `$CLAUDE_PLUGIN_ROOT` resolves to the *owning* plugin, so it cannot reach this file from a vertical plugin; `scripts/generate-settings.sh` emits `COGNI_WORKSPACE_PLUGIN` for every discovered plugin, and `docs/ecosystem-overview.md` already requires `manage-workspace` to have run before any plugin is used, so the variable is present. **The read must be fail-soft**: an overlay that cannot resolve the variable says so once, applies itself alone, and carries on — matching the convention already used for optional Python deps and theme fallback. A missing canonical register degrades the doctrine; it never blocks the work.
```
sdh07: **"so the variable is present" is false, and it is the premise the whole overlay mechanism rests on.**

`scripts/generate-settings.sh` writes `env[plugin_var] = path` only **`if path:`** (line ~104), and only for plugins in that run's set. Its own comment at line ~100 says so: "a plugin supplied without a path writes no `_PLUGIN` key this run". Verified by running the script directly:

- `--plugins '["cogni-workspace","cogni-consult"]'` (bare name strings — the non-dict branch sets `path = ""`) writes `env` = `{PROJECT_AGENTS_OPS_ROOT, COGNI_WORKSPACE_ROOT, COGNI_WORKSPACE_PYTHON_VENV, COGNI_CONSULT_ROOT}` — **zero `_PLUGIN` keys**. `manage-workspace/SKILL.md:88` explicitly permits a hand-edited JSON list after the confirm step.
- An `--update` whose list omits cogni-workspace **prunes** `COGNI_WORKSPACE_PLUGIN` while `COGNI_WORKSPACE_ROOT` survives beside it. Arm 1 spares ROOT because it is assigned unconditionally at line ~76; PLUGIN has no such protection. `tests/test-generate-settings-prune.sh` case `P6` exercises this exact scenario and asserts only the three unconditional keys, so the regression is invisible to CI.

Also note it is stated as "for every discovered plugin", but `COGNI_WORKSPACE_PLUGIN` comes solely from cogni-workspace's own entry.

This doc-level assurance is what stops anyone writing a guard or a `workspace-status` check for it. `#1820` seeds the workspace keyspace as always-owned and adds case `Q6`, verified red without the fix.

**Separately, the "only sanctioned use" framing declares an exception where the repo already has a mechanism.** The repo-root `CLAUDE.md` states the markets pattern as "Read sites call `cogni-workspace/scripts/get-market-config.py` to join them" — read sites call a script, never a raw path, precisely so resolution and degradation are code. This PR's shape is identical (canonical + per-plugin overlay + a join at read time) but the join is a `cat` and the degradation is a paragraph.

---

cogni-workspace/output-styles/workspace-advisor.md
```diff
@@ -1,18 +1,37 @@
...
+response should read as if it were going to someone who will act on it. This is
+the general register for cross-plugin insight-wave work; a plugin with its own
+domain voice ships its own.
 
-This file carries stance only. It is language-neutral: the anchors below apply in
-whatever language the session resolves to, which the workspace `language` settings
-key and the session-start language hook govern — not this register.
+This file carries stance only. The register — wording, lexicon, orthography,
+table and announcement rules, in whatever language the session resolves to —
+lives in `$CLAUDE_PLUGIN_ROOT/references/user-facing-output.md` — the canonical
```
sdh07: **This pointer resolves for no one: nothing in cogni-workspace ever loads that file.**

`grep -rn user-facing-output cogni-workspace/{skills,agents,hooks,commands,scripts}` returns nothing — zero of the plugin's skills cite it. And `cogni-consult/CLAUDE.md:144`, which this PR left standing, states the constraint flatly: "An output style still does not load `references/`".

The asymmetry is the reverse of how it reads. cogni-consult documents the identical gap as **temporary** — "it closes the moment any consult-* skill loads" — because nine consult skills load the overlay. On the workspace side there is no loader at all, so the gap is permanent and undocumented: a user who picks Workspace Advisor for cross-plugin work gets stance only, forever. No lexicon, no orthography, no table contract, no announcement budget.

That is strictly worse than silence, because the sentence reads as wiring while wiring nothing — the same silent-configuration failure mode `cogni-workspace/CLAUDE.md:73` already catalogues for copied styles.

**It also costs real content.** The pre-PR file's four Behavioral Anchors included "Use concise, professional language". Three of the four survive inline; the conciseness anchor was replaced by this pointer. Net effect: activating Workspace Advisor now yields *less* brevity discipline than before, and the new Stance/Structure lists never restate brevity or compression.

Fix: keep the anchor inline, and say what selecting the style does not deliver rather than implying it does.

**One more on this file:** 15 of its 32 non-blank lines are byte-identical to `cogni-consult/output-styles/strategy-advisor.md` — the frontmatter key, the opener, all three headings, 4 of 5 Stance bullets, 3 of 4 Structure bullets. `cogni-consult/CLAUDE.md:144` argues its single style is right because "there is no longer a second doctrine to keep in step"; this creates one, across two independently-versioned plugins, with nothing pinning them. Styles cannot compose, so a full copy is structurally required — but the pair should at least name each other.

---

cogni-consult/skills/consult-design-thinking/SKILL.md
```diff
@@ -502,7 +502,7 @@ or by reading the field manifests directly).
 artifact, one next-step. Six is the ceiling on the whole close: a trailing offer
 spends one of the substance lines rather than adding a seventh, so a close that
 makes an offer carries at most two other substance lines. At most one table, and
-only at four or five rows — section (d) of
+only at four or five rows — section (d) of the register headed by
```
sdh07: **"the register headed by" is a semantic hint, not a resolvable pointer — and these are the only two citation sites the PR repaired, out of ~13.**

The overlay's `##` headings at this head are exactly (a), (c), (f) and (g). This line cites **section (d)**, and :507 cites **section (e)** of "that same register" — both absent from the file the code span names. A model resolving the close budget opens that file, finds neither, and must notice the preamble and run the `cat` itself; when the variable does not resolve, the overlay's fail-soft clause directs it to "apply this overlay alone and carry on", so the four-row table floor and the announce-XOR-report rule — the two most frequently applied rules in the plugin, stated here as hard close-budget constraints — evaporate while the surrounding sentences still read as binding.

Same shape at `consult-personas/SKILL.md:215` and `:218`.

`tests/test_reference_pointers.sh` case `pointers-resolve` extracts `$CLAUDE_PLUGIN_ROOT/references/...` tokens and tests `[ -e "$PLUGIN_DIR/$rest" ]` — **file existence only, never a section anchor**. I ran it: all six cases pass. So CI signs off on a whole new class of unresolvable pointer, and a reviewer reading green CI concludes pointer integrity was checked.

That guard gap is the root cause of this entire finding class, and it is one grep to close: for every `section (X) of <path>` citation, assert `^## (X)` exists in that path. `#1820` adds it as case `sections-resolve` — **verified red against this head**, naming this file, `consult-personas` and `engagement-list-rendering.md`. Note the matcher has to tolerate a short intervening clause, precisely because the citations read "section (d) of the register headed by `<path>`"; a stricter form matches none of them and ships green over the defect.

---

sdh07 (commented):
## Review: the split moved four sections and left their citations behind

**Not merged.** The seam this PR cuts on is the right one, and the (c) rule-vs-rows division is genuinely well judged. But sections (b), (d), (e) and (h) moved out of `cogni-consult/references/user-facing-output.md` while the citations naming them stayed pointed at it. **Every one of those pointers still resolves as a file**, which is why all 15 checks and all 14 root guards are green over them: no guard in the repo resolves the cited *section*. `tests/test_reference_pointers.sh` tests `[ -e <file> ]` and nothing about the letter.

Fixes are up as a draft at **#1820**, based on this branch so its diff is only the repairs — take it, cherry-pick from it, or ignore it.

Every claim below and in the inline comments is backed by a command run against a worktree at `7114f32`.

### Two claims in the PR body that don't hold

- **"144 suites green"** — it is 142/144 locally. The two failures (`cogni-consult/tests/test_resolve_assumptions.sh`, `tests/test_release_bundle_wiki.sh`) fail identically on a fresh `main` worktree, so they are pre-existing and environment-dependent, **not** this PR's fault. Worth correcting the number rather than the code.
- **"Every dangling section citation was chased"** — two were fixed (`consult-design-thinking`, `consult-personas`); at least eleven were not.

### Findings outside the diff, so not inline-commentable

**The nine Step 0 blocks are the main-loop entry to the register, and they are now wrong** — `consult-action-fields:54`, `consult-dashboard:57`, `consult-design-thinking:75`, `consult-personas:50`, `consult-project-plan:68`, `consult-publish:65`, `consult-resume:46`, `consult-scope:40`, `consult-setup:43`:

> The register that output follows — scope, state lexicon, table rules, step announcements — is `$CLAUDE_PLUGIN_ROOT/references/user-facing-output.md`

Three of those four are no longer in that file. Because the pointer asserts this file *is* the register rather than an overlay, a model has no reason to look further — and Step 0 never mentions the `cat`. Result: tables under four rows, announcements both made and reported, unbounded 25/45-word budgets. `test_step0_register_block.sh` case `register-same-*` pins this paragraph **byte-identical across all nine**, so the guard now actively enforces the false enumeration until all nine are edited together.

The same blocks also say section (f) is owned by the overlay and "edit them there", which contradicts the overlay's own "Edit them in the canonical file first". Two maintainers following the two instructions edit two different files.

**Other stale citations:**

| Site | Cites | Problem |
|---|---|---|
| `references/engagement-list-rendering.md:59` | "rule 8 of the table contract" in the overlay | (d) moved. An `unreadable` action field flattens to `—` — exactly the missing-vs-unmeasured conflation rule 8 exists to forbid, so a failed read reports as "nothing to measure" |
| `references/subagent-output-contract.md:13` | the overlay as the half to keep in step | The rules it mirrors are canonical (g) and (h). A maintainer editing canonical (h) has no pointer to the agent-side twin; one following this line finds no (h) and concludes there is nothing to reconcile. The hook keeps injecting the stale half verbatim |
| `references/interaction-language.md:48`, `:78` | the overlay owns whether a slug may appear | Now canonical (d).7. `:73` also still says "a fourth copy of the specifics" — the count is now eleven |
| `scripts/orthography-drift-scan.py:4`, `README.md:167`, `CLAUDE.md:186` | overlay §(a) as the home of the ß/ss rule | Text is now canonical-only. The scan's own rationale points at a file that no longer carries the rule it enforces |
| `CLAUDE.md:63` | "The register itself lives in references/user-facing-output.md" | Contradicts `:144`, which this PR rewrote, and `README.md:213`, which it updated. The developer guide now contradicts itself and the README |

**`README.md:236,238` still market the plugin as it was before the split.** The dependency table reads `| cogni-workspace | No |` and the prose says "cogni-consult is standalone as an orchestrator". After this PR, four register sections exist only in cogni-workspace. Install cogni-consult + cogni-knowledge — the documented supported configuration — and the fail-soft branch fires permanently, including for the brevity budgets the plugin's own close-budget rules cite as hard constraints. That doctrine shipped inside the plugin before this PR.

**`scripts/generate-settings.sh` has a live bug this PR makes load-bearing.** An `--update` that deselects cogni-workspace prunes `COGNI_WORKSPACE_PLUGIN` while `COGNI_WORKSPACE_ROOT` survives beside it, because arm 1 spares ROOT only by virtue of its unconditional assignment. Case `P6` walks this exact scenario and asserts only the three unconditional keys, so CI cannot see it. Details in the inline comment on `cogni-workspace/CLAUDE.md:85`.

### The guard gap is the finding that matters most

Every item above shares one root cause: **no guard resolves a cited section letter.** Two suites read as coverage and are not:

- `test_reference_pointers.sh` — file existence only. Its scope note documents which pointer *forms* it skips but says nothing about section-level resolution, so the gap reads as coverage.
- `test_step0_register_block.sh` — `OWNER` is the cogni-consult overlay; nothing in the suite opens cogni-workspace. It watched three of the four rule families named by the paragraph it pins leave the file that paragraph points at, and reported green.

#1820 closes both, each with a red-proving arm, and the section guard is verified red against this head.

### Not raised as defects — your call, not a fix branch's

- **The canonical file has exactly one consumer.** Several passes argued the split's machinery is paid for now against a second reader that does not exist, and that deferring it until a second plugin needs it is one `git mv` away. That is an argument for reverting, not repairing, so #1820 leaves it alone.
- **The `cat` costs a second serial round trip on Step 0 of all nine skills**, and no read-once clause exists anywhere in the chain, so a two-skill session re-issues it. Measured: `main` delivered the whole doctrine in one read of 16564 B; this head needs the 10110 B overlay plus an 11292 B `cat` that cannot be batched, since no SKILL.md names the canonical path. The split is also +29.2% bytes / ~+1300 tokens for the same doctrine, because both halves grew navigational scaffolding rather than one file being sliced in two.

---

PR #1826, opened 2026-09-04
render-html-slides: deterministic parse + producer field alignment

### LINKED ISSUE #1795: cogni-workspace: render-html-slides parses the brief deterministically and reads the fields the producer actually writes
## Plugin
cogni-workspace

## Epic context
Phase 3 (checks and the HTML renderer) of "EPIC cogni-workspace: one presentation-brief grammar for every slide renderer — Claude Design first". It makes the third renderer a thin adapter over the shared parser and closes the producer/renderer field mismatches that silently drop content today.

## Current behavior
`skills/render-html-slides/SKILL.md` Phase 1 (`:58-72`) is an LLM parse of the brief markdown into `slide-data.json`, and `scripts/generate-html-slides.py` (1,948 lines) renders from that JSON only — so there is no fixture anywhere in the path and no way to test the parse. Phase 0 (`:49`) pins `version: "4.0"`, which a v4.1 brief fails.

The renderer also reads different field names than the producer writes, and each mismatch fails silently:

- `render_four_quadrants` reads only `Quadrant-{i}` (`generate-html-slides.py:341`) while the producer writes `Q1:` — so the Buying Center slide renders four empty cards.
- The closing slide reads `Headline` / `Key-Takeaway` / `CTA` (`:550`) while the brief emits `Title` / `Subtitle` / `Metadata`.
- Three-options reads `Title` / `Subtitle` / `Bullets` / `Recommended` (`:427`) while the brief emits `Name` / `Price` / `Badge` / `Features`.
- Timeline-steps reads `Steps[]{Title,Detail,Bullets}` or `Step-N{Label,Date}` while the brief emits `Step-N{Number,Label,Description,Duration}`.
- `Slide-Title` and `Source` are never read at all.
- IS/DOES/MEANS badges come from `--language` rather than the brief's `Label:` field.
- A `references` renderer exists in `LAYOUT_RENDERERS` (`:620`) but nothing emits that layout.
- `Icon:` is unused.

## Desired behavior
**Deterministic parse.** Phase 1 runs `parse-brief.py --emit slide-data` and falls back to the LLM parse only when the parser returns `success: false`. Phase 0 accepts `4.0` and `4.1`. The existing theme fallback to the ecosystem picker (`SKILL.md:53-56`) stays.

**Field alignment in `generate-html-slides.py`,** with old names kept as accepted aliases so no existing brief regresses:

- `Slide-Title` is used when present, with the H2 headline as fallback.
- `Source` is rendered as a footer link.
- `Q1`..`Q4` are accepted alongside `Quadrant-1`..`Quadrant-4`.
- Closing slide renders `Title` / `Subtitle` / `Metadata`, with `Headline` / `Key-Takeaway` / `CTA` still accepted.
- Three-options renders `Name` / `Price` / `Badge` / `Features` (a non-empty `Badge` means recommended), with the old names still accepted.
- Timeline-steps renders `Step-N{Number,Label,Description,Duration}` as well as `Steps[]`.
- IS/DOES/MEANS badges come from the brief's `Label:` first, `--language` as fallback.
- `slide_kind: references` routes to the existing references renderer.
- `render_generic_slide` skips the non-content keys `intent`, `visual`, `cta`, `Slide-Kind`, `Source` and `Slide-Title` rather than printing them.

**Old → new, user-visible strings**

- `references/01-layout-renderers.md` closing-slide block `Headline / Key-Takeaway / CTA` → `Title / Subtitle / Metadata (Headline, Key-Takeaway and CTA accepted as aliases)`.
- `references/01-layout-renderers.md` three-options block `Title / Subtitle / Bullets / Recommended` → `Name / Price / Badge / Features (old names accepted)`.
- `SKILL.md:49` `version: "4.0"` → `version: "4.0" or "4.1"`.

## Motivation
The HTML renderer is the one render path that runs entirely inside this repo, so it is where a producer/renderer mismatch is both most visible and most fixable — and today it renders a four-card stakeholder slide as four empty cards, drops every `Source` footer, and mislabels the closing slide, all without a single error. Moving Phase 1 onto the shared parser also means this renderer stops being the only consumer whose parse cannot be tested.

## Acceptance criteria
- [ ] A new eval case renders `cogni-workspace/libraries/EXAMPLE_BRIEF.md` through the parser and asserts by shape: `status: ok`, a `slides` count equal to the `^## Slide` heading count, four non-empty `quad-card` elements on the Buying Center slide, the closing subtitle present, and one source footer per slide that carries a `Source` field. Falsified by empty quadrant cards, a missing closing subtitle or a missing footer.
- [ ] `python3 cogni-workspace/skills/render-html-slides/evals/run.py` exits 0 with existing cases 1–4 byte-identical to their pre-change outputs — old field names keep precedence where both old and new are present. Falsified by any byte difference in those four outputs.
- [ ] `grep -n "parse-brief.py" cogni-workspace/skills/render-html-slides/SKILL.md` matches in Phase 1, and the fallback rule (LLM parse only on `success: false`) is stated there. Falsified by no script call, or by a fallback that fires on a successful parse.
- [ ] A `version: "4.1"` brief passes Phase 0 and a `version: "4.0"` brief still does. Falsified by a rejection of either.
- [ ] `references/01-layout-renderers.md` documents the producer's field names for closing-slide, three-options and timeline-steps, marking the old names as accepted aliases. Falsified by the reference still naming only the old contract.
- [ ] `render_generic_slide` emits no visible output for `intent`, `visual`, `cta`, `Slide-Kind`, `Source` or `Slide-Title`. Falsified by any of those keys appearing as rendered slide body text.
- [ ] Every import in `generate-html-slides.py` resolves to the standard library. Falsified by a third-party import.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-workspace` exits 0 with `data.total` greater than 0, and `python3 scripts/check-external-dispatch.py` and `python3 scripts/check-breadcrumbs.py` exit 0.
- [ ] Scope boundary: the `{source_dir}/cogni-visual/` output directory name is unchanged, the two-section speaker-notes format is unchanged, and `python3 scripts/check-version-bump.py` reports no `version-touched` finding.

_Affected components: cogni-workspace/skills/render-html-slides/SKILL.md (Phase 0, Phase 1, Phase 4), skills/render-html-slides/references/01-layout-renderers.md, skills/render-html-slides/scripts/generate-html-slides.py, skills/render-html-slides/evals/run.py_

_Complexity hint: medium_

## Depends on
- #1791 cogni-workspace: add a stdlib parse-brief.py that turns a presentation brief into slide-data.json, an outline, or metadata

## Planning provenance
- Repo filed into: cogni-work/insight-wave
- Plugin: cogni-workspace
- Ownership verdict: owned
- Evidence: .claude-plugin/marketplace.json in the CWD repo lists cogni-workspace
- CWD repo: cogni-work/insight-wave

claude-cogni-work:
Closes #1795.

## Summary

- **Phase 1 is deterministic.** `render-html-slides` now runs the shared `parse-brief.py` (`--emit slide-data`, plus `--emit metadata` for the block that emitter does not carry) and falls back to the LLM free-parse **only** when the envelope returns `success: false`. The free-parse prose is **relocated** into an explicit `#### LLM fallback parse` subsection, not deleted.
- **The renderer reads what the producer writes**, with old names keeping precedence so no existing brief regresses: `Slide-Title`, a `Source` footer link, `Q1`..`Q4`, closing `Title`/`Subtitle`/`Metadata`, three-options `Name`/`Price`/`Badge`/`Features`, timeline `Step-N{Number,Label,Description,Duration}`, and IS/DOES/MEANS badges from each box's `Label:`. `render_generic_slide` stops printing six non-content keys.
- **Two new eval cases** — one end-to-end through the parser, one on direct slide-data.
- `references/01-layout-renderers.md` documents the producer names and corrects the claim this change makes false.

## The two decisions I took from the files rather than the issue text

**1. `slide_kind: references` routes conditionally.** `EXAMPLE_BRIEF.md` slide 13 declares `Slide-Kind: references` while carrying `Layout: two-columns-equal` with `Left-Column`/`Right-Column` and **no** `References`/`Citations` field. Routing it unconditionally resolves the reference list to `[]` and destroys three visible source strings — re-committing the exact silent-content-drop class this issue exists to close. Routing therefore fires only when a reference payload is actually present. See `## Open questions`.

**2. The `Q1`..`Q4` alias is proven on direct slide-data, not end-to-end.** `parse-brief.py`'s `_normalize_quadrants` rewrites `Q{n}` to `Quadrant-{n}` *upstream*, so an end-to-end `EXAMPLE_BRIEF` assertion renders four populated cards **with zero renderer change** — a vacuous green. Eval case 6 builds its slide-data directly so the assertion can actually fail. Confirmed by mutation M1 below, which reddens **only** case 6.

## One acceptance criterion was already satisfied at base

The AC asks that a `version: "4.1"` brief pass Phase 0 and a `version: "4.0"` brief still do. `SKILL.md:49` already read ``version: "4.0"`` or ``"4.1"`` at `origin/main` — a sibling landed it ahead of this issue. **This diff deliberately leaves line 49 untouched**; it is a preserve-invariant, not work. Please grade it as passing rather than as a missing edit.

## Guard verification — the new cases can actually fail

Each mutation was applied to `generate-html-slides.py` alone, the suite re-run, then the file restored:

| Mutation | Suite result |
|---|---|
| M1 — drop the `Q{i}` arm of the quadrant alias | case 6 red (case 5 stays green — the vacuous-green point) |
| M2 — make `slide_kind` routing unconditional | cases 5 + 6 red |
| M3 — drop `Subtitle` from the closing takeaway chain | cases 5 + 6 red |
| M4 — disable the `Source` footer | cases 5 + 6 red |
| M5 — revert the IS/DOES/MEANS `Label` priority | cases 5 + 6 red |
| M6 — revert the `Slide-Title` H2 override | case 6 red |
| restored | 6/6 green |

## Regression evidence

`generate-html-slides.py` was run at `origin/main` and at HEAD over an old-name-only slide-data covering all six affected layouts, in both `en` and `de`:

- **The slide markup is byte-identical in both languages**, and **no line is removed** from the output.
- The only difference is **+26 lines of new CSS** (`.source-footer`, `.source-footer-raised`, `.source-footer a`, `.timeline-number`), which is unavoidable for a new visual element — the stylesheet is emitted unconditionally. Worth knowing when reading the AC's "byte-identical" wording: it holds exactly for the slide markup, and the stylesheet gains rules without losing any.

`_first_present()` is presence-based, not truthiness-based, so an authored-but-empty old name keeps rendering empty rather than falling through to a newer alias.

## Gates

- `evals/run.py` — 6/6, exit 0
- `run-plugin-tests.py --filter cogni-workspace` — 21/21 suites, exit 0
- `check-external-dispatch.py`, `check-breadcrumbs.py`, `check-version-bump.py` — exit 0, no `version-touched` finding
- `check-frontmatter.sh`, `check-breadcrumbs.sh` (cogni-workspace) — exit 0
- Every import in both scripts is stdlib; no `plugin.json` in the diff

No `*/tests/test[-_]*.sh` and no `check-*` guard file is touched, so no `## Mutation recipe` section applies; the mutations above are recorded for the reviewer's benefit.

## Review notes

- **Pre-existing, not introduced here.** The skill-quality gate flagged the `## Features` section (~19 lines) as user-facing copy duplicating the Parameters table and Phase 5/6. Marked `introduced_by_change: false`, so it is annotated rather than re-planned. Body is 17,559 chars against a 21,000 cap (78%).
- The skill-quality gate returned `needs_revision` on three auto-fixable findings, all applied: the LLM-fallback branch now also assembles `metadata` (it previously would have produced the exact "Presentation"-titled deck the deterministic branch warns about), the merge step now names its mechanism, and a `## Additional Resources` section now names all three bundled references.
- Token-scope preflight reports `over_scoped` — extra `gist`, `read:org`, `workflow`, with `workflow` classified **forbidden**. Cleared with the settled `--allow-overscoped` opt-out, flagged here because the posture is a runner-level fact a reviewer may want to revisit, not something this PR changes.

## Open questions

- **`parse-brief.py:381` states, on the field it sets: "The layout is the Layout field. Slide-Kind never overrides it."** This PR's `effective_layout()` makes `Slide-Kind` override `Layout` at *render* time (guarded on payload presence). The parser's own comment stays literally true — the parsed `layout` field is untouched — but the two files now express opposite routing intents, and the issue's AC explicitly asked for the `slide_kind` routing. If the parser's rule is meant to bind consumers too, the right fix is to teach `render_references_slide` a `Left-Column`/`Right-Column` shape instead and drop `effective_layout()` entirely.
- **The defect class is narrowed, not closed.** This PR fixes the reported instances; a quality pass found `Icon` (7 occurrences in `EXAMPLE_BRIEF.md`), `visual.chart` (real chart data), and the per-slide `cta` block still authored and still read nowhere — two of them now *deliberately* suppressed by `NON_CONTENT_FIELDS`, so they degrade silently by design. The rejected alternative plan's answer was an unread-authored-key report in the output envelope (precedent: `theme_slug_resolution`). I did **not** add it — it is a new public output-envelope key and outside this issue's acceptance bar — but it is the change that would close the class.
- **Alias ownership is split.** `parse-brief.py` already owns the quadrant alias with a warn-and-drop rule; this PR adds a second, silent owner in the renderer. Consolidating into the parser would also benefit the pptx consumer. Genuinely a separate PR.
- Not filed as issues: the three items above are recorded here rather than on the board, because they surfaced from the quality pass rather than from planned deferred scope. Say the word and I will file them.

### DIFF
--- a/cogni-workspace/skills/render-html-slides/SKILL.md
+++ b/cogni-workspace/skills/render-html-slides/SKILL.md
@@ -29,7 +29,7 @@ all in a single `.html` file.
 | `brief_path` | auto-discovered | Path to presentation-brief.md |
 | `theme` | from brief frontmatter | Path to theme.md (or omit for pick-theme) |
 | `design_variables` | derived from theme | Pre-computed design-variables.json path |
-| `output_path` | `{brief_dir}/{slug}-slides.html` | HTML output path |
+| `output_path` | `{brief_dir}/{slug}-slides.html` | HTML output path. `{slug}` is the deck title Phase 1 assembles into `metadata.title`, lowercased with every run of non-alphanumerics collapsed to a single `-`. |
 | `transition` | `fade` | Slide transition: `fade`, `slide`, `none` |
 | `aspect_ratio` | `16:9` | Slide aspect ratio: `16:9`, `4:3` |
 | `language` | from brief frontmatter | `en` or `de` |
@@ -48,6 +48,7 @@ all in a single `.html` file.
    - `type: presentation-brief` (must match)
    - `version: "4.0"` or `"4.1"` (must match one of these — the 4.1 delta is content-only, so nothing this renderer parses changed)
 5. Extract metadata: `theme`, `theme_path`, `language`, `customer`, `provider`, `arc_type`, `governing_thought`
+6. Resolve `brief_dir` — the project directory that *contains* `cogni-visual/`, which every artifact path below is anchored on. A brief discovered inside a `cogni-visual/` folder (where `story-to-slides` writes it) resolves to that folder's **parent**, not the brief's own directory — otherwise `{brief_dir}/cogni-visual/` doubles the path. An explicit `brief_path` outside any `cogni-visual/` folder resolves to the brief's own directory.
 
 **Theme resolution** (3-stage, same as enrich-report):
 1. If `design_variables` parameter provided → use directly, skip to Phase 1
@@ -62,10 +63,13 @@ Parse the presentation-brief.md body into structured slide data. The brief uses 
 **Parse deterministically.** Run the shared parser rather than reading the brief yourself — it is the same parser the other brief consumers use, so this renderer cannot drift away from them:
 
 ```bash
+mkdir -p "{brief_dir}/cogni-visual"
 python3 "${CLAUDE_PLUGIN_ROOT}/scripts/parse-brief.py" \
   --brief "{brief_path}" --emit slide-data \
   --output "{brief_dir}/cogni-visual/slide-data.json"
 ```
+
+`mkdir -p` runs first because the parser opens `--output` directly without creating its parent.
 
 `--brief` is required (omitting it exits with `--brief is required`). The `{"success": ..., "data": ..., "error": ...}` envelope prints on **stdout**, while `--output` writes only the payload (`{version, slides, warnings}`) to the file.
 
@@ -87,7 +91,9 @@ top-level key alongside `slides`, and write the file out again. Leaving the pars
 `version` and `warnings` keys in place is harmless — the renderer reads only `metadata` and
 `slides`. Surface any `warnings` to the user.
 
-**Fall back to the LLM parse only when the parser fails** — that is, when an envelope comes back with `success: false`, or the script cannot be run at all. A successful parse is authoritative and is never second-guessed or "corrected" by re-reading the brief.
+**Fall back to the LLM parse only when the parser fails to parse** — that is, when an envelope comes back with `success: false` for a parse reason, or the script cannot be run at all. A successful parse is authoritative and is never second-guessed or "corrected" by re-reading the brief.
+
+**A `cannot write output:` envelope is not a parse failure and must not trigger the fallback.** Fix the write — the parent directory is missing or unwritable — and re-run. Falling back on it would hide the broken write behind a silently LLM-parsed deck, which is exactly the outcome the deterministic path exists to retire.
 
 #### LLM fallback parse
 
@@ -104,11 +110,13 @@ For each slide section, extract:
 - `citations` — extracted `<sup>[N](url)</sup>` patterns
 
 Also assemble the top-level `metadata` block by the same mapping the deterministic path uses
-above: the renderer reads `metadata.title` / `.customer` / `.provider` / `.generated` on both
-paths, and omitting it silently yields the title "Presentation" and today's date.
+above — the renderer reads it on both paths, and omitting it carries the same silent
+"Presentation"/today's-date default.
 
-Write the parsed data to `{brief_dir}/cogni-visual/slide-data.json`, in the shape both paths
-share — see the `slide-data.json shape` section of `references/01-layout-renderers.md`.
+Run `mkdir -p "{brief_dir}/cogni-visual"` here too — this branch writes into the same directory
+and must not depend on the deterministic block above having run. Then write the parsed data to
+`{brief_dir}/cogni-visual/slide-data.json`, in the shape both paths share — see the
+`slide-data.json shape` section of `references/01-layout-renderers.md`.
 
 **Parsing rules:**
 - Each slide is separated by `---` (horizontal rule) in the brief
@@ -318,4 +326,4 @@ Apply re-render changes FIRST (they produce a fresh HTML file), then apply HTML-
 
 **Self-contained:** Single HTML file. All CSS inline. All JS inline. Only external dependency is Mermaid.js CDN (conditional, only included when diagrams are present).
 
-**Refinement Loop:** After viewing slides, give feedback on specific slides in natural language. Text and style fixes are applied directly to the HTML (instant). Structural changes (layout swaps, slide additions) trigger a targeted re-render via the Python script. Max 3 rounds by default.
+**Refinement Loop:** Natural-language feedback per slide after viewing — text and style fixes edit the HTML directly, structural changes trigger a targeted re-render. Channels and round cap: Phase 6.
--- a/cogni-workspace/skills/render-html-slides/SKILL.md
+++ b/cogni-workspace/skills/render-html-slides/SKILL.md
@@ -59,6 +59,40 @@ all in a single `.html` file.
 
 Parse the presentation-brief.md body into structured slide data. The brief uses `## Slide N: {headline}` sections with fenced YAML blocks.
 
+**Parse deterministically.** Run the shared parser rather than reading the brief yourself — it is the same parser the other brief consumers use, so this renderer cannot drift away from them:
+
+```bash
+python3 "${CLAUDE_PLUGIN_ROOT}/scripts/parse-brief.py" \
+  --brief "{brief_path}" --emit slide-data \
+  --output "{brief_dir}/cogni-visual/slide-data.json"
+```
+
+`--brief` is required (omitting it exits with `--brief is required`). The `{"success": ..., "data": ..., "error": ...}` envelope prints on **stdout**, while `--output` writes only the payload (`{version, slides, warnings}`) to the file.
+
+**Then add the `metadata` block.** `--emit slide-data` carries no `metadata` key, but the renderer reads `metadata.title` / `.customer` / `.provider` / `.generated` — skip this and every deck silently renders with the title "Presentation" and today's date. Run:
+
+```bash
+python3 "${CLAUDE_PLUGIN_ROOT}/scripts/parse-brief.py" --brief "{brief_path}" --emit metadata
+```
+
+That returns `{version, frontmatter, generation_metadata, cta_summary, unowned_sections, warnings}` — a source to map from, not a ready-made block. Assemble `metadata` and merge it into the written JSON as a top-level key:
+
+- `customer`, `provider`, `language`, `arc_type`, `governing_thought` — from `frontmatter`
+- `generated` — from `frontmatter.generated`, else today's date
+- `title` — slide 1's `fields.Title`, else slide 1's `headline` (the frontmatter carries no title)
+- `subtitle` — slide 1's `fields.Subtitle`
+
+Read `{brief_dir}/cogni-visual/slide-data.json` back, add the assembled `metadata` object as a
+top-level key alongside `slides`, and write the file out again. Leaving the parser's own
+`version` and `warnings` keys in place is harmless — the renderer reads only `metadata` and
+`slides`. Surface any `warnings` to the user.
+
+**Fall back to the LLM parse only when the parser fails** — that is, when an envelope comes back with `success: false`, or the script cannot be run at all. A successful parse is authoritative and is never second-guessed or "corrected" by re-reading the brief.
+
+#### LLM fallback parse
+
+Used only on the failure condition above. Parse the brief body directly.
+
 For each slide section, extract:
 - `number` — slide number
 - `headline` — the assertion headline from the H2 heading
@@ -69,38 +103,12 @@ For each slide section, extract:
 - `diagram_mermaid` — the Diagram field content if it contains Mermaid syntax
 - `citations` — extracted `<sup>[N](url)</sup>` patterns
 
-Write the parsed data to `{brief_dir}/cogni-visual/slide-data.json`:
+Also assemble the top-level `metadata` block by the same mapping the deterministic path uses
+above: the renderer reads `metadata.title` / `.customer` / `.provider` / `.generated` on both
+paths, and omitting it silently yields the title "Presentation" and today's date.
 
-```json
-{
-  "metadata": {
-    "title": "...",
-    "subtitle": "...",
-    "customer": "...",
-    "provider": "...",
-    "language": "de",
-    "arc_type": "why-change",
-    "generated": "2026-02-09",
-    "governing_thought": "..."
-  },
-  "slides": [
-    {
-      "number": 1,
-      "headline": "Assertion headline",
-      "layout": "title-slide",
-      "fields": {
-        "Title": "...",
-        "Subtitle": "...",
-        "Metadata": "..."
-      },
-      "speaker_notes": null,
-      "bottom_banner": null,
-      "diagram_mermaid": null,
-      "citations": []
-    }
-  ]
-}
-```
+Write the parsed data to `{brief_dir}/cogni-visual/slide-data.json`, in the shape both paths
+share — see the `slide-data.json shape` section of `references/01-layout-renderers.md`.
 
 **Parsing rules:**
 - Each slide is separated by `---` (horizontal rule) in the brief
@@ -108,8 +116,8 @@ Write the parsed data to `{brief_dir}/cogni-visual/slide-data.json`:
 - Nested structures (Hero-Stat-Box, Context-Box, Detail-Grid) must be preserved as nested objects
 - Speaker-Notes is a multi-line YAML string (after `|`) — preserve the full text
 - Mermaid diagrams appear in the `Diagram:` field as multi-line strings
-- IS/DOES/MEANS labels come from the `Label:` field inside each box — but the Python script handles localization, so just pass the text content
-- Bottom-Banner can be a dict with `Text:` key or a plain string. Write it to the top-level `bottom_banner` field — the renderer emits it as a shared footer on **every** layout. The legacy nested form (`Bottom-Banner` inside `fields`) is also accepted.
+- IS/DOES/MEANS labels come from each box's `Label:` field — pass it through. The renderer prefers that `Label:` for the badge and falls back to its own `--language` localization only when it is absent
+- Bottom-Banner can be a dict with a `Text:` key or a plain string. Write it to the **top-level** `bottom_banner` field, not inside `fields` (the legacy nested form is also accepted) — footer contract in `references/01-layout-renderers.md`.
 
 ### Phase 2: Theme → Design Variables
 
@@ -194,7 +202,7 @@ If error, read the error message and attempt to fix the input data. Common issue
 
 Tier-1 tokens are wired today. Pass `theme_slug: cogni-work` to import that theme's canonical CSS custom properties. Tier-3 deck-component primitives (`title-slide.html`, `content-slide.html`, etc.) are the next increment — the loader infrastructure is in place at `cogni-workspace/scripts/load-theme-component.py` (see `cogni-workspace/references/theme-component-loader.md`), but `cogni-work` does not yet ship a `tiers.components.deck` family, so component-substitution is gated on a follow-up issue. Today's behavior: every layout renderer uses its inline template; tomorrow's behavior (after deck primitives ship): the renderer prefers theme-supplied primitives and falls back to inline on miss.
 
-**Backwards-compat contract.** Omitting `--theme-slug` preserves the legacy (pre-Theme-System-v2) rendering path byte-for-byte — `theme.md`-derived design variables still apply; only the tier-1 `tokens.css` import is skipped. With `--theme-slug` set, themes without `tiers.tokens` (and tier-0 themes generally) exercise the same fallback path — there is no failure case for unmigrated themes. `evals/run.py` enforces this with three regression cases: tier-0 baseline, tier-1 cogni-work tokens.css imported, tier-0 `_template` with `--theme-slug` set (graceful fallback).
+**Backwards-compat contract.** Omitting `--theme-slug` preserves the legacy (pre-Theme-System-v2) rendering path byte-for-byte — `theme.md`-derived design variables still apply; only the tier-1 `tokens.css` import is skipped. With `--theme-slug` set, themes without `tiers.tokens` (and tier-0 themes generally) exercise the same fallback path — there is no failure case for unmigrated themes. `evals/run.py` enforces this with dedicated regression cases — among them a tier-0 baseline, tier-1 cogni-work tokens.css imported, and tier-0 `_template` with `--theme-slug` set (graceful fallback).
 
 **Fallback diagnostics.** When `--theme-slug` is set, the output envelope's `theme_slug_resolution` field names *why* the tier-1 path was (or wasn't) taken. It is always a **bare reason code** — exact-matchable: `imported` on success, or a fallback code — `manifest_missing` (tier-0 theme), `manifest_unreadable`, `tokens_tier_absent`, `tokens_css_missing`, or `themes_dir_unresolved` (workspace not found — e.g. `$COGNI_WORKSPACE_ROOT` unset or unexported). The field is `null` when `--theme-slug` is omitted. A companion `theme_slug_resolution_detail` field carries an optional human-readable hint for the cases that warrant one (today, `themes_dir_unresolved`); it is `null` otherwise. Machines branch on `theme_slug_resolution`; humans read `theme_slug_resolution_detail`. This distinguishes "this theme is tier-0" from "I couldn't find the workspace" without re-walking the resolution. Add `--verbose` to additionally write a one-line `themes_dir=… manifest_path=… tokens_css=… resolution=…` diagnostic (using the bare reason code) to stderr (off by default; the stdout JSON contract is unchanged).
 
@@ -284,6 +292,13 @@ Apply re-render changes FIRST (they produce a fresh HTML file), then apply HTML-
 3. Increment refinement counter
 4. If counter < `max_refinements`: return to Step 6.1
 5. If counter >= `max_refinements`: inform the user the refinement cap is reached
+
+## Additional Resources
+
+- **`references/01-layout-renderers.md`** — per-layout field contracts (including the accepted
+  aliases for each slot), the Bottom-Banner footer contract, and the shared `slide-data.json` shape
+- **`references/02-slide-navigation.md`** — navigation, transitions and keyboard behaviour
+- **`references/03-speaker-notes.md`** — speaker-notes parsing and the notes-panel format
 
 ## Features
 
--- a/cogni-workspace/skills/render-html-slides/evals/run.py
+++ b/cogni-workspace/skills/render-html-slides/evals/run.py
@@ -300,11 +300,20 @@ def case_5_example_brief_end_to_end():
         failures.append("case 5: no closing-slide found in the parsed fixture")
     else:
         subtitle = closing["fields"].get("Subtitle", "")
-        if subtitle and subtitle not in html:
+        # Pinned: without this the check below no-ops the moment the producer
+        # stops emitting Subtitle at all, which is the regression it guards.
+        if not subtitle:
+            failures.append("case 5: closing slide carries no Subtitle — the fixture must parse one")
+        elif subtitle not in html:
             failures.append("case 5: closing subtitle {!r} absent from the render".format(subtitle))
 
     # One source footer per Source-carrying slide, each an actual link.
     with_source = [s for s in slides if s.get("source") or s.get("fields", {}).get("Source")]
+    # Pinned: a bare count comparison comes out 0 == 0 and passes when the
+    # producer emits no Source at all, so require the fixture to carry some.
+    if len(with_source) < 2:
+        failures.append("case 5: {} Source-carrying slide(s) parsed, expected at least 2".format(
+            len(with_source)))
     if html.count('class="source-footer') != len(with_source):
         failures.append("case 5: {} source footer(s) for {} Source-carrying slide(s)".format(
             html.count('class="source-footer'), len(with_source)))
@@ -402,6 +411,23 @@ def case_6_field_aliases_direct():
              "fields": {"Source": "[SRCTEXT](https://example.invalid/doc)",
                         "Left-Column": {"Headline": "L", "Bullets": ["b"]},
                         "Right-Column": {"Headline": "R", "Bullets": ["c"]}}},
+            # Slide 6 covers only the Step-N map shape; the Steps[] list form is
+            # the other live shape and was untested.
+            {"number": 13, "layout": "timeline-steps", "headline": "steps list form",
+             "fields": {"Steps": [{"Title": "LISTSTEPA", "Detail": "LISTDETA",
+                                   "Bullets": ["listbullet-a"]},
+                                  {"Title": "LISTSTEPB", "Detail": "LISTDETB"}]}},
+            # Slide 6 uses Label; Title is the older name that must win over it.
+            {"number": 14, "layout": "timeline-steps", "headline": "steps old name wins",
+             "fields": {"Step-1": {"Number": "02", "Title": "OLDSTEPTITLE",
+                                   "Label": "NEWSTEPLABEL", "Detail": "OLDSTEPDETAIL",
+                                   "Duration": "NEWSTEPDUR"}}},
+            # Slide 7 tests only the Label-wins direction of the badge contract.
+            # With no Label anywhere, --language de must supply the defaults.
+            {"number": 15, "layout": "is-does-means", "headline": "idm no label",
+             "fields": {"IS-Box": {"Text": "nolabel-is"},
+                        "DOES-Box": {"Text": "nolabel-does"},
+                        "MEANS-Box": {"Text": "nolabel-means"}}},
         ],
     }
     # German, so a badge falling back to --language would read IST/MACHT/BEDEUTET.
@@ -447,10 +473,22 @@ def case_6_field_aliases_direct():
     for needle in ("01", "STEPLABEL", "STEPDESC", "STEPDUR"):
         want(6, needle)
 
+    # The Steps[] list form is the other live shape and must render the same slots.
+    for needle in ("LISTSTEPA", "LISTDETA", "listbullet-a", "LISTSTEPB", "LISTDETB"):
+        want(13, needle)
+    # Within a Step-N map the older Title/Detail win over Label/Duration.
+    want(14, "OLDSTEPTITLE")
+    want(14, "NEWSTEPLABEL", present=False, why="Title must keep precedence over Label")
+    want(14, "OLDSTEPDETAIL")
+    want(14, "NEWSTEPDUR", present=False, why="Detail must keep precedence over Duration")
+
     # The authored badge beats the --language default.
     for needle in ("CUSTOMIS", "CUSTOMDOES", "CUSTOMMEANS"):
         want(7, needle)
     want(7, "IST", present=False, why="--language fallback must not override an authored Label")
+    # ...and the untested direction: with no Label at all, --language de supplies them.
+    for needle in ("IST", "MACHT", "BEDEUTET"):
+        want(15, needle, why="Label-less box must fall back to the --language default badge")
 
     # slide_kind routing: WITH a payload it routes...
     want(8, "references-list")
--- a/cogni-workspace/skills/render-html-slides/evals/run.py
+++ b/cogni-workspace/skills/render-html-slides/evals/run.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python3
-"""evals/run.py — Regression eval for render-html-slides Theme System v2 wiring.
+"""evals/run.py — Regression eval for render-html-slides.
 
-Stdlib-only. Three test cases, run end-to-end against the real
-``generate-html-slides.py`` script:
+Stdlib-only. Every case runs end-to-end against the real
+``generate-html-slides.py`` script; ``main()``'s ``cases`` list is the
+authoritative roster. The Theme System v2 cases are:
 
 1. **Tier-0 baseline (regression).** No ``--theme-slug`` flag. Asserts the
    rendered HTML contains the inline ``:root`` token block and does NOT
@@ -31,6 +32,7 @@ on stdout for machine consumption.
 
 import json
 import os
+import re
 import subprocess
 import sys
 import tempfile
@@ -70,14 +72,19 @@ MINIMAL_SLIDE_DATA = {
 }
 
 
-def run_render(theme_slug=None, themes_dir=None):
-    """Invoke generate-html-slides.py with optional --theme-slug. Returns the
-    rendered HTML body as a string and the JSON status the script printed."""
+def run_render(theme_slug=None, themes_dir=None, slide_data=None, language=None):
+    """Invoke generate-html-slides.py. Returns the rendered HTML body as a
+    string and the JSON status the script printed.
+
+    Defaults to ``MINIMAL_SLIDE_DATA``; pass ``slide_data`` to render a
+    different deck.
+    """
     with tempfile.TemporaryDirectory() as tmpdir:
         tmp = Path(tmpdir)
         slide_data_path = tmp / "slide-data.json"
         output_path = tmp / "out.html"
-        slide_data_path.write_text(json.dumps(MINIMAL_SLIDE_DATA), encoding="utf-8")
+        payload = MINIMAL_SLIDE_DATA if slide_data is None else slide_data
+        slide_data_path.write_text(json.dumps(payload), encoding="utf-8")
 
         cmd = [
             sys.executable,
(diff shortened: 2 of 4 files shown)


claude-cogni-work:
## Ship-ready

Draft PR handed off for independent merger review. Issue #1795 → `svc:resolved`; this PR carries `svc:needs-review`.

**No review verdict was formed on this path.** No `svc:review-*` label, no `post-verdict`, no `reviewed-head-sha` marker — the independent review is the merger's to run.

### Handoff preflight — clean

`check-preflight.sh` exit 0, `status: ok`, `clean: true`, base `origin/main`, head `origin/fix/issue-1795-deterministic-parse-field-alignment`.

- touched: 4 paths, all `cogni-workspace`
- instruments resolved 7/7 (min 7); gating_applicable 4, gating_verdicts 4
- **passed 5** — `run-tests`, `frontmatter`, `breadcrumbs`, `version-bump`, `skill-length`
- **skipped 2, both `not_applicable`** — `agent-length` (no touched `agents/*.md`); `mutation-recipe` (no touched `*/tests/test[-_]*.sh`, and no `check-*.sh|.py` guard surviving at head)
- violations 0, advisories 0, contract_breaches 0

### Author self-grade against the issue-time contract — 23/23 satisfied

Graded against the contract committed on the **issue** at `2026-09-04T06:57:22Z` (`origin: triage`), written before this branch existed.

**graded 23 · satisfied 23 · violated 0 · unclear 0 · fixed 0 · residual none**

The grader re-ran the eval suite, the three repo gates, and its own renders of `EXAMPLE_BRIEF.md` and synthetic slide-data at base and head rather than reading the diff impressionistically, and independently reproduced mutation M1.

### Three things worth the merger's attention

1. **Item 4 is a preserve-invariant, not an edit.** `SKILL.md:49` already accepted `"4.0"` and `"4.1"` at base — a sibling landed it ahead of this issue. The diff deliberately leaves line 49 untouched, and the item is satisfied by the property still holding at head. An "expected edit is missing" reading of this item would be wrong.

2. **Item 17 has a disclosed, deliberate deviation.** Old-name-only slide-data renders **byte-identical slide markup** at base and head in both `en` and `de`, with no line removed — but the stylesheet gains 26 lines, because the CSS is emitted unconditionally and a new visual element needs rules. Both the author grader and the PR body treat this as satisfied-with-disclosure rather than silently. If you read the AC's "byte-identical" as covering the stylesheet too, this is the one item to argue about.

3. **One narrow eval gap the grader noted while still passing item 8.** Case 5 asserts four non-empty `quad-label` elements but does not separately assert `quad-bullets`. The bullets path is identical regardless of which key resolved the quadrant dict, so it carries no regression risk — recorded here rather than left for you to find.

### Open questions carried in the PR body

The PR body's `## Open questions` section raises three things this PR deliberately does **not** settle: a routing-intent contradiction with a comment in `parse-brief.py`, the defect class being narrowed rather than closed (`Icon`, `visual.chart` and `cta` are still authored and still unread), and alias ownership now being split between the parser and the renderer. None blocks the merge; all three would be separate PRs, and none has been filed as an issue.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1826
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer

**Summary:** Two independent perspectives converge on the same un-waived major defect: the deterministic Phase 1 parse writes to "{brief_dir}/cogni-visual/slide-data.json" but SKILL.md never creates that directory, so parse-brief.py's bare open() raises OSError, the resulting success:false envelope is silently caught by the skill's own new fallback rule, and the deterministic path the issue exists to install would never actually run on a fresh brief directory - verified empirically (zero makedirs in parse-brief.py, both sibling producers mkdir -p first, render-html-slides has zero mkdir occurrences). All 23 committed contract items grade satisfied; every other finding is minor and none gate.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (10)

- **[major]** (auto-fixable) `cogni-workspace/skills/render-html-slides/SKILL.md:62-70 (also reported as SKILL.md:64-68)` — The deterministic Phase 1 parse writes to --output "{brief_dir}/cogni-visual/slide-data.json", but nothing in SKILL.md creates that directory (no mkdir appears anywhere in the file, and Phase 1 is the first writer into cogni-visual/). parse-brief.py opens the output path directly with a bare open(args.output, "w") - no os.makedirs anywhere in the script - so a missing parent raises OSError, converted to {"success": false, "error": "cannot write output: [Errno 2] ..."}. SKILL.md's own newly-stated fallback rule (around line 90) then treats any success:false envelope as the LLM-fallback trigger, so the failure is invisible by design and every deck on a fresh brief directory silently renders through the LLM parse the issue was filed to retire. Both sibling producers guard this explicitly (story-to-slides/SKILL.md:392, story-to-web/SKILL.md:449 both mkdir -p first); render-html-slides/SKILL.md has zero mkdir occurrences.
  - *Suggested fix:* Add `mkdir -p "{brief_dir}/cogni-visual"` as the first line of the Phase 1 bash block, immediately before the parser invocation, and note it applies to the LLM-fallback write path too. Optionally state that a `cannot write output` envelope is an environment fault to fix, not a legitimate reason to fall back.

- **[minor]** `PR diff` — Diff is 662 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier - consider splitting if the scope allows.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/SKILL.md Phase 0 (lines 42-56)` — {brief_dir} is used as the anchor for every artifact path but is never defined, and the two readings diverge materially: the producer writes the brief into cogni-visual/, so the literal dirname of the discovered brief is already cogni-visual, making {brief_dir}/cogni-visual/ a doubled path.
  - *Suggested fix:* Add a line to Phase 0 step 5 defining brief_dir as the project root that contains cogni-visual/, without touching the artifact token (contract item 23 pins it literally).

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/evals/run.py:297-315` — Two of case 5's assertions are derived from the parsed data rather than pinned, so they cannot fail if the producer side stops emitting the field entirely (the closing-subtitle check no-ops when Subtitle is absent; the footer count compares 0 == 0 when no slide carries Source).
  - *Suggested fix:* Assert len(with_source) >= 2 (or that a closing Subtitle was parsed at all) before comparing.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/evals/run.py:377-379` — Eval coverage for the timeline and is-does-means contracts is one-sided: only the new Step-N map shape is exercised, and only the Label-wins direction of item 14 is tested. Both behaviours are preserved by construction in the code, so this is a guard-coverage gap rather than a regression.
  - *Suggested fix:* Add a Steps[] list-form slide, an old-shape Step-N{Title,Detail} slide, and a Label-less is-does-means slide asserting IST/MACHT/BEDEUTET to case 6.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/references/01-layout-renderers.md:128-131` — The three-options alias list names Name/Price/Features but not Label, which render_three_options also accepts for the title slot ahead of Name.
  - *Suggested fix:* Add Label to the three-options alias sentence, and consider one direction for all three blocks.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/references/01-layout-renderers.md:216-252` — The `slide-data.json shape` block claims to describe the shape both Phase 1 paths write, but omits the five top-level keys parse-brief.py emits per slide (slide_kind, source, cta, intent, visual). No content is lost - both readers accept the nested fields[...] form.
  - *Suggested fix:* Add "slide_kind": null and "source": null to the example slide object and mention the remaining three as parser-emitted passthroughs.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/scripts/generate-html-slides.py:13` — The module docstring still reads 'Input: slide-data.json (parsed from presentation-brief.md by LLM)'. After this change the LLM parse is the failure-only fallback and parse-brief.py is the primary producer.
  - *Suggested fix:* Reword to name `parse-brief.py --emit slide-data` as the producer, with the LLM parse as the documented fallback.

- **[minor]** (waived) `cogni-workspace/skills/render-html-slides/SKILL.md Features (lines 303-321)` — PRE-EXISTING (introduced_by_change: false). The ## Features block restates content the body and bundled references already own.
  - *Justification:* Pre-existing whole-skill drift, introduced_by_change: false - annotated, not gating.

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/render-html-slides/scripts/generate-html-slides.py:186 (NON_CONTENT_FIELDS)` — `Layout` is a seventh non-content key that still leaks into the generic fallback renderer as a visible labelled paragraph. The contract's own risk_watchlist explicitly scopes it out of the AC.
  - *Justification:* Explicitly scoped out of the AC by the contract's own risk_watchlist.

### Positive observations

- All 23 committed contract items graded satisfied with independent re-verification, including item 4 as a preserve-invariant and item 17's disclosed CSS-only deviation.
- The payload-gated effective_layout() is the correct resolution of the contract's sharpest trap - the silent-content-drop class is closed rather than re-committed.
- Mutation M1 confirms case 6 is the only case that can fail on the quadrant alias - the anti-vacuous-guard split is real.
- _first_present() tests presence, not truthiness, with old-first ordering at every call site.
- Eval cases 1-4 are byte-identical to base and the fixture was not edited to make an assertion pass.
- Every factual claim Phase 1 makes about the parser holds against the source.

### Recommendation

---

claude-cogni-work:

**PR:** https://github.com/cogni-work/insight-wave/pull/1826

**Summary:** All four perspectives (the synthetic oversized-diff check, plugin-validator, skill-reviewer, and the adapted skill-quality-reviewer) returned pass; seven of ten findings are explicitly waived with non-empty opt_out_justification and every un-waived finding is minor, so Rule 5b closes the PR as pass with ten advisory/waived minors surfaced for the record. This revise cycle closed the prior verdict's major (Phase 1 wrote into a directory nothing created, so a `cannot write output:` envelope silently routed every deck through the LLM fallback) plus six of its seven minors; the seventh, the 400-line diff bound, is declined as out of revise scope and re-surfaces here.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (10)

- **[minor]** `PR diff` — Diff is 730 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/SKILL.md §Phase 1, the 'Then add the metadata block' merge step (lines 89-92)` — The deterministic path instructs the model to read slide-data.json back, splice in a hand-assembled metadata object, and rewrite the file — but gives no pointer to the canonical merged shape. The LLM-fallback branch twelve lines later does point at it. The primary path is where hand-assembled JSON surgery is most error-prone, so it is the path that most needs the reference.
  - *Suggested fix:* Append to the sentence beginning 'Read {brief_dir}/cogni-visual/slide-data.json back': ' The merged result must match the shape both paths share — see the `slide-data.json shape` section of `references/01-layout-renderers.md`.' Touches no artifact token. [best_practice_ref: skill-creator §Writing Patterns (defining output formats) / §Progressive Disclosure]

- **[minor]** (waived) `cogni-workspace/skills/render-html-slides/SKILL.md §Phase 3 → 'Theme System v2 (tier-aware rendering)' (line 211)` — PRE-EXISTING (introduced_by_change: false). The paragraph is forward-looking maintenance narrative rather than executable instruction ('are the next increment', 'gated on a follow-up issue', 'tomorrow's behavior'). The executing model only needs today's behavior; roadmap state belongs in git history or references/theme-component-loader.md, which the Parameters table already points at.
  - *Justification:* Pre-existing whole-skill drift, introduced_by_change: false — annotated, not gating.

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/render-html-slides/SKILL.md:102-110` — The LLM fallback enumerates 8 per-slide keys, but the shape both paths are told to share is 13 (SLIDE_KEYS adds slide_kind, source, cta, intent, visual). The pointer to the shape reference is there, but the enumeration above reads as complete. No user-observable fault today: the renderer resolves slide_kind/source via a legacy nested fallback and reads none of cta/intent/visual.
  - *Justification:* Quality observation only. The renderer's legacy nested fallback means the shorter fallback enumeration produces a correctly-rendered deck today, and the section already points at the authoritative shape reference.

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/render-html-slides/SKILL.md:109` — The two Phase 1 paths disagree on diagram_mermaid population: the fallback says 'if it contains Mermaid syntax' while the deterministic path assigns it unconditionally (parse-brief.py:386). Risk is low because the grammar defines Diagram: only for the three Mermaid layouts, but it is a residual alignment gap in a PR whose stated theme is producer field alignment.
  - *Justification:* No reachable fault for in-grammar briefs, and the deterministic path — now the default — is the one that is correct; the clause only describes the fallback branch.

- **[minor]** (auto-fixable) `cogni-workspace/skills/render-html-slides/SKILL.md:63` — The rationale clause 'it is the same parser the other brief consumers use, so this renderer cannot drift away from them' asserts a state that does not exist at this head. A repo-wide search returns parse-brief.py references only in the script itself, cogni-workspace/tests/test-parse-brief.sh, this skill's SKILL.md, its evals/run.py, and prose in generate-html-slides.py / references/01-layout-renderers.md. render-html-slides is the sole consumer, so the stated anti-drift invariant is aspirational, not actual — a future author would believe parser changes are already coordinated across consumers when they are not.
  - *Suggested fix:* Reword to the true, still-motivating form, or drop the trailing 'so this renderer cannot drift away from them' clause and keep 'Run the shared parser rather than reading the brief yourself.'

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/render-html-slides/evals/run.py:45` — PLUGIN_ROOT = SKILL_DIR.parent.parent carries the trailing comment '# cogni-visual/' but resolves to cogni-workspace/. Pre-existing, but this diff makes new code depend on that variable (EXAMPLE_BRIEF, PARSER), so a reader of the new case-5 setup is now actively misled.
  - *Justification:* Pre-existing comment outside the diff's own hunks; cosmetic.

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/render-html-slides/references/01-layout-renderers.md:262-266` — The sentence naming cta/intent/visual as unread is true only for the lowercase keys — render_closing_slide does read an uppercase CTA from fields (_first_present(f, ("CTA","Metadata"), "") at line 655), a different case-sensitive key. A case-insensitive grep makes the doc look false.
  - *Justification:* Prose precision on a true-as-written claim; no behavioural consequence.

- **[minor]** (waived) (auto-fixable) `cogni-workspace/skills/render-html-slides/scripts/generate-html-slides.py:1853-1872 (assemble_html title guard) with effective_layout at 274-292` — The Slide-Kind: references reroute silently drops a rerouted slide's authored H2 — reproduced directly: a slide headlined MYHEADLINE renders as data-layout="references" with the headline absent. The payload-less reroute is guarded (case 6 slide 9 proves teeth) but the branch that does fire loses the headline.
  - *Justification:* Unreachable from every producer template in the repo (verified against both story-to-slides reference templates and EXAMPLE_BRIEF slide 13), so no deck can hit it today; latent rather than live.

- **[minor]** (waived) `cogni-workspace/skills/render-html-slides/scripts/generate-html-slides.py:887-903,1233-1238 (generate_css)` — AC2's literal falsifier is tripped: base-vs-head render of cases 1-4 differs by 26 added lines / +590 bytes of unconditional .source-footer / .source-footer-raised / .timeline-number CSS. Markup is byte-identical (0 removed lines, no diff outside the <style> block); evals cannot catch it because cases 1-4 assert substrings, never a pinned byte baseline.
  - *Justification:* Zero user-observable effect: 0 removed lines, no diff outside the <style> block, no element in cases 1-4 references the added classes. The CSS is required by the source-footer feature the same issue asks for. Literal-vs-intent gap on a falsifier, not a regression.

### Positive observations

- Both new eval cases proven red at base by executing the head run.py against a copy of the pre-change renderer — case 5 fails on absent closing subtitle, 0-of-2 source footers and all three badges; case 6 fails on 30+ assertions. Guards with teeth. (plugin-dev:plugin-validator)
- Case 6 exists precisely because the author noticed case 5 would be a vacuous green (parse-brief.py normalizes Q{n} upstream) — the reasoning a reviewer usually supplies, written down by the author. (plugin-dev:plugin-validator)
- Two assertions carry explicit '# Pinned:' comments guarding against their own vacuity — the two failure modes that most often turn a shape assertion into a no-op. (plugin-dev:plugin-validator)
- The AC1 slide count is derived from the fixture itself rather than hardcoded, so the case cannot go stale. (plugin-dev:plugin-validator)
- Every operator claim SKILL.md makes about parse-brief.py checks out against the script, including the cannot-write-output string and the absent parent-dir creation that motivates mkdir -p. (plugin-dev:plugin-validator)
- _first_present tests presence, not truthiness, which is what makes the old-names-keep-precedence half of AC2 actually hold. (plugin-dev:plugin-validator)
- Repo gates run rather than trusted: run-plugin-tests 21/21, check-external-dispatch 0, check-breadcrumbs 0, check-version-bump 0 violations with mode=pr. (plugin-dev:plugin-validator)
- Scope fences respected: cogni-visual/ unchanged, speaker-notes untouched, no plugin.json version line modified. (plugin-dev:plugin-validator)
- The cannot-write-output carve-out is the sharpest thing in this diff: both a parse failure and a write failure return success:false from the same emit(), so without it the fallback would fire on a broken write and silently hide it — a textbook fail-open. (plugin-dev:skill-reviewer)
- The fallback rule is stated as a positive precondition rather than a negative, which closes AC1's falsifier directly. (plugin-dev:skill-reviewer)
- The metadata gap is handled honestly — the consequence is stated explicitly rather than left as a silent trap, and the four-bullet mapping matches the real frontmatter of EXAMPLE_BRIEF.md. (plugin-dev:skill-reviewer)
- mkdir -p is justified rather than asserted, and the LLM branch runs its own so the two paths do not couple — independence that is easy to omit. (plugin-dev:skill-reviewer)
- The 4.0/4.1 acceptance carries its reason inline and is corroborated by the parser warning rather than rejecting on the 4.0 shape. (plugin-dev:skill-reviewer)
- Length is well within budget (18,650 / 22,500 at complexity 11) despite seven phases, because per-layout detail lives in the three references/ files. (plugin-dev:skill-reviewer)
- AC1 satisfied and hardened beyond the literal ask: the cannot-write-output envelope is explicitly carved out as not-a-parse-failure, with the why spelled out. (cogni-service:skill-quality-reviewer)
- AC2 satisfied: Phase 0 step 4 accepts 4.0 or 4.1 and states the reason rather than asserting a bare rule. (cogni-service:skill-quality-reviewer)
- AC3 satisfied: the cogni-visual/ output directory name is unchanged at 5 loci and the two-section speaker-notes format is intact. (cogni-service:skill-quality-reviewer)
- The change explains its own why at each rule rather than stacking MUSTs — the reasoned-instruction register skill-creator asks for. (cogni-service:skill-quality-reviewer)
- The fallback branch is defensively self-contained: it re-runs its own mkdir -p with the reason stated, so the two Phase 1 paths do not couple. (cogni-service:skill-quality-reviewer)
- Progressive disclosure holds — 18,650 against a 22,500 cap, and all three referenced files exist on disk, so no pointer dangles. (cogni-service:skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

Addressed the `NEEDS_REVISION` verdict ([#issuecomment-5537672553](https://github.com/cogni-work/insight-wave/pull/1826#issuecomment-5537672553), reviewed head `c2b506e`) in commit `6a20d281`. Fresh verdict: ✅ **PASS** ([#issuecomment-5538324469](https://github.com/cogni-work/insight-wave/pull/1826#issuecomment-5538324469)).

All 8 un-waived findings are accounted for — 7 fixed, 1 declined on the record.

### Resolved

| Finding | Where | What changed |
|---|---|---|
| **major** — Phase 1 wrote into `{brief_dir}/cogni-visual/` but nothing created it, so `parse-brief.py`'s bare `open()` returned `cannot write output:` under `success: false`, which the new fallback rule read as the LLM-fallback trigger — the deterministic parse failed invisibly on every fresh brief directory | `SKILL.md:62-70` | Added `mkdir -p "{brief_dir}/cogni-visual"` to the Phase 1 bash block **and** to the LLM-fallback write branch (so neither path depends on the other having run), and carved `cannot write output:` out of the fallback trigger as an environment fault to fix and re-run. Verified against the source: `parse-brief.py:518` opens bare, `:522` emits that exact string, and the file had **0** `mkdir` occurrences. |
| minor — `{brief_dir}` anchored every artifact path but was never defined; the `dirname` reading doubles the path | `SKILL.md` Phase 0 | New Phase 0 step 6 defines it as the project directory *containing* `cogni-visual/`, naming the doubling trap. The pinned artifact token is untouched (5 occurrences). |
| minor — two of case 5's assertions were derived from parsed data, so they no-op if the producer stops emitting the field | `evals/run.py:297-315` | Pinned both: case 5 now fails unless a closing `Subtitle` was actually parsed, and unless `len(with_source) >= 2` before the footer comparison. Each proved red under its own mutation. |
| minor — timeline / is-does-means coverage was one-sided (only the `Step-N` map shape, only the Label-wins direction) | `evals/run.py:377-379` | Added three slides to case 6: a `Steps[]` list-form timeline, an old-shape `Step-N{Title,…}` precedence slide, and a Label-less is-does-means slide asserting `IST`/`MACHT`/`BEDEUTET` under `--language de`. All three proved red under their own mutation, each reddening only its own slide. |
| minor — three-options alias list omitted `Label` | `references/01-layout-renderers.md:128-131` | Names the implemented precedence `Title` → `Label` → `Name`, stated once for all three alias groups. |
| minor — the `slide-data.json shape` block omitted five parser-emitted keys | `references/01-layout-renderers.md:216-252` | Example now carries all thirteen `SLIDE_KEYS` keys, with `SLIDE_KEYS` named as the single contract rather than the count being restated in prose. |
| minor — stale module docstring naming the LLM as producer | `scripts/generate-html-slides.py:13` | Reworded to name `parse-brief.py --emit slide-data` as the producer, LLM free-parse as the documented fallback. |

### Declined

- **minor — "Diff is 662 lines (bound: 400); consider splitting."** Not applied, and not silently dropped. The PR is already pushed and contract-graded at `c2b506e`; a retroactive split would mean rewriting or re-opening PRs and would strand the review contract committed against this one, and the finding is not marked auto-fixable — diff size is a plan-time decomposition signal, not a revise-cycle delta. Stated plainly: this revision *adds* ~68 lines, so the diff is now 730 and the bound moves further away. That is the cost of closing the major, and the fresh verdict re-surfaces the finding at the new count rather than burying it.

### Gates

- `evals/run.py` — 6/6, exit 0 · `run-plugin-tests.py --filter cogni-workspace` — 21/21, exit 0
- `check-external-dispatch.py`, `check-breadcrumbs.py`, `check-version-bump.py` — exit 0 (`status: ok`, 0 violations, `plugins_touched: [cogni-workspace]`)
- `check-frontmatter.sh`, `check-breadcrumbs.sh` (cogni-workspace) — exit 0
- `check-token-scope.sh --strict` — `over_scoped` (`gist`, `read:org`, `workflow`), cleared with the settled `--allow-overscoped` opt-out. Unchanged runner-level posture, same as the original resolve.

### Review notes

- A `/simplify` pass and a `skill-simplifier` prose pass ran before the commit. Three converging cleanup findings were applied: the alias precedence was stated three ways, the `slide-data.json` example needed a paragraph to explain its own incompleteness, and the `mkdir` obligation was carried across ~30 lines as prose instead of being placed in the branch that needed it.
- **One claim I wrote and then falsified myself:** a first draft said the renderer reads the five keys past `citations` only through the nested `fields` object. That is false — `generate-html-slides.py:283` reads top-level `slide_kind` first and `:235` does the same for `source`. Corrected before commit; `cta`/`intent`/`visual` are the ones read in neither form.
- Body length 18,650 / 22,500 chars (82.9%), below the 90% approach zone, so no net-growth budget engaged.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1826
**Perspectives consulted:** cogni-service:service-review, cogni-service:contract-grading-reviewer, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer

**Summary:** All 5 perspectives passed against the 23-item committed review contract — a triage-origin checklist adopted under the by-chronology diff-blindness proof. 23/23 items graded satisfied by both Bash-capable perspectives, independently (not on trust): the eval suite re-run 6/6, mutations M1/M2/M6 reproduced, a base-vs-head old-name-only render diff in both en and de, and a live end-to-end render of the real EXAMPLE_BRIEF.md. No item graded violated anywhere, no INTENT-UNCLEAR finding, so no split judgment. The 7 remaining findings are all minor and all category:preference — merger-added additions off the contract locus, demoted to advisory — so none gate.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (7)

- **[minor]** (advisory-addition) `PR diff` — Diff is 730 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/render-html-slides/SKILL.md §Phase 1, metadata mapping bullets (lines 84-88)` — The title/subtitle bullets say to read 'slide 1's fields.Title' / 'slide 1's headline', but the surrounding text describes the payload of --emit metadata, which carries no slides key at all. The source of 'slide 1' is only implied by the read-back instruction one paragraph later. A reader taking the bullets at face value looks for a title inside the --emit metadata payload, does not find one, and falls back to the generator's 'Presentation' default — precisely the silent degradation this change exists to prevent.
  - *Suggested fix:* Rewrite the lead-in to name both sources — the frontmatter block and the first entry of the slides array in the slide-data.json just written — and change the two slide-derived bullets to say 'the first slide's fields.Title, else its headline'. [best_practice_ref: skill-creator §Writing Style]

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/render-html-slides/SKILL.md §Phase 2 (lines 137-175) and §Phase 3 Theme System v2 (lines 209-215)` — PRE-EXISTING (introduced_by_change: false). Two blocks of reference-grade detail sit inline in the body while a populated references/ directory exists: Phase 2 inlines a ~40-line design-variables.json example then names its authoritative .schema.json two lines below, and the Theme System v2 subsection carries the full reason-code catalogue while already pointing at theme-component-loader.md. The keyboard map is also spelled twice.
  - *Justification:* Pre-existing whole-skill drift, introduced_by_change: false — annotated, not gating. Body is 18650/22500 chars, not over budget.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/render-html-slides/SKILL.md:102-119` — The LLM-fallback extraction list enumerates 8 per-slide keys while the shape it points at publishes 13, adding slide_kind, source, cta, intent and visual. The two lists disagree, so a fallback-parsed deck is likely to omit the top-level slide_kind/source the deterministic path emits. No content is actually lost because the renderer accepts the legacy nested form for both — hence preference, not defect — but the same paragraph claims the fallback writes 'in the shape both paths share', so the enumerations should agree.
  - *Suggested fix:* Either extend the fallback bullet list with the five newer top-level keys, or drop the enumeration entirely and defer wholly to the reference's shape block.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/render-html-slides/SKILL.md:63` — The rationale sentence asserts a present-tense fact the repo does not support: 'it is the same parser the other brief consumers use, so this renderer cannot drift away from them.' At head, a repo-wide search for parse-brief returns only 6 files — the script itself, this SKILL.md, references/01-layout-renderers.md, generate-html-slides.py, evals/run.py and tests/test-parse-brief.sh. No other skill or script invokes it, so there are no 'other brief consumers' today. The motive is sound; only the claim about existing peers is unsupported.
  - *Suggested fix:* Restate the rationale in terms of the shared contract rather than asserting peer consumers that do not yet exist at head.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/render-html-slides/SKILL.md:79-82` — Nesting-level imprecision on the second parse invocation. Line 74 establishes that the {success,data,error} envelope prints on stdout and only --output writes the bare payload. The --emit metadata call at :79 runs WITHOUT --output, so its result arrives as the envelope — but :82 says 'That returns {version, frontmatter, generation_metadata, cta_summary, unowned_sections, warnings}', describing the payload one level below where it actually appears. An executor reading .frontmatter off stdout instead of .data.frontmatter gets nothing, and the failure mode is a silently empty metadata block — exactly the 'Presentation'/today's-date degradation this section exists to prevent.
  - *Suggested fix:* Say 'That prints an envelope whose `data` is {version, frontmatter, ...}', or add --output to the metadata call as well, mirroring the first invocation.

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-workspace/skills/render-html-slides/scripts/generate-html-slides.py:NON_CONTENT_FIELDS` — `Layout` is a seventh non-content key that render_generic_slide still emits as visible body text. Verified by direct render: a slide with an unrecognized layout whose fields carry `Layout: LAYV` produces a visible labelled paragraph alongside the legitimately-suppressed six.
  - *Justification:* Pre-existing at base — render_generic_slide skipped only Speaker-Notes and Bottom-Banner before this diff, so `Layout` printed identically at 4104c560. Not a regression the diff introduces. The contract's risk_watchlist names it explicitly and rules it out of scope; checklist item 16 enumerates exactly six keys, all six suppressed.

### Positive observations

- item 7/19 proven non-vacuous by reproduced mutation: reverting the Q{i} alias arm reddens ONLY case 6 (direct slide-data) while case 5 (through the parser) stays green — matching the PR body's recorded M1 entry, independently reproduced. (cogni-service:contract-grading-reviewer)
- item 8 verified by live render of the real EXAMPLE_BRIEF.md: status ok, 13 slides == 13 '## Slide' headings, 4 quad-cards on the Buying Center slide, closing subtitle present, 2 source footers matching the 2 Source-carrying slides. (cogni-service:contract-grading-reviewer)
- item 15 verified: slide 13 (Slide-Kind references, no References payload) stays data-layout=two-columns-equal with all three source strings intact — the silent-drop class the issue was filed against does not occur. (cogni-service:contract-grading-reviewer)
- item 17 verified by an independent base-vs-head render diff over an old-name-only fixture in both en and de: 0 lines removed, 26 added, every added line inert CSS never matched by old-name-only slides. (cogni-service:contract-grading-reviewer)
- item 21 verified at head: run-plugin-tests 21/21, check-external-dispatch clean, check-breadcrumbs clean, check-version-bump clean with no version touch. (cogni-service:contract-grading-reviewer)
- item 14 verified: slide 7 renders IS/DOES/MEANS from the authored Label under --language de, and eval case 6 slide 15 covers the Label-less direction asserting IST/MACHT/BEDEUTET. (cogni-service:contract-grading-reviewer)
- The payload-guarded effective_layout() is the correct resolution of the contract's #2 risk, documented in both the code docstring and the PR body, and the eval encodes the guard in BOTH directions (case 6 slide 8 payload-present routes, slide 9 payload-absent does not) so neither direction can rot. (plugin-dev:plugin-validator)
- _first_present() tests PRESENCE, not truthiness, with the reason stated — closing the precedence-inversion regression the risk_watchlist warned about. (plugin-dev:plugin-validator)
- Eval case 5 derives the expected slide count from the fixture instead of hardcoding 13, so the assertion tracks the fixture. (plugin-dev:plugin-validator)
- Two case-5 assertions are explicitly PINNED against going vacuous, each commented with the failure mode it guards. (plugin-dev:plugin-validator)
- New markup got new CSS: .source-footer, .source-footer-raised, .source-footer a and .timeline-number all have rules, and the raised variant is actually exercised. (plugin-dev:plugin-validator)
- The PR body's mutation table is honest — M1, M2 and M6 all reproduced independently and matched exactly, including the subtle M1 claim that case 5 stays green. (plugin-dev:plugin-validator)
- The PR body pre-emptively flags the one place the AC's 'byte-identical' wording does not literally hold (+26 CSS lines) rather than letting a reviewer discover it; the independent diff confirmed the characterization precisely. (plugin-dev:plugin-validator)
- The Parameters table now defines `{slug}`, previously an undefined token in the default output_path. (plugin-dev:plugin-validator)
- The worktree was left byte-clean after each mutation, with the final eval re-run 6/6 green at the unmodified head. (plugin-dev:plugin-validator)
- Phase 1 is the contract's escalation answered in prose; the `cannot write output:` carve-out is a genuine over-delivery, since that envelope is also success:false and treating it as a fallback trigger would have hidden a broken write behind an LLM-parsed deck. (plugin-dev:skill-reviewer)
- The mkdir -p justification is a precise verified claim rather than a decorative one: parse-brief.py:518 opens --output with no os.makedirs. (plugin-dev:skill-reviewer)
- The base's inline slide-data JSON block was removed and replaced with a reference pointer, and a new ## Additional Resources section names all three reference files — base had zero reference pointers. All three targets exist on disk. (plugin-dev:skill-reviewer)
- Two prose claims this change would otherwise have falsified were corrected in the same diff: the IS/DOES/MEANS parsing rule, and the Phase 3 backwards-compat sentence loosened from a hard-counted 'three regression cases'. (plugin-dev:skill-reviewer)
- Contract item 1 satisfied: Phase 1 invokes parse-brief.py with BOTH required flags, so the issue's shorthand falsifier does not apply. (cogni-service:skill-quality-reviewer)
- Contract item 2 satisfied: the fallback is stated as an envelope condition and closes the loophole explicitly — no 'if the output looks wrong' escape hatch. (cogni-service:skill-quality-reviewer)
- Line 96 anticipates the exact mis-routing a reader would otherwise make, with the reason spelled out — explain-the-why applied to a non-obvious edge, not a stacked MUST. (cogni-service:skill-quality-reviewer)
- Contract item 4 preserve-invariant holds, and item 23's SKILL.md clause holds — {brief_dir}/cogni-visual/ unchanged at every locus. (cogni-service:skill-quality-reviewer)
- The LLM fallback branch was kept consistent with the new deterministic path by reference rather than by copy, re-stating only the branch-local mkdir -p requirement with the reason it cannot depend on the deterministic block having run. (cogni-service:skill-quality-reviewer)
- Progressive disclosure is real: all three references/0*.md files exist on disk and the shape and Bottom-Banner contracts are routed to them rather than inlined. (cogni-service:skill-quality-reviewer)

### Recommendation

---

Issue #1877, opened 2026-09-06
cogni-knowledge: wiki-verifier's grounding normalization strips numbered markers only, so an author-date marker depresses the grounded signal


sdh07:
## Plugin
cogni-knowledge

## Current behavior
`agents/wiki-verifier.md:87` computes the per-citation `grounded` signal by normalizing the
cited `draft_sentence` and the claim's `excerpt_quote` and requiring the normalized excerpt to
be a contiguous substring of the normalized sentence. That normalization is specified as
"NFC, case- and whitespace-folded, inline `[N]`/`<sup>` citation markers stripped".

Author-date citation rendering has now shipped, so a draft planned with
`--citation-format apa|mla|harvard` carries markers of the shape `([Author, Year](url))`,
`([Author](url))` and `([Author Year](url))`. None of those is an `[N]` or a `<sup>`, so the
normalization leaves the marker in the compared string. The excerpt then has to cover a
string that still contains the marker, so `grounded` can come back `false` on a citation that
is in fact grounded — and `verify-store.py merge` aggregates that into the headline grounding
rate.

Scope is the `grounded` signal only. The **verdict** (`verbatim` / `paraphrase` /
`unsupported` / `synthesis`) is unaffected: it is scored against the claim, not by
marker-stripping. So this depresses a reported metric; it does not mis-verify a citation.

## Why it was not fixed alongside the author-date work
`agents/wiki-verifier.md` was explicitly fenced out of the parent change. Its acceptance
criteria required, verbatim, that the file be unchanged, on the stated rationale that the
verifier "scores the verbatim `draft_sentence` and never re-parses marker shape, so it is
already family-agnostic". That rationale is not accurate at base — `:87` does strip markers —
but the disposition was pre-resolved, so the residual was shipped knowingly and recorded
rather than silently absorbed.

## Desired behavior
The verifier's grounding normalization is citation-family-aware, so an author-date marker is
stripped from the compared string exactly as a numbered marker is.

The primitive already exists and needs no new code: `_knowledge_lib.strip_citation_markers`
dispatches on `citation_family`, and `citation-store.py` and `verify-store.py` already call it.
The verifier is an LLM agent, so the change is to its prose — and it needs a channel for the
format, which it does not have today: its dispatch threads `PROJECT_PATH`, so reading
`.metadata/plan.json::citation_format` in Phase 0 is the same shape the `revisor` already
uses.

## Acceptance criteria
[...]

Issue #1755, opened 2026-09-02
cogni-knowledge: give URL-less citations a native author-date marker instead of the numbered <sup>[N]</sup> fallback


claude-cogni-work:
## Plugin
cogni-knowledge

## Current behavior
#1749 wires `apa`/`mla`/`harvard` as author-date citations end to end, but deliberately keeps the
family-blind hard rule at `references/citation-formats.md:43`: a source with no URL at all — a
synthesis, distilled (`dcl-NNN`) or question-node (`acl-NNN`) page whose `sources:` are `wiki://`
backlinks — renders a plain `<sup>[N]</sup>` superscript, with the reference-list entry carrying the
`[[<slug>]]` wikilink.

That rule is family-blind at base, so keeping it is documented rather than invented, and it is what
lets #1749 ship without reversing a merged sibling's design decision. But under the author-date
family it leaves two bounded residuals:

1. The `[N]` numeral is **display-only**. It cannot be resolved against the alphabetical, un-numbered
   author-date reference list, because #1749's AC 5 requires the numbered renumber pass to be
   bypassed on that family.
2. `strip_citation_markers(text, "apa")` does **not** strip `<sup>[N]</sup>` — the author-date
   pattern (`scripts/_knowledge_lib.py:698-700`) is scheme-anchored by design. So such a sentence's
   prefilter core retains the marker, and `verify-store.py prefilter` declines to classify it
   verbatim and routes it to the LLM verifier. That is the prefilter's documented fail-safe posture
   (`skills/knowledge-verify/SKILL.md:205`) — a hit-rate cost, never a wrong verdict.

This is not an edge case: the composer *prefers* the distilled or question-node citation on a
converged fact, so URL-less citations are a routine path.

## Desired behavior
URL-less citations carry a native author-date marker under `apa`/`mla`/`harvard`, so no numbered
`[N]` marker survives anywhere in an author-date draft and the prefilter can strip it.

## Motivation
Closing this needs a **strip-only** author-date pattern admitting the destination-less
`([Author, Year])` form — mirroring the numbered family's own split between the permissive
`_SUP_MARKER_RE` (`scripts/_knowledge_lib.py:630`, strip) and the strict `_INLINE_CITATION_URL_RE`
(:653, extract).

That reverses the single-pattern rationale #1748 recorded at `scripts/_knowledge_lib.py:702-708`
("strip and extract share ONE pattern, deliberately"). Overturning a merged sibling's explicitly
[...]

Issue #1748, opened 2026-09-01
cogni-knowledge: add citation-family-aware primitives to _knowledge_lib.py and make CITATION_FAMILY load-bearing


claude-cogni-work:
## Plugin
cogni-knowledge

## Epic context
Child 2 of 3 under tracker #1744 (`--citation-format apa|mla|harvard` is accepted and persisted but silently renders IEEE). The maintainer ruling chose Option 1 — honour the formats — which requires every marker primitive to understand the author-date shape, not just the numbered one. Runs in parallel with child 1 (ingest metadata); both are prerequisites of child 3 (the end-to-end flip).

This child is pure library code with **no consumer** until child 3 wires it in, so it cannot change rendering. Until child 3 lands, `apa`/`mla`/`harvard` keep falling through to numbered IEEE and every prose site documenting that fall-through stays accurate.

## Use case
`scripts/_knowledge_lib.py:556` already defines `CITATION_FAMILY`, classifying `ieee`/`chicago` as `numbered` and `apa`/`mla`/`harvard` as `author_date`. Nothing reads it: grep across `cogni-knowledge/` returns the definition itself, three assertions in `tests/test_knowledge_lib.sh:1061-1063`, and CHANGELOG entries — zero functional consumers. Every marker primitive is hardcoded to the numbered `<sup>[N](url)</sup>` shape:

- `_SUP_CITATION_RE` (line 613) — `<sup>\[(\d+)\]`
- `_SUP_MARKER_RE` (line 617) — whole-marker match
- `strip_inline_citation_markers` (line 620)
- `_INLINE_CITATION_URL_RE` (line 640) and `extract_inline_citation_urls` (line 645)
- `renumber_inline_citations` (line 843)

Per `references/citation-formats.md:39` the author-date inline shapes are `([Author, Year](url))` (APA), `([Author](url))` (MLA), `([Author Year](url))` (Harvard). None of them contains `[N]`, so every primitive above returns nothing useful for them. Until these primitives exist, no consumer can be made family-aware — which is why this is a prerequisite rather than part of the flip.

## Proposed solution
Add author-date counterparts inside `scripts/_knowledge_lib.py`, dispatched on `CITATION_FAMILY`:

[...]

Issue #1747, opened 2026-09-01
cogni-knowledge: capture author / publication-date metadata at ingest, with a legacy-base derivation fallback


claude-cogni-work:
## Plugin
cogni-knowledge

## Epic context
Child 1 of 3 under tracker #1744 (`--citation-format apa|mla|harvard` is accepted and persisted but silently renders IEEE). The maintainer ruling on #1744 chose Option 1 — honour the formats — so APA/MLA/Harvard need a real author-date tuple per source. No `author:` or publication-year field exists anywhere in the pipeline today, so this metadata must exist before any author-date marker can render. Runs in parallel with child 2 (`_knowledge_lib.py` primitives); both are prerequisites of child 3 (the end-to-end flip).

This child is deliberately invisible to rendering. Until child 3 lands, `apa`/`mla`/`harvard` keep falling through to numbered IEEE and every prose site documenting that fall-through stays accurate.

## Use case
A user runs `knowledge-plan --citation-format apa` and expects author-date citations. APA, MLA and Harvard each need an (author, year) tuple per cited source. `agents/source-ingester.md`'s Phase-3 frontmatter template (lines 117-131) carries only `title`, `publisher`, `fetch_method`, `fetched_at`, `content_hash` — no author, no publication year. Without this field pair, no author-date marker can be rendered at all.

The second half of the problem is the installed base. #1744's acceptance criterion 6 requires that a base already persisting `citation_format: apa` keeps working with no migration step demanded of the user — but every page ingested before this change lands has no `author:` and no `year:`. Something must resolve a usable tuple for those pages without erroring and without prompting.

## Proposed solution
Two additive pieces, no consumer wiring.

1. **Optional frontmatter at ingest.** Add optional `author:` and `year:` (or `published_date:`) keys to the Phase-3 wiki-page frontmatter template in `agents/source-ingester.md` (lines 117-131) and to the YAML frontmatter rules immediately below it. Both keys are **optional** — a required key would break every page already on disk and every fetch where no byline is discoverable. If byline detection is better placed at claim extraction, add the corresponding emission to `agents/claim-extractor.md`; if it is not, leave that agent untouched and say so in the PR body.

[...]

PR #1878, opened 2026-09-06
cogni-knowledge: render apa/mla/harvard as real author-date citations end to end

### LINKED ISSUE #1744: cogni-knowledge: --citation-format apa|mla|harvard is accepted and persisted but silently renders IEEE
## Plugin
cogni-knowledge

## Current behavior
`knowledge-plan` accepts `--citation-format apa`, `mla` and `harvard`, validates them, and persists the choice into `plan.json::citation_format` — then every downstream surface silently renders the numbered IEEE shape instead.

The fall-through is documented at each surface, but only in maintainer-facing prose:

- `skills/knowledge-plan/SKILL.md:34` — the flag table marks `apa`/`mla`/`harvard` as "staged author-date".
- `agents/wiki-composer.md:169` — "`apa`/`mla`/`harvard` (staged author-date): render the **numbered** IEEE shape for now".
- `skills/knowledge-finalize/SKILL.md:1095` — "Does NOT yet render **author-date** `apa`/`mla`/`harvard` — those are accepted + persisted but fall through to the numbered IEEE string".
- `scripts/_knowledge_lib.py:545` — "the staged author-date follow-up (parsed + accepted here so a base can persist …)".

Nothing surfaces this to the user at the point of choice. A user who runs `knowledge-plan --citation-format apa` gets an accepted flag, a persisted setting, and IEEE output, with no warning at any step.

This was the one genuinely live item surfaced by an audit of all 13 forward-looking items in `references/absorption-roadmap.md`; the other twelve were shipped, moot, superseded or unfileable. It is filed here on its own merits as a user-visible defect, not as a roadmap carry-over.

## Desired behavior
The accepted set and the honoured set agree. Either direction closes it:

1. **Honour the formats.** Render author-date inline citations and reference lists for `apa` / `mla` / `harvard`, which requires making the `knowledge-finalize` renumber pass and the verify / reviewer / revisor scans citation-family-aware rather than numbered-only.
2. **Reject them at the boundary.** `knowledge-plan` fails fast on an unimplemented format, naming `ieee` and `chicago` as the wired set. Cheap, honest, and reversible when option 1 lands.

Option 2 is the smaller change and removes the silent-wrong-output class immediately. Option 1 is the feature. A maintainer decision on which to ship is the first step — this issue should not be resolved by picking one unilaterally.

## Motivation
A flag that is accepted, validated and persisted, then ignored, is worse than an unsupported flag: the user has positive confirmation their choice was recorded. The defect is silent at runtime and only discoverable by reading maintainer prose in four separate files.

## Ruling (maintainer, 2026-09-01)

**Option 1 — honour the formats.** `apa`, `mla` and `harvard` render as real author-date citations end to end. Option 2 (reject at the boundary) is rejected: `scripts/_knowledge_lib.py:545` accepts and persists the value *specifically so a base can hold it*, so failing fast at `knowledge-plan` would break bases already carrying `apa` in order to fix a rendering defect. A warn-only variant was considered and also rejected — it makes the fall-through honest but leaves the requested format unimplemented indefinitely.

**Scope note for triage.** This ruling makes the issue feature-sized rather than a defect fix. Honouring author-date requires the `knowledge-finalize` renumber pass and the verify / reviewer / revisor citation scans to become citation-family-aware rather than numbered-only, so a `decompose` strategy routing the parent to `service-plan` is a legitimate outcome. The severity label is left at `S3-minor` because user impact is unchanged; the complexity is what moved.

**Interim behaviour is unchanged and must stay honest.** Until this lands, `apa`/`mla`/`harvard` continue to fall through to numbered IEEE, and the four prose sites that document the fall-through stay accurate. No partial state may ship in which a format is advertised as rendered while it still falls through.

## Acceptance criteria
- [ ] A draft planned with `--citation-format apa` renders author-date inline citations and an author-date reference list end to end. Falsified by numbered `[N]` markers surviving anywhere in the output for that format.
- [ ] The same holds for `mla` and `harvard`. Falsified by any of the three rendering the IEEE numbered shape.
- [ ] `ieee` and `chicago` output is unchanged. Falsified by any diff in the rendered inline markers or reference-list shape for either format.
- [ ] The set of `citation_format` values `knowledge-plan` accepts equals the set rendered as requested. Falsified by any value being accepted and persisted while rendering a different citation family.
- [ ] The `knowledge-finalize` renumber pass and the verify / reviewer / revisor citation scans are citation-family-aware, not numbered-only. Falsified by an author-date draft whose citations score as unsupported purely because a scan expected `[N]`.
- [ ] An existing base persisting `citation_format: apa` keeps working — it renders author-date after the change, with no migration step required of the user. Falsified by an error, a prompt, or a silent reset to `ieee` on such a base.
- [ ] The four prose sites that document the staged fall-through (`skills/knowledge-plan/SKILL.md:34`, `agents/wiki-composer.md:169`, `skills/knowledge-finalize/SKILL.md:1095`, `scripts/_knowledge_lib.py:545`) agree with the shipped behaviour. Falsified by any of them still describing a fall-through that no longer exists.
- [ ] No path outside `cogni-knowledge/` is changed, and no `version` field changes in the diff.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` passes.

## Tracks

- [ ] #1747 cogni-knowledge: capture author / publication-date metadata at ingest, with a legacy-base derivation fallback
- [ ] #1748 cogni-knowledge: add citation-family-aware primitives to _knowledge_lib.py and make CITATION_FAMILY load-bearing
- [ ] #1749 cogni-knowledge: render apa/mla/harvard as real author-date citations end to end and reconcile the fall-through prose

## Close gate

- [ ] #1747 cogni-knowledge: capture author / publication-date metadata at ingest, with a legacy-base derivation fallback
- [ ] #1748 cogni-knowledge: add citation-family-aware primitives to _knowledge_lib.py and make CITATION_FAMILY load-bearing
- [ ] #1749 cogni-knowledge: render apa/mla/harvard as real author-date citations end to end and reconcile the fall-through prose

This issue was decomposed by cogni-service triage on 2026-09-01 and stays open as the tracker. The
maintainer ruling above is unchanged and binding on every child; the acceptance criteria above are
carried by #1749, which is the child that flips user-visible behaviour. Until #1749 lands,
`apa`/`mla`/`harvard` keep falling through to numbered IEEE and the prose sites documenting that
fall-through stay accurate.

### LINKED ISSUE #1749: cogni-knowledge: render apa/mla/harvard as real author-date citations end to end and reconcile the fall-through prose
## Plugin
cogni-knowledge

## Epic context
Child 3 of 3 under tracker #1744 (`--citation-format apa|mla|harvard` is accepted and persisted but silently renders IEEE). This is the child that flips user-visible behaviour and carries #1744's user-facing acceptance criteria. It depends on child 1 (author/date metadata at ingest) and child 2 (citation-family-aware primitives in `_knowledge_lib.py`) — both must be merged before this branches, because the composer needs an author-date tuple to render and the consumers need family-aware helpers to call.

**SCOPE FENCE — do not split this child.** Its boundary is fixed by #1744's maintainer ruling, not by PR size. A PR that lands the composer's author-date rendering without finalize, reviewer, revisor, citation-store and verify-store in the same commit set produces a format that renders author-date markers and is then silently mis-validated, mis-scored, or rejected by the revisor's integrity check — strictly worse than today's honest IEEE fall-through, and exactly the partial state the ruling forbids. The forbidden state moves from the rendering layer to the validation layer; it does not disappear. This child is large on purpose.

## Current behavior
`knowledge-plan` accepts `--citation-format apa|mla|harvard`, validates them via `normalize_citation_format` (`scripts/_knowledge_lib.py:583`), and persists the choice into `plan.json::citation_format` — then every downstream surface renders the numbered IEEE shape instead. The user gets positive confirmation their choice was recorded and IEEE output, with no warning at any step.

Six surfaces are hardcoded to the numbered `<sup>[N](url)</sup>` shape:

- `agents/wiki-composer.md:166-169` — reference-string format branches on `CITATION_FORMAT`, but line 169 makes `apa`/`mla`/`harvard` render the numbered IEEE shape outright.
- `references/finalize-compose-subprocess.md:284` — the actual `renumber_inline_citations(body)` call site the finalize deposit runs (`skills/knowledge-finalize/SKILL.md:1095` documents it; this reference executes it).
- `agents/wiki-reviewer.md:126-131` — the citation-density gate counts `<sup>\[\d+\]\([^)]+\)</sup>` and `<sup>\[\d+\]</sup>` only.
- `agents/revisor.md:130` — the read-back citation-integrity check counts inline numbered markers and treats a non-`<sup>[N]` marker as a regression, so an author-date draft fails it.
- `scripts/citation-store.py:56,133,208` — imports `extract_inline_citation_urls` to cross-check every inline citation URL against the ingest manifest at build time. Numbered-only means an author-date draft's URLs are invisible: silent validation bypass or outright manifest rejection.
- `scripts/verify-store.py:67,382` — calls `strip_inline_citation_markers` in the verbatim prefilter. Numbered-only means the marker never strips and prefilter hit-rate degrades (fail-soft: falls through to the LLM verifier).

Five prose sites document the fall-through, all currently accurate: `skills/knowledge-plan/SKILL.md:34`, `agents/wiki-composer.md:169`, `skills/knowledge-finalize/SKILL.md:1095`, `scripts/_knowledge_lib.py:542-548`, and `references/citation-formats.md:37-39` (the fifth site, which #1744's body does not name).

## Desired behavior
`apa`, `mla` and `harvard` render real author-date citations end to end. `ieee` and `chicago` are untouched. The accepted set equals the rendered set.

Exact user-visible strings that change:

- `agents/wiki-composer.md:169` — "**apa/mla/harvard** (staged author-date): render the **numbered** IEEE shape for now — the numbered renumber pass in knowledge-finalize requires it until the author-date follow-up lands." becomes a per-format author-date rendering rule naming the inline shape and the alphabetical un-numbered reference list for each of APA / MLA / Harvard, with no "for now" and no staged language.
- `skills/knowledge-plan/SKILL.md:34` — "`ieee`/`chicago` (wired) or `apa`/`mla`/`harvard` (staged author-date — see …)" becomes all five listed as wired, with the numbered vs author-date distinction stated as a rendering family, not a staging status.
- `skills/knowledge-finalize/SKILL.md:1095` — "Does NOT yet render **author-date** apa/mla/harvard — those are accepted + persisted but fall through to the numbered IEEE string" becomes a statement that finalize assembles the reference list per citation family and bypasses the numbered renumber pass on the author-date family.
- `scripts/_knowledge_lib.py:542-548` — the `VALID_CITATION_FORMATS` comment's "`apa`/`mla`/`harvard` are the staged author-date follow-up (parsed + accepted here so a base can persist the choice, but the composer falls back to numbered rendering until the format-aware finalize rework lands)" becomes a comment stating all five formats render as requested, with the family split describing marker shape only. The `CITATION_FAMILY` comment at lines 553-555 ("NOT yet wired into finalize's numbered renumber pass" / "The author_date branch is the named P2 follow-up") changes the same way.
- `references/citation-formats.md:37-39` — the "APA / MLA / Harvard (staged — author-date family)" heading and its sentence "That inline shape has no [N], so it is **not compatible** with the current numbered pipeline: knowledge-finalize's renumber_inline_citations pass and the <sup>[N] scans in wiki-verifier / wiki-reviewer / revisor all assume numbered markers." becomes a wired-status description of the author-date family. Two independent corrections in this sentence: drop "staged", **and** drop `wiki-verifier` from the list of numbered-assuming scans — it is already family-agnostic today (it scores the verbatim `draft_sentence` and carries no `<sup>[N]` regex), so that clause is false at base, not merely stale after this change.

`agents/wiki-verifier.md` is **not** an affected file and must not be changed, despite #1744's acceptance criterion 5 grouping "verify" with reviewer and revisor.

## Motivation
A flag that is accepted, validated and persisted, then ignored, is worse than an unsupported flag: the user has positive confirmation their choice was recorded. #1744's ruling rejected the cheaper Option 2 (fail fast at the boundary) because `scripts/_knowledge_lib.py:545` accepts and persists the value specifically so a base can hold it — failing fast would break bases already carrying `apa` in order to fix a rendering defect. A warn-only variant was also rejected: it makes the fall-through honest but leaves the requested format unimplemented indefinitely.

## Acceptance criteria
- [ ] A draft planned with `--citation-format apa` renders author-date inline citations and an author-date reference list end to end. Evidence: an assertion in `tests/test_compose_contract.sh` plus `tests/test_finalize_contract.sh` over an apa-format fixture. Falsified by any numbered `[N]` marker surviving anywhere in that fixture's rendered output.
- [ ] The same holds for `mla` and `harvard`, each with its own inline shape. Falsified by any of the three rendering the IEEE numbered shape, or by all three rendering one shared shape when `references/citation-formats.md` specifies three distinct ones.
- [ ] `ieee` and `chicago` output is unchanged. Evidence: the pre-existing numbered-format assertions in `tests/test_compose_contract.sh` and `tests/test_finalize_contract.sh` pass with their expected values unedited, including `test_finalize_contract.sh:128` (`finalize-49-renumbers-inline-n-markers`). Falsified by any diff in the rendered inline markers or reference-list shape for either format, or by an existing numbered assertion's expected value being edited to accommodate the change.
- [ ] The set of `citation_format` values `knowledge-plan` accepts equals the set rendered as requested. Evidence: every member of `VALID_CITATION_FORMATS` (`scripts/_knowledge_lib.py:549`) has a rendering branch reachable from the composer's format dispatch. Falsified by any value being accepted and persisted while rendering a different citation family.
- [ ] Finalize's reference-list assembly is family-aware and bypasses the numbered renumber pass on the author-date family. Evidence: the `renumber_inline_citations(body)` call at `references/finalize-compose-subprocess.md:284` is guarded by a family branch, asserted by `tests/test_finalize_contract.sh`. Falsified by an author-date draft's markers being rewritten by the renumber pass, or by a numbered draft skipping it.
- [ ] The reviewer's citation-density gate counts author-date markers. Evidence: an assertion in `tests/test_reviewer_contract.sh` over the `agents/wiki-reviewer.md:126-131` gate. Falsified by a fully-cited author-date section scoring as a citation-density deficit purely because the scan expected `[N]`.
- [ ] The revisor's read-back citation-integrity check accepts author-date markers. Evidence: the check at `agents/revisor.md:130` is family-aware. Falsified by an author-date draft whose retained citations all carry a marker still returning `write_failed`.
- [ ] `citation-store.py`'s build-time ingest-manifest gate sees an author-date draft's URLs. Evidence: an assertion in `tests/test_citation_store.sh` that the `--ingest-manifest` cross-check resolves every inline URL of an author-date fixture. Falsified by that fixture's URLs being invisible to the gate (silent bypass) or by the manifest rejecting a legitimately-cited source.
- [ ] `verify-store.py`'s verbatim prefilter strips author-date markers. Evidence: an assertion in `tests/test_verify_store.sh` covering the `scripts/verify-store.py:382` call. Falsified by an author-date marker surviving into the compared core string.
- [ ] An existing base persisting `citation_format: apa` keeps working, rendering author-date with no migration step required of the user. Evidence: a fixture whose source pages carry no `author:`/`year:` frontmatter resolves through child 1's derivation helper. Falsified by an error, a prompt, or a silent reset to `ieee` on such a base.
- [ ] All five prose sites agree with the shipped behaviour, with the exact old-to-new replacements named in `## Desired behavior` applied at `skills/knowledge-plan/SKILL.md:34`, `agents/wiki-composer.md:169`, `skills/knowledge-finalize/SKILL.md:1095`, `scripts/_knowledge_lib.py:542-555`, and `references/citation-formats.md:37-39`. Evidence: `grep -rn 'staged author-date\|staged — author-date\|not yet render\|NOT yet wired' cogni-knowledge/` returns nothing. Falsified by any site still describing a fall-through that no longer exists, or by `references/citation-formats.md` still naming `wiki-verifier` among the numbered-assuming scans.
- [ ] `agents/wiki-verifier.md` is unchanged. Evidence: `git diff --name-only origin/main...HEAD` does not list `cogni-knowledge/agents/wiki-verifier.md`. Falsified by any diff to that file — it scores the verbatim `draft_sentence` and never re-parses marker shape, so it is already family-agnostic and needs no change.
- [ ] Scope boundary: `git diff --name-only origin/main...HEAD` lists only paths beginning `cogni-knowledge/`, and `git diff origin/main...HEAD -- '*plugin.json'` is empty. Falsified by any changed path outside that prefix, or by any added/removed line matching `"version"` anywhere in the diff.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-knowledge` exits 0. Falsified by any non-zero exit or any FAIL line in its output.

## Depends on
- #1747 cogni-knowledge: capture author / publication-date metadata at ingest, with a legacy-base derivation fallback
- #1748 cogni-knowledge: add citation-family-aware primitives to _knowledge_lib.py and make CITATION_FAMILY load-bearing

_Affected components: cogni-knowledge/agents/wiki-composer.md, cogni-knowledge/agents/wiki-reviewer.md, cogni-knowledge/agents/revisor.md, cogni-knowledge/skills/knowledge-finalize/SKILL.md, cogni-knowledge/skills/knowledge-plan/SKILL.md, cogni-knowledge/references/citation-formats.md, cogni-knowledge/references/finalize-compose-subprocess.md, cogni-knowledge/scripts/_knowledge_lib.py, cogni-knowledge/scripts/citation-store.py, cogni-knowledge/scripts/verify-store.py, cogni-knowledge/tests/test_compose_contract.sh, cogni-knowledge/tests/test_finalize_contract.sh, cogni-knowledge/tests/test_reviewer_contract.sh, cogni-knowledge/tests/test_citation_store.sh, cogni-knowledge/tests/test_verify_store.sh_

_Complexity hint: high_

sdh07:
Closes #1749
Closes #1744

`knowledge-plan --citation-format apa` (or `mla`, `harvard`) was accepted, validated and
persisted into `plan.json::citation_format` — and then every downstream surface rendered the
numbered IEEE shape anyway. The user got positive confirmation their choice was recorded and
silently wrong output, discoverable only by reading maintainer prose in a dozen files.

The maintainer ruling on #1744 chose **Option 1 — honour the formats**. Children #1747
(author/date metadata at ingest) and #1748 (citation-family-aware primitives) are already
merged, so this PR is #1749: the end-to-end flip. It closes the epic's gate, hence both
`Closes` lines.

Seven commits, one per layer. They are **one commit set on purpose** — #1749's SCOPE FENCE
forbids splitting, because a composer emitting author-date markers against validation that is
still numbered-only is strictly worse than the honest IEEE fall-through it replaces.

## What changed

| Layer | Change |
|---|---|
| `_knowledge_lib.py` | new pure `author_date_reference_entry` — the three canonical bibliography strings, per format |
| `wiki-composer.md` | three distinct inline shapes + an alphabetical, un-numbered reference list |
| `finalize-compose-subprocess.md` | family-branched `refs` assembly; `renumber_inline_citations` **guarded**, not removed |
| `wiki-reviewer.md` / `revisor.md` | density gate and read-back check count the draft's own family |
| `citation-store.py` | all three URL-extraction sites onto the family dispatcher, `--citation-format` threaded from both build call sites |
| `verify-store.py` | verbatim-prefilter strip onto the family dispatcher, `--citation-format` from a new `knowledge-verify` Step 0.7 |
| 12 prose sites | reconciled |

Rendered output, produced by actually executing the finalize subprocess over a two-source
fixture (this is `finalize-203`, not a mock-up):

```
APA      acme.de. (2019). "Acme Studie". [https://acme.de/s](…) — [[acme-page]]
         Zimmermann, Ada. (2024). "Zeta Report". Zeta Institute. [https://zeta.eu/r](…) — [[zeta-page]]
MLA      acme.de. "Acme Studie." 2019. …
         Zimmermann, Ada. "Zeta Report." Zeta Institute, 2024. …
Harvard  acme.de (2019) "Acme Studie". Available at: …
         Zimmermann, Ada (2024) "Zeta Report". Zeta Institute. Available at: …
```

Alphabetical order (`acme` before `Zimmermann`) is the **reverse** of citation order, so the
sort is exercised, not assumed. The `acme-page` fixture carries **no** `author:` /
`published_date:` — it resolved through the `publisher:` / `fetched_at:` surrogates, which is
#1744 AC 6's no-migration path asserted rather than assumed.

Running the same harness against `origin/main`'s reference file renders `**[1]** Zeta
Institute, "Zeta Report".` for `apa` — the defect, reproduced — while `ieee` and `chicago`
diff **byte-identically** between base and head.

## Design decisions worth a reviewer's attention

**Three distinct bibliography strings, not one shared author-date string.** A single string
satisfies "not numbered" while leaving MLA and Harvard unimplemented — the same
accepted-but-rendered-as-something-else defect, moved one layer down. `klib-51` pins each
format by full-string equality **and** asserts the three pairwise distinct, so a collapse onto
one shape cannot pass.

**URL-less citations keep the plain `<sup>[N]</sup>` marker under the author-date family.**
Synthesis, distilled (`dcl-NNN`) and question-node (`acl-NNN`) citations carry no external
URL. `references/citation-formats.md`'s hard rule — an un-linked `(Publisher, 2026)` is a
format violation — is family-blind and already covers this, so the arm is *documented*, not
invented. A destination-less `([Author, Year])` would also neither strip nor extract (the
#1748 pattern is scheme-anchored), so the revisor would return `write_failed`. Both
marker-counting sites keep their URL-less arm and apply it to this family, which is what keeps
the revisor's marker count equal to its record count. Two residuals are accepted and already
tracked by the deferred #1755: the `[N]` numeral is display-only against an un-numbered list,
and such a sentence falls through the prefilter to the LLM verifier. Widening the strip
pattern would reverse #1748's explicitly-argued single-pattern rationale — a human go/no-go,
not this PR's call.

**Both new flags are free text normalized, not `argparse choices=`.** `normalize_citation_format`
exists to map the deprecated `wikilink` alias onto `ieee`, and `knowledge-binding.py` writes
`research_defaults.citation_format` **unnormalized**, so a persisted value can carry it.
`choices=` would hard-exit before the mapping could run, failing a build over a value the
pipeline supports. This deliberately differs from the sibling `--prose-density` on the same
`verify-store.py` parser; the rationale is in a code comment at both sites. Omitting either
flag yields `ieee` → `numbered` → the existing numbered primitives, so every pre-existing call
site is byte-identical (`citation-store-28`).

## Mutation recipes — all run, all `guard_verified`

The harness is not vendored here; it lives in the managed-service repo. `$R` is this worktree.

```bash
M=~/GitHub/dev/managed-service/cogni-service/scripts/mutation-check.sh
```

| case | `--file` | `--expr` | verdict |
|---|---|---|---|
| `klib-51` | `cogni-knowledge/scripts/_knowledge_lib.py` | `'s/if fmt == "mla":/if fmt == "__never__":/'` | `guard_verified` |
| `compose-contract-155-harvard-inline-shape` | `cogni-knowledge/agents/wiki-composer.md` | `'s/\(\[Author Year\]\(url\)\)/([Author, Year](url))/g'` | `guard_verified` |
| `finalize-199-renumber-family-guarded` | `cogni-knowledge/references/finalize-compose-subprocess.md` | `'s/if family == "numbered":\n    body = renumber_inline_citations\(body\)/body = renumber_inline_citations(body)/'` | `guard_verified` |
| `finalize-203-author-date-deposit-behavioral` | `cogni-knowledge/references/finalize-compose-subprocess.md` | `'s/refs = build_author_date_reference_list\(author_date_entries\)/refs = [x["rendered"] for x in author_date_entries]/'` | `guard_verified` |
| `reviewer-47-reads-citation-format-from-plan` | `cogni-knowledge/agents/wiki-reviewer.md` | ``'s/Capture `citation_format` \(default `ieee`\)/Capture the run format/'`` | `guard_verified` |
| `verify-contract-110-revisor-counts-author-date-markers` | `cogni-knowledge/agents/revisor.md` | `'s/\(\[Author, Year\]\(url\)\)/(numbered only)/'` | `guard_verified` |
| `citation-store-26` / `-27` | `cogni-knowledge/scripts/citation-store.py` | `'s/extract_citation_urls\(draft_text, citation_format\)/extract_citation_urls(draft_text)/; s/extract_citation_urls\(r.get\("draft_sentence"\) or "", citation_format\)/extract_citation_urls(r.get("draft_sentence") or "")/'` | both `guard_verified` |
| `citation-store-25` | `cogni-knowledge/scripts/_knowledge_lib.py` | `'s/    return \[a or b for a, b in _AUTHOR_DATE_CITATION_URL_RE.findall\(text\)\]/    return ["https:\/\/nicht-ingestiert.invalid\/x"] + [a or b for a, b in _AUTHOR_DATE_CITATION_URL_RE.findall(text)]/'` | `guard_verified` |
| `verify-store-38-prefilter-strips-author-date` | `cogni-knowledge/scripts/verify-store.py` | `'s/core = strip_citation_markers\(\n                    draft_sentence, citation_format\n                \)/core = strip_citation_markers(draft_sentence)/'` | `guard_verified` |

Every `--test` is `bash cogni-knowledge/tests/<that suite>.sh`; every `--expr` is a `perl -0pi`
substitution, never a `/.../d` delete form.

**Two findings this surfaced, both fixed rather than papered over:**

1. The family-blind mutation makes `citation-store-25` (the positive) go **green** — which is
   exactly the silent-bypass failure mode: a positive case cannot tell a working gate from a
   disabled one. That is why each positive here is paired with a negative, and why `-25`'s own
   recipe instead makes the gate *over*-fire.
2. `citation-store-27`'s first fixture was a genuine **`vacuous_guard`** — it failed on the
   slug leg, so the prose leg it was written to prove was never exercised. Rebuilt with a
   record whose slug leg *agrees* and whose own marker points at a different, genuinely
   ingested URL: the one shape only the prose leg can catch.

## Two defects in the issue's own acceptance criteria — reported, not worked around

**AC 11's evidence grep cannot see one of the five sites it names.**
`grep -rn 'staged author-date\|staged — author-date\|not yet render\|NOT yet wired' cogni-knowledge/`
is case-sensitive, and `skills/knowledge-finalize/SKILL.md:1095` reads "Does **NOT** yet
render". At base the predicate returned 5 lines and `:1095` was not among them; the `-i`
variant returned 6. Both variants now return exactly **one** line, and it is not a prose claim: it is
`tests/test_compose_contract.sh:444`, where `compose-contract-158`'s own
`assert_not_grep 'staged author-date'` carries the phrase as its search pattern — the
assertion that pins the retirement. **Zero live prose sites remain.** I am not claiming
that grep proves all five named sites moved: `:1095` was verified by direct inspection,
and its rewrite is in commit 3.

**The "five prose sites" inventory undercounts.** Twelve sites described the fall-through.
Seven match none of AC 11's four literal phrases: `references/citation-formats.md:5` (its
"there is no code-level inline-formatting logic" claim is falsified by this change), the
`:11-15` wiring-status rows, `:17`'s "all work unchanged" sentence (a numbered-family
statement that reads as a global claim), `skills/knowledge-compose/SKILL.md:61`/`:68`/`:164`/`:465`,
`skills/knowledge-setup/SKILL.md:40`, `skills/knowledge-plan/SKILL.md:111`,
`scripts/knowledge-binding.py:113`/`:814`, and `CLAUDE.md:96`. All reconciled in commit 7.
`references/citation-formats.md` also no longer names `wiki-verifier` among the
numbered-assuming scans, per AC 11.

## Accepted residual — and its follow-up

`agents/wiki-verifier.md` is **not** touched (AC 12, graded `[fidelity: literal]`). Its
grounding normalization at `:87` *does* strip `[N]`/`<sup>` markers, so the AC's stated
rationale for dropping `wiki-verifier` from the citation-formats list ("never re-parses marker
shape") is not accurate at base. Implementing AC 11 and AC 12 exactly as written therefore
ships a real residual: an author-date marker survives into the compared string and can depress
the per-citation `grounded` signal and the merged grounding rate. **Verdict assignment is
unaffected** — only the `grounded` metric. Filed as **#1877**.

## Review notes

**Skill / agent length.** Five units were already over their advisory char caps at base; each
grew, none by much. Recorded here as the content-floor escape hatch — the additions are the
family branches and their rationale, and there is no redundant prose in the touched
paragraphs to trade against them.

| unit | base | head | advisory cap |
|---|---|---|---|
| `skills/knowledge-finalize/SKILL.md` | 119 323 | 119 451 | 43 500 |
| `skills/knowledge-verify/SKILL.md` | 39 410 | 41 222 | 33 000 |
| `skills/knowledge-compose/SKILL.md` | 53 964 | 54 274 | 36 000 |
| `skills/knowledge-setup/SKILL.md` | 38 665 | 38 704 | 31 500 |
| `agents/wiki-composer.md` | 57 339 | 60 605 | 28 000 |
| `agents/wiki-reviewer.md` | 33 287 | 35 046 | 28 000 |
| `agents/revisor.md` | 30 885 | 32 286 | 28 000 |

`knowledge-verify` and `knowledge-finalize` are already in
`scripts/baselines/skill-spec-baseline.json` on the length arm, so no baseline entry is added.
`knowledge-compose` (469 lines) and `knowledge-setup` (453) stay under the guard's 500-line
ceiling, so neither becomes a new un-baselined finding. No `description` moved past 1024.

**`finalize-125` is reworded, not deleted, and its pattern and case id are byte-unchanged.**
Its premise — "the named follow-up, NOT wired this round" — flipped false. AC 3 fences
pre-existing *numbered* assertions; this is a staged-status assertion whose subject this issue
deliberately changes. `finalize-49`, `finalize-124`, `compose-contract-78`, `reviewer-16` and
`reviewer-17` are all untouched.

**A pre-existing assertion caught a real slip mid-work.** `finalize-125` greps lowercase
`author-date`; my first rewrite of `:1095` capitalised it, and the suite went red. Worth
noting because it demonstrates that assertion is load-bearing rather than decorative.

## Verification

- `python3 scripts/run-plugin-tests.py` — **158/158 suites**, 0 failures (full repo, not just the plugin).
- `check-skill-spec` · `check-breadcrumbs` · `check-case-id-pairing` · `check-result-line-plainness` · `check-code-span-nesting` · `check-version-bump` · `check-mcp-tool-grant` · `check-marketplace-manifest-sync` · `check-output-style-placement` · `check-attribution-path-integrity` · `check-readme-inventory-sync` · `check-workflow-yaml` · `check-retirement-ledger` — all clean.
- `check-frontmatter.sh cogni-knowledge` — clean, 37 files.
- The executed `python3 -c` body in `references/finalize-compose-subprocess.md` `ast.parse`s (362 lines), and the diff introduces **no** new apostrophe into that shell-single-quoted block.
- Case-id uniqueness checked by **running** all seven touched suites and grouping emitted first tokens — 672 result lines, zero duplicates.
- `git diff --name-only origin/main...HEAD` lists only `cogni-knowledge/**`, never `wiki-verifier.md`; `git diff origin/main...HEAD -- '*plugin.json'` is empty; no added or removed line matches `"version"`.

## Human eyeball, please

The one thing no gate can decide: whether the three bibliography strings above are defensible
APA / MLA / Harvard renderings — field order, punctuation, `n.d.` for a missing year. The
canonical-per-format shape was chosen deliberately over one shared author-date string; one
rendered example per format is in the table near the top.

---

## Review round 1 — both items addressed (commit `4473c4ee`)

**1 (blocking) — MLA `n.d..`.** Reproduced exactly as reported, and the blast radius confirmed
independently: `concept-store.py` and `question-store.py` write neither date key, and
`finalize-compose-subprocess.md`'s own synthesis frontmatter emits only `created:`/`updated:`,
so the doubled period was the **default** rendering for every cited synthesis, concept, entity
and question node under `mla`. Fixed with the suggested `endswith(".")` guard; a real year
never ends in a period, so a dated entry is byte-unchanged.

The assertion that pinned the typo now pins the fix **in both directions** — the correct shape
*and* the absence of the doubled one — in the with-publisher form, the bare-year form, and
across all three formats.

`finalize-203`'s fixture gains a third page. Rather than dropping `fetched_at:` from
`acme-page` (which would have cost the surrogate/no-migration assertion), it adds the
real-world shape: a **synthesis** page with no author, no publisher and neither date key —
author-less, URL-less and year-less at once. It asserts the entry leads with its title, sorts
last as a block, renders `n.d.`, carries no link, and never doubles the period. Rendered:

```
MLA      acme.de. "Acme Studie." 2019. [https://acme.de/s](…) — [[acme-page]]
         Zimmermann, Ada. "Zeta Report." Zeta Institute, 2024. […] — [[zeta-page]]
         "Vorherige Synthese." n.d. — [[vorherige-synthese]]
```

Reverting the fix now reddens **both** levels — `guard_verified` at the primitive and end to
end:

```bash
--file cogni-knowledge/scripts/_knowledge_lib.py --expr 's/        segments.append\(tail if tail.endswith\("."\) else tail \+ "."\)/        segments.append(tail + ".")/'
# --case klib-51                                  -> guard_verified
# --case finalize-203-author-date-deposit-behavioral -> guard_verified
```

**2 (optional) — taken.** The counter-argument about symmetry is fair, but the deciding fact
runs the other way: the instruction to angle-bracket a paren-bearing destination *under
author-date* is one **this branch introduced** (`wiki-composer.md:146`), so that gap is newly
created rather than inherited. The numbered pattern's identical gap genuinely predates the
branch and is the shape `ieee`/`chicago` output is pinned against, so it is left alone — with
the reason now stated inline so nobody "fixes" it by accident.

The widened pattern was validated against the canonical `_AUTHOR_DATE_CITATION_URL_RE` across
nine cases (plain, angle-bracketed, `file://` with a space, all three format shapes, prose
link, `[[N]]`, bare `<sup>`, two markers in one sentence) — identical match counts on every
one. `reviewer-46` is `guard_verified`; a new `reviewer-48` pins the bracketed arm's presence
separately, since it is the half a first pass drops.

**AC-11 grep observation** — you are right and the body overstated it. Corrected above: the
command returns **1**, and that one line is `compose-contract-158`'s own `assert_not_grep`
pattern literal. My original run filtered `/tests/` out and I did not say so. Zero live prose
sites remain.

**MLA author-disambiguation** — noted, and I agree it is out of scope: the inline shape is
specified at base in `references/citation-formats.md:39`, so narrowing it is its own decision
rather than something to slip into this PR.

**Re-verified after the fix:** `run-plugin-tests.py` full repo **158/158**, all twelve repo
guards clean, the executed `python3 -c` body still `ast.parse`s with **0** apostrophes added,
case-id uniqueness re-checked by running the three touched suites, and the diff fences still
hold (`cogni-knowledge/**` only, `wiki-verifier.md` untouched, no `version` line).


sdh07:
## Review — one blocking fix, one optional

The engineering is sound and the evidence discipline is the strongest part of this PR: every positive paired with a negative control, the finalize case genuinely executing the subprocess rather than grepping it, and the two self-reported findings (the family-blind mutation greening `citation-store-25`, and `-27`'s first fixture being a real `vacuous_guard`) fixed rather than papered over. I re-ran the acceptance evidence independently and it holds — AC 12 (`wiki-verifier` untouched), AC 13 (scope + no version line), `run-plugin-tests.py --filter cogni-knowledge` green, all 16 CI checks green, and no numbered-only consumer of `extract_inline_citation_urls` / `strip_inline_citation_markers` / `renumber_inline_citations` left anywhere outside `_knowledge_lib.py` and the suites.

One rendering defect has to be fixed before merge. It lands squarely in the "human eyeball, please" question you flagged — *punctuation, `n.d.` for a missing year* — and the answer there is no.

---

### 1. Blocking — MLA renders a double period `n.d..` on every year-less entry

`author_date_reference_entry`'s MLA arm appends a sentence period to a year segment that already carries one:

```python
segments.append(((publisher + ", " + year) if publisher else year) + ".")
```

`year` has already defaulted to the literal `"n.d."` above, so the MLA branch — and only the MLA branch — concatenates two periods. APA (`(n.d.)`) and Harvard (`(n.d.)`) put the year inside parentheses and are unaffected.

Reproduced:

```
apa      '"Vorherige Synthese". (n.d.).'
mla      '"Vorherige Synthese." n.d..'      <-- double period
harvard  '"Vorherige Synthese" (n.d.).'
```

**This is not an exotic edge.** `resolve_author_year` returns `("", "")` for any page carrying neither `published_date:` nor the `fetched_at:` surrogate, and the synthesis frontmatter that `finalize-compose-subprocess.md` itself emits carries only `created:` / `updated:` — no date key `resolve_author_year` reads. `concept-store.py` and `question-store.py` write no `fetched_at:` either. So under `mla` the double period is the **default** rendering for every cited synthesis, concept, entity and question-node reference entry, plus any legacy source page missing both date keys. Those URL-less page kinds are first-class citizens in this PR's own design discussion, which is what makes this worth blocking on rather than deferring.

`finalize-203` does not catch it: both fixture pages resolve a year (`2024` explicitly, `2019` through the `fetched_at:` surrogate), so the year-less MLA path is never exercised end to end.

Worse, the unit case **pins the typo as the expected value** — `tests/test_knowledge_lib.sh:1517`:

```python
assert ', n.d..' in e("mla", "A", None, "T", "P", url)
```

so the assertion that should have caught this is currently asserting it. Fixing the code alone will redden `klib-51`; the assertion has to move with it.

**Suggested fix** — only append the period when the tail does not already end in one:

```python
tail = (publisher + ", " + year) if publisher else year
segments.append(tail if tail.endswith(".") else tail + ".")
```

and at `test_knowledge_lib.sh:1517`, assert the correct shape *and* the absence of the doubled one, so the fix is pinned in both directions:

```python
mla_nd = e("mla", "A", None, "T", "P", url)
assert 'P, n.d. ' in mla_nd and 'n.d..' not in mla_nd, mla_nd
```

Please also add the year-less page to the `finalize-203` fixture (a third source, or drop `fetched_at:` from `acme-page`) so the path is covered behaviourally and not only at the primitive. The other three `n.d.` assertions (`:1516` apa, `:1518` harvard) and the `no publisher -> bare year` case at `:1524` are unaffected by the change — I checked.

---

### 2. Optional — the reviewer's author-date density regex drops the angle-bracketed arm

`agents/wiki-reviewer.md`'s new density pattern is

```
\(\[[^\[\]]+\]\((?:https?|file)://[^)]+\)\)
```

while the canonical `_AUTHOR_DATE_CITATION_URL_RE` it mirrors carries a `<…>` alternative that this one omits. The composer is explicitly instructed two paragraphs earlier to emit `([Author, Year](<url>))` when the destination contains parentheses, so that marker is invisible to the density count:

```
([Doe, 2024](https://a.eu/r))                                    reviewer=1  lib=1
([Doe, 2024](<https://en.wikipedia.org/wiki/Foo_(bar)>))         reviewer=0  lib=1
([the annex](#section-3))                                        reviewer=0  lib=0   (correctly ignored)
```

A section citing a paren-bearing URL under `apa`/`mla`/`harvard` therefore scores as a citation-density deficit purely because of the marker shape — narrower than, but the same class as, the defect AC 6 exists to close.

**Marked optional deliberately, and here is the honest counter-argument:** the *numbered* pattern one line above (`<sup>\[\d+\]\([^)]+\)</sup>`) has exactly the same gap at base, so the new arm is symmetric with its neighbour rather than a regression, and `wiki-reviewer` is advisory / non-blocking / fail-soft — the cost is a depressed score, never a wrong verdict. Your call whether to close it here (adding `<((?:https?|file)://[^>]+)>|` as a first alternative, matching the lib pattern, and extending `reviewer-47`'s sibling case) or to leave both arms consistent and file it. Either is defensible; I would not hold the PR for it.

---

### Observations — no action needed

- **AC 11's grep returns 1, not 0.** The PR body states "Both variants now return **zero**". Re-running the AC's own command returns one line: `tests/test_compose_contract.sh:444`, which is `compose-contract-158`'s own `assert_not_grep 'staged author-date'` pattern literal. That is the correct way to pin the retirement and is not a live prose claim — but a reviewer re-running the stated evidence gets `1` and has to work out why. Worth one clause in the body rather than a code change.
- **MLA's inline shape `([Author](url))` cannot disambiguate two sources by the same author** — real MLA disambiguates with a short title. Noting it because you asked for eyes on the format shapes; the shape is specified at base in `references/citation-formats.md`, so changing it is out of scope here and would be its own decision.
- The scope fence held, `finalize-125`'s reword is correctly argued (pattern and case id byte-unchanged, premise genuinely flipped), the `--citation-format` flag reaches all three real call sites (`knowledge-compose:229`, `knowledge-verify:219`, `knowledge-verify:360`), and the free-text-normalize-over-`argparse choices=` decision is right for the reason given — a persisted `wikilink` would hard-exit a build under `choices=`.

Fix item 1 (code + the two test sites) and push; I will re-review and merge.

sdh07:
## Both items addressed — `4473c4ee`

Thank you for the MLA catch. I reproduced both findings independently before touching anything;
your analysis holds on every point, including the blast radius.

### 1 — blocking: MLA `n.d..`

Reproduced, and the "not an exotic edge" claim is confirmed from the writers' side rather than
taken on trust: `concept-store.py` and `question-store.py` write **neither** date key
(`grep -c` = 0 for both `published_date` and `fetched_at`), and
`finalize-compose-subprocess.md:296-297` emits only `created:`/`updated:` on the synthesis
frontmatter — so `resolve_author_year` returns no year for every synthesis, concept, entity and
question node the pipeline writes. Under `mla` the doubled period was the default, not an edge.

You are also right that my own assertion was the problem, not just silent about it: it pinned
the typo as the expected value. That is the more embarrassing half of this finding and the part
worth fixing carefully, so the replacement pins the fix **in both directions** — correct shape
*and* absence of the doubled one — in the with-publisher form, the bare-year form
(`A. "T." n.d.`, where the tail *is* the year), and across all three formats.

Applied your suggested guard verbatim. A real year never ends in a period, so a dated entry is
byte-unchanged.

On the fixture: I did not drop `fetched_at:` from `acme-page`, because that page is what carries
the surrogate / no-migration assertion and removing the key would have cost it. Instead the
fixture gains a **third** page in the real-world shape — a synthesis page with no author, no
publisher and neither date key, so it is author-less, URL-less and year-less at once. It now
asserts title-first, sorts-last-as-a-block, `n.d.`, no link, and no doubled period:

```
MLA   acme.de. "Acme Studie." 2019. […] — [[acme-page]]
      Zimmermann, Ada. "Zeta Report." Zeta Institute, 2024. […] — [[zeta-page]]
      "Vorherige Synthese." n.d. — [[vorherige-synthese]]
```

Reverting the fix now reddens **both** levels, which is the property that was missing before:

```bash
--file cogni-knowledge/scripts/_knowledge_lib.py \
--expr 's/        segments.append\(tail if tail.endswith\("."\) else tail \+ "."\)/        segments.append(tail + ".")/'
#   --case klib-51                                     -> guard_verified
#   --case finalize-203-author-date-deposit-behavioral  -> guard_verified
```

I checked your note on the three unaffected assertions (`:1516` apa, `:1518` harvard,
`:1524` bare-year) and it is correct — all three still pass unedited.

### 2 — optional: taken

Your counter-argument about symmetry is fair, but one fact decides it the other way: the
instruction to angle-bracket a paren-bearing destination **under author-date** is one this
branch introduced (`wiki-composer.md:146`). So that gap is newly created here, not inherited —
which is exactly what makes it in scope, while the numbered pattern's identical gap is not. I
left the numbered arm alone (it is the shape `ieee`/`chicago` output is pinned against) and
stated that reason inline, so the asymmetry reads as deliberate rather than as an oversight
someone should tidy up.

The widened pattern was validated against the canonical `_AUTHOR_DATE_CITATION_URL_RE` across
nine cases — plain, angle-bracketed, `file://` with a literal space, all three format shapes,
prose link, `[[N]]`, bare `<sup>`, two markers in one sentence — with identical match counts on
every one. `reviewer-46` is `guard_verified`; `reviewer-48` is new and pins the bracketed arm's
presence separately, since that is the half a first pass drops.

### Observations

- **AC-11 grep.** You are right and the body overstated it. My original run filtered `/tests/`
  out and I did not say so. The command returns **1**, and that line is `compose-contract-158`'s
  own `assert_not_grep 'staged author-date'` pattern literal. The body now says so explicitly:
  one hit, it is the assertion pinning the retirement, zero live prose sites remain.
- **MLA author-disambiguation.** Agreed, and agreed it is out of scope. The inline shape is
  specified at base in `references/citation-formats.md:39`; narrowing it is its own decision,
  not something to slip into this PR. Say the word if you want it filed.

### Re-verified after the fix

`run-plugin-tests.py` full repo **158/158**, all twelve repo guards clean, the executed
`python3 -c` body still `ast.parse`s with **0** apostrophes added by this branch, case-id
uniqueness re-checked by *running* the three touched suites, and the diff fences hold —
`cogni-knowledge/**` only, `wiki-verifier.md` untouched, no `version` line.

sdh07:
## Re-review — both items land. Merging.

I re-derived every claim in `4473c4ee` rather than reading the diff and taking the reasoning on trust.

**1 — MLA `n.d..`** — fixed and pinned in both directions. Swept all 48 combinations of `{author} × {year: '', None, '   ', 2024} × {publisher} × {url}` across the three formats: no doubled period survives anywhere, and a dated entry is byte-unchanged (`A. "T." P, 2024.`). The synthesis-shaped entry now renders `"Vorherige Synthese." n.d.` — the case that was broken.

Adding a third fixture page rather than stripping `fetched_at:` from `acme-page` was the right call, and better than what I suggested: my version would have traded the surrogate assertion for the year-less one, where yours holds both and adds author-less, URL-less and sorts-last-as-a-block on the same page. That is the shape the pipeline actually writes.

**2 — reviewer marker** — the widened pattern now counts `([Doe, 2024](<https://…/Foo_(bar)>))` while still ignoring `([the annex](#section-3))` and `[[3]]`, so neither load-bearing property was relaxed to buy the new arm. Leaving the numbered pattern alone with the reason stated inline is the right boundary — that shape is what `ieee`/`chicago` are pinned against, so widening it is its own change with its own regression surface.

### Verified independently

- **Mutation, re-run here, not read from your table** — reverting the MLA guard reddens `klib-51` *and* `finalize-203`, and narrowing the reviewer pattern back reddens `reviewer-46`. Three recipes, all `guard_verified`. The two-level redness is the property that was missing before: the primitive case alone could have been satisfied by a fixture that never reached the year-less path.
- 81/81 `cogni-knowledge` suites green locally; 16/16 CI green **against `4473c4ee` itself**, checked by SHA rather than by the PR-level summary.
- `check-case-id-pairing`, `check-result-line-plainness`, `check-code-span-nesting`, `check-skill-spec`, `check-breadcrumbs`, `check-version-bump` — all clean. `reviewer-48` is unique, verified by running the suite and grouping emitted first tokens, not by grepping call sites.
- Scope holds: zero paths outside `cogni-knowledge/`, empty `*plugin.json` diff, `wiki-verifier.md` untouched, no `"version"` line moved.
- `closingIssuesReferences` carries both #1749 and #1744, so neither strands on merge.

### Carried forward, not blocking

The AC 11 evidence grep still returns 1 — `compose-contract-158`'s own `assert_not_grep` pattern literal, which is the correct way to pin the retirement. Worth a clause in the body if it is ever re-read as evidence; not worth a commit.

The residual you fenced out is real and correctly filed: `wiki-verifier`'s grounding normalization at `:87` strips `[N]`/`<sup>` but not an author-date marker, so the marker survives into the compared string and can depress the `grounded` signal. Verdict assignment is untouched, and #1877 owns it. AC 12 graded `[fidelity: literal]` was the right call given the AC's stated rationale is inaccurate at base — reporting that rather than quietly widening scope is the behaviour I want.

Squash-merging. Epic #1744 closes with it; #1747 and #1748 are already closed.

PR #1906, opened 2026-09-07
feat(cogni-workspace): open the slides brief on the answer, close it on a sources unit

### LINKED ISSUE #1901: cogni-workspace: the slides brief opens on the answer and closes on a sources unit
## Plugin
cogni-workspace

## Current behavior
The slides derivation in `skills/text-to-narrative/references/design-brief-template.md` reads: "`cover` (the title; the subtitle goes to `talk_track`) -> `bluf` from the TL;DR -> one unit per element in arc order ... -> a `metric` unit ... -> a closing `bluf`". `tests/fixtures/design-brief/slides-en.md` matches it: unit 1 is `type: cover`, unit 2 is `type: bluf`, unit 8 (the last) is `type: bluf`, `climax: 7`. So the deck opens on a title rather than on the answer, which contradicts the narrative's own answer-first TL;DR contract.

Separately, Rendering Contract clause 3 for slides already tells the renderer that "a sources slide built from the Sources block stays last in the deck", but no unit represents it. The deck's last slide is therefore unaddressable by the brief, `--max-units` can silently drop the material that carries it, and the checker cannot assert anything about it. `TYPE_ENUM` in `scripts/check-design-brief.py` carries eight values and no `sources`, and `check_density_slides` requires a `slide_points` list on every slides unit.

## Desired behavior
Slide 1 **is** the Executive TL;DR as `type: bluf`, using the narrative title as its headline or kicker with the subtitle integrated there; no standalone `cover` is emitted for slides. The final unit is `type: sources`, built by the renderer from the trailing `**Sources**` block verbatim and carrying no invented bullets — so a `sources` unit is exempt from the `slide_points` requirement. `cover` stays in the enum for the other targets; only the slides derivation stops emitting it.

The checker asserts, on the slides target only: unit 1 has `type: bluf`, and the last unit has `type: sources`; both findings name the unit number.

Exact strings that change:

- `scripts/check-design-brief.py`: `TYPE_ENUM = ("cover", "bluf", "two-column", "table", "timeline", "quote", "metric", "roles")` -> the same tuple with `"sources"` appended.
- `scripts/check-design-brief.py`: `BRIEF_VERSION = "1.0"` -> `BRIEF_VERSION = "1.1"`.
- `references/design-brief-template.md` frontmatter table: `| version | all | "1.0", quoted |` -> `| version | all | "1.1", quoted |`.
- `references/design-brief-template.md` type-enum paragraph: "one of `cover`, `bluf`, `two-column`, `table`, `timeline`, `quote`, `metric`, `roles`" -> the same list plus `sources`, with one sentence stating that the trailing source register is `sources` and carries no invented bullets.
- `references/design-brief-template.md` slides derivation: "`cover` (the title; the subtitle goes to `talk_track`) -> `bluf` from the TL;DR -> ..." -> "`bluf` carrying the Executive TL;DR, with the narrative title as headline or kicker and the subtitle integrated there -> one unit per element ... -> a `metric` unit ... -> a closing `bluf` -> a `sources` unit built from the `**Sources**` block verbatim".
- `references/design-brief-template.md` overflow rule: "When the count exceeds `--max-units`, re-merge split elements starting with the smallest `## Composition` share; never drop the cover, the BLUF or the close." -> re-merge split elements first, then drop the optional metric unit; never drop slide 1, an arc element, the decision close where it adds distinct content, or the final `sources` unit.
- `references/design-brief-template.md` climax rule: "`climax` defaults to the `metric` unit when present, else the last element unit." -> defaults to the decision close, else the metric unit, else the last element unit — never the `sources` unit.
- `SKILL.md` Phase 7 step 3: "slides: cover, BLUF, one or two per element, a metric unit, a close" -> "slides: a BLUF opening on the TL;DR, one or two per element, a metric unit, a close, a sources unit".
- `libraries/presentation-intent.md`: "One of: `cover`, `bluf`, `two-column`, `table`, `timeline`, `quote`, `metric`, `roles`" -> the same list plus `sources`, **or** one sentence stating why the render chain's vocabulary deliberately omits it. That file shares this type vocabulary with the brief, so leaving it silent is the drift this bullet exists to prevent.
- `tests/fixtures/design-brief/*.md` frontmatter: `version: "1.0"` -> `version: "1.1"` in **all five** fixtures.

**One correction to the source spec, from reading the code.** The spec asked that the non-slides fixtures be "unaffected". They cannot be: `check_frontmatter_type` compares every brief's `version` against the single `BRIEF_VERSION` constant regardless of target, and `document-en.md`, `infographic-en.md` and `web-en.md` all declare `version: "1.0"`. Bumping the constant without touching them turns all three red on `frontmatter-type`. The resolution adopted here: bump the `version:` line in all five fixtures, and change **nothing else** in the three non-slides ones — their unit shape, copy and checker results stay as they are.

**Non-goals — a hard fence.** This adopts one change found in an external, host-neutral export of this skill. That export dropped every plugin integration to run standalone, and a diff that strips them here is wrong. Do **not** rename `copywriter` to "the language pass" anywhere. Do **not** replace Pass 4's `readability.sh` + `tests/fixtures/copywriter/readability.yml` band with a qualitative review. Do **not** replace the `libraries/*-copywriting.md` and `libraries/*-presets.md` ceiling homes in `references/density-ceilings.md` with a phrase like "bundled ceiling policy" — `tests/test-brief-density-sync.sh` compares the reference against exactly those homes. Do **not** rewrite `${CLAUDE_PLUGIN_ROOT}` to `<skill-root>`, drop `allowed-tools`, or delete `evals/evals.json`. Do **not** add `agents/openai.yaml` or `assets/icon.svg`. No version bump of `cogni-workspace/.claude-plugin/plugin.json` or the root `.claude-plugin/marketplace.json` in the branch — that is the post-merge workflow's, and is unrelated to the brief `version` field this issue does change.

Sibling issues in this batch also edit `references/design-brief-template.md` and `scripts/check-design-brief.py` (a `visual_intent` block and an `evidence_status` scalar). Regenerating the slides fixtures must preserve whichever of those has already landed rather than reverting to the pre-change unit shape.

## Motivation
The deck currently opens on a title while the narrative it is cut from opens on an answer, so the brief contradicts its own source contract at slide 1. And a sources slide the Rendering Contract mandates but no unit represents is the one part of the deck the brief cannot protect: `--max-units` prunes units, and material nothing addresses is exactly what gets pruned silently.

## Acceptance criteria
- [ ] The brief version constant in `scripts/check-design-brief.py`, the template's frontmatter table row and all five fixtures under `tests/fixtures/design-brief/` state the same version, one increment above the pre-change value. Any two of the three disagreeing falsifies this.
- [ ] `sources` is accepted as a unit `type` by the checker and documented in the template's type-enum paragraph; a `type: sources` unit that the checker reports as out-of-enum falsifies this.
- [ ] Two separate findings exist and each carries its unit number. On a copy of `tests/fixtures/design-brief/slides-en.md` whose unit 1 `type:` line is changed to `two-column`, `python3 skills/text-to-narrative/scripts/check-design-brief.py` exits 1 and emits a finding whose `unit` field is 1; on a copy whose last unit `type:` line is changed to `metric`, the same command exits 1 and emits a differently-named finding whose `unit` field is that unit's number. A mutation producing no finding, two mutations producing the same check name, or a finding whose `unit` is null, falsifies this.
- [ ] A `type: sources` unit carrying no `slide_points` produces no finding, while a non-`sources` slides unit carrying no `slide_points` still does. A change that exempts every unit falsifies this.
- [ ] In both `tests/fixtures/design-brief/slides-en.md` and `tests/fixtures/design-brief/slides-de.md`: `grep -n '^type: cover$'` returns nothing; the `type:` line of unit 1 reads `type: bluf` and that unit's `slide_points` carry the TL;DR copy; the `type:` line of the last unit reads `type: sources`; the frontmatter `climax:` value is not that last unit's number; and `check-design-brief.py` exits 0 on each fixture. A surviving `type: cover` line, a `climax:` naming the sources unit, or a non-zero exit falsifies this.
- [ ] The three non-slides fixtures differ from their pre-change contents only in the `version:` line, and each still passes every check. A diff touching any other line of `document-en.md`, `infographic-en.md` or `web-en.md` falsifies this.
- [ ] `libraries/presentation-intent.md`'s slide-type list either carries `sources` or states in one line why the render chain's vocabulary omits it. A file where the two lists disagree with no stated reason falsifies this.
- [ ] `bash cogni-workspace/tests/test-text-to-narrative-brief.sh` exits 0 and covers the slide-1 and last-unit assertions with red mutants; each `FAIL: <id>` arm has a reachable same-id `PASS:` twin and a unique first-token id in the suite's `ttn-NN-<discriminator>` shape.
- [ ] Each new assertion is proven load-bearing by a mutation recorded in the suite's header block in its existing `--file / --expr / --test / --case` form — disabling the unit-1 type assertion turns only the slide-1 case red, and disabling the last-unit assertion turns only the last-unit case red.
- [ ] `python3 scripts/run-plugin-tests.py --filter cogni-workspace` exits 0.
- [ ] The diff touches only `cogni-workspace/skills/text-to-narrative/references/design-brief-template.md`, `.../scripts/check-design-brief.py`, `.../SKILL.md`, `cogni-workspace/libraries/presentation-intent.md`, `cogni-workspace/tests/fixtures/design-brief/*.md` and `cogni-workspace/tests/test-text-to-narrative-brief.sh`. A change to `cogni-workspace/.claude-plugin/plugin.json`, the root `.claude-plugin/marketplace.json`, `references/density-ceilings.md` or any arc contract falsifies this.

_Affected components: cogni-workspace/skills/text-to-narrative/references/design-brief-template.md, cogni-workspace/skills/text-to-narrative/scripts/check-design-brief.py, cogni-workspace/skills/text-to-narrative/SKILL.md, cogni-workspace/libraries/presentation-intent.md, cogni-workspace/tests/fixtures/design-brief/, cogni-workspace/tests/test-text-to-narrative-brief.sh_

_Complexity hint: medium_

claude-cogni-work:
## What changed

The slides derivation opened on a `cover` title unit while the narrative it is cut from opens on
an answer, and the sources slide the Rendering Contract mandates had no unit representing it.

- **Slide 1 is the answer.** `type: bluf` carrying the Executive TL;DR, with the narrative title
  as its headline and the subtitle folded into the talk track. No standalone `cover` is emitted
  for slides; `cover` stays in the enum for the infographic and web targets.
- **The deck closes on `type: sources`**, built by the renderer from the trailing `**Sources**`
  block verbatim. It carries no invented bullets and is the one unit exempt from the
  `slide_points` floor — the exemption is keyed on the *type*, not on "last unit" and not on the
  target.
- **Two separately-named checks**, `slides-open-bluf` and `slides-close-sources`, each guarded on
  the slides target and each naming the offending unit number, so a deck that loses its
  answer-first opening and one that loses its source register are attributable apart.
- **`BRIEF_VERSION` 1.0 → 1.1**, matched by the template's frontmatter row and all five fixtures.
  The three non-slides fixtures change *only* that line: `check_frontmatter_type` grades every
  target against the single constant, so leaving them would have turned all three red.

## Verification

| Gate | Result |
|---|---|
| `test-text-to-narrative-brief.sh` | 36/36 green |
| `run-plugin-tests.py --filter cogni-workspace` | 33/33 suites green |
| `run-plugin-tests.py --filter cogni-consult` | 14/14 suites green |
| Both new assertions proven load-bearing | by mutation, below |

Each new assertion was verified by **running** its recorded recipe, not by inspection:

- disabling `if t != "bluf":` → **only** `ttn-19-slides-open-bluf` goes red
- disabling `if t != "sources":` → **only** `ttn-20-slides-close-sources` goes red
- disabling the `type: sources` exemption → both green slides fixtures go red, which is the
  exemption's teeth

The second mutation is the informative one: the mutant still exits 1 (a `metric` unit with no
`slide_points` trips `density-slides` anyway), so only the unit-number assertion in
`has_fail_unit` distinguishes it. A `has_fail`-only test would have passed under the mutation.

## Mutation recipe

```bash
bash ~/GitHub/dev/managed-service/cogni-service/scripts/mutation-check.sh --root . \
  --file cogni-workspace/skills/text-to-narrative/scripts/check-design-brief.py \
  --expr 's{if t != "bluf":}{if False:}' \
  --test 'bash cogni-workspace/tests/test-text-to-narrative-brief.sh' --case ttn-19-slides-open-bluf
```

```bash
bash ~/GitHub/dev/managed-service/cogni-service/scripts/mutation-check.sh --root . \
  --file cogni-workspace/skills/text-to-narrative/scripts/check-design-brief.py \
  --expr 's{if t != "sources":}{if False:}' \
  --test 'bash cogni-workspace/tests/test-text-to-narrative-brief.sh' --case ttn-20-slides-close-sources
```

Both were replayed on this branch and returned `verdict: guard_verified` (`mutated_result: red`,
`restored_result: green`). Each was additionally applied by hand with the whole suite re-run: the
first reddens **only** `ttn-19-slides-open-bluf`, the second **only** `ttn-20-slides-close-sources`
— so each assertion is proven load-bearing and neither recipe is a blunt instrument.

A third recipe is recorded in the suite header for the `slide_points` exemption
(`s{if f.get\("type"\) == "sources":}{if False:}`); it reddens the two green slides cases rather
than one named case, which is the exemption's teeth, and is annotated as such in the header.

## Open questions

**1. `libraries/presentation-intent.md` took the issue's second option, deliberately.** The issue
allowed either adding `sources` to the slide-type list or stating in one line why the render
chain's vocabulary omits it. I first added the tag — and it turned `pi-sync-03` red in
`cogni-consult/tests/test-presentation-intent-sync.sh`. That list sits inside the byte-synchronized
`<!-- PRESENTATION-INTENT:SHARED -->` block mirrored at `cogni-consult/references/publish-routing.md`,
so option (a) obliges a cross-plugin edit — which this issue's own file fence ("A change to ...
falsifies this") forbids. The two options are therefore not equivalent here: only option (b) is
reachable without violating the acceptance criteria. The file now states the reason, outside the
shared block. **Flagging in case the intent was to widen the fence rather than take option (b).**

**2. The acceptance bar was under-specified against the base code**, in four ways the
triage-time contract records: the checker's own module docstring pinned version `"1.0"` and
enumerated the check set; `SKILL.md` Phase 7 and the template's derivation still instructed a
`cover` opening, so the skill would have told Claude to emit exactly the brief the changed checker
rejects; the presentation-intent list carries the cross-plugin coupling above; and the
pre-existing `ttn-12-no-styling-keys` case anchors its mutant on the literal `type: cover` string
this change deletes — it would have failed as "mutant could not be built", not as a checker
regression. All four are fixed here, but none was in the stated criteria.

**3. `design.dark_slides` moved `[1, 8]` → `[1, 7]`** in both slides fixtures. Nothing checks it
(only `climax` is range-checked), but leaving it would have darkened the new sources slide. Called
out because it is a judgment call the criteria do not cover.

## Not batched

`batch-compose` selected this issue as the anchor of a 5-member `cogni-workspace` batch. The issue
carries `svc:no-batch` and the maintainer ruling on it is explicit — *"Not batched — one PR
each."* The anchor slot bypasses the composer's `lifecycle_excluded` filter, so the composed batch
was discarded and this ships standalone. **That bypass looks like a real gap in `batch-compose`**
rather than a policy choice: `svc:no-batch` is honoured for members and ignored for the anchor.

Closes #1901


sdh07:

_The merger formed this expectation from issue #1901 before checking out the PR head — the review grades the diff against it._

**Expected approach:** A correct fix flips the slides-target derivation at four coupled layers, all readable at base. (1) check-design-brief.py bumps BRIEF_VERSION (base line 55 = 1.0) to 1.1, adds sources to TYPE_ENUM while keeping cover (base line 59 is an 8-member tuple with no sources), registers two differently-named slides-only entries in the CHECKS tuple (base lines 638-650, 11 entries) whose findings carry the offending unit number in the unit field, and exempts the sources unit from the slide_points-list-missing failure in check_density_slides (base lines 458-460) without weakening it for any other slide. (2) The five fixtures under tests/fixtures/design-brief/ all move version from 1.0 to 1.1 (line 3 of each); the two slides fixtures additionally restructure so unit 1 stops being type: cover (slides-en.md:51, slides-de.md:50) and becomes the TL;DR type: bluf - the copy currently in Slide 2 (slides-en.md:60-65), verbatim selections from the narrative TL;DR paragraph (tests/fixtures/narrative-output/corporate-visions-en.md:17) - and the deck gains a trailing type: sources unit; climax (slides-en.md:34 = 7, slides-de.md:33 = 7) must not name that last unit, which a green run cannot prove because check_unit_numbering (lines 418-420) only range-checks climax against len(units). (3) tests/test-text-to-narrative-brief.sh gains a red-mutant case per new check with recipes in its header (base lines 22-29), and its existing ttn-12-no-styling-keys mutant (lines 168-169) keys on the literal type: cover string in slides-en.md while mutate() (line 88) asserts the transform changed the text, so removing the cover unit turns that case red unless the transform is re-pointed. (4) libraries/presentation-intent.md must take its sources note OUTSIDE the PRESENTATION-INTENT:SHARED block (base lines 30-71) - clause 3's type list at lines 52-53 sits inside it and is byte-compared against cogni-consult/references/publish-routing.md:94-135 by cogni-consult/tests/test-presentation-intent-sync.sh, and that file is not in changed_files.

**Acceptance checklist:**
- [ ] [literal] grep -n '^BRIEF_VERSION' check-design-brief.py returns exactly one line whose value is the quoted string 1.1 (base line 55 reads 1.0).
- [ ] [purpose - sources becomes an accepted unit type while cover stays available to the non-slides targets that still use it] grep -n '^TYPE_ENUM' shows a tuple containing BOTH sources and cover (base line 59: 8 members, no sources). A diff that adds sources by replacing cover fails this item.
- [ ] [literal] grep -rn '^version:' tests/fixtures/design-brief/ returns exactly five lines, every one carrying the quoted string 1.1 (base: all five are 1.0).
- [ ] [purpose - the three non-slides fixtures are version-bump collateral only] the base..head diff for document-en.md, infographic-en.md and web-en.md shows exactly one removed and one added line per file, both on the version: line. Any other hunk in those three files fails this item.
- [ ] [purpose - the change is fenced to the text-to-narrative brief surface and its own guards] every path in the base..head diff matches cogni-workspace/skills/text-to-narrative/**, cogni-workspace/libraries/**, or cogni-workspace/tests/**; no changed path falls under cogni-consult/**, docs/**, wiki/**, or any .claude-plugin/plugin.json.
- [ ] [purpose - no issue-tracking or version breadcrumb leaks into the authoring files] no leaked issue-ref or version tokens appear in the changed SKILL.md, reference or library prose.
- [ ] [purpose - the one touched SKILL.md keeps a schema-valid header] SKILL.md frontmatter stays valid against the required-key schema and the 1024-char description cap.
- [ ] [purpose - the type-vocabulary edit reaches the library without colliding with the byte-for-byte cross-plugin sync guard] bash cogni-consult/tests/test-presentation-intent-sync.sh exits 0. Confirm by evidence: the diff for presentation-intent.md changes no line between the PRESENTATION-INTENT:SHARED:START and :END markers (base lines 30-71, containing clause 3's type list at 52-53), and publish-routing.md is untouched.
- [ ] [purpose - the pre-existing suite is the gate and every case must survive the fixture restructure] bash tests/test-text-to-narrative-brief.sh exits 0 with zero lines beginning FAIL:. In particular all five ttn-01-green-* cases pass, the executable proof the restructured slides fixtures are still clean under every ceiling.
- [ ] [purpose - the styling-key red mutant stays buildable after the cover unit leaves slides-en.md] the suite prints PASS: ttn-12-no-styling-keys, and the diff shows its transform (base lines 168-169, keyed on the literal type: cover string) re-pointed to a string that still occurs in its fixture. mutate() asserts the transform changed the text, so a stale key is a hard FAIL, not a silent skip.
- [ ] [purpose - a slides brief that does not open on the answer is caught by a check that names the offending unit] the CHECKS tuple gains a named entry asserting unit 1 is type: bluf. Evidence: mutating slides-en.md unit 1 to a non-bluf type yields exit 1 with a finding whose check equals that new name and whose unit equals 1, and --list-checks lists that name.
- [ ] [purpose - a deck that does not close on sources is caught by a SECOND, differently-named check that also names its unit, and neither new check fires on a non-slides target] the CHECKS tuple gains a second entry with a distinct name asserting the last unit is type: sources. Evidence: mutating the last unit away from sources yields exit 1 with a finding whose check equals that second name and whose unit equals that unit's number; and ttn-01-green-web, -infographic and -document all pass unchanged, proving the two assertions are slides-only.
- [ ] [purpose - each new assertion is backed by a suite case whose guard is proven load-bearing] the suite gains one red-mutant case per new check, and its header records a mutation-check.sh recipe per new case. Each recipe: names the harness at a path that resolves where it is READ (never ${CLAUDE_PLUGIN_ROOT}, which replays to a false green); passes --file naming the MUTATED file relative to --root, never the harness path; supplies an --expr that is a single-quoted perl -0pi substitution; and pins --case to the label the suite actually prints, so the case resolves rather than grading nothing as case_not_found.
- [ ] [purpose - the sources unit is exempt from slide_points without weakening that requirement for any other slide] the exemption added to check_density_slides (base lines 458-460) is predicated on the unit's TYPE value, not on unit position alone. Evidence: both slides fixtures are green with the sources unit carrying no slide_points, AND a non-sources slide stripped of slide_points still yields exit 1 with a density-slides fail.
- [ ] [literal] AC bullet 5 clause (b) - in BOTH slides-en.md and slides-de.md the type: line of unit 1 reads exactly type: bluf (base: type: cover at :51 and :50), AND that unit's slide_points carry the Executive TL;DR copy - each line a verbatim substring of the narrative TL;DR paragraph. A unit 1 reading type: bluf but carrying cover copy (only the title line, as base slides-en.md:53 does) fails this item.
- [ ] [purpose - the deck closes on a sources unit in both language fixtures] in BOTH slides fixtures the highest-numbered ## Slide N unit carries type: sources and is the last unit before the Sources block. Evidence: grep shows sources on the final unit only; ttn-01-green-slides and -slides-de pass, proving numbering runs 1..N without gaps.
- [ ] [literal] AC bullet 5 clause (a) - in BOTH slides fixtures the frontmatter climax integer is NOT the number of that fixture's last (sources) unit. Base: slides-en.md:34 climax 7, slides-de.md:33 climax 7, both carrying 8 units. Grade by reading the two values directly from the diff - a green run does NOT adjudicate this, because check_unit_numbering only range-checks climax against 1 <= climax <= len(units), so climax equal to the last unit's number is in range and passes.
- [ ] [purpose - the sources unit reproduces the brief's own sources material and invents no copy] neither slides fixture's sources unit introduces a bullet or copy line not already in that brief's trailing Sources block or a fixed localized label (English Sources / German Quellen headline). Evidence: copy-frozen-numbers stays green, and the unit's segment carries no invented slide_points.
- [ ] [purpose - the checker's own header stops asserting a version and a check inventory its code falsifies] the module docstring (base lines 24-38, line 28 reads version is not 1.0) names 1.1 and lists both new check names in its Checks block. Evidence: --list-checks returns data.checks with 13 entries including both new names (base: 11).
- [ ] [purpose - the prose a brief author follows no longer instructs a cover unit and does instruct the sources close] grep-anchored: SKILL.md:299 no longer derives slides as cover, BLUF, ... and names the sources close; design-brief-template.md:89 no longer instructs never drop the cover and names the sources unit; :70 (type enum sentence) includes sources; :37 (version row) reads 1.1; :151 (What the checker proves) names the two new assertions. Wording may differ so long as each of the five sites no longer makes the superseded claim.

**Risk watchlist:**
- Removing type: cover from slides-en.md silently kills ttn-12-no-styling-keys, whose base transform keys on that exact literal and whose mutate() asserts the text changed. A restructure without re-pointing turns the suite red - or worse, an author deletes the case and removes no-styling-keys coverage entirely.
- Adding sources to clause 3's type list inside presentation-intent.md's PRESENTATION-INTENT:SHARED block breaks cogni-consult/tests/test-presentation-intent-sync.sh, which byte-compares it against publish-routing.md - a file outside this issue's scope.
- climax is the item no green run catches. check_unit_numbering range-checks it against the unit count only, so a fixture that appends a sources unit and leaves climax pointing at a renumbered slide passes every check while being semantically wrong. Grade it by reading the integers.
- A too-broad exemption: implementing the slide_points exemption as 'the last unit is exempt' rather than 'a sources unit is exempt' disarms the guard for any deck whose last slide is a normal close.
- Collapsing the two new checks into one name, or naming them so similarly a grader cannot tell them apart.
- A new check registered in CHECKS but not scoped to slides turns ttn-01-green-web/-infographic/-document red. Conversely, adding the check outside the CHECKS tuple breaks the --list-checks contract that ttn-16 compares against checks_run.
- Building the sources unit as slide_points from the Sources entries would blow slide_points_max_lines: 4 on slides-de.md (5 entries) and slide_point_words_max: 10 on every entry, and drag URL/year digits through copy-frozen-numbers.
- Judgment-only: exactly how much of the Sources block the sources UNIT surfaces (a bare localized headline vs a rendered list) is underdetermined by the issue. Both readings satisfy every checkable item; a human may want to eyeball which shipped.
- design.dark_slides: [1, 8] in both slides fixtures is un-gated by any check but now points at a renumbered deck - slide 8 may become the sources slide. Cosmetic, ungradable, worth a glance.
- The trailing note: lines are hoisted from self.units[-1] by the Brief constructor (base lines 292-297). Appending a sources unit after them leaves the notes parsed as a field of the second-to-last unit. Nothing checks it.

- A base-grounded regeneration derives at least four substantive gating items the issue's Scope paragraph does not name: (1) the ttn-12-no-styling-keys mutant keys on the literal type: cover in slides-en.md and breaks when that unit disappears; (2) editing clause 3's type list in presentation-intent.md lands inside the PRESENTATION-INTENT:SHARED block and breaks the byte-for-byte guard against out-of-scope cogni-consult/references/publish-routing.md; (3) TYPE_ENUM must retain cover as well as gain sources, since _type_ok is shared by the slides, infographic and web targets; (4) the checker's module docstring hardcodes version is not 1.0 at line 28 and becomes a false claim after the bump. (recorded — does not hold the PR)

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1906

**Summary:** All 4 perspectives passed and the contract-grading-reviewer graded all 20 committed checklist items satisfied with zero findings; the 6 minor findings raised by the other three perspectives are all merger-added preferences (2 already opted-out, 4 demoted via the advisory-addition predicate) and none gate this PR.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (6)

- **[minor]** (waived) (advisory-addition) `cogni-workspace/skills/text-to-narrative/SKILL.md (whole file: body 31700 chars vs 30000 cap at complexity 16)` — Body exceeds its complexity-scaled character budget by ~5.7%. Pre-existing whole-skill drift: the base file was 33035 bytes and the head is 33067, a +32 char delta, so this change did not introduce the overage.
  - *Justification:* Pre-existing whole-skill drift (introduced_by_change: false); annotates but does not gate this diff.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/text-to-narrative/SKILL.md §Phase 7 step 3 (line 299), read against step 5 (line 301)` — Step 3 now derives the slides deck as "... a close, a sources unit", while step 5 still instructs that the brief ends with "the narrative's **Sources** block verbatim". A brief author reading SKILL.md alone cannot tell whether the new sources unit and that trailing block are the same artifact or two, nor what the unit may contain.
  - *Suggested fix:* Extend the slides clause on line 299 to name the relation: "a close, and a sources unit built from the narrative's own **Sources** block - the block is still reproduced verbatim at the end of the brief (step 5); the unit surfaces it inside the deck and invents no entry".

- **[minor]** (waived) (auto-fixable) (advisory-addition) `cogni-workspace/skills/text-to-narrative/references/design-brief-template.md:70` — The pick-by-content list still opens with the unqualified clause 'the title unit is `cover`', which a slides-brief author reads before reaching the qualifier two sentences later. The paragraph is target-general so the statements are reconcilable, but the ordering makes the reader hold a superseded instruction briefly. Does not violate item 20 - that site's superseded claim was the enum omitting `sources`, and the enum now carries it.
  - *Justification:* Prose-ordering precision with no user-observable fault: the paragraph states the correct rule explicitly two sentences on, the authoritative derivation at :89 says no standalone cover unit is emitted for slides, and the checker enforces it mechanically.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/skills/text-to-narrative/scripts/check-design-brief.py:429` — The new check's docstring reads 'Silent on every other target, which still uses cover'. After this diff no fixture and no documented derivation emits cover: infographic-en's first block is two-column and web-en's is bluf, at base as well as head, and the template's infographic and web grammars never name a cover unit. 'Available' is accurate; 'still uses' overstates. Does not touch item 2, which requires only that TYPE_ENUM retain cover - it does.
  - *Suggested fix:* Reword to the availability claim the template already makes: 'Silent on every other target, whose type vocabulary still admits cover.'

- **[minor]** (advisory-addition) `cogni-workspace/tests/fixtures/design-brief/slides-en.md:138` — The sources unit's talk_track is authored prose rather than a verbatim narrative selection, while SKILL.md step 3 states 'Copy is frozen: every line on the brief is a verbatim selection from the narrative'. Item 18 is satisfied on its own terms (headline is the fixed localized label, no invented slide_points, copy-frozen-numbers green), and the contract's risk_watchlist flags this exact area as judgment-only, so this is the eyeball it asks for, not a violation. The bare-headline choice is also correct against watchlist item 7.
  - *Suggested fix:* Optional: drop the sources unit's talk_track (no element: key, so the >=150-word floor does not apply) or make it a fixed localized renderer instruction the template names explicitly.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-workspace/tests/test-text-to-narrative-brief.sh:30-34` — The diff adds three new suite cases but the header records three recipes, the third pinned to the pre-existing ttn-01-green-slides. The new case ttn-20-slides-points-still-required has no recorded recipe of its own. Its guard IS load-bearing - independently verified with --expr 's{elif not isinstance\(points, list\) or not points:}{elif False:}' --case ttn-20-slides-points-still-required returning guard_verified - so this is a bookkeeping gap in the recorded list, not an unproven guard, and item 13's purpose clause is met (it asks for one recipe per new CHECK, and both new checks have one).
  - *Suggested fix:* Add one header line: --expr 's{elif not isinstance\(points, list\) or not points:}{elif False:}' --case ttn-20-slides-points-still-required

### Positive observations

- Every risk on the contract's watchlist was actually handled, including the four the escalation flagged as underdetermined by the issue. ttn-12-no-styling-keys was re-pointed from 'type: cover' to 'type: bluf' rather than deleted, so no-styling-keys coverage survives and mutate()'s changed-text assertion still holds. (plugin-dev:plugin-validator)
- The presentation-intent note was placed at lines 120-128, outside the SHARED block at 30/71; the byte-compared region is untouched, publish-routing.md is not in the diff, and test-presentation-intent-sync.sh exits 0 with 11 PASS. The note states WHY the vocabularies differ rather than just asserting the omission. (plugin-dev:plugin-validator)
- The two new checks are genuinely distinguishable, not one check with two messages: separate functions, separate CHECKS entries, distinct failure text, and separate suite cases graded by a new has_fail_unit helper requiring the finding's unit field to equal the expected number - so 'names the offending unit number' is executably enforced, not just claimed. (plugin-dev:plugin-validator)
- The exemption keys on f.get('type') == 'sources', the type-predicated form the watchlist named, paired with a mutant that strips slide_points from a non-sources slide and asserts a density-slides fail. The mutant transform asserts nb != blk, so a stale re-point is a hard error rather than a silent no-op. (plugin-dev:plugin-validator)
- The fixture restructure is coherent beyond what any check reads: dark_slides updated [1,8]->[1,7] in both fixtures to follow the renumbering (watchlist item 10, explicitly ungated), and the four trailing note: lines kept after the appended sources unit so the Brief constructor's hoist from units[-1] still resolves (watchlist item 11). (plugin-dev:plugin-validator)
- No collateral breakage outside the suite under test: thirteen further cogni-workspace suites all exit 0. A repo-wide grep for 'type: design-brief' finds only the five fixtures, so the BRIEF_VERSION bump strands no other document. (plugin-dev:plugin-validator)
- The five-site sweep was done consistently, not site-by-site: SKILL.md:299, template :37, :70, :89 and :151 all move together, and a tree-wide grep for 'cover' turns up no site left behind - the usual de-drift-misses-half failure mode did not occur. (plugin-dev:skill-reviewer)
- Template :151 is a checkable claim and it holds against the code: both new checks fail with the offending unit number after an early return on non-slides targets. The documentation asserts exactly what the guard does. (plugin-dev:skill-reviewer)
- The presentation-intent.md note was added at lines 121-124, OUTSIDE the shared block; the clause-3 type list at :52 is unchanged, so the byte-for-byte cross-plugin sync guard is not disturbed, and the note explains WHY sources is absent rather than silently omitting it. (plugin-dev:skill-reviewer)
- The slide_points exemption is predicated on the unit's TYPE value, not position, and the elif arm still fires for every other slide - the too-broad 'last unit is exempt' shape was avoided. (plugin-dev:skill-reviewer)
- No breadcrumb leakage across the changed SKILL.md, references and library. (plugin-dev:skill-reviewer)
- The derivation change is internally consistent with the ceilings: 'seven to twelve units' sits inside units_min 5 / units_max_default 15, so a mandatory sources unit does not push the documented range outside the enforced one. (plugin-dev:skill-reviewer)
- Contract item 20, SKILL.md site: satisfied. Line 299 no longer derives slides as "cover, BLUF, ..." and does name the sources close; a grep for cover over the whole file returns zero hits. (cogni-service:skill-quality-reviewer)
- Contract item 6 (no breadcrumb leaks): clean. (cogni-service:skill-quality-reviewer)
- Contract item 7 (frontmatter): untouched; description remains under the 1024-char cap. (cogni-service:skill-quality-reviewer)
- The edit is minimal and correctly located: a single derivation clause in the one phase that owns the brief's unit grammar, deferring the authoritative rule to references/design-brief-template.md rather than restating it. (cogni-service:skill-quality-reviewer)
- Every one of the 20 committed acceptance-checklist items graded satisfied, including the two clauses of issue AC bullet 5 that forced the contract to be re-derived rather than adopted: the climax integer does not name the sources unit, and unit 1's slide_points carry the Executive TL;DR copy. (cogni-service:contract-grading-reviewer)
- No checklist item was unadjudicable — the grading produced no INTENT-UNCLEAR finding, so there is no merger-vs-author intent divergence on this diff. (cogni-service:contract-grading-reviewer)

### Recommendation

---

Issue #1960, opened 2026-09-13
cogni-publishing: design-render support for themes without tokens/*.json


claude-cogni-work:
## Context

Deferred from #1946 during service-resolve (full implementation of the parent shipped the rest). Tracked here so the scope is not silently dropped.

**Deferred item:** Render tier-0 (`theme.md`-only) themes by deriving the required token roles. #1946 design-render reads a theme's authoritative `tokens/*.json`; only `themes/cogni-work` ships them today, so the other bundled themes fail with `invalid-theme` naming the missing token role. Deriving tokens for them belongs to the theme lifecycle (`manage-themes`), not the renderer.

**Parent issue:** #1946
**Plan record:** https://github.com/cogni-work/insight-wave/issues/1946#issuecomment-5656717706

## Acceptance criteria

- [ ] Every bundled theme renders through design-render or fails with an actionable `invalid-theme` naming the missing token role
- [ ] Derived tokens are written through the existing token compiler, never hand-edited

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-09-14T02:06:57Z` (GitHub clock).

Released automatically if triage parks, decomposes, or closes this issue; carried
into resolution otherwise. A claim older than the freshness TTL reads as a strand
and becomes pickable again.

claude-cogni-work:
## Triage Report

- **Severity:** S4-enhancement
- **Component:** cogni-publishing / design-render theme resolution + manage-themes token tier
- **Root cause hypothesis:** n/a (enhancement)
- **Affected files:** `cogni-publishing/themes/{boardroom,clean-slate,editorial,signal}/` (new `tokens/*.json` + generated `tokens.css`, `manifest.json` `tiers.tokens`), a new stdlib derivation script under `cogni-publishing/scripts/` feeding `cogni-publishing/scripts/generate-tokens-css.py`, `cogni-publishing/skills/manage-themes/SKILL.md` (Operation 7 entry point), `cogni-publishing/scripts/render_core.py` (`resolve_theme` no-`tokens/` finding), a suite under `cogni-publishing/tests/`
- **Complexity:** medium
- **Resolution strategy:** feature-build-code

**Findings from the investigation (base `645c1b79`):**

[...]

Issue #1959, opened 2026-09-13
cogni-publishing: embed bundled theme font faces in design-render HTML


claude-cogni-work:
## Context

Deferred from #1946 during service-resolve (full implementation of the parent shipped the rest). Tracked here so the scope is not silently dropped.

**Deferred item:** Embed theme-shipped licensed font faces so brand typefaces render instead of documented fallbacks. #1946 bundles no font file: a brand face a theme names (for example DM Sans in `cogni-work`) resolves to the documented generic chain in `cogni-publishing/references/font-fallbacks-v1.json`, and provenance records the substitution. This follow-up lets a theme ship licensed faces that design-render embeds locally.

**Parent issue:** #1946
**Plan record:** https://github.com/cogni-work/insight-wave/issues/1946#issuecomment-5656717706

## Acceptance criteria

- [ ] A theme may ship font files that design-render embeds as local `@font-face` sources with no remote request
- [ ] The licence text sits under cogni-publishing and a root NOTICE entry resolves (`scripts/check-attribution-path-integrity.py`)
- [ ] Provenance records `substituted: false` for an embedded face, and layout uses its metrics

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-09-14T01:06:45Z` (GitHub clock).

Released automatically if triage parks, decomposes, or closes this issue; carried
into resolution otherwise. A claim older than the freshness TTL reads as a strand
and becomes pickable again.

claude-cogni-work:
## Triage Report

- **Severity:** S4-enhancement
- **Component:** cogni-publishing / design-render (font resolution, HTML adapter, fidelity checks, provenance) + repo-root `NOTICE`
- **Root cause hypothesis:** n/a (enhancement — deferred scope from #1946, not a defect)
[...]

Issue #1958, opened 2026-09-13
cogni-publishing: provision the design-render measurement runtime in CI


claude-cogni-work:
## Context

Deferred from #1946 during service-resolve (full implementation of the parent shipped the rest). Tracked here so the scope is not silently dropped.

**Deferred item:** Provision the pinned design-render browser runtime in CI so the runtime-gated cases run on every PR. #1946 renders HTML in Python stdlib and uses a separately lockfile-pinned `playwright-core` runtime (`cogni-publishing/runtime/`) only to measure a rendered page. The CI "Plugin test suites" job installs no Node, so the two runtime-gated `drnd` cases (isolated offline browser load, DOM geometry re-render) print `SKIP:` there; their PASS evidence currently comes from a provisioned local run.

**Parent issue:** #1946
**Plan record:** https://github.com/cogni-work/insight-wave/issues/1946#issuecomment-5656717706

## Acceptance criteria

- [ ] The Plugin test suites job provisions the runtime from `cogni-publishing/runtime/package-lock.json` in a step outside the suite (no npx, no global browser)
- [ ] The isolated-load and DOM-geometry `drnd` cases print PASS in CI instead of SKIP
- [ ] No suite touches the network

claude-cogni-work:
## Claimed for triage

An autonomous drain claimed this issue at pick time and is triaging it now.
Claimed at `2026-09-14T00:06:43Z` (GitHub clock).

Released automatically if triage parks, decomposes, or closes this issue; carried
into resolution otherwise. A claim older than the freshness TTL reads as a strand
and becomes pickable again.

claude-cogni-work:
## Triage Report

- **Severity:** S4-enhancement
- **Component:** cogni-publishing / design-render measurement runtime + repo CI (`.github/workflows/lint.yml`, job "Plugin test suites (discover and run tests/*.sh)")
- **Root cause hypothesis:** n/a (enhancement — CI coverage gap, not a defect)
- **Affected files:** `.github/workflows/lint.yml`; likely `cogni-publishing/tests/test-design-render.sh` (only if the author makes the SKIP→PASS guarantee structural, see below); prose that states the runtime is operator-provisioned only or that CI SKIPs drnd-34/35: `cogni-publishing/CLAUDE.md`, `cogni-publishing/README.md`, `cogni-publishing/references/design-render.md`
- **Complexity:** medium
- **Resolution strategy:** feature-build-code

### Precondition — lands only after PR #1961

[...]

PR #1961, opened 2026-09-13
cogni-publishing: render branded HTML from resolved compositions

### LINKED ISSUE #1946: cogni-publishing: render branded HTML from resolved compositions
## Plugin
cogni-publishing

## Epic context
Phase 5 of "Roadmap: establish cogni-publishing and retire obsolete workspace responsibilities" implements the browser target as a sibling renderer over the shared composition.

## Use case
Users need a portable branded HTML presentation with readable figures and source links from the frozen composition.

## Proposed solution
Add the HTML adapter and the HTML branch of design-render. Resolve pattern geometry and CSS from the semantic composition and versioned design system; insert original strings programmatically. Implement reusable HTML components and one bounded SVG figure path for the proof patterns. Bundle required assets for portable output, define font fallback behavior, and expose rendering through stdlib JSON-envelope wrappers around a separately pinned runtime. Add basic structural/fidelity checks now; comprehensive visual verification lands in the proof child. Medium renderer PR, limited to the agreed pattern set; presentation navigation is included only as needed for the deliverable.

## Alternatives considered
Ad hoc generated HTML per request avoids reusable constraints. Using HTML as an obligatory PPTX conversion input would couple the targets and impair editable output.

## Acceptance criteria
- [ ] The HTML rendering entry point consumes validated composition and theme artifacts and emits a target plan, artifact, and provenance manifest. Evidence: fixture invocation fails if any required output is missing.
- [ ] Rendered visible strings, notes, and sources retain the original content without normalization that changes meaning. Evidence: DOM/content extraction against the frozen brief detects omissions or changed strings.
- [ ] All proof patterns render with their declared semantics and brand tokens. Evidence: DOM/SVG assertions verify comparison relationships, chart values, conceptual labeling, and source-register placement.
- [ ] Citation markers resolve to supplied URLs and source records. Evidence: link extraction rejects missing or substituted targets.
- [ ] Output opens with packaged assets without requiring the wider ecosystem or a model API. Evidence: isolated browser loading reports no missing required asset requests.
- [ ] Reading order, alternative descriptions, and non-color distinctions are represented. Evidence: DOM/accessibility inspection fails on a missing declared description or content order mismatch.
- [ ] Font resolution and substitutions are explicit and recorded. Evidence: a missing-font fixture either uses a documented fallback with recomputed layout or fails clearly; silent substitution fails.
- [ ] Re-rendering a captured plan under the same pinned runtime preserves logical content and geometry. Evidence: normalized plan and DOM/geometry comparison ignores irrelevant timestamps but fails material drift.
- [ ] The wrapper remains stdlib-only and runtime absence yields an actionable JSON failure. Evidence: minimal-environment invocation and dependency lock inspection fail on an implicit install or unpinned required package.
- [ ] Fidelity checks detect a damaged renderer. Evidence: the HTML suite prints a frozen-copy case and a documented installed cogni-service/scripts/mutation-check.sh recipe mutates original-string insertion, failing that exact case with repository --root, mutated relative --file, and single-quoted Perl -0pi --expr.

_Affected components: new cogni-publishing design-render skill, HTML target adapter/components, isolated renderer runtime, fixtures and tests_

_Complexity hint: medium_

## Planning provenance
- Repository: cogni-work/insight-wave
- Existing plugin: cogni-workspace; destination: cogni-publishing
- Ownership verdict: owned
- Evidence: the CWD .claude-plugin/marketplace.json lists cogni-workspace
- CWD repository: cogni-work/insight-wave

## Depends on
- #1945 cogni-publishing: compose frozen briefs with reusable visual patterns

## Blocks
- #1948 cogni-publishing: verify and prove the two-brand HTML and PPTX path

claude-cogni-work:
Closes #1946.

## Summary

- New **`design-render`** skill with its HTML target. It turns a validated `semantic-composition@2` plus a theme into `target-plan.json` (a new `target-resolved-plan@2`), a single self-contained `index.html`, and `provenance.json` (`render-provenance@1`).
- Rendering is Python 3 stdlib only and needs no Node, model API, network or cogni-workspace. A separately pinned `playwright-core` runtime (committed `runtime/package.json` + `package-lock.json`) is used only by `measure` / `render --measure`: an offline isolated load, DOM geometry, clipped-copy detection and the fonts the platform actually used.
- `validate-publishing.py` gains `target-resolved-plan@2` (it accepts `semantic-composition@2` + `normalized-brief@1` + `pattern-library@1`) and a `check-plan` command. `validate` routes a plan@2 chain to the same check.
- Every original string reaches the page through one insertion function, as literal text, never as markup. The render runs the independent fidelity checks on its own page before it writes anything, so a damaged renderer is refused rather than shipped. The checks (`check-html`, `check-provenance`, `compare`, `check-runtime-lock`) are also standalone commands the proof child can reuse.

## Changes

- `cogni-publishing/skills/design-render/SKILL.md`, `scripts/design-render.py` (CLI wrapper), `scripts/render_core.py` (theme and font resolution, plan@2 geometry, provenance), `scripts/html_adapter.py` (HTML components, one bounded SVG path for charts and system figures), `scripts/render_checks.py` (html.parser extraction, pattern/citation/a11y/asset/token checks, comparator, lock inspection).
- `references/design-render.md` (normative), `target-resolved-plan-v2.schema.json`, `render-provenance-v1.schema.json`, `font-fallbacks-v1.json`. Amended: `artifact-contracts.md` (compatibility row, references, `check-plan`), `configuration-and-renderer-boundary.md`.
- `validate-publishing.py`: plan@2 in `SUPPORTED`/`COMPATIBLE`, `check_plan`, the `check-plan` subcommand and the `validate` routing. The mutation-keyed one-line bodies and the register line are byte-identical; no environment, subprocess, socket or urllib use was added (pubc-16 now also isolation-audits `check-plan`).
- `runtime/`: the exact-pinned manifest, the integrity-hashed lockfile, `measure.mjs`, an operator-run `provision.sh`, and a `.gitignore` covering `node_modules/`, `browsers/` and `.provisioned.json`.
- Fixtures under `tests/fixtures/render/`: a German direct brief with umlauts, ß, numbers with units and edge strings (`<b>`, `<script>`, `&amp;`, quotes, a non-breaking space) plus its `compose` output; a fixed fixture `cogni-work` theme; a captured plan@2. New suite `tests/test-design-render.sh` (`drnd-NN-*`, 39 cases).
- Inventory: `docs/command-reference.md` (skills line, and the stale "no local renderer" bullet), root `README.md` counts (3→4, roll-up 91→92), cogni-publishing `README.md`/`CLAUDE.md`, and a one-paragraph delivery note in `docs/architecture/publishing-transition.md`.

## Scope decisions

- Rendering does not need the browser runtime, so a plain `render` succeeds without it. Runtime absence is a single actionable failure envelope, returned before anything is written, only for the operations that need the runtime (`measure`, `render --measure`). The wrapper resolves the runtime only from `runtime/.provisioned.json` and re-verifies the interpreter version, the installed package version and the lockfile digest. It never looks anything up on PATH and never installs.
- Fonts: no font files are bundled. A requested face that is not shipped falls through the theme's own stack to a generic family documented in `font-fallbacks-v1.json`. Layout uses that resolved face's metrics, and provenance records `substituted: true` with the chain. A stack with no resolvable member fails as `font-unresolved`. Copy is set in the resolved chain, never in a skipped family, so a face installed locally cannot substitute silently.
- Copy is plain text. Markdown characters stay visible, and nothing is truncated, clamped or hidden; boxes grow instead. Chart value labels are the brief's numeric literal plus the unit, exactly as written, with no locale formatting.
- The page adds no visible text of its own beyond a closed vocabulary documented in `design-render.md`: `[n]` links for direct-brief sources and relationship-kind tokens.
- The source register must be the last unit (`register-not-last`), per the brief's rendering contract.
- `--target pptx` is rejected as `unsupported-target`; that target belongs to the sibling PPTX issue.
- No page script: sections sit in DOM order and use CSS scroll-snap only.

## Deviations from acceptance criteria

- Isolated browser load and DOM geometry re-render: in CI (no Node) these two cases print `SKIP:` naming the absent runtime, as review-contract item 22 permits, and never `PASS:`. Their PASS evidence comes from the provisioned local run below. Provisioning the runtime in CI is filed as #1958.
- Brand faces (for example DM Sans) are never embedded. Every brand theme therefore renders with a recorded, documented fallback, which is the criterion's fallback arm. Embedding licensed font files is filed as #1959.

## Local evidence (revision head `1be222f3`)

- The portability scan now reads only loadable surfaces. `drnd-37` renders frozen copy containing "Risk profile:", "cogni-workspace", "/tmp/" and "node_modules" and passes `check-html`. `drnd-38` proves genuine `file:`, absolute-path and `node_modules/` references are still refused. `drnd-39` covers a negative chart value drawn left of the shared zero baseline.
- `bash cogni-publishing/tests/test-design-render.sh` without the runtime (the CI shape): **37 passed, 0 failed, 2 skipped** under Python 3.14.2, and the same under `/usr/bin/python3` 3.9.6, where `drnd-02` now passes through its stdlib-origin fallback. The two skips are `drnd-34` and `drnd-35`. Emitted first tokens: 39 unique ids.
- Against the pre-revision `render_checks.py` and `html_adapter.py`, `drnd-37` and `drnd-39` go red. With the scan made blind, `drnd-38` goes red.
- Both recorded recipes replayed with the installed harness at this head: `mutated_result: red`, `restored_result: green`, `verdict: guard_verified`.
- `python3 scripts/run-plugin-tests.py --filter cogni-publishing` green; `check-skill-spec`, `check-case-id-pairing`, `check-result-line-plainness`, `check-mutation-recipe-harness`, `check-breadcrumbs`, `check-code-span-nesting`, `check-command-reference-sync`, `check-readme-inventory-sync`, `check-marketplace-manifest-sync`, `check-attribution-path-integrity` and `check-version-bump` (base `origin/main`) all exit 0. Folded descriptions: design-render 943, publishing-validate 949 characters.

## Local evidence (head `7acf9972`)

- `bash cogni-publishing/tests/test-design-render.sh` with the runtime provisioned (`bash cogni-publishing/runtime/provision.sh`; node v22.19.0, playwright-core 1.63.0, Chrome Headless Shell 153.0.8010.12): **35 passed, 0 failed, 0 skipped**, including `drnd-34-browser-isolated-load` and `drnd-35-browser-geometry-rerender`.
- Same suite with the provisioning record hidden (the CI shape): **33 passed, 0 failed, 2 skipped**; the two skips are exactly `drnd-34` and `drnd-35`, each naming the absent runtime.
- Offline measurement of all three fixture pages: 0 blocked and 0 failed requests, 0 clipped copy elements.
- `python3 scripts/run-plugin-tests.py`: **157/157 suites passed**. `--filter cogni-publishing`: 8/8.
- Emitted first tokens of the drnd suite grouped by running it: 35 unique ids, no duplicates.
- The lockfile's sha512 integrity matches an independently downloaded `playwright-core-1.63.0.tgz`.
- `check-skill-spec`, `check-command-reference-sync`, `check-readme-inventory-sync`, `check-marketplace-manifest-sync`, `check-version-bump` (base `origin/main`), `check-breadcrumbs`, `check-case-id-pairing`, `check-result-line-plainness`, `check-mutation-recipe-harness`, `check-attribution-path-integrity`, `check-code-span-nesting`, `check-mcp-tool-grant`, `check-output-style-placement`, `check-workflow-yaml` and `cogni-workspace/scripts/check-skill-names.sh` all exit 0. cogni-service `check-frontmatter.sh`, `measure-skill-length.sh` (66 body lines) and `check-structure-all.sh` are clean.

## Test plan

- [x] `bash cogni-publishing/tests/test-design-render.sh` — every case PASS; the two runtime-gated cases print `SKIP:` without the runtime
- [x] `bash cogni-publishing/runtime/provision.sh`, then rerun the suite — the runtime-gated cases print `PASS:`
- [x] `bash cogni-publishing/tests/test-publishing-contracts.sh` (25/25), `bash cogni-publishing/tests/test-design-compose.sh` (70/70) and `python3 scripts/run-plugin-tests.py --filter cogni-publishing` are green
- [x] The mutation recipe below replays red then green
- [ ] Merger: replay the recipe and the guard battery on the pushed head

## Mutation recipe

Damages original-string insertion in the HTML adapter. The frozen-copy case must go red, then green once the file is restored. Replayed locally with the installed harness: `mutated_result: red`, `restored_result: green`, `verdict: guard_verified`.

```bash
bash "$HOME/.claude/plugins/marketplaces/managed-service/cogni-service/scripts/mutation-check.sh" --root . --file cogni-publishing/scripts/html_adapter.py --expr 's/return escape\(value, quote=True\)/return escape(value.upper(), quote=True)/' --test 'bash cogni-publishing/tests/test-design-render.sh' --case drnd-10-frozen-copy
```

A second recipe makes the portability scan read copy text; the prose-paths case must go red. Replayed the same way at `1be222f3` with verdict `guard_verified`.

```bash
bash "$HOME/.claude/plugins/marketplaces/managed-service/cogni-service/scripts/mutation-check.sh" --root . --file cogni-publishing/scripts/render_checks.py --expr 's/if node\.tag == "style":/if True:/' --test 'bash cogni-publishing/tests/test-design-render.sh' --case drnd-37-prose-paths-render
```

## Follow-up issues

- #1958 — Provision the design-render measurement runtime in CI so the runtime-gated cases run on every PR. `drnd-36-python-floor-compiles` likewise prints `SKIP:` on CI's 3.12+ runner, so the Python 3.9 floor is enforced only on hosts with a 3.9–3.11 interpreter until CI provisions one alongside the runtime.
- #1959 — Embed theme-shipped licensed font faces so brand typefaces render instead of documented fallbacks
- #1960 — Render tier-0 (theme.md-only) themes by deriving the required token roles
- #1963 — Make drawn SVG label geometry agree with the target-resolved-plan@2 estimates (latent; every current fixture agrees)

## Open questions

None.

Plan record: https://github.com/cogni-work/insight-wave/issues/1946#issuecomment-5656717706


claude-cogni-work:

#1946 end-to-end to a draft PR. Handing off to the merger identity for
never reviews its own work."

`svc:needs-review` applied to this PR; #1946 marked `svc:resolved`.

Handoff preflight: rc=0 — 6 passed, 0 failed, 1 skipped (agent-length: not applicable, no agent touched). One advisory: the skill-length arm measures the base checkout, which does not contain the new `design-render/SKILL.md`; measured directly on the head it is 66 body lines, under the cap.

### Author-side contract grading (advisory — not a review verdict)

Graded 26 checklist items from the issue-time contract (`origin:
triage`, derived 2026-09-13T16:19:28Z) — the author checking its own diff against a
specification it did not write. Satisfied: 26. Violated: 0
(fixed in-session: 0; residual: 0). Unclear: 0.

Advisory only — this carries no review-verdict label and does not substitute
for the merger's independent review.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1961
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:contract-grading-reviewer

**Summary:** Merged verdict is needs_revision: plugin-dev:plugin-validator raised an un-waived major finding — html_adapter.py:339 uses a nested f-string that only compiles on Python 3.12+, so design-render crashes with an unhandled traceback (no JSON envelope) on the repo's stock Python 3.9/3.10/3.11, silently raising the Python floor; the orchestrator independently confirmed this by compiling all 118 tracked *.py files with /usr/bin/python3 3.9.6 and reproducing exactly this failure. All 26 committed acceptance-checklist items for issue #1946 are otherwise satisfied (contract-grading-reviewer, pass), CI is 18/18 green, and every other perspective passed with only preference-level authoring nits, most of which are merger-added preferences on this gatekeeper review and are non-gating.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (14)

- **[major]** (auto-fixable) `cogni-publishing/scripts/html_adapter.py:339` — A nested f-string (`f"<li>{self.copy_element('span', f'trailer#{b['index']}', notes[b['index']])}</li>"`) reuses the single quote inside a nested f-string, which only Python 3.12+ (PEP 701) accepts. On older Python, importing html_adapter fails before main() runs, so every design-render.py command prints a SyntaxError traceback on stderr, nothing on stdout, and exits 1 — breaking the plugin's one-envelope, nothing-on-stderr contract and silently raising the repo's Python floor from 3.9 to 3.12 (CI stays green only because ubuntu-latest ships 3.12+). Orchestrator-verified: compiling all 118 tracked *.py files with /usr/bin/python3 3.9.6 fails on exactly this file; patching only this line in a scratch copy renders successfully and byte-identically to the Python 3.14 output.
  - *Suggested fix:* Avoid the nested same-quote f-string, e.g. `self.copy_element('span', 'trailer#' + str(b['index']), notes[b['index']])`, or build the key in a local variable first. Leave the `text()` body untouched (the drnd-10 recipe mutates it). Optionally add a suite case that compiles the render modules with the oldest supported interpreter so the floor cannot move silently again.

- **[minor]** (advisory-addition) `cogni-publishing/runtime/provision.sh:21` — provision.sh records the operator's on-PATH Node absolute path and version into .provisioned.json at provisioning time, but only playwright-core and its pinned Chromium build are reproducible across machines via the lockfile — the Node interpreter itself is a per-machine recorded value, not a pin.
  - *Suggested fix:* Either record an exact supported Node major in package.json/provision.sh and have check-runtime-lock verify it, or state explicitly in references/design-render.md that the interpreter is a per-machine recorded host dependency, not part of the pin.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/runtime/provision.sh:26` — `npm ci` and the playwright-core install print progress to stdout ahead of the final JSON envelope, and failures are plain text on stderr — this operator script's stdout is not a single parseable envelope, unlike the repo's script-output convention. It is operator-run and outside scripts/, so non-blocking.
  - *Suggested fix:* Redirect npm/playwright install output to stderr or a log file so stdout carries only the final envelope, and emit a `success:false` envelope on the node-missing path.

- **[minor]** (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md:3` — Most trigger phrases use pipeline terms and the description has no 'applies whenever ... even when the user does not name the skill' clause (unlike the sibling design-compose description), so the skill may under-trigger on plain requests. By hand count the description is already ~973-974 of the 1024 folded characters, leaving little room.
  - *Suggested fix:* Trim the capability sentence (e.g. drop 'plus a provenance record of fonts, pins and fingerprints') and use the freed space for an applies-whenever clause such as 'applies whenever a semantic-composition@2 with an html target needs to become a page.' Avoid every retired phrase in cogni-workspace/references/retired-trigger-phrases.tsv. Re-measure with check-skill-spec.py.

- **[minor]** (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md:19,41-51` — Workflow step 1 says to confirm `check-composition`, but the Commands block never shows that command (only design-compose). No remedy is given for `measure-failed` or `render-incomplete`, both defined in design-render.md. `compare` also accepts `--expected-html/--actual-html` and `check-runtime-lock` accepts `--runtime-dir`, neither shown. Completeness polish only, since references/design-render.md covers all of it.
  - *Suggested fix:* Either add the check-composition invocation or note that `render` runs the same validation. Add a Read-the-result bullet for `measure-failed` / `render-incomplete`: report the error verbatim and re-run; never hand-write the missing output.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md: body §Workflow step 3 (line 21)` — Step 3 lists what a success report must contain (three paths, unit count, and the font substitution disclosure when applicable) but gives no worked example, so the user-facing wording of the no-silent-substitution guarantee is likely to drift run to run.
  - *Suggested fix:* Add a short worked example after step 3, e.g. 'Rendered 7 units to out/index.html ... Font: the theme asks for Inter, which ships nowhere; rendered with system-ui via the chain Inter → system-ui → sans-serif, recorded as a substitution.'

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md: body §Commands (lines 30, 33) and §Read the result` — `check-provenance` and `check-runtime-lock` appear in the Commands block, but no workflow step says when to run them, and their codes (silent-substitution, font-unrecorded, output-digest, range-pin, lock-mismatch, install-tracked) have no action class — they reach the model only via the line-53 catch-all.
  - *Suggested fix:* Add a workflow step: 'To audit a delivered bundle, run check-provenance on its provenance.json; to confirm the runtime pin before provisioning/measuring, run check-runtime-lock.' Add a Read-the-result bullet for a provenance/lock finding directing to re-render or re-provision from the committed lockfile, never hand-editing provenance.json or the lock.

- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:43,49` — The finding code `copy-changed` sits in two remedy groups that say different things: line 43 treats it as a composition finding (fix through design-compose), line 49 treats the same code from check-html as a fidelity finding (re-render). `copy-changed` is genuinely emitted by both validate-publishing.py and render_checks.py's `frozen-copy` check. `fingerprint-mismatch` has the same ambiguity (composition code vs. provenance code at render_checks.py:541). Routing by code alone can send the agent to re-compose when the page itself failed.
  - *Suggested fix:* Route by `data.check` or the emitting command rather than by `data.code` alone: a `copy-changed`/`fingerprint-mismatch` from composition validation or check-plan is a composition finding; the same code from check-html (`check: frozen-copy`) or check-provenance means re-render.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md:50` — The provisioning command `bash ${CLAUDE_PLUGIN_ROOT}/runtime/provision.sh` does not quote the variable, unlike every other invocation in the Commands block. A plugin root path containing a space will hit a word-splitting failure.
  - *Suggested fix:* Write it as `bash "${CLAUDE_PLUGIN_ROOT}/runtime/provision.sh"`.

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md: body §Read the result, `runtime-missing / runtime-unpinned` bullet (line 50)` — This bullet stacks two 'never' rules with no stated reason ('never install anything yourself, and never substitute a browser found elsewhere') and uses second person. Separately, the provisioning command is unquoted while the Commands block quotes the same variable.
  - *Suggested fix:* Replace with wording that explains why (a measurement is only reproducible against the lockfile-pinned engine, and an implicit install would break the stdlib-only, no-install contract), drop second person, and quote the provision.sh invocation.

- **[minor]** (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md:3` — The frontmatter description was not updated for the new `check-plan` capability: it still says the skill validates only a normalized-brief@1 → semantic-composition@1 → target-resolved-plan@1 chain, with no trigger phrase for checking a renderer's target-resolved-plan@2. The body now routes plan@2 grading here (lines 19, 22), but the skill-selection text does not say so. A routing-precision gap, not a failure — the existing 'validate the composition and target plan' trigger and design-render/SKILL.md:34 still get users to the command.
  - *Suggested fix:* Widen the clause to cover 'a renderer's target-resolved-plan@2 against its brief and composition, for version compatibility and dangling references'; optionally add a quoted trigger such as "check the render plan". Keep under 1024 folded characters and avoid retired phrases.

- **[minor]** (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md:43` — The Exit 0 bullet says 'For `validate`, `data` lists the three artifacts and their counts,' but line 22 now routes a plan@2 chain to check_plan, whose return shape (valid, artifact_id, target, content_fingerprint, units, slots plus an artifacts list) differs from the @1 summary. The bullet does not say so.
  - *Suggested fix:* Add: 'for a plan@2 chain it also carries the `check-plan` summary.'

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md: body §Read the result, closing reference paragraph (line 47)` — `check-plan` can return layout findings (`reordered-unit`, `typography-relaxed`) whose rules live in references/design-render.md §target-resolved-plan@2, but the skill only points to artifact-contracts.md and configuration-and-renderer-boundary.md, leaving no pointer to the broken rule when a user asks why a plan was rejected on layout.
  - *Suggested fix:* Append a pointer at line 47 to references/design-render.md as the home of the target-resolved-plan@2 layout rules (slot reading order, typography roles, measured faces), to read when a check-plan finding needs explaining.

- **[minor]** (waived) `PR diff` — Diff is 4532 changed lines against a 400-line reviewable bound. Larger PRs are harder to review thoroughly, so a blind pass is riskier.
  - *Justification:* Advisory by design, and structurally unclosable by a revising agent: the only remedy is splitting already-pushed history, which this repo's no-force-push contract forecloses. Waived at emit time so it renders with the (waived) audit line but never enters a non-waived gating partition.

### Positive observations

- Every repo guard exits 0 against the PR head: check-version-bump, check-breadcrumbs, check-skill-spec (92 skills, no new baseline entries), check-command-reference-sync, check-readme-inventory-sync, check-marketplace-manifest-sync, check-case-id-pairing, check-result-line-plainness, check-mutation-recipe-harness, check-attribution-path-integrity, check-code-span-nesting, check-mcp-tool-grant and check-output-style-placement. (plugin-dev:plugin-validator)
- Full sweep in a git-initialised copy of the head: 157/157 suites pass; test-design-render.sh prints 35 uniquely-named drnd-NN-<discriminator> lines, 33 pass, only drnd-34/35 SKIP (runtime absent, deferred to #1958). (plugin-dev:plugin-validator)
- Replaying the recorded mutation recipe through the installed managed-service harness gives verdict guard_verified — drnd-10-frozen-copy is red when mutated, green once restored; independently reproduced by the orchestrator. (plugin-dev:plugin-validator; orchestrator-verified)
- The wrapper answers every failure with a JSON envelope, stages outputs in a temp directory, writes nothing on rejection, and resolves the runtime only from the provisioning record with lock digest, package version and interpreter version checks. (plugin-dev:plugin-validator)
- design-render/SKILL.md is a lean, well-formed skill (71 lines, ~7144 body chars against a 15000 cap); every command, flag and resource it names exists in the PR head, no claimed trigger phrase is retired, and there are no breadcrumbs. (plugin-dev:skill-reviewer)
- The anti-patch stance holds throughout: never edit the page/plan/composition to make a check pass, re-render from inputs, never install the runtime, never use a browser found elsewhere. (plugin-dev:skill-reviewer)
- publishing-validate's five-line edit is fully accurate: check-plan's flags, the plan@2 routing inside validate, and the success-data fields all match validate-publishing.py at 7acf9972. (plugin-dev:skill-reviewer)
- Progressive disclosure is done well for both skills — schemas, code catalogues and fallback data live in references/, each with a 'Read it when' pointer; body sizes are well under their caps and no breadcrumbs are introduced. (cogni-service:skill-quality-reviewer)
- Against the review contract, none of the 26 items targets publishing-validate's edit directly and the edit breaks none of them — stays inside cogni-publishing/**, adds no breadcrumb, leaves trigger phrases unchanged, touches no version value. (cogni-service:skill-quality-reviewer)
- All 26 committed acceptance-checklist items for issue #1946 are satisfied, verified by reading the render source, running the drnd suite plus test-publishing-contracts.sh/test-design-compose.sh at the PR head, and executing every relevant repo guard clean; PR CI is 18/18 green. (cogni-service:contract-grading-reviewer)

### Recommendation

---

claude-cogni-work:
## Revision posted

**Commit:** 9120032439b81422a84db2a717b975806ab59896
**Addressed:** the merger NEEDS_REVISION verdict on 7acf9972. 13 findings fixed, 1 waived. The gating major (the Python 3.12-only nested f-string) gets a byte-identical rewrite, and a new drnd-36 case pins the Python 3.9 floor. The gating routing minor now routes by emitting command, data.artifact and data.check. All advisory minors are fixed.
**New verdict:** ⏳ DEFERRED_TO_MERGER. This author lane runs no self-verdict; the independent review belongs to the merger (see the prior verdict at https://github.com/cogni-work/insight-wave/pull/1961#issuecomment-5657282574).

### Findings resolved (14)

- **[major]** `cogni-publishing/scripts/html_adapter.py:339` — The nested same-quote f-string only compiles on Python 3.12+, so on 3.9–3.11 every design-render.py command crashed with no envelope.
  - *Resolution:* The key is now `'trailer#' + str(b['index'])`. That is valid on 3.9 and byte-identical, because `copy_element` applies `attr(str(key))`, and `text()` is untouched. A 3.10+/3.12+ syntax scan of render_core.py, render_checks.py, design-render.py, validate-publishing.py and generate-tokens-css.py found nothing else. The new case `drnd-36-python-floor-compiles` compiles all six render-path modules in memory under the oldest 3.9–3.11 interpreter on the host. It prints SKIP, never PASS, where there is none, as on CI's 3.12+ runner. The floor is also stated in cogni-publishing/CLAUDE.md. Verified here: `/usr/bin/python3` 3.9.6 compiles all six, and its render of the narrative fixture is byte-identical to the Python 3.14 render.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/runtime/provision.sh:21` — The Node interpreter is a per-machine recorded value, not a pin.
  - *Resolution:* I took the documentation option. references/design-render.md §Runtime boundary now states that Node (20 or newer, on PATH or named by NODE) is a per-machine host dependency outside the pin: the lockfile makes playwright-core and its browser build reproducible, and the recorded interpreter binds only later measurements on the same machine. I declined to pin a Node major checked by check-runtime-lock, because that is a new runtime contract rather than a fix.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/runtime/provision.sh:26` — The script's stdout was not a single parseable envelope.
  - *Resolution:* `npm ci` and the browser install now write to stderr. A fixed-message `fail()` helper emits one `success:false` envelope on stdout for a missing node, a missing npm, an npm ci failure and a browser-install failure. I exercised the missing-node path; it returns one parseable envelope and exits 1.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:3` — The description had no applies-whenever clause.
  - *Resolution:* I dropped the provenance-record clause and added "It applies whenever a semantic-composition@2 with an html target needs to become a page." That brings the description to 1004 of 1024 characters (check-skill-spec and drnd-01 are green). The "even when the user does not name the skill" tail would push it to about 1049, so I left it out. No quoted trigger phrase was added or removed.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:19,41-51` — The check-composition invocation, the measure-failed / render-incomplete remedy and the optional flags were missing.
  - *Resolution:* Step 1 now names check-composition as the design-compose command and says that render repeats it. There is a new bullet for measure-failed / render-incomplete: report the error verbatim and re-run, and never hand-write the missing output. The Commands block shows `--expected-html/--actual-html` and `--runtime-dir`.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md: body §Workflow step 3 (line 21)` — There was no worked example of the success report.
  - *Resolution:* Added one after step 3. It covers the units, the three outputs, and a DM Sans → system-ui substitution named with the documented chain system-ui, sans-serif (matching font-fallbacks-v1.json).
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md: body §Commands (lines 30, 33) and §Read the result` — check-provenance and check-runtime-lock had no workflow step and no action class.
  - *Resolution:* New workflow step 6 covers both. A new provenance/lock finding bullet says to re-render or re-provision from the committed lockfile, and never to hand-edit provenance.json, package.json or the lockfile.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:43,49` — copy-changed and fingerprint-mismatch were routed by code alone.
  - *Resolution:* Findings now route by emitting command, `data.artifact` and `data.check`, and the section says so up front. A composition finding (`semantic_composition`) goes to design-compose. A check-plan finding (`target_resolved_plan`, check `slot-content` / `content-fingerprint`) means a stale plan, so re-render. A check-html copy-changed under `frozen-copy`, or a check-provenance fingerprint-mismatch under `fingerprint`, also means re-render. This refines the suggestion: check_plan validates the composition first (validate-publishing.py `check_plan`), and its own plan-side findings carry artifact `target_resolved_plan` (`plan_error`). So not every check-plan finding is a composition finding.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:50` — The provisioning command was unquoted.
  - *Resolution:* It is now `bash "${CLAUDE_PLUGIN_ROOT}/runtime/provision.sh"`.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md: body §Read the result, runtime-missing / runtime-unpinned bullet (line 50)` — The bullet stacked unexplained "never" rules in second person.
  - *Resolution:* Rewritten without second person. It now gives the reasons: a measurement is reproducible only against the lockfile-pinned engine, and an implicit install or a foreign browser would break the no-install contract. It keeps the quoted command.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:3` — The description did not cover target-resolved-plan@2.
  - *Resolution:* The chain clause is widened to "…target-resolved-plan@1 chain, or a renderer's target-resolved-plan@2, for version compatibility…". I added the quoted trigger "check the render plan", which is not in retired-trigger-phrases.tsv. The description is 983 of 1024 characters.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:43` — The Exit 0 bullet omitted the plan@2 return shape.
  - *Resolution:* It now says that validate lists the three artifacts, with their counts for a plan@1 chain and with the check-plan summary (valid, artifact_id, target, content_fingerprint, units, slots) for a plan@2 chain.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md: body §Read the result, closing reference paragraph (line 47)` — There was no pointer to the plan@2 layout rules.
  - *Resolution:* Added a pointer to references/design-render.md §target-resolved-plan@2, to read when a check-plan finding such as reordered-unit or typography-relaxed needs explaining.
  - *Commit:* 9120032439b81422a84db2a717b975806ab59896
- **[minor]** (waived) `PR diff` — Diff is 4532 changed lines against a 400-line reviewable bound.
  - *Justification:* Advisory by design, and structurally unclosable by a revising agent: the only remedy is splitting already-pushed history, which this repo's no-force-push contract forecloses. Waived at emit time so it renders with the (waived) audit line but never enters a non-waived gating partition.

### Verification at 91200324

- `bash cogni-publishing/tests/test-design-render.sh`: 34 passed, 0 failed, 2 skipped. It prints 36 unique result-line ids; drnd-36 passes on this host; drnd-34/35 SKIP because the runtime is absent. The PR body's "35 cases" is now 36.
- `python3 scripts/run-plugin-tests.py --filter cogni-publishing`: every suite passes.
- These all exit 0: check-skill-spec, check-breadcrumbs, check-case-id-pairing, check-result-line-plainness, check-code-span-nesting, check-mutation-recipe-harness, check-version-bump, check-command-reference-sync, check-readme-inventory-sync, check-marketplace-manifest-sync, check-attribution-path-integrity, check-mcp-tool-grant, check-output-style-placement and check-skill-names.
- Handoff preflight: rc=0, with 6 passed, 0 failed and 1 skipped (agent-length, not applicable).
- Both mutation checks end `guard_verified`, red when mutated and green when restored. The drnd-36 check below reintroduces the 3.12-only f-string, and the recorded drnd-10 recipe replays the same way:

```bash
bash "$HOME/.claude/plugins/marketplaces/managed-service/cogni-service/scripts/mutation-check.sh" --root . --file cogni-publishing/scripts/html_adapter.py --expr "s/'trailer#' \+ str\(b\['index'\]\)/f'trailer#{b['index']}'/" --test 'bash cogni-publishing/tests/test-design-render.sh' --case drnd-36-python-floor-compiles
```

drnd-36 can only discriminate on a host with a Python 3.9–3.11 interpreter. On CI's 3.12+ runner it prints SKIP by design.

Auto-revise triggered by bot-authored verdict findings — no human feedback since the last automated review. See the linked verdict comment above for the original findings.

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1961
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:skill-quality-reviewer, cogni-service:contract-grading-reviewer

**Summary:** plugin-dev:plugin-validator's major finding — check_assets in cogni-publishing/scripts/render_checks.py:252 runs an unanchored `file:` pattern via re.search over the whole rendered page, so valid frozen copy containing ordinary words like "profile:" (or mentions of cogni-workspace/node_modules/a Unix path) is wrongly refused as local-reference — was re-confirmed present at the reviewed head by the orchestrator, so despite all 26 committed contract items being satisfied and clean skill-quality reviews across both changed skills, the PR needs revision before merge.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (14)

- **[major]** `cogni-publishing/scripts/render_checks.py:252` — check_assets runs its local-path regexes (r"file:", r"(?<![A-Za-z0-9])/(?:Users|home|private|tmp|var|opt)/", r"cogni-workspace", r"node_modules", r"ms-playwright") with re.search over the whole serialized page, visible copy included. `file:` has no scheme anchor, so it matches inside the ordinary English word "profile:". Because render runs check_html on its own page before writing, a valid, validated brief whose frozen copy says "Risk profile: ...", mentions cogni-workspace or node_modules, or quotes a Unix path is refused as `local-reference` under check `assets`, and nothing is written. Reproduced from direct-costs-v1.json by changing only answer.body, normalizing and recomposing from a stripped draft: "Risk profile: ..." gave rc=1 local-reference 'file:'; "Built with cogni-workspace, ..." gave rc=1 local-reference 'cogni-workspace'; "Logs sit under /tmp/ ..." gave rc=1 local-reference '/tmp/'; the control body gave rc=0 with all three outputs. The SKILL routes a fidelity finding to "Re-render from the inputs; never patch the page", so the only way out is to edit frozen copy. That defeats the copy-is-frozen premise and issue AC 1 (the entry point consumes a validated composition and emits the three outputs). It also goes beyond the plugin's own normative rule (design-render.md:98 bans absolute or file paths as references, not as prose).
  - *Suggested fix:* Scope the path, runtime-directory and workspace scans to the places a reference can load from: attribute values other than navigational <a href>, <style> text and inline style attributes. Never scan text nodes or data-copy subtrees. Anchor the scheme test as a URL scheme at the start of an attribute or url() value (for example ^\s*file:), not a bare substring. Add a drnd case whose frozen copy contains "Risk profile:", "cogni-workspace" and "/tmp/" and must render and pass check-html.

- **[minor]** (waived) `PR diff` — Diff is 4578 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.
  - *Justification:* Advisory by design, and structurally unclosable by a revising agent: the only remedy is splitting already-pushed history, which this repo's no-force-push contract forecloses. Waived at emit time so it renders with the (waived) audit line but never enters a non-waived gating partition.

- **[minor]** `cogni-publishing/scripts/html_adapter.py:186` — The bounded chart path scales bars by `abs(float(item['value']))`, so a negative data point is drawn as a bar identical in direction to a positive one of the same size. Only the value label carries the sign. Every point in one chart also shares one peak regardless of unit. Labels stay correct (item 13 holds), but the figure can visually misstate the data it sources.
  - *Suggested fix:* Draw negative values from a zero baseline in the opposite direction, or reject mixed-sign or mixed-unit series with a named finding until the pattern library declares how to draw them.

- **[minor]** (advisory-addition) `cogni-publishing/scripts/html_adapter.py:229` — SVG <text> for entity nodes and chart labels never wraps. build_plan/layout_slot estimates entity slot height with multi-line wrapping (render_core.py:393-400), while system() draws each node one line tall. For long labels the recorded plan geometry and the drawn SVG disagree, and label text can run past its node or over its bar. The two agree for every current fixture, so this is latent. The proof child owns visual quality.
  - *Suggested fix:* Either break SVG labels into <tspan> lines using the same estimate_lines width the plan uses, or make the plan lay out entity nodes one line tall so the plan matches the drawing.

- **[minor]** (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md:3` — By a hand count, the folded description is about 1004 of the 1024 characters the spec allows, which leaves about 20 characters of headroom. The HTML-only scope is stated three times: 'renders its HTML target', then 'It applies whenever a semantic-composition@2 with an html target needs to become a page.', then 'It renders the html target only.' The PPTX sibling is expected to extend this same SKILL.md with its own trigger phrases, and with this little room it will have to rewrite the description rather than add to it. check-skill-spec's hard description arm would go red first.
  - *Suggested fix:* Drop the redundant sentence 'It applies whenever a semantic-composition@2 with an html target needs to become a page.' (about 89 characters). Keep one scope statement ('It renders the html target only.'), or fold it into the lays-out sentence as '...and renders it as one self-contained HTML page...'. This leaves roughly 110 characters of headroom for the sibling target's phrases without touching any of the eight quoted triggers.

- **[minor]** (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md:3` — Most trigger phrases use the artifact's own terms ('render this composition', 'lay out the semantic composition for the browser'). A user who asks for the output in plain words, such as a standalone branded web page from the composition, reaches the skill only through the prose sentences. The retired ledger rules out the obvious alternatives ('html presentation', 'browser presentation', 'render slides as html'), so a natural phrase has to be picked carefully.
  - *Suggested fix:* Optional, once headroom is recovered: add one non-retired plain-language phrase such as "export the composition as a standalone HTML page". Check it against cogni-workspace/references/retired-trigger-phrases.tsv before adding.

- **[minor]** (advisory-addition) `cogni-publishing/skills/design-render/SKILL.md:53` — The sentence says '`data.findings[].check` names the rule' but then lists finding codes. In render_checks.py, finding(code, check, ...) emits `copy-omitted` and `copy-changed` under check `frozen-copy` (lines 170-186) and `description-missing` under check `description` (lines 313-317). So `copy-omitted`, `citation-missing` and `description-missing` are codes, not check names. An agent that routes on `check` as this line tells it to would look for values that never appear there. The routing rule at line 44 (route by emitting command) still sends it to the right class, so nothing breaks for the user.
  - *Suggested fix:* Reword to: 'A fidelity finding from `check-html` — each `data.findings[]` entry carries a `code` under a `check`: for example `copy-changed` or `copy-omitted` under `frozen-copy`, `description-missing` under `description`, and `citation-missing`, `chart-semantics`, `remote-asset`, `css-literal`, ….'

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md:18-20` — The new `check-plan` row sits between two cross-skill handoff rows (design-compose, then design-render). In-skill operations and routes to other skills are now interleaved, so the table no longer reads as 'this skill's operations first, then where else to go'.
  - *Suggested fix:* Move the `check-plan` row directly under the `resolve-config` row, so the in-skill operations (normalize ×2, validate, resolve-config, check-plan) come first and the two handoff rows (design-compose, design-render) close the table.

- **[minor]** (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md:3` — The description still ends 'pattern-bound semantic-composition@2 work belongs to design-compose'. Now that this skill validates a renderer's target-resolved-plan@2, and `check_plan` runs `validate_composition` on the @2 composition first (validate-publishing.py:1621), the sentence sits a little awkwardly next to the new capability. The new trigger 'check the render plan' also overlaps design-render, which lists the same `check-plan` command (design-render/SKILL.md:37) and claims 'check the rendered HTML'. Either route works, so this is a routing-precision nit, not a fault. The folded description is also about 983 of the spec's 1024 characters, which leaves little room for future edits before the hard skill-spec arm trips.
  - *Suggested fix:* Clarify the boundary sentence, e.g. 'It does not write, compose or render briefs; authoring a pattern-bound semantic-composition@2 belongs to design-compose and rendering to design-render.' Consider tightening another clause to recover headroom under the 1024-character cap.

- **[minor]** (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md:43` — The Exit 0 bullet is now one long sentence that covers four commands, with a nested em-dash aside and a parenthesized field list for the `validate` plan@2 case. It is accurate (the fields match `check_plan`'s return at validate-publishing.py:1751-1752 plus the `artifacts` list added at :544-546), but it is hard to scan.
  - *Suggested fix:* Split the Exit 0 bullet into one sub-bullet per command (normalize / validate @1 / validate @2 / check-plan / resolve-config).

- **[minor]** (auto-fixable) `cogni-publishing/skills/publishing-validate/SKILL.md:44 (Read the result, Exit 1 bullet)` — The Exit 1 bullet still says to explain the fix in the author terms and stop because no downstream artifact exists for a rejected input. With check-plan (and validate on a plan@2 chain) the rejected input is itself a downstream artifact, and check_plan raises with data.artifact set to semantic_composition or target_resolved_plan (validate-publishing.py:1591); the same code (e.g. copy-changed) appears in both classes. The skill never mentions data.artifact, so a plan finding is routed back to the brief author when the fix is to re-render. design-render SKILL.md lines 44-47 already routes these by data.artifact.
  - *Suggested fix:* After the sentence ending "no downstream artifact exists for a rejected input.", append: " For check-plan, or validate on a target-resolved-plan@2 chain, also quote data.artifact and route by it, not by the code alone: semantic_composition means fix the composition through the design-compose skill; target_resolved_plan means the plan no longer matches its brief and composition, so re-render it through the design-render skill and never edit the plan by hand." (ref: skill-creator §Writing Style (explain the why; keep instructions correct for every path the skill routes to))

- **[minor]** (advisory-addition) `cogni-publishing/skills/publishing-validate/SKILL.md:47` — The new pointer says to read design-render.md §target-resolved-plan@2 'when a `check-plan` finding such as `reordered-unit` or `typography-relaxed` needs explaining'. That section (design-render.md:82) covers reading order and says each slot records its `type_role` and `measured_with`, but it never mentions the type floor or either finding code. The rule that actually explains `typography-relaxed` for a plan slot (a canvas slot set below the pattern's minimum role) is artifact-contracts.md:124. The finding table row at artifact-contracts.md:145 only describes the composition-side `type_floor` case. A reader who follows the pointer for `typography-relaxed` lands on a section that doesn't explain it.
  - *Suggested fix:* Point the finding-code explanation at artifact-contracts.md (the `check-plan` paragraph after its command table), and keep design-render.md §target-resolved-plan@2 for the layout rules. For example: '... holds the plan@2 layout rules — slot reading order, typography roles and measured faces; `artifact-contracts.md` explains what `check-plan` reports as `reordered-unit`, `copy-changed`, `reference-omitted` and `typography-relaxed`.'

- **[minor]** `cogni-publishing/tests/test-design-render.sh:206` — drnd-02 asserts `mod in sys.stdlib_module_names`, an attribute that only exists from Python 3.10. With python3 resolving to 3.9 (the render path's declared floor per cogni-publishing/CLAUDE.md, and /usr/bin/python3 3.9.6 on macOS), the suite prints `AttributeError: module 'sys' has no attribute 'stdlib_module_names'` and `FAIL: drnd-02-stdlib-only` (33 passed, 1 failed, 2 skipped), even though the render path itself runs correctly on 3.9. CI's 3.12+ runner hides this.
  - *Suggested fix:* Fall back when the attribute is absent, for example derive the stdlib set from sysconfig paths or a frozen list on older interpreters. Or print an explicit SKIP line (never PASS) for drnd-02 when sys.version_info < (3, 10).

- **[minor]** (advisory-addition) `cogni-publishing/tests/test-design-render.sh:739` — drnd-36-python-floor-compiles is the guard the revision added for the merger's Python-floor finding. It prints SKIP on CI, whose runner has no 3.9 to 3.11 interpreter, so the floor is enforced only on hosts that happen to have one. A future PR that reintroduces 3.12-only syntax in the render path would still pass CI. This is stated honestly in the suite header and does not breach item 22, which covers the fidelity cases.
  - *Suggested fix:* Record a follow-up to provision a 3.9 interpreter in the plugin-test CI job (outside this PR's workflow scope), or state the local-only enforcement next to the other runtime follow-ups.

### Positive observations

- All 26 committed contract items graded satisfied by the contract-grading reviewer and independently by plugin-validator, corroborated by live guard runs. (contract-grading-reviewer, plugin-validator)
- The recorded mutation recipe replayed via the installed managed-service harness: mutated red, restored green, verdict guard_verified on drnd-10-frozen-copy; under the mutation the renderer refuses its own page, so a damaged renderer is refused rather than shipped. (contract-grading-reviewer, plugin-validator)
- Manifest and structure are clean: version 0.0.4 equals the merge base in both manifests; check-structure-all.sh over all 9 plugins and check-frontmatter.sh are clean. (plugin-validator)
- The full repo guard battery exits 0 at the PR head, and all 18 CI checks pass. (plugin-validator)
- The PR merges cleanly with current main (e50f28a5, the role-token change to the same validator); in the merged tree the publishing contracts suite passes 27/27, the render suite 34 with 2 skipped, and run-plugin-tests --filter cogni-publishing 8/8. (plugin-validator)
- Nothing is forked: composition validation reuses validate-publishing.py and theme tokens reuse generate-tokens-css.py; all validation and fidelity checks run before any write, staged then moved into place. (plugin-validator)
- The runtime boundary is well built: playwright-core exact-pinned with a sha512 lock, install dirs gitignored and untracked, provisioned only by an operator script, resolved only from the provisioning record, measure.mjs runs offline. (plugin-validator)
- The render path genuinely runs on Python 3.9.6: render, compare, check-plan and check-runtime-lock all succeed. (plugin-validator)
- design-render SKILL.md: third-person description with eight quoted trigger phrases, none retired; commands match the wrapper's argparse; progressive disclosure through a 'Read it when' table; font substitution is never silent. (skill-reviewer, skill-quality-reviewer)
- publishing-validate SKILL.md: the now-false 'validate covers the @1 chain only' sentence is replaced accurately; the new check-plan Run-it line and Exit 0 fields match the code. (skill-reviewer, skill-quality-reviewer)
- No maintainer breadcrumbs in either SKILL.md; both bodies are well under their length caps. (skill-reviewer, skill-quality-reviewer)

### Recommendation

---

claude-cogni-work:
## Deferred follow-up issues filed

Scope deferred during this revision has been filed as tracked follow-ups so it
is not silently dropped:

- #1963 — design-render: make drawn SVG label geometry agree with the target-resolved-plan@2 estimates (entity nodes and chart labels wrap as the plan assumes)

claude-cogni-work:
## Revision posted

**Commit:** 1be222f3d6c73366769c2bfa683532967639384e
**Addressed:** the merger NEEDS_REVISION verdict on `91200324`: 12 findings fixed, 1 deferred to a tracked follow-up, 1 kept waived. The gating major was that the portability scan refused frozen copy such as "Risk profile:", "cogni-workspace" or "/tmp/". The scan is now scoped to the surfaces a page can load from, and a new recorded recipe proves the scope. The chart now draws negative values from a shared zero baseline, and drnd-02 passes honestly on the Python 3.9 floor.
**New verdict:** ⏸️ DEFERRED_TO_MERGER — this author lane runs `--defer-review`, so the merger forms the fresh verdict; the prior verdict is https://github.com/cogni-work/insight-wave/pull/1961#issuecomment-5657716134

### Findings resolved (14)

- **[major]** `cogni-publishing/scripts/render_checks.py:252` — check_assets scanned the whole serialized page, copy included, with an unanchored `file:` and bare workspace, runtime and path patterns.
  - *Resolution:* the scan now reads only `reference_surfaces(root)`: `<style>` text and attribute values other than navigational `<a href>` and the identity attributes `id`, `class`, `data-*` and `aria-*`. Text nodes are never read. `file:` is anchored as a URL scheme, so it never matches inside a word. `cogni-workspace`, `node_modules` and `ms-playwright` match only as path segments. The drive-letter pattern now matches a single backslash, as intended. The src/href/url()/@import/`<link>`/aria arms are unchanged.
  - `drnd-37-prose-paths-render` renders the prose case. It passes `check-html`, including a scanned `<meta name="description" content="Risk profile: low">`.
  - `drnd-38-local-reference-refused` proves real references are still refused on each surface: a `file:` URL in an attribute, a `/Users/` path in `<style>` text and a `node_modules/` path in a style attribute.
  - A second recorded recipe mutates the scope line `if node.tag == "style":`, and `drnd-37` must fail: verdict `guard_verified`. Against the pre-revision code, `drnd-37` is red; with the scan made blind, `drnd-38` is red. `design-render.md` now states the scan scope.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/scripts/html_adapter.py:186` — bars were scaled by abs(value), so a negative point was drawn like a positive one.
  - *Resolution:* chart marks now share one zero baseline. Positive values extend right, negative values extend left, and a baseline line is drawn when a negative value exists. An all-positive series draws byte-identical marks; the page gains one token-only `.baseline` CSS rule.
  - `check_patterns` independently asserts that the marks share one baseline, taking each sign from the brief, never the page.
  - `drnd-39-chart-negative-baseline` covers a -2.1 point, whose literal label is shown. A doctored page that draws it rightward from the baseline fails check `chart`.
  - The mixed-unit half needed no change: a chart with two units is already rejected upstream as `mismatched-unit`, before render. `design-render.md`'s Figures rule states both.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:3` — the description was about 1004/1024 characters and stated the HTML-only scope three times.
  - *Resolution:* dropped the redundant "It applies whenever…" sentence and tightened the standalone clause. "It renders the html target only." is the single scope statement. The folded length is now 943 characters, and all eight earlier quoted triggers are kept.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:3` — no plain-language trigger for a standalone web page.
  - *Resolution:* added "export the composition as a standalone HTML page". It appears in no row of `cogni-workspace/references/retired-trigger-phrases.tsv` and in no other SKILL.md; drnd-01 stays green.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/design-render/SKILL.md:53` — the line said `data.findings[].check` names the rule, then listed codes.
  - *Resolution:* reworded: each finding carries a `code` under a `check`, for example `copy-changed`/`copy-omitted` under `frozen-copy` and `description-missing` under `description`. Added `local-reference`, with a note that copy naming a path is never one.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:18-20` — the `check-plan` row sat between the two handoff rows.
  - *Resolution:* moved it directly under `resolve-config`, so the design-compose and design-render handoff rows close the table.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:3` — an awkward boundary sentence, and about 983/1024 characters.
  - *Resolution:* the closing sentence now reads "It never writes briefs; composing a semantic-composition@2 belongs to design-compose, rendering to design-render.", naming both boundaries. The standalone clause is tightened. The folded length is now 949 characters, and all eight quoted triggers are unchanged.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:43` — the Exit 0 bullet was one hard-to-scan sentence.
  - *Resolution:* split into one sub-bullet per command (normalize, validate @1, validate @2, check-plan, resolve-config), with the same facts and fields.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:44 (Read the result, Exit 1 bullet)` — plan findings were routed back to the brief author, and `data.artifact` was never mentioned.
  - *Resolution:* appended the suggested routing: for `check-plan` or `validate` on a plan@2 chain, quote `data.artifact`. `semantic_composition` goes to design-compose. `target_resolved_plan` means re-render through design-render, never a hand edit.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/skills/publishing-validate/SKILL.md:47` — the pointer sent finding-code explanations to a section that never mentions them.
  - *Resolution:* design-render.md §target-resolved-plan@2 is kept for the layout rules. Finding codes now point at `artifact-contracts.md`, whose `typography-relaxed` row is widened to cover the plan-side case, a canvas slot set below the pattern's minimum role.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/tests/test-design-render.sh:206` — drnd-02 used `sys.stdlib_module_names`, which exists only from 3.10.
  - *Resolution:* a real fallback, not a SKIP. On an interpreter without the attribute, a module counts as stdlib when it is built in or frozen, or resolves under sysconfig's stdlib paths outside any `site-packages`/`dist-packages` segment. An unresolvable module fails closed. Probed on `/usr/bin/python3` 3.9.6: `json`, `re`, `_json` and `ast` are accepted; `pip`, `setuptools` and a nonexistent module are rejected. The segment rule exists because Apple's framework `site-packages` sits under its stdlib directory. drnd-02 passes on 3.9.6 and on 3.14.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** `cogni-publishing/tests/test-design-render.sh:739` — drnd-36 SKIPs on CI, so the floor is enforced only on hosts with a 3.9–3.11 interpreter.
  - *Resolution:* took the reviewer's second remedy and stated the local-only enforcement next to the runtime follow-up, in `cogni-publishing/CLAUDE.md` and in the PR body's follow-up entry for #1958. Provisioning a 3.9 interpreter in CI belongs with that runtime-provisioning follow-up; this lane did not edit that issue's scope.
  - *Commit:* 1be222f3d6c73366769c2bfa683532967639384e
- **[minor]** (waived) `cogni-publishing/scripts/html_adapter.py:229` — SVG labels never wrap while the plan estimates multi-line entity slots (latent).
  - *Justification:* deferred to tracked follow-up #1963 with a five-item acceptance checklist. It is latent: plan and drawing agree for every current fixture. A real fix changes the plan layout rules for the entities and series slots, and possibly the captured plan fixture that drnd-25 pins. That is a deliberate layout-rule change, not a fidelity defect in this renderer.
- **[minor]** (waived) `PR diff` — the diff is 4578 lines against a 400-line bound.
  - *Justification:* advisory by design, and a revising agent cannot close it: the only remedy is splitting already-pushed history, which the no-force-push contract forecloses. It was waived at emit time and stays waived.

Also in this commit: `cogni-publishing/README.md` records the second render recipe beside the first, so the plugin's recipe list stays complete. The PR body now carries this head's evidence, the second recipe, #1963 and the drnd-36 note.

Auto-revise triggered by bot-authored verdict findings — no human feedback
since the last automated review. See the linked verdict comment above for
the original findings.

## Follow-up issues

Scope deferred during this revision was filed as tracked follow-ups so it is not
silently dropped:
- #1963 — design-render: make drawn SVG label geometry agree with the target-resolved-plan@2 estimates

sdh07:

**PR:** https://github.com/cogni-work/insight-wave/pull/1961
**Perspectives consulted:** cogni-service:service-review, plugin-dev:plugin-validator, plugin-dev:skill-reviewer, cogni-service:skill-quality-reviewer, cogni-service:contract-grading-reviewer

**Summary:** All 7 dispatched perspectives passed, including the 26-item committed review contract for issue #1946; the only findings are minor — a YAML colon-space defect duplicated across both new SKILL.md descriptions, a waived oversized-PR advisory, and a handful of merger-added preference nits demoted to advisory-addition.

### Verdicts by perspective

| Perspective | Verdict |
|---|---|

### Findings (10)

- **[minor]** (auto-fixable) (advisory-addition) `cogni-publishing/scripts/design-render.py:1` — The four new render scripts (design-render.py, render_core.py, html_adapter.py, render_checks.py) and runtime/provision.sh are tracked as 100644. Sibling entry points such as validate-publishing.py and select-theme.py are 100755. Every documented call goes through python3/bash, so nothing breaks, but design-render.py has a shebang and still cannot run directly.
  - *Suggested fix:* Run `git update-index --chmod=+x` on cogni-publishing/scripts/design-render.py and cogni-publishing/runtime/provision.sh, or drop the shebang on design-render.py, so the mode matches the entry-point convention.

- **[minor]** (waived) `PR diff` — Diff is 4775 lines (bound: 400). PRs this large are harder to review thoroughly, so a blind pass is riskier — consider splitting if the scope allows.
  - *Justification:* Advisory by design, and structurally unclosable by a revising agent: the only remedy is splitting already-pushed history, which this repo's no-force-push contract forecloses. Waived at emit time so it renders with the (waived) audit line but never enters a non-waived gating partition.

- **[minor]** (advisory-addition) `skills/design-render/SKILL.md:19` — Workflow step 1 tells the reader to confirm check-composition, but the Commands block does not include that command's invocation, so the reader has to open design-compose to run it. The step also says render repeats the same validation, which makes the step optional in practice.
  - *Suggested fix:* Either add the one-line `validate-publishing.py check-composition --brief ... --composition ...` invocation to the Commands block, or reword step 1 as optional ('Optionally pre-check with check-composition; render repeats it').

- **[minor]** (auto-fixable) `skills/design-render/SKILL.md:3` — The description is an unquoted YAML plain scalar that contains a colon followed by a space: "Rendering is standalone: no model API, ...". YAML 1.2 forbids that sequence inside a plain scalar, so strict loaders (PyYAML, strict js-yaml, the Agent Skills reference validator) reject the frontmatter with 'mapping values are not allowed here'. Neither repo instrument catches it: check-frontmatter.sh and check-skill-spec.py both read frontmatter line by line. The same PR brings the identical defect into publishing-validate/SKILL.md:3. At base that sentence used an em dash and no colon.
  - *Suggested fix:* Replace ': ' with ' — ' ('Rendering is standalone — no model API, network, Node or cogni-workspace; ...'), matching the convention the sibling descriptions used at base. Apply the same one-character fix to publishing-validate/SKILL.md:3. Verified: PyYAML 6.0.3 safe_load raises ScannerError on both touched SKILL.md frontmatters at head 1be222f3.

- **[minor]** (advisory-addition) `skills/design-render/SKILL.md:3` — The description is about 943 characters folded, leaving only about 80 characters under the 1024 cap. About 190 of those characters are implementation claims (frozen-copy insertion, standalone/no Node, optional measurement runtime) that the body's Boundaries section already states. The contract's risk watchlist expects the PPTX sibling to land in this same SKILL.md and will need room for its own trigger phrases.
  - *Suggested fix:* Cut the description down to the trigger sentence plus one what-it-produces sentence, and leave the fidelity/standalone claims in the Boundaries section, freeing roughly 190 characters for the sibling target's triggers.

- **[minor]** (advisory-addition) `skills/design-render/SKILL.md:48` — The invalid-theme bullet covers only three causes (slug mismatch, no tokens/, missing role). The normative reference (references/design-render.md:32) lists two more: no theme.md, and a token that does not compile or has the wrong unit (theme-tokens, theme-token-unit), for which the bullet's advice is the wrong fix.
  - *Suggested fix:* Add a clause covering a token that does not compile or carries the wrong unit — fix it through manage-themes — or point the bullet at design-render.md for the full sub-check list.

- **[minor]** (auto-fixable) (advisory-addition) `skills/publishing-validate/SKILL.md body §Read the result, Exit 1 bullet (line 49)` — The bullet routes check-plan and plan@2 validate failures by data.artifact, but names only semantic_composition and target_resolved_plan. check_plan calls validate_composition first, which can return data.artifact 'normalized_brief' (most often caused by passing the whole normalize envelope as --brief) or 'pattern_library'; neither value is given a route despite the skill just having been told to route by data.artifact.
  - *Suggested fix:* Append routing guidance for `normalized_brief` (save the `data` envelope alone and re-run) and `pattern_library` (report rather than work around) after the existing 'never edit the plan by hand' sentence.

- **[minor]** (advisory-addition) `skills/publishing-validate/SKILL.md:3` — The description still opens with 'check or normalize a brief before anything renders it', but the same edit makes the skill validate a renderer's target-resolved-plan@2, which is renderer output checked after layout. The opening clause understates the scope and a question about an already-rendered plan may read as out of scope.
  - *Suggested fix:* Reword the opening clause, e.g. 'This skill should be used when the user wants to check or normalize a brief, or verify an artifact chain or render plan against it —', staying under the 1024-character folded cap.

- **[minor]** (advisory-addition) `skills/publishing-validate/SKILL.md:3` — By a manual estimate, the folded description is about 950 characters against the 1024-character hard cap in scripts/check-skill-spec.py, leaving roughly 70 characters of headroom before a later edit could trip the hard, unbaselined arm.
  - *Suggested fix:* No change needed now. On the next edit, cut an enumerated detail already carried by the body table (e.g. the '(Pyramid, SCQA, MECE)' parenthetical) before adding a trigger phrase.

- **[minor]** (advisory-addition) `skills/publishing-validate/SKILL.md:3,18,32` — check-plan is now documented in two skills: publishing-validate has the new trigger/table row/command line, and design-render also lists `validate-publishing.py check-plan` in its own Commands block. The trigger phrases are distinct and a mis-route does no harm, but it is a second copy of the invocation that can drift if the flags change.
  - *Suggested fix:* Keep publishing-validate as the single home for the check-plan invocation; have design-render's Commands block and its 'A plan finding' bullet refer to publishing-validate's check-plan instead of repeating the command line.

### Positive observations

- Every diff path falls in the four admitted groups (cogni-publishing/**, docs/command-reference.md, root README.md counts, one paragraph in docs/architecture/publishing-transition.md); no other plugin, workspace or workflow tree is touched. (plugin-dev:plugin-validator)
- All 8 cogni-publishing suites pass at head 1be222f3 (drnd: 37 PASS, 2 legitimate SKIP for the absent browser runtime), and every relevant repo guard (skill-spec, command-reference-sync, readme-inventory-sync, marketplace-manifest-sync, breadcrumbs, case-id-pairing, result-line-plainness, mutation-recipe-harness, attribution-path-integrity, code-span-nesting, output-style-placement, mcp-tool-grant, version-bump) exits 0. (plugin-dev:plugin-validator)
- Both recorded mutation recipes (drnd-10-frozen-copy, drnd-37-prose-paths-render) return guard_verified when replayed on a scratch copy with the installed managed-service harness, and a trial merge of current main (645c1b79) into the PR head auto-merges without conflict with all suites staying green. (plugin-dev:plugin-validator)
- Every referenced path resolves, every new schema/lockfile/fixture parses, and render_core delegates validation to the real publishing validators via importlib rather than forking a schema. (plugin-dev:plugin-validator)
- design-render's description is third-person with nine specific quoted trigger phrases and no collision with retired render-chain phrases; every referenced file and documented subcommand/flag exists and matches the argparse definitions. (plugin-dev:skill-reviewer)
- design-render's body is a lean 77 lines / 9534 chars under the 15000 cap, with full detail pushed to references/ behind a 'Read it when' resource table, and error-handling routes ambiguous codes by command and data.artifact/check rather than by code alone. (plugin-dev:skill-reviewer)
- The worked success example names the actual font substitution used by the bundled theme, so substitution is never silent to the user. (plugin-dev:skill-reviewer)
- item-18 and item-24 hold at the skill level: the font fallback chain is machine-readable and documented, and none of the nine trigger phrases overlaps a retired phrase in cogni-workspace/references/retired-trigger-phrases.tsv. (cogni-service:skill-quality-reviewer)
- Rules explain their reasoning rather than stacking commands (never-edit, lockfile-pinned runtime, routing by artifact/check), and there are no maintainer breadcrumbs — the @2 and -v1/-v2 tokens are genuine contract-version and schema-filename values. (cogni-service:skill-quality-reviewer)
- Routing between publishing-validate and design-render is clean in both directions; neither skill claims the other's operation, and the new 'check the render plan' trigger phrase is unique and not in the retired-trigger-phrases ledger. (plugin-dev:skill-reviewer)
- Every cross-reference in the publishing-validate edit (design-render.md heading, artifact-contracts.md codes, argparse flags) resolves at the PR head, and Exit-1 guidance routes check-plan findings by data.artifact while forbidding hand-editing the plan, matching the plugin's own convention. (plugin-dev:skill-reviewer)
- Every new claim in the publishing-validate diff checks out against the PR head (check-plan flags, exit-0 summary keys, plan@2 chain validation, finding codes); the body stays lean at 5558 of 12000 chars with an imperative voice and reasoned 'never' rules. (cogni-service:skill-quality-reviewer)
- All 26 committed acceptance-checklist items for issue #1946 are graded satisfied: scope stays inside the four allowed groups, no version was touched, the design-render skill and its stdlib-only HTML renderer are well-formed, the pinned runtime is exact-versioned and gitignored, and the full drnd-* fidelity suite (37 pass, 2 legitimate browser-runtime skips) plus every relevant repo guard ran clean against the PR head. (cogni-service:contract-grading-reviewer)

### Recommendation

---

